PR #2, opened 2022-02-07
Benchmarks

cmyr:
This is a mess of code containing benchmarks, as well as various things I implemented in order to benchmark them, or try to understand differences in some of the results.

You can run the benchmarks locally by checking this out and doing `cargo bench`; otherwise there is a summary available in [benchmarks.md](https://github.com/cmyr/explore-font-types/blob/benchmarks/toy-types/benchmarks.md).

I'm in the process of writing up a summary of what I'm taking away from this, which I will share elsewhere. :)


rsheeter:
Is this still something you want to merge or can we close it?

cmyr:
this is ancient, and definitely not mergeable, although it might be nice to keep the branch around.

PR #28, opened 2022-09-24
Add some fuzzing

madig:
How do I make this do something useful


rsheeter:
Are you planning to proceed with this? - if not plz close.

madig:
I don't know what fuzzing would be useful so eh.

cmyr:
I definitely think there's some interesting possibilities here, but let's wait until things stabilize a bit more.

PR #54, opened 2022-10-14
Add docs/codegen-tour.md

cmyr:
This is a big brain dump explaining the various bits of code that we generate, and how they fit together.

[Rendered](https://github.com/cmyr/fontations/blob/codegen-tour/docs/codegen-tour.md)

### DIFF
--- a/docs/codegen-tour.md
+++ b/docs/codegen-tour.md
@@ -15,7 +15,9 @@ require refinement.
 
 - [overview](#overview)
 - [`read-fonts`](#read-fonts)
-    - [scalars and `BigEndian<T>`](#scalars-detour)
+    - [the code we don't generate](#what-we-dont-generate)
+        - [scalars and `BigEndian<T>`](#scalars-detour)
+        - [`FontData`](#font-data)
     - [tables and records](#tables-and-records)
     - [tables](#read-tables)
         - [`FontRead` and `FontReadWithArgs`](#font-read-args)
@@ -54,9 +56,22 @@ We will examine each of these crates separately.
 
 ## <a id="read-fonts"></a> `read-fonts`
 
-### <a id="scalars-detour"></a> A brief detour on scalars and `BigEndian<T>`
+### <a id="what-we-dont-generate"></a> The code we *don't* generate
 
-#### a description of the problem
+Although this writeup is focused specifically on the code we generate, that code
+is closely entwined with code that we hand-write. This is a general pattern: we
+manually implement some set of types and traits, which are then used in our
+generated code.
+
+All of the types which are used in codegen are reexported in the
+[`codegen_prelude`][read-prelude] module; this is glob imported at the top of
+evey generated file.
+
+We will describe various of these manually implemented types as we encounter
+them throughout this document, but before we get started it is worth touching on
+two cases: `FontData` and scalars / `BigEndian<T>`.
+
+#### <a id="scalars-detour"></a> Scalars and `BigEndian<T>`
 
 Before we dive into the specifics of the tables and records in `read-fonts`, I
 want to talk briefly about how we represent and handle the [basic data types](ot-data-types)
@@ -89,7 +104,7 @@ I have taken a slightly different approach, which tries to be more ergonomic and
 intuitive to the user, at the cost of having a slightly more complicated
 implementation.
 
-#### `BigEndian<T>` and `Scalar`
+##### `BigEndian<T>` and `Scalar`
 
 Our design has two basic components: a trait, `Scalar` and a type
 `BigEndian<T>`, which look like this:
@@ -131,7 +146,7 @@ trait Scalar {
 }
 ```
 
-But this is not []currently something we can express with Rust's generics,
+But this is not currently something we can express with Rust's generics,
 although [it should become possible eventaully](generic-const-exprs).
 
 In any case: what this lets us do is avoid having two separate sets of types for
@@ -142,7 +157,7 @@ advantage that we can define new types in our generated code that implement
 useful for things like custom enums and flags that are defined at various points
 in the spec.
 
-### `FixedSize`
+##### `FixedSize`
 
 In addition to these two traits, we also have a [`FixedSize`][] trait, which is
 implemented for all scalar types (and later, for structs consisting only of
@@ -159,6 +174,22 @@ pub trait FixedSize: Sized {
 This is implemented for both all the scalar values, as well as all their
 `BigEndian` equivalents; and in both cases, the value of `RAW_BYTE_LEN` is the
 size of the raw (big-endian) representation.
+
+#### <a id="font-data"></a> `FontData`
+
+The [`FontData`][] struct is at the core of all of our font reading code. It
+represents a pointer to raw bytes, augmented with a bunch of methods for safely
+reading scalar values from that raw data.
+
+It looks approximately like this:
+
+```rust
+pub struct FontData<'a>(&'a [u8]);
+```
+
+And can be thought of as a specialized interface on top of a Rust byte
+slice.This type is used extensively in the API, and will show up frequently in
+subsequent code snippets.
 
 ### <a id="tables-and-records"></a> tables and records
 
@@ -260,6 +291,17 @@ cases, however, this is not possible. A simple example is the [`loca` table][loc
 the data for this table cannot be interpreted correctly without knowing the
 number of glyphs in the font (stored in the `maxp` table) as well as whether the
 format is long or short, which is stored in the `head` table.
+
+> ***note***:
+>
+> The `FontRead` trait is similar the 'sanitize' methods in HarfBuzz: that is to
+> say that it does not parse the data, but only ensures that it is well-formed.
+> Unlike 'sanitize', however, `FontRead` is not recursive (it does not chase
+> offsets) and it does not in anyway modify the structure; it merely returns an
+> error if the structure is malformed.
+>
+> We will likely want to change the name of this method at some point, to
+> clarify the fact that it is not exactly *reading*.
 
 In either case, the generated table code is very similar.
 
@@ -846,3 +888,5 @@ clarify.
 [`FontWrite`]: https://docs.rs/write-fonts/latest/write_fonts/trait.FontWrite.html
 [`FixedSize`]: https://docs.rs/font-types/latest/font_types/trait.FixedSize.html
 [generic-const-exprs]: https://github.com/rust-lang/rust/issues/60551#issuecomment-917511891
+[read-prelude]: https://github.com/cmyr/fontations/blob/main/read-fonts/src/lib.rs#L42
+[`FontData`]: https://docs.rs/read-fonts/latest/read_fonts/struct.FontData.html
--- a/docs/codegen-tour.md
+++ b/docs/codegen-tour.md
@@ -16,6 +16,7 @@ require refinement.
 - [overview](#overview)
 - [`read-fonts`](#read-fonts)
     - [scalars and `BigEndian<T>`](#scalars-detour)
+    - [tables and records](#tables-and-records)
     - [tables](#read-tables)
         - [`FontRead` and `FontReadWithArgs`](#font-read-args)
         - [versioned tables](#versioned-tables)
@@ -53,22 +54,6 @@ We will examine each of these crates separately.
 
 ## <a id="read-fonts"></a> `read-fonts`
 
-In the [`read-fonts`][] crate, we make a distinction between *table* objects and
-*record* objects, and we generate different code for each.
-
-The distinction between a *table* and a *record* is blurry, but the
-specification offers two "general criteria":
-
-> - Tables are referenced by offsets. If a table contains an offset to a
-> sub-structure, the offset is normally from the start of that table.
-> - Records occur sequentially within a parent structure, either within a
-> sequence of table fields or within an array of records of a given type. If a
-> record contains an offset to a sub-structure, that structure is logically a
-> subtable of the record’s parent table and the offset is normally from the start
-> of the parent table.
->
-> ([The OpenType font file][otff])
-
 ### <a id="scalars-detour"></a> A brief detour on scalars and `BigEndian<T>`
 
 #### a description of the problem
@@ -84,6 +69,11 @@ exactly match the memory layout (including byte ordering) of the underlying font
 file; this is necessary for us to take advance of zerocopy semantics (see the
 [zerocopy section](#zerocopy) below.)
 
+In addition to endianness, it is also sometimes the case that types will be
+represented by a different number of bytes in the raw file than when are
+manipulating them natively; for instance `Offset24` is represented as three
+bytes on disk, but represented as a `u32` in native code.
+
 This leads us to a situation where we require two distinct types for each
 scalar: a native type that we will use in our program logic, and a
 'raw' type that will represent the bytes in the font file (as well as some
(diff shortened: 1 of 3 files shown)


docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+the [`write-fonts`][] crate is focused on compilation. The two crates contain a
+parallel `tables` module, with a nearly identical set of type definitions: for
+instance, [both crates][read-name-record] [contain a][write-name-record] `tables::name::NameRecord` type.
+
+We will examine each of these crates separately.
+
+## <a id="read-fonts"></a> `read-fonts`
+
+In the [`read-fonts`][] crate, we make a distinction between *table* objects and
+*record* objects, and we generate different code for each.
```
rsheeter, line 57: It would be helpful to explain what the difference is, and if they are *always* different. Is it *only* offset resolution or is there more to it?

If a table and a record have identical fields and none of them are offsets to sub-structures would they be identical?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+> sub-structure, the offset is normally from the start of that table.
+> - Records occur sequentially within a parent structure, either within a
+> sequence of table fields or within an array of records of a given type. If a
+> record contains an offset to a sub-structure, that structure is logically a
+> subtable of the record’s parent table and the offset is normally from the start
+> of the parent table.
+>
+> ([The OpenType font file][otff])
+
+### <a id="scalars-detour"></a> A brief detour on scalars and `BigEndian<T>`
```
rsheeter, line 72: Suggestion: make the discussion of basic data types the first ting in the doc, before we hit table vs record.

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+
+#### a description of the problem
+
+Before we dive into the specifics of the tables and records in `read-fonts`, I
+want to talk briefly about how we represent and handle the [basic data types](ot-data-types)
+of which records and tables are composed.
```
rsheeter, line 78: I feel like there should be a link to https://learn.microsoft.com/en-us/typography/opentype/spec/otff here somewhere?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+exactly match the memory layout (including byte ordering) of the underlying font
+file; this is necessary for us to take advance of zerocopy semantics (see the
+[zerocopy section](#zerocopy) below.)
+
+This leads us to a situation where we require two distinct types for each
+scalar: a native type that we will use in our program logic, and a
+'raw' type that will represent the bytes in the font file (as well as some
+mechanism to convert between them.)
+
+There are various ways we could express this in Rust. The most straightforward
```
rsheeter, line 92: Are there other options we considered beyond these two?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+would be to just have two parallel sets of types: for instance alongside the
+`F2Dot14` type, we might have `RawF2Dot14`, or `F2Dot14Be`. Another option might
+be to have types that are generic over byte-order, such that you end up with
+types like `U16<BE>` and `U16<LE>`.
+
+There's one additional complication around these values: certain types in the
```
rsheeter, line 98: This feels like it would fit very naturally mixed into the section above. E.g.

There are two cases where we need to alter the raw value of a field on access:

1. Endian conversion
   *  The font format uses big endian ([spec](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types)). We typically want the value in host endian, such as little endian for x86.
1. Word size
   * Our target machines tend to use 4 byte words so when the font uses 3 bytes for a value, such as uint24 or Offset24, we convert to i32

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+spec are stored as three bytes, (`uint24` and `Offset24`) but in native code it
+is easier and more efficient to represent these as 32-bit integers. This means
+that the raw and normal types may not have the same size.
+
+I have taken a slightly different approach, which tries to be more ergonomic and
+intuitive to the user, at the cost of having a slightly more complicated
+implementation.
```
rsheeter, line 105: IMHO you could delete "at the cost of having a slightly more complicated implementation" since you directly discuss the implementation below.

rsheeter: I suggest it would be helpful to compare/contrast with HarfBuzz.

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+
+```rust
+trait Scalar {
+    const RAW_SIZE: usize;
+    fn from_raw(bytes: [u8; Self::RAW_SIZE]) -> Self;
+    fn to_raw(self) -> [u8; Self::RAW_SIZE];
+}
+```
+
+But this is not currently something we can express with Rust's generics.
```
rsheeter, line 149: Will it be in future? Is there an RFC or in flight proposal that would permit this?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+/// A trait for types that have a known, constant size.
+pub trait FixedSize: Sized {
+    /// The raw (encoded) size of this type, in bytes.
+    const RAW_BYTE_LEN: usize;
+}
+```
+
+This is implemented for both all the scalar values, as well as all their
+`BigEndian` equivalents; and in both cases, the value of `RAW_BYTE_LEN` is the
+size of the raw (big-endian) representation.
```
rsheeter, line 175: suggest link to where all this is implemented

rsheeter: oh wait you already did!

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+
+### <a id="read-tables"></a> tables
+
+Conceptually, a table object is additional type information laid over a
+`FontData` object (a wrapper around a rust byte slice (`&[u8]`), essentially
```
rsheeter, line 180: FontData just popped out of nowhere, I was scared, should we have a brief discussion about FontData first?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+    pub fn format(&self) -> u16 {
+        self.0.read_at(0)
+    }
+}
+```
+
+In practice, what we generate is slightly different: instead of
+generating a struct for the table itself (and wrapping the data directly)
+we generate a 'marker' struct, which defines the type of the table, and then we
+combine it with the data via a `TableRef` struct.
```
rsheeter, line 199: I feel like there is a missing "We do this <because reasons>"

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+        let range = self.shape.format_byte_range();
+        self.data.read_at(range.start)
+    }
+}
+```
+
+To the user these two API are equivalent (you have a type `MyTable`, on which
+you can call methods to read fields) but the 'marker' pattern potentially allows
+for us to do some fancy things in the future (involving various cases where we
+want to store a type separate from a lifetime).
```
rsheeter, line 238: I suppose this is the why, but ... it's pretty impenetrable for a Rust noob. I think we want it to make sense to a Rust noob.

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+#### <a id="font-read-args"></a> `FontRead` & `FontReadWithArgs`
+
+After generating the type definitions, the next thing we generate is an
+implementation of one of [`FontRead`][] or [`FontReadWithArgs`][]. The
+`FontRead` trait is used if a table is self-describing: that is, if the data in
+the table can be fully interpreted without any external information. In some
+cases, however, this is not possible. A simple example is the [`loca` table][loca-spec]:
+the data for this table cannot be interpreted correctly without knowing the
+number of glyphs in the font (stored in the `maxp` table) as well as whether the
+format is long or short, which is stored in the `head` table.
```
rsheeter, line 258: I immediately find myself wondering why we distinguish FontRead from FontReadWithArgs instead of always having args, just sometimes it's empty?

I assume the answer is ergonomics but ... wouldn't it be nice if we had just one FontRead type? Would reading a table with no args, say  `head` then look like `.read(data, Default::default())` or some such?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+
+In either case, the generated table code is very similar.
+
+For the purpose of illustration, let's imagine we have a table that looks like
+this:
+
+```rust
+table Foob {
```
rsheeter, line 266: I suggest doing a real example might be helpful for a reader familiar with fonts and less familiar with Rust. I think many readers will fit that description. OS/2 with some fields ommitted perhaps?

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+}
+```
+
+This generates the following code:
+
+```rust
+impl<'a> FontRead<'a> for Foob<'a> {
+    fn read(data: FontData<'a>) -> Result<Self, ReadError> {
+        let mut cursor = data.cursor();
+        let version: u16 = cursor.read()?;
```
rsheeter, line 285: Suggest annotation, such as to describe

1. Which fields we actually read, e.g. always version if there is a version
1. Why we don't coalesce consecutive fixed size advances
1. The relationship to HarfBuzz sanitize, which for at least some readers will be very helpful

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+Finally, having finished with each field, we call the `finish` method on the
+cursor: this performs a final bounds check, and instantiates the table with the
+provided marker.
+
+> ***note***:
+>
+> The `FontRead` trait is currently doing a bit of a double duty: in the case of
+> tables, it is expected to perform a very minimal validation (essentially just
+> bounds checking) but in the case of records it serves as an actual parse
+> function, returning a concrete instance of the type. It is possible that these
```
rsheeter, line 341: suggest explaining why

docs/codegen-tour.md
```diff
@@ -0,0 +1,843 @@
...
+
+#### <a id="versioned-tables"></a> versioned tables
+
+As hinted at above, for tables that are versioned (which have a version field,
+and which have more than one known version value we do not generate a distinct
+table per version; instead we generate a single table. For fields that are
+available on all versions of a table, we generate getters as usual. For fields
+that are only available on certain versions, we generate getters that return an
+`Option` type, which will be `Some` in the case where that field is present for
+the current version.
```
rsheeter, line 352: I immediately wonder if it would be nice to access all the fields at a version as a group after *one* Option resolution. If I want to code against say COLRv1 I'd rather do one check then have access to all the fields, e.g.

```
if let Some(colrv1) = colr_table.v1_fields {
  // colrv1 has non-Option fields, I can use them all
}
```

docs/codegen-tour.md
cmyr: I hope this is explained in the subsequent sections, where I walk through each case separately, let me know if that is unclear :)

docs/codegen-tour.md
cmyr: I agree this makes sense.

docs/codegen-tour.md
cmyr: I believe that the link in this section should resolve to that page, and specifically the data types table?

docs/codegen-tour.md
cmyr: I considered three, total: `u16`/`RawU16`, `U16<BE>`/`U16<LE>`, and `u16`/`BigEndian<u16>`.

docs/codegen-tour.md
cmyr: I'm not sure I know enough about harfbuzz? My understanding is that harfbuzz takes advantage of C++'s more liberal overloading to perform conversion when the value is accessed/read, but I'm not sure if there's anything else interesting going on?

docs/codegen-tour.md
cmyr: Yes, I believe this will be possible when we get the [`generic_const_exprs`](https://github.com/rust-lang/rust/issues/76560) feature, although that currently has no clear timeline for stabilization.

docs/codegen-tour.md
cmyr: Hmm, where exactly? I'm staring at this and I'm not seeing it. :/

docs/codegen-tour.md
cmyr: Unfortunately explaining the full rationale for this requires some familiarity with intermediate-to-advanced Rust. I could certainly go into more detail, but providing enough detail to fully explain the situation to a Rust noob would likely end up being another thousand words or so, and it ultimately is out of scope of the sort of high-level overview I'm trying to provide, here.

The idea I'm trying to express in this section is that it doesn't really matter *exactly* what types are involved in representing a table, right now; in either case the overall logic will be very similar.

docs/codegen-tour.md
cmyr: Yes, as you say `FontRead` is the overwhelmingly common case, and requiring the user to always pass in some empty arguments would significantly degrade the API.

In practice, `FontReadWithArgs` is more-or-less exclusively called from generated code, and the user just needs to call a getter:

https://github.com/cmyr/fontations/blob/f968425082d91d539db2f4b073f1927688681705/read-fonts/generated/generated_gpos.rs#L2073-L2078

docs/codegen-tour.md
cmyr: for the first two points, are they answered by the section below? I walk through this line by line, explaining what code we generate for each field.

👍 on sanitize

docs/codegen-tour.md
cmyr: The point of these blockquote notes is to point out things that I think are a bit awkward, e.g. I really don't have a good explanation of why 😒

docs/codegen-tour.md
cmyr: Yea, my sense here is that we should see how the ergonomics feel in practice, and we can revisit as needed.

This also gets complicated when you think of multiple versions. Does `color_table.v2_fields` also return `v1` fields?

I *could* imagine a world where we have a sort of 'upcasting' ability, where you can do, like:

```rust
if let Some(color_v1) = color_table.as_color_v1() {
    // color_v1 is a separate type that includes non-option getters
}
```

I think this pattern could be interesting, if we end up unsatisfied with the current ergonomics.

docs/codegen-tour.md
rsheeter: Is it necessary to make me work for it? - as reader I would have appreciated a terse explanation right away.

docs/codegen-tour.md
rsheeter: To an extent I see making the explicit refs as a means to credit the giants whose shoulders we stand on :)

docs/codegen-tour.md
rsheeter: I would posit someone who bothers to read this IS interested in exactly what types and why. Maybe drop the makes sense to a Rust noob requirement if that lets it be both terse and informative. Right now it reads to me as hand wavily saying "there's some advanced stuff I can't be bothered to explain."

docs/codegen-tour.md
rsheeter: If it's wrapped up in a getter then I think I come back to why do we need to specialize the no args case?

docs/codegen-tour.md
rsheeter: I think that would be worth noting?

rsheeter (approved):
Couple of questions or nits noted in comments, overall LGTM, very helpful, ty for writing this up

rsheeter:
I just realized we never merged this. Despite having open comments, some of which we may in time wish to address, I think we should. So I will.

PR #63, opened 2022-10-19
Add CPAL

rsheeter:
WIP addition of CPAL; opening PR because I like having the diff handy. Currently fails as follows:

Notes from the experience:

```shell
# preprocessor instructions are curiously out of band

# preprocessor is destructive ... and can't capture all the data we need
# https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-type-array
# has consts, what do I do with those?

# had to hand-add #![parse_module(read_fonts::tables::colr)] to get past
    Error: 
    × parsing failed
    ╭─[resources/codegen_inputs/cpal.rs:1:1]
    1 │ /// [CPAL (Color Palette Table)](https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-table-header) table
    · ▲
    · ╰── expected #![parse_module(..)] attribute
    2 │ table Cpal0 {
    ╰────

# The preprocessor left me with *two* tables, table Cpal0 and table Cpal1
# based on colr.rs I manually edited this to one table Cpal
# Not sure if this was user error or not

# Indentation in result of preprocessor is inconsistent

# having to manually edit every offset sucked
# I would have preferred to just write the schema and include what the offset target is, no destructive preproc step

# This is confusing

    × parsing failed
        ╭─[resources/codegen_inputs/cpal.rs:41:1]
    41 │ /// [CPAL (Color Record)](https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-entries-and-color-records) record
    42 │ record ColorRecords {
        ·      ──────┬─────
        ·            ╰── This record contains an array, and so must have a lifetime
    43 │     /// Color records for all palettes
        ╰────

# Apparently to make a Something Array you declare a *table* with read args?
    # and those args are primitive typed (any u16 will do!) and thus very easy to get wrong?

# Apparently since COLR was added the path pattern for generated files changed so my include!("../../generated/generated_colr.rs"); failed?

# Apparently this means "add read_offset_with"
error[E0277]: the trait bound table_ref::TableRef<'_, PaletteLabelEntryArrayMarker>: read::FontRead<'_> is not satisfied
   --> read-fonts/src/tables/../../generated/generated_cpal.rs:179:24
    |
179 |             .map(|x| x.resolve(data))?
    |                        ^^^^^^^ the trait read::FontRead<'_> is not implemented for table_ref::TableRef<'_, PaletteLabelEntryArrayMarker>
    |
```


cmyr:
want to rebase this on `main`?

cmyr:
now... why isn't CI running?

cmyr:
> Notes from the experience:
> 
> ```shell
> # preprocessor instructions are curiously out of band

Can you elaborate?

> # preprocessor is destructive ... and can't capture all the data we need

Unless I'm missing something, the preprocessor is only destructive if you pipe it to an existing file? 

The way I would think about the preprocessor is that it is a very dumb tool that makes it easier to sketch out the codegen input using copy/paste, instead of requiring you to write each entity manually.

> # https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-type-array
> # has consts, what do I do with those?

You can actually define a set of bitflags in your codegen input, which would look like:

```rust
flags u32 PaletteFlags {
    /// palette is appropriate to use when displaying the font on a light background such as white.
    USABLE_WITH_LIGHT_BACKGROUND = 0x0001,
    /// palette is appropriate to use when displaying the font on a dark background such as black.
    USABLE_WITH_DARK_BACKGROUND  = 0x0002,
}
```

> # had to hand-add #![parse_module(read_fonts::tables::colr)] to get past
>     Error: 
>     × parsing failed
>     ╭─[resources/codegen_inputs/cpal.rs:1:1]
>     1 │ /// [CPAL (Color Palette Table)](https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-table-header) table
>     · ▲
>     · ╰── expected #![parse_module(..)] attribute
>     2 │ table Cpal0 {
>     ╰────

yes, this is something you always need to manually add; it is about the relationship between `write-fonts` and `read-fonts`. The codegen code does not know anything about the module structure of the crates.

> # The preprocessor left me with *two* tables, table Cpal0 and table Cpal1
> # based on colr.rs I manually edited this to one table Cpal
> # Not sure if this was user error or not

this is a relic from when we used to generate a distinct table object for each version.

> # Indentation in result of preprocessor is inconsistent

yes, this is annoying

> # having to manually edit every offset sucked
> # I would have preferred to just write the schema and include what the offset target is, no destructive preproc step

This is definitely one of the major annoyances, and one of the things that would be nice about a machine-readable spec that included more info.

> # This is confusing
> 
>     × parsing failed
>         ╭─[resources/codegen_inputs/cpal.rs:41:1]
>     41 │ /// [CPAL (Color Record)](https://learn.microsoft.com/en-us/typography/opentype/spec/cpal#palette-entries-and-color-records) record
>     42 │ record ColorRecords {
>         ·      ──────┬─────
>         ·            ╰── This record contains an array, and so must have a lifetime
>     43 │     /// Color records for all palettes
>         ╰────
> 
> # Apparently to make a Something Array you declare a *table* with read args?
>     # and those args are primitive typed (any u16 will do!) and thus very easy to get wrong?

This is a consequence of us not expecting offsets to point to bare arrays, something that has only come up (so far) in COLR/CPAL.


> # Apparently since COLR was added the path pattern for generated files changed so my include!("../../generated/generated_colr.rs"); failed?

I've opened #66 which should prevent this from happening again.

> # Apparently this means "add read_offset_with"
> error[E0277]: the trait bound table_ref::TableRef<'_, PaletteLabelEntryArrayMarker>: read::FontRead<'_> is not satisfied
>    --> read-fonts/src/tables/../../generated/generated_cpal.rs:179:24
>     |
> 179 |             .map(|x| x.resolve(data))?
>     |                        ^^^^^^^ the trait read::FontRead<'_> is not implemented for table_ref::TableRef<'_, PaletteLabelEntryArrayMarker>
>     |
> ```

The clue here is: `FontRead` is not implemented for my type, but the type exists. This must mean either that it is not a table, or it is a table that does not implement `FontRead` (which means it should implement `FontReadWithArgs`)

read-fonts/src/tests/test_fonts/README.md
```diff
@@ -0,0 +1,6 @@
+The ttx files are the masters. To update the binaries:
+
+```shell
+# in the ttx dir
```
andyjgf, line 4: the test_fonts dir?

resources/codegen_plan.toml
```diff
@@ -13,6 +13,11 @@ mode = "parse"
 source = "resources/codegen_inputs/colr.rs"
 target = "read-fonts/generated/generated_colr.rs"
 
+[[generate]]
+mode = "parse"
+source = "resources/codegen_inputs/cpal.rs"
+target = "read-fonts/generated/generated_cpal.rs"
+
 [[generate]]
```
andyjgf, line 21: Do you also need a block with mode = "compile"? I have one in my PR.

read-fonts/src/tests/test_fonts/README.md
rsheeter: Fixed, ty

resources/codegen_plan.toml
rsheeter: Hm. It observably compiles but ... that does seem consistent with others. Curiously colr doesn't have one.

rsheeter: Fixed

andyjgf (approved):

read-fonts/src/tests/test_data.rs
```diff
@@ -329,3 +329,20 @@ pub mod post {
...
+    pub fn sample() -> Cpal<'static> {
+        return FontRef::new(COLR_GRADIENT_RECT).unwrap().cpal().unwrap();
+    }
+}
+
+pub mod test_fonts {
+    use crate::FontData;
+
+    pub static COLR_GRADIENT_RECT: FontData<'static> =
+        FontData::new(include_bytes!("test_fonts/linear_gradient_rect_colr_1.ttf"));
```
cmyr, line 347: I would keep test data out of the `src` directory; maybe `resources/test_fonts`?

(including these in the src directory will cause them to be automatically included in the `.crate` that is uploaded to crates.io/downloaded by any users)

read-fonts/src/tests/test_fonts/README.md
```diff
@@ -0,0 +1,6 @@
+The ttx files are the masters. To update the binaries:
+
+```shell
+# in the test_fonts dir
+for f in $(ls *.ttx); do ttx -o ${f/.ttx/.ttf} $f; done
```
cmyr, line 5: Maybe we should just include a `rebuild.sh` script? I might also keep the ttx & ttf in separate folders, but that's just me.

resources/codegen_plan.toml
cmyr: fwiw: the 'compile' entry is only necessary if you are also adding support for the module in `write-fonts`.

resources/codegen_inputs/cpal.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Array. Set to 0 if no array is provided.
+    #[available(1)]
+    #[nullable]
+    #[read_offset_with($num_palettes)]
+    palette_labels_array_offset: BigEndian<Offset32<PaletteLabelArray>>,
+    /// Offset from the beginning of CPAL table to the Palette Entry 
+    /// Labels Array. Set to 0 if no array is provided.
+    #[available(1)]
+    #[nullable]
+    #[read_offset_with($num_palettes)]
```
cmyr, line 40: should this be `$num_palette_entries`?

cmyr (commented):
A few notes:

the `compile` entry in the codegen plan isn't necessary until you are adding support in `write-fonts`.

The ttx for testing looks like a very useful approach, I can see us adopting that broadly.

There are a few more observations inline.

read-fonts/src/tests/test_data.rs
rsheeter: TIL! Will fix.

resources/codegen_plan.toml
```diff
@@ -13,6 +13,16 @@ mode = "parse"
...
+[[generate]]
+mode = "parse"
+source = "resources/codegen_inputs/cpal.rs"
+target = "read-fonts/generated/generated_cpal.rs"
+
+[[generate]]
+mode = "compile"
+source = "resources/codegen_inputs/cpal.rs"
+target = "write-fonts/generated/generated_cpal.rs"
+
```
cmyr, line 25: oh late note but we can leave this out until we're actually generating the `write-fonts` module.

cmyr (approved):
a few fixups in my previous review, but overall this looks good!

read-fonts/src/tables/cpal.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    #[test]
+    fn read_sample() {
+        let table = test_data::cpal::sample();
+        assert_eq!(table.version(), 0);
+        assert_eq!(table.num_palette_entries(), 2);
+        assert_eq!(table.num_palettes(), 2);
+        assert_eq!(table.num_color_records(), 4);
+
+        if let Some(Ok(marker)) = table.color_records_array() {
```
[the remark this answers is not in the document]
rsheeter: Makes sense, updated so

read-fonts/src/tests/test_fonts/README.md
rsheeter: Moved to separate folders. As rebuild is a one-liner I did not add an sh. I could be convinced to do so.

read-fonts/src/tests/test_data.rs
rsheeter: Fixed

resources/codegen_inputs/cpal.rs
rsheeter: Yes!

PR #66, opened 2022-10-19
[ci] Ensure codegen is up-to-date

cmyr:
This adds a github actions task that runs the codegen tool and ensures it does not produce any changes; this ensures that generated files which have been commited match the commited inputs.

This might cause issues when the files have been generated in windows? If this is the case we should figure out why, and get windows producing the same outputs as the unixes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -79,6 +79,30 @@ jobs:
           command: test
           args: --all-targets --all-features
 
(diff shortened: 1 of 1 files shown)


cmyr:
Okay I have manually tested this by adding a commit that changed the codegen plan, and verified that CI failed in that case, so I think this is good to go.

rsheeter:
LGTM, merging

PR #67, opened 2022-10-19
Add font collection (ttc) support

dfrg:
Initial support for reading ttc files. Also adds an index command line flag to otexplorer.

The API probably needs some adjustments. `FontRef::new` should likely take an `index` parameter to make this consistent.

### DIFF
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -81,6 +81,7 @@ pub enum FileRef<'a> {
 }
 
 impl<'a> FileRef<'a> {
+    /// Creates a new reference to a file representing a font or font collection.
     pub fn new(data: FontData<'a>) -> Result<Self, ReadError> {
         Ok(if let Ok(collection) = CollectionRef::new(data) {
             Self::Collection(collection)
@@ -89,20 +90,15 @@ impl<'a> FileRef<'a> {
         })
     }
 
+    /// Returns an iterator over the fonts contained in the file.
     pub fn fonts(&self) -> impl Iterator<Item = Result<FontRef<'a>, ReadError>> + 'a + Clone {
-        let copy = self.clone();
-        let len = match self {
-            Self::Collection(collection) => collection.len(),
-            _ => 1,
+        let (iter_one, iter_two) = match self {
+            Self::Font(font) => (Some(Ok(font.clone())), None),
+            Self::Collection(collection) => (None, Some(collection.iter())),
         };
-        (0..len).filter_map(move |ix| match &copy {
-            Self::Font(font) => Some(Ok(font.clone())),
-            Self::Collection(collection) => collection.get(ix),
-        })
+        iter_two.into_iter().flatten().chain(iter_one)
     }
 }
-
-const TTC_HEADER_TAG: Tag = Tag::new(b"ttcf");
 
 /// Reference to the content of a font collection file.
 #[derive(Clone)]
@@ -112,6 +108,7 @@ pub struct CollectionRef<'a> {
 }
 
 impl<'a> CollectionRef<'a> {
+    /// Creates a new reference to a font collection.
     pub fn new(data: FontData<'a>) -> Result<Self, ReadError> {
         let header = TTCHeader::read(data)?;
         if header.ttc_tag() != TTC_HEADER_TAG {
@@ -121,32 +118,32 @@ impl<'a> CollectionRef<'a> {
         }
     }
 
+    /// Returns the number of fonts in the collection.
     pub fn len(&self) -> u32 {
         self.header.num_fonts()
     }
 
+    /// Returns true if the collection is empty.
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
 
-    pub fn get(&self, index: u32) -> Option<Result<FontRef<'a>, ReadError>> {
-        if index >= self.len() {
-            return None;
-        }
+    /// Returns the font in the collection at the specified index.
+    pub fn get(&self, index: u32) -> Result<FontRef<'a>, ReadError> {
         let offset = self
             .header
             .table_directory_offsets()
-            .get(index as usize)?
+            .get(index as usize)
+            .ok_or(ReadError::InvalidCollectionIndex(index))?
             .get() as usize;
-        match TableDirectory::read(self.data.slice(offset..)?) {
-            Ok(table_directory) => Some(FontRef::with_table_directory(self.data, table_directory)),
-            _ => None,
-        }
+        let table_dir_data = self.data.slice(offset..).ok_or(ReadError::OutOfBounds)?;
+        FontRef::with_table_directory(self.data, TableDirectory::read(table_dir_data)?)
     }
 
+    /// Returns an iterator over the fonts in the collection.
     pub fn iter(&self) -> impl Iterator<Item = Result<FontRef<'a>, ReadError>> + 'a + Clone {
         let copy = self.clone();
-        (0..self.len()).filter_map(move |ix| copy.get(ix))
+        (0..self.len()).map(move |ix| copy.get(ix))
     }
 }
 
@@ -158,10 +155,12 @@ pub struct FontRef<'a> {
 }
 
 impl<'a> FontRef<'a> {
+    /// Creates a new reference to a font.
     pub fn new(data: FontData<'a>) -> Result<Self, ReadError> {
         Self::with_table_directory(data, TableDirectory::read(data)?)
     }
 
+    /// Returns the data for the table with the specified tag, if present.
     pub fn table_data(&self, tag: Tag) -> Option<FontData<'a>> {
         self.table_directory
             .table_records()
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -71,6 +71,63 @@ pub(crate) mod codegen_prelude {
 
 include!("../generated/font.rs");
 
+const TTC_HEADER_TAG: Tag = Tag::new(b"ttcf");
+
+pub struct FontFileRef<'a> {
+    pub data: FontData<'a>,
+}
+
+impl<'a> FontFileRef<'a> {
+    pub fn new(data: FontData<'a>) -> Self {
+        Self { data }
+    }
+
+    pub fn is_collection(&self) -> bool {
+        self.ttc_header().is_some()
+    }
+
+    pub fn num_fonts(&self) -> u32 {
+        if let Some(ttc_header) = self.ttc_header() {
+            ttc_header.num_fonts()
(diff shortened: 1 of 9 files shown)


read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+            _ => 1,
+        };
+        (0..len).filter_map(move |ix| match &copy {
+            Self::Font(font) => Some(Ok(font.clone())),
+            Self::Collection(collection) => collection.get(ix),
+        })
+    }
+}
+
+const TTC_HEADER_TAG: Tag = Tag::new(b"ttcf");
```
cmyr, line 105: Maybe this should live in `font-types`? that's where we put the consts for the sfnt versions.

We could also consider ignoring the spec in this case and treating this as a `u32`, so that it's consistent with the defined sfnt versions, which would let us do something nice, like:

```rust
// in FileRef
pub fn new(data: FontData<'a>) -> Result<Self, ReadError> {
    match data.read_at::<u32>(0)? {
        TTC_HEADER_TAG => CollectionRef::new(data).map(Self::Collection),
        TT_SFNT_VERSION | CFF_SFNT_VERSION => FontRef::new(data).map(Self::Collection),
        other => Err(ReadError::InvalidSfnt(other),
    }
}
```
We could even continue to use the `Tag` type to represent the field elsewhere, and just use a separate constant here.

I'm not sure this is particularly worth changing, but wanted to point it out.

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+/// Reference to the content of a font or font collection file.
+pub enum FileRef<'a> {
+    /// A single font.
+    Font(FontRef<'a>),
+    /// A collection of fonts.
+    Collection(CollectionRef<'a>),
+}
+
+impl<'a> FileRef<'a> {
+    pub fn new(data: FontData<'a>) -> Result<Self, ReadError> {
```
cmyr, line 84: I think we should start adopting the policy of documenting all public API.

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+        Ok(if let Ok(collection) = CollectionRef::new(data) {
+            Self::Collection(collection)
+        } else {
+            Self::Font(FontRef::new(data)?)
+        })
+    }
+
+    pub fn fonts(&self) -> impl Iterator<Item = Result<FontRef<'a>, ReadError>> + 'a + Clone {
```
cmyr, line 92: this `+ Clone` bound seems out of place, and this should compile without it?

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+        let copy = self.clone();
+        let len = match self {
+            Self::Collection(collection) => collection.len(),
+            _ => 1,
+        };
+        (0..len).filter_map(move |ix| match &copy {
+            Self::Font(font) => Some(Ok(font.clone())),
+            Self::Collection(collection) => collection.get(ix),
+        })
+    }
```
cmyr, line 102: Just for fun, my personal preferred way of writing this sort of thing is,

```rust
        let (iter_one, iter_two) = match self {
            FileRef::Font(font) => (Some(Ok(font.clone())), None),
            FileRef::Collection(collection) => (None, Some(collection.iter())),
        };
        iter_two.into_iter().flatten().chain(iter_one)
```

this is basically just at trick to ensure we have a single concrete return type.

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+            .table_directory_offsets()
+            .get(index as usize)?
+            .get() as usize;
+        match TableDirectory::read(self.data.slice(offset..)?) {
+            Ok(table_directory) => Some(FontRef::with_table_directory(self.data, table_directory)),
+            _ => None,
+        }
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = Result<FontRef<'a>, ReadError>> + 'a + Clone {
```
cmyr, line 147: again don't think we need `Clone`?

And also for fun (it's fine either way) the alternative to the inner clone would be,

```suggestion
    pub fn iter<'b: 'a>(&'b self) -> impl Iterator<Item = Result<FontRef<'a>, ReadError>> + 'a {
```

although I honestly might prefer your version just for avoiding the confusion of an extra lifetime in the signature.

resources/codegen_inputs/font.rs
```diff
@@ -29,3 +29,19 @@ record TableRecord {
     /// Length of the table.
     length: BigEndian<u32>,
 }
+
+/// [TTC Header](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#ttc-header)
+table TTCHeader {
+    ttc_tag: BigEndian<Tag>,
+    #[version]
+    version: BigEndian<Version16Dot16>,
```
cmyr, line 37: This should be `MajorMinor`, not `Version16Dot16`.

read-fonts/src/table_ref.rs
```diff
@@ -6,6 +6,7 @@ use crate::{
     offset::{Offset, ResolveOffset},
 };
 
+#[derive(Clone)]
```
cmyr, line 9: I had originally avoided adding this intentionally to enforce some abstract notion of purity, but... it's probably fine? The only thing we're going to be cloning is the marker, but we might want to discourage having this be used widely?

If so, and if we need this, I previously had a private method called `sneaky_copy` that i used in the traversal code.

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+
+    pub fn len(&self) -> u32 {
+        self.header.num_fonts()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn get(&self, index: u32) -> Option<Result<FontRef<'a>, ReadError>> {
```
cmyr, line 132: I could maybe make an argument for not doing the `Option<Result<_>>` dance here, and just say that passing an invalid index is a valid error condition (in a way that it is not an error condition of a nullable offset is null).

read-fonts/src/lib.rs
```diff
@@ -71,6 +71,86 @@ pub(crate) mod codegen_prelude {
...
+        }
+        let offset = self
+            .header
+            .table_directory_offsets()
+            .get(index as usize)?
+            .get() as usize;
+        match TableDirectory::read(self.data.slice(offset..)?) {
+            Ok(table_directory) => Some(FontRef::with_table_directory(self.data, table_directory)),
+            _ => None,
+        }
```
cmyr, line 144: Could we make the `table_directory_offsets` actually be offsets, and then implement `FontRead` for `FontRef`? 

Also I'm confused by the `None` branch at the bottom; if we can't read the data we should be returning `Some(Err(_))`, I think.

cmyr (commented):
Went over this with a nit-toothed comb, some comments are likely more useful than others; use your judgement. :)

read-fonts/src/lib.rs
dfrg: That was my first thought but was trying to avoid any direct parsing. I can go either way on this.

read-fonts/src/lib.rs
dfrg: 👍

read-fonts/src/lib.rs
dfrg: I think there’s a general notion that iterators in Rust should be `Clone` if possible. The standard library follows this practice

read-fonts/src/lib.rs
dfrg: Interesting! I haven’t seen this one before

read-fonts/src/lib.rs
dfrg: Yeah, adding a lifetime to `&self` can be annoying for many reasons and often breaks method chaining forcing you to store temporaries in a variable

resources/codegen_inputs/font.rs
dfrg: Thanks! I wasn’t aware of this type

read-fonts/src/table_ref.rs
dfrg: I’m not sure. As above, I like having `Clone` available if it doesn’t break any invariants. I would personally make all of these types `Copy` as well but I found this to be a decent middle ground :)

read-fonts/src/lib.rs
dfrg: This seems reasonable to me

read-fonts/src/lib.rs
dfrg: Table offsets in font collections are based at the root of the file and not the table directory, so the default offset resolution stuff might not work here?

Good catch on the branch. I saw that but forgot to fix it.

read-fonts/src/lib.rs
cmyr: huh, TIL.

read-fonts/src/lib.rs
cmyr: The root of the file is also the location of the ttc header, right? So if the header is resolving these offsets it should work.

Alternatively if you have an offset you can always provide your own data, like:

```rust
let some_offset = my_table.some_offset();
let resolved = some_offset.resolve(my_font_data);
```

which still might be easier than manually indexing etc.

cmyr (approved):
Looks great; only lingering thing is whether or not we want offsets in TtcHeader? If it works then great, and if not then this lgtm. :)

read-fonts/src/lib.rs
dfrg: Ah, yes. I can change that. We just can’t implement `FontRead` for `FontRef` in the case of collections because the base of data and table_directory don’t align in this case.

dfrg:
So I looked at making table_directory_offsets `[BigEndian<Offset32<TableDirectory>>]`. This requires removing the `#[skip_from_obj]` attributes on `TableDirectory` and `TableRecord` which leads to a chain of compile errors in the writing code.

Accounting for those is beyond the scope of this PR so I'm going to leave this as is for now and we can revisit it in the future.

read-fonts/src/lib.rs
dfrg: Those offsets are relative to the start of the file. This is so that fonts in a collection can share tables (typically `cmap`). I was actually pleasantly surprised that `FontRef` contained the data field because I would have add to add it for collections to work.

Issue #68, opened 2022-10-19
Less attributes, more types?


rsheeter:
Feels like I should be able to more directly tell it what I mean. Instead of this:

```rust
    #[nullable]
    #[read_offset_with($num_color_records)]
    color_records_array_offset: BigEndian<Offset32<ColorRecordArray>>,
// ...
#[read_args(num_color_records: u16)]
table ColorRecordArray {
    /// Array of 32-bit flag fields that describe properties of each 
    /// palette. See below for details.
    #[count($num_color_records)]
    color_records: [ColorRecord],
}
```

Let me write something akin to:

```rust
    #[nullable]
    color_records_array_offset: BigEndian<OffsetTo<[ColorRecord; num_color_records]>>,
```

Being able to spell nullable directly (NullableOffsetTo? OptionalOffsetTo?) would be nice too. IMHO this moves the codegen inputs closer to the schema we've mentioned before.

I would love to try to hack this together if it makes sense. Pointers most appreciated :)

rsheeter:
Similarly it would be nice to write arrays without duplicative annotations. Instead of having to relate num_tables and encoding_records twice:

```rust
table Cmap {
    /// Table version number (0).
    version: BigEndian<u16>,
    /// Number of encoding tables that follow.
    #[compile(array_len($encoding_records))]
    num_tables: BigEndian<u16>,
    #[count($num_tables)]
    encoding_records: [EncodingRecord],
}
```

Let me just say it once, for example:

```rust
table Cmap {
    /// Table version number (0).
    version: BigEndian<u16>,
    /// Number of encoding tables that follow.
    num_tables: BigEndian<u16>,
    encoding_records: [EncodingRecord; num_tables],
}
```

cmyr:
Going to close this, I think the things that are easily expressed in types mostly are right now.

PR #70, opened 2022-10-20
[exploratory] Playing with [T; len] style arrays

rsheeter:
Exploratory for now, see #68. The most interesting part is the modified form of resources/codegen_inputs/cmap.rs.


rsheeter:
If we do decide to make this change it would make more sense to start from scratch, likely after #81 is merged, so closing

PR #72, opened 2022-10-20
Document per-file execution in codegen readme

cmyr:


rsheeter (approved):
Ty! This would have saved me a good hour yesterday. I still don't know how to get the codegen tool to actually print any sort of useful help about it's args, I had to read the code and the code was in a macro so then I had to read the (bad) documentation of the macro and how it related to what you actually enter on CLI.

PR #75, opened 2022-10-20
Support offsets to arrays

### LINKED ISSUE #74: Support offsets directly to arrays
This pattern is used quite widely in CPAL & COLR, and will clean up the generated code for these tables significantly.

cmyr:
With this change, you can now have an offset where the inner type is either a) a scalar (`BigEndian<T>`) type, or a record. I believe this covers all the existing cases?

I've updated COLR to use the new syntax.

I've also added support in traversal and `otexplorer`, although I haven't played around much to see what the formatting looks like; we may want to tweak it.

This required some fairly deep changes in our parsing code since we need to distinguish between whether an offset's inner type is an array or not, and then if it *is* we also need to know if it's a scalar. Anyway, it wasn't so bad.

- closes #74 
- relevant to #68

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -1,15 +1,15 @@
 //! raw parsing code
 
-use std::collections::HashMap;
+use std::{collections::HashMap, ops::Deref};
 
 use proc_macro2::{Span, TokenStream};
 use quote::{quote, ToTokens};
 use syn::{
-    braced, bracketed, parenthesized,
+    braced, parenthesized,
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
     spanned::Spanned,
-    token, Attribute, Token,
+    Attribute, Token,
 };
 
 pub(crate) struct Items {
@@ -288,7 +288,7 @@ pub(crate) enum FieldType {
 #[derive(Debug, Clone)]
 pub(crate) enum OffsetTarget {
     Table(syn::Ident),
-    Array(syn::TypeSlice),
+    Array(Box<FieldType>),
 }
 
 /// A representation shared between computed & varlen arrays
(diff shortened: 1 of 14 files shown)


font-codegen/src/parsing.rs
```diff
@@ -270,7 +270,7 @@ impl InlineExpr {
 pub(crate) enum FieldType {
     Offset {
         typ: syn::Ident,
-        target: Option<syn::Ident>,
+        target: Option<OffsetTarget>,
```
rsheeter, line 273: I'm struggling to grasp what an Offset without a target means, perhaps a comment explaining why it's not `target: OffsetTarget`?

rsheeter (commented):
Plz update CPAL too as that is now merged

resources/codegen_inputs/test.rs
```diff
@@ -9,6 +9,11 @@ table KindsOfOffsets {
     /// An offset that is nullable, but always present
     #[nullable]
     nullable_offset: BigEndian<Offset16<Dummy>>,
+    /// count of the array at array_offset
```
rsheeter, line 12: What's the deal with test.rs in codegen inputs? Is it a workaround for lack of a good way to test or ...?

write-fonts/src/offsets.rs
```diff
@@ -159,6 +161,33 @@ where
...
 
+#[cfg(feature = "parsing")]
+impl<'a, const N: usize, T: Scalar> From<Result<&'a [BigEndian<T>], ReadError>>
+    for OffsetMarker<Vec<T>, N>
+where
+    BigEndian<T>: Copy,
+{
+    fn from(from: Result<&'a [BigEndian<T>], ReadError>) -> Self {
+        OffsetMarker::new_maybe_null(from.ok().map(|vec| vec.iter().map(|x| x.get()).collect()))
+    }
```
rsheeter, line 172: This is harder to read than NullableOffsetMarker::new below. Possible to restructure to be easier to read?

read-fonts/src/traversal.rs
```diff
@@ -149,6 +157,7 @@ impl<'a> FieldType<'a> {
     ///
     /// The `resolver` argument is a function that takes an offset and resolves
     /// it.
+    //TODO: rename this to array_of_offsets?
```
rsheeter, line 160: +1 to the todo

font-codegen/src/parsing.rs
```diff
@@ -1109,4 +1179,26 @@ mod tests {
...
+
+    #[test]
+    fn offset_target() {
+        let array_target = make_path_seg("Offset16<[BigEndian<u16>]>");
+        assert!(get_offset_target(&array_target.arguments)
+            .unwrap()
+            .is_some());
+
+        let path_target = make_path_seg("Offset16<SomeType>");
+        assert!(get_offset_target(&path_target.arguments).unwrap().is_some());
```
rsheeter, line 1196: Why assert is_some? Wouldn't it make sense to assert_eq! the expected thing?

rsheeter (approved):
LGTM, much nicer.

edit: please do the same to cpal

font-codegen/src/parsing.rs
cmyr: This is a fairly little-used codepath, but sometimes the target of an offset is not a type we can easily represent in codegen (for instance in the top-level table directory, where the offset can be any different table) or is a type that we haven't implemented yet.

read-fonts/src/traversal.rs
cmyr: agree, going to do this in a separate PR to avoid cluttering this one up with a bunch of renames.

font-codegen/src/parsing.rs
cmyr: (I did a little audit: this is used in a few places where there are variation tables we don't support yet, intentionally in the root `TableDirectory`, and by accident in cmap14)

resources/codegen_inputs/test.rs
rsheeter: Maybe worth putting that explanation into the comments of test.rs or a README.md for codegen_inputs?

font-codegen/src/parsing.rs
rsheeter: The "the target of an offset..." explanation would make a magnificent comment. Definitely would have helped me as reader.

resources/codegen_inputs/test.rs
cmyr: It is mentioned in the codegen readme, but could definitely also be mentioned in the input file itself.

PR #77, opened 2022-10-21
Update CPAL to use offsets to arrays

cmyr:
~This is on top of #75, which will be merged first.~

I'm encouraged by how much code this deletes.


rsheeter:
Good demo of the power of codegen :)

PR #78, opened 2022-10-21
[traversal] Rename offset_array -> array_of_offsets

cmyr:
This becomes confusing once we also support offsets _to_ arrays.


rsheeter:
The hardest part is the names! Ty.

PR #88, opened 2022-10-25
Rename #[phantom] to #[generic_offset]

cmyr:
This better expresses this attribute's purpose.


rsheeter:
> This better expresses this attribute's purpose.

vigorous +1, ty

PR #90, opened 2022-10-26
Add convenience init methods to AnchorTable

cmyr:
Constructing these otherwise is a bit annoying. We might consider generating this sort of code at some point, but we wouldn't want it for all multi-format types; this is perhaps another interesting example where we would want to augment a future schema?

Actually this maybe clarifies part of how I've been thinking about schema/codegen, and which I've struggled to express coherently, thus far. Basically, there are two related and heavily overlapping tasks performed by the current codegen inputs: the first is that they serve as a description of the tables in the spec, and the second is that *they serve as a description of the code we wish to generate*, and these things do not perfectly line up.

### DIFF
--- a/write-fonts/src/layout/gpos.rs
+++ b/write-fonts/src/layout/gpos.rs
@@ -74,6 +74,25 @@ impl<'a> FontRead<'a> for PositionLookupList {
     }
 }
 
+impl AnchorTable {
+    /// Create a new [`AnchorFormat1`] table.
+    pub fn format_1(x_coordinate: i16, y_coordinate: i16) -> Self {
+        Self::Format1(AnchorFormat1 {
+            x_coordinate,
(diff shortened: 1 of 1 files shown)


dfrg (approved):
Simple enough so I'll give it a checkmark. I _am_ curious as to whether anything actually generates `AnchorFormat2` tables in `GPOS`. @anthrotype any idea?

anthrotype:
the [Adobe FEA spec](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#anchor-format-b) supports those and fontTools.feaLib implemements them so one could use them, though I've never seen them in the wild

cmyr:
yea, these are motivated by my work on FEA related stuff. 🤷

PR #92, opened 2022-10-26
Update parser for glyf table

dfrg:
This is basically just some cleanup of the hand-written glyf outline parsing code with a few additions (transforms and instructions for composite components).

I've added some docs, a basic test, and made some small stylistic changes as well.

Also (hopefully) fixes the newline issue on Windows.

### DIFF
--- a/read-fonts/src/font_data.rs
+++ b/read-fonts/src/font_data.rs
@@ -24,6 +24,7 @@ pub struct FontData<'a> {
 /// # Note
 ///
 /// call `finish` when you're done to ensure you're in bounds
+#[derive(Debug, Default, Clone, Copy)]
 pub struct Cursor<'a> {
     pos: usize,
     data: FontData<'a>,
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -1,7 +1,5 @@
 //! The [glyf (Glyph Data)](https://docs.microsoft.com/en-us/typography/opentype/spec/glyf) table
 
-use font_types::ReadScalar;
-
 use crate::{FontData, ReadError};
 
 /// 'glyf'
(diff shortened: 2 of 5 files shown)


read-fonts/src/font_data.rs
```diff
@@ -24,6 +24,7 @@ pub struct FontData<'a> {
 /// # Note
 ///
 /// call `finish` when you're done to ensure you're in bounds
+#[derive(Debug, Default, Clone, Copy)]
```
cmyr, line 27: so... this is not an issue with this patch, but more an observation: something that has been bugging me about `Cursor`, so far, is that it is serving two purposes: it was initially designed specifically for the validation case, and then later on I started using it in places for actually constructing items.

These do feel like two distinct tasks, sort of? The validating cursor does need to be able to read scalars, but it also has a bunch of API for skipping items etc that might not be useful for a more general purpose cursor. Does any of this resonate with you?

read-fonts/src/tables/glyf.rs
```diff
@@ -55,34 +54,33 @@ impl<'a> SimpleGlyph<'a> {
...
+#[derive(Clone)]
+struct PointIter<'a> {
     end_points: &'a [BigEndian<u16>],
     cur_point: u16,
-    flags: OldCursor<'a>,
-    x_coords: OldCursor<'a>,
-    y_coords: OldCursor<'a>,
+    flags: Cursor<'a>,
+    x_coords: Cursor<'a>,
+    y_coords: Cursor<'a>,
```
cmyr, line 74: in case this wasn't obvious: this glyf code predates the current codegen code, and the current implementation of Cursor. I ported it over at some point but found it annoying to express with the existing cursor, and ended up just using the simple cursor I had implemented at the time for that purpose.

cmyr (approved):
Looks good, the impl here was definitely half-baked.

My only comments are really about `Cursor`, and whether or not it feels fit for purpose here? Maybe it would be nice to have a simpler `Cursor` in `read-fonts` that doesn't have any of the validation logic, and which doesn't error?

read-fonts/src/font_data.rs
dfrg: I didn't find it particularly cumbersome to use for this purpose and I did actually need to use the `read_array` method for grabbing the composite bytecode instructions, so I was happy that it was there. It's probably worth revisiting these types after we have a greater portion of the spec implemented and a better understanding of what is necessary.

I'll keep this in mind as we add new tables.

read-fonts/src/tables/glyf.rs
dfrg: The only real difference seems to be the Option/Result distinction. We may want to surface read errors in the glyph parser at some point. I just figured I'd get rid of `OldCursor` while I was in the code.

dfrg:
Thanks! We'll probably want to give this another look down the road, but I think it's in decent shape now and sufficient to build a real truetype scaler.

PR #93, opened 2022-10-26
Add extern record and scalar notation, use in gpos

### LINKED ISSUE #91: Add syntax for declaring external record or scalar
With the removal of `BigEndian`, we no longer have a mechanism whereby the user can communicate in a codegen file whether an unknown type is a scalar or a record. (It is currently possible to use unknown types that are tables, since tables are always referenced via an offset.)

This is a *fairly* uncommon case, but it does exist, and should be supported.

The information we will need to be able to express for these types is the following:
- The name of the type in the read crate
- The name of the type in the write crate (these will generally be the same?)
- possibly metadata/attributes that explain how to read the type? ~For instance if we're declaring a custom record, it is probably because it is non-trivial, which probably means it needs external arguments in order to be parsed. For instance `ValueRecord` requires a `ValueFormat` argument. ~ (Actually this is unnecessary? you'll need to have this metadata at the place where the type is used, but not at the declaration site. It's possible that other attributes may be required, though?)

What should syntax like this look like?

I think the best approach is to match as closely as possible the existing syntax. I think we can do this by introducing a new keyword, `extern` (which is already a rust keyword, although it is no longer used very often).

So basically I think that `ValueRecord` should be declared as,

```rust
extern record ValueRecord;
```

And a custom scalar (let's imagine `ValueFormat` required a custom implementation) would look like,

```rust
extern scalar ValueFormat;
```

And if required, we could add attributes to this.

At parse time, I think we would probably want to parse this into a new category of top-level item.

rsheeter:
Fixes #91

### DIFF
--- a/font-codegen/src/lib.rs
+++ b/font-codegen/src/lib.rs
@@ -88,7 +88,7 @@ pub(crate) fn generate_parse_module(items: &Items) -> Result<proc_macro2::TokenS
             Item::Format(item) => table::generate_format_group(item)?,
             Item::RawEnum(item) => flags_enums::generate_raw_enum(item),
             Item::Flags(item) => flags_enums::generate_flags(item),
-            Item::Extern(..) => TokenStream::new(),
+            Item::Extern(..) => Default::default(),
         };
         code.push(item_code);
     }
--- a/font-codegen/src/lib.rs
+++ b/font-codegen/src/lib.rs
@@ -1,6 +1,7 @@
 //! Generating types from the opentype spec
 
 use log::debug;
+use proc_macro2::TokenStream;
 use quote::quote;
 
 mod error;
@@ -87,6 +88,7 @@ pub(crate) fn generate_parse_module(items: &Items) -> Result<proc_macro2::TokenS
             Item::Format(item) => table::generate_format_group(item)?,
             Item::RawEnum(item) => flags_enums::generate_raw_enum(item),
             Item::Flags(item) => flags_enums::generate_flags(item),
+            Item::Extern(..) => TokenStream::new(),
         };
         code.push(item_code);
     }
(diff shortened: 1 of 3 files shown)


font-codegen/src/lib.rs
```diff
@@ -87,6 +88,7 @@ pub(crate) fn generate_parse_module(items: &Items) -> Result<proc_macro2::TokenS
             Item::Format(item) => table::generate_format_group(item)?,
             Item::RawEnum(item) => flags_enums::generate_raw_enum(item),
             Item::Flags(item) => flags_enums::generate_flags(item),
+            Item::Extern(..) => TokenStream::new(),
```
cmyr, line 91: we can take advantage of the fact that `TokenStream` impls `Default` to write,

```suggestion
            Item::Extern(..) => Default::default(),
```

which... I mean honestly they're both fine. This saves us one line of importing, though!

font-codegen/src/parsing.rs
```diff
@@ -365,6 +366,18 @@ pub(crate) struct BitFlags {
     pub(crate) variants: Vec<RawVariant>,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) enum ExternTarget {
```
cmyr, line 370: maybe `ExternType`, to match `FieldType`? I also might move the ident out of here and into `Extern` directly, since every possible extern item will have a name.

font-codegen/src/parsing.rs
```diff
@@ -519,6 +534,24 @@ impl Parse for RawEnum {
...
+        let target = match input.parse::<syn::Ident>()?.to_string().as_str() {
+            "scalar" => ExternTarget::Scalar(input.parse()?),
+            "record" => ExternTarget::Record(input.parse()?),
+            _ => {
+                return Err(syn::Error::new(
+                    input.span(),
+                    "expected one of 'scalar' 'record'.",
+                ))
+            }
+        };
```
cmyr, line 549: instead of this, we can use [custom keywords](https://docs.rs/syn/latest/syn/macro.custom_keyword.html). We already have a custom keyword for 'record', and we could add one for 'scalar'; search for "mod kw" in this file.

cmyr (approved):
that looks nice and easy; a few notes are inline.

font-codegen/src/parsing.rs
rsheeter: Will do, good catch

font-codegen/src/parsing.rs
rsheeter: Done, ty

PR #94, opened 2022-10-26
Update ttx testing workflow

cmyr:
I want to add a new test of this style, and this patch came out of my playing around to get that working.

A few little changes, mostly polish, probably over-engineering. The heart of this is that there's now a script for rebuilding the outputs, and this script makes a few tweaks to the invocation to make it less likely to produce diffs unless the actual contents change.

~Note: this doesn't muck around with installing fonttools; I don't actually expect the outputs to change much between versions, and I record the version only as a debugging aid.~


anthrotype (commented):
I suggest you also do `$PIP install --upgrade pip` to upgrade pip installer itself so one won't see annoying warning messages from pip saying its not up to date with the latest one. The venv module doesn't automatically upgrade pip after it creates the venv. In general it's advisable to use the latest pip available before installing anything, less bugs usually 

```diff
diff --git a/resources/test_fonts/rebuild.sh b/resources/test_fonts/rebuild.sh
index 27bfda8..c1067ad 100755
--- a/resources/test_fonts/rebuild.sh
+++ b/resources/test_fonts/rebuild.sh
@@ -28,6 +28,7 @@ fi
 if [ ! -d "$VENV_DIR" ]; then
     echo "Setting up venv"
     python3 -m venv .venv
+    $PIP install --upgrade pip
 fi

 echo "Installing fonttools"
```

resources/test_fonts/README.md
```diff
@@ -1,6 +1,15 @@
-The ttx files are the masters. To update the binaries:
+# test files
+
+This directory contains files used for testing. The masters are the ttx files;
+these are human readable/editable. From these, we generate the binary ttf
+files that are the actual test inputs.
+
+## rebuilding
+To update the binaries, run the `./resources/test_fonts/rebuild.sh` from the
+repo root. This script will install the correct version of fonttools, and then
```
rsheeter, line 9: It's a papercut that you have to run from root, should work regardless of where I run it from

resources/test_fonts/rebuild.sh
```diff
@@ -0,0 +1,39 @@
...
+# their xml/ttx representations.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SRC_DIR=./resources/test_fonts/ttx
+OUT_DIR=./resources/test_fonts/ttf
+VENV_DIR=./.venv
+PIP=$VENV_DIR/bin/pip
```
rsheeter, line 13: Establishing the venv already sets pip to point to the right place, what's the point of having this?

rsheeter: (similar for ttx)

rsheeter (approved):
LGTM with nits

resources/test_fonts/rebuild.sh
anthrotype: because he has not activated the venv so the venv's bin directory is not in the environment's PATH variable. I suggested Colin to **not** activate the venv but instead invoke the venv scripts (pip, ttx, etc.) directly using their relative paths, it works both ways.

resources/test_fonts/README.md
cmyr: totally agree, but is there an easy way for me to resolve all the paths I need to access, if I'm starting in an arbitrary working directory? I don't do much shell scripting.

resources/test_fonts/rebuild.sh
cmyr: yes I call the executables directly, instead of activating the venv. This was per @anthrotype's suggestion, but I can see it being nice in the case where the user has an active venv that I don't want to clobber?

anthrotype: It should work either way, because activating a venv from within another venv would place the former at the beginning of the $PATH, and calling activate from within the shell script won't modify the environment of the parent shell that's invoking the script (unless you do source rebuild.sh). I think I prefer the relative path approach without the venv activation, it's more explicit what executable you are actually invoking

resources/test_fonts/README.md
anthrotype:
```bash
SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```

I'm no shell scripter either but according to stack overflow that's how you get a bash script's directory. You could then either prefix that to all relative paths, or cd or pushd to it and popd at the end

https://stackoverflow.com/a/246128

resources/test_fonts/rebuild.sh
anthrotype: On the other hand, an argument in favour of doing `source .venv/bin/activate` would be if the python code you are invoking in turn relies on something else being on the $PATH (e.g. it may invoke some other tool as a subprocess and expect that to be on the $PATH and you may want it to invoke the copy which is inside the venv's bin instead of the one from the global environment). But this unfrequent and definitely is not the case for ttx

resources/test_fonts/README.md
cmyr: okay thank you for doing my googling for me, that ended up being a nice bit of cleanup. :)

anthrotype (approved):

anthrotype:
> Note: this doesn't muck around with installing fonttools

nit: edit the PR description above saying that in fact we _do_ muck around installing fonttools, but safely in its venv

PR #95, opened 2022-10-27
Add STAT

cmyr:
This is based off of #94.

This only adds the `read-fonts` impl; i'll do `write-fonts` as a followup. (I need STAT in the fea parser)


dfrg (approved):
This looks good to me! It seems there’s no way to propagate version information to validate use of format 4 which is gated on version 1.2. I think this is fine, but should there be a comment somewhere about it?

Also not specific to this particular PR, but we should probably add a newtype for name identifiers at some point.

cmyr:
my thinking about format 4 was basically: does this actually matter, in any way? as in, if we encounter a 1.1 table, that just means we shouldn't encounter format 4 subtables, which is fine. If we encounter a version 1.1 subtable and it has format4 tables, that will work 'incorrectly', but I think this is fine?

PR #96, opened 2022-10-27
Rationalize from obj conversions

cmyr:
This cleans up our code for converting the parse types to the compile types.

Previously we were writing a lot of different bespoke impls for these conversions, using a variety of different patterns; this patch replaces a lot of those custom implementations with recursive calls to the `FromObjRef`/`FromTableRef` methods.

This simplifies our codegen code somewhat, but the main advantage is that it is more sustainable as we add new types; one particular wart of the previous implementation was that relied on a blanket impl of `From` to convert offset types, and this was going to break once we had offsets to arrays of records and not just offsets to arrays of scalars, which I ran into in STAT.

The one place where we still need bespoke impls is in handling arrays of offsets; this is because the getters for these types currently return an iterator, but there is no way to write a blanket impl that refers to the iterator's `Item` type, since it is an associated type and not a generic parameter. 

I have an idea for how we could replace the iterator business, which would look something like,

```rust
pub trait ArrayOfOffsets<'a, T> {
    fn len(&self) -> usize;
    fn get(&self, index: usize) -> Result<T, ReadError>;
    fn iter(&self) -> impl Iterator<Item=Result<T, ReadError>>;
}

impl MyTable<'a> {
    fn some_array_of_offsets(&self) -> impl ArrayOfOffsets<'a, OffsetTarget<'a>> { .. }
}
```

but the problem here is the `iter` method, which has a method with a generic param, which means we can't use it with impl Trait until we get GATs? I think? In any case, this is an improvement for now, and unblocks STAT.

edit: and we're getting GATS on November 3rd, so this API change should be possible then.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -1115,6 +1115,10 @@ impl Field {
     fn from_obj_requires_offset_data(&self, in_record: bool) -> bool {
         match &self.typ {
             _ if self.attrs.to_owned.is_some() => false,
+            FieldType::Offset {
+                target: Some(OffsetTarget::Array(_)),
+                ..
+            } => true,
             FieldType::Offset { .. } => in_record,
             FieldType::ComputedArray(_) | FieldType::VarLenArray(_) => true,
             FieldType::Struct { .. } => true,
@@ -1142,13 +1146,19 @@ impl Field {
             FieldType::Scalar { .. } => quote!(obj.#name()),
             FieldType::Struct { .. } => quote!(obj.#name().to_owned_obj(offset_data)),
             FieldType::Offset {
-                target: Some(_), ..
+                target: Some(target),
+                ..
             } => {
                 let offset_getter = self.offset_getter_name().unwrap();
-                if self.is_version_dependent() && !self.is_nullable() {
-                    quote!(obj.#offset_getter(#pass_offset_data).map(|obj| obj.into()))
-                } else {
-                    quote!(obj.#offset_getter(#pass_offset_data).into())
+                match target {
+                    // in this case it is possible that this is an array of
+                    // records that could contain offsets
+                    OffsetTarget::Array(_) => {
+                        quote!(obj.#offset_getter(#pass_offset_data).to_owned_obj(offset_data))
+                    }
+                    OffsetTarget::Table(_) => {
+                        quote!(obj.#offset_getter(#pass_offset_data).to_owned_table())
+                    }
                 }
             }
             FieldType::Array { inner_typ } => match inner_typ.as_ref() {
@@ -1158,7 +1168,7 @@ impl Field {
                 FieldType::Offset { .. } => {
                     let offset_getter = self.offset_getter_name().unwrap();
                     let getter = quote!(obj.#offset_getter(#pass_offset_data));
-                    let converter = quote!(.map(|x| x.into()).collect());
+                    let converter = quote!(.map(|x| x.to_owned_table()).collect());
                     if self.attrs.available.is_some() {
                         quote!(#getter.map(|obj| obj #converter))
                     } else {
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -1120,7 +1120,7 @@ impl Field {
             FieldType::Struct { .. } => true,
             FieldType::Array { inner_typ } => match inner_typ.as_ref() {
                 FieldType::Offset { .. } => in_record,
-                FieldType::Struct { .. } => true,
+                FieldType::Struct { .. } | FieldType::Scalar { .. } => true,
                 _ => false,
             },
             _ => false,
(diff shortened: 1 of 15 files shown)


rsheeter:
The codegen code is alarmingly untested. What would it take to make it possible to unit test it? - for example, I could imagine parsing some snippet of content then passing it to one of the small processor functions and verifying it build sthe right thing.

cmyr:
I broadly agree with you, but there are a few sorts of implicit testing going on:

- codegen generally works by implementing traits or creating instances of types that are defined manually in the host crates. This imposes a major constraint on exactly what codegen can do, if it wants to compile: the generated code needs to match the types/traits, which significantly reduces the 'search space' (not exactly the term i'm looking for, but <something> space) of the possibly-generated-code. In this particular instance `FromObj` we're always starting with some type `T` and generating some type `U`, on a per-field basis. In general there is a fairly small number of `TokenStream`s that we can generate which will compile.
- the generated code *is* tested by the unit tests that we write in the various table modules. If our generated code is incorrect (for instance if a field is of the wrong type in the codegen inputs) then even if parsing is successful, it should cause the affected fields to have incorrect data, which we should catch. Thorough unit tests written *against* the generated  code should be our primary tool of ensuring that it is correct.

That said, there is a particular set of cases that we do not currently test, which I would like to, and that is *false positives*; situations where we successfully parse and generate code where actually the input was malformed in an unexpected way. Testing this would require us to have some mechanism where we can generate some codegen unit and then verify that it *doesn't* compile, which we are not currently able to do. An example of a test like this that I would like is a test where we declare types with fields with an unknown type.

cmyr:
Also note that in this particular PR, the bulk of the changes are the consequence of additions to `resources/codegen_inputs/test.rs`; I wanted to ensure that I had examples there of all the patterns I was dealing with.

rsheeter:
> the generated code is tested by the unit tests that we write in the various table modules

We aren't very dilligent about writing table tests. Let's assume - https://github.com/googlefonts/fontations/issues/119 - we will have tests for all tables. At that point I would say those are integration tests. We should still have unit tests. For example, https://github.com/googlefonts/fontations/pull/93 should have tested it could parse the intended syntax and understand it in the expected way. I would like us to start pushing back on PRs without unit tests as standard practice. If there's a Good Reason something can't have one, fine, but that should be unusual.

font-codegen/src/fields.rs
```diff
@@ -1115,12 +1115,16 @@ impl Field {
...
         match &self.typ {
             _ if self.attrs.to_owned.is_some() => false,
+            FieldType::Offset {
+                target: Some(OffsetTarget::Array(_)),
+                ..
+            } => true,
             FieldType::Offset { .. } => in_record,
             FieldType::ComputedArray(_) | FieldType::VarLenArray(_) => true,
             FieldType::Struct { .. } => true,
             FieldType::Array { inner_typ } => match inner_typ.as_ref() {
```
rsheeter, line 1125: Do you have to have a nested match here or could you have multiple parent level matches?

font-codegen/src/parsing.rs
```diff
@@ -1060,20 +1060,13 @@ fn build_type_map(items: &Items) -> HashMap<String, FieldType> {
...
     } else {
-        Err(syn::Error::new(
-            field_name.span(),
-            format!(
-                "I don't know what this is, do you? {}: {}",
-                field_name.to_string().as_str(),
-                ident
-            ),
-        ))
+        Err(syn::Error::new(field_type.span(), "Error: undeclared type"))
```
rsheeter, line 1069: Why are we reducing the error information? - seeing the name *and* type together, formed how you'd likely find them in code, is useful

rsheeter:
We got GATs, do you want to use here or attack that in a subsequent PR?

rsheeter (approved):
I still don't like how we're approaching testing but I don't think it makes sense to block this on that so LGTM

font-codegen/src/parsing.rs
cmyr: When the error is printed, it will show in the source tree where it occurs, and so the name and type (as well as the file/line number etc) are visible in the final message.

font-codegen/src/fields.rs
cmyr: Unfortunately we need to nest because the inner type is in a `Box`, and so we can't match it directly without the `as_ref` call. :(

font-codegen/src/parsing.rs
cmyr: to clarify, here is the old message and the new:

old:

```
Error: 
  × parsing failed
    ╭─[resources/codegen_inputs/test.rs:59:1]
 59 │     value: u16,
 60 │     other_thing: ExternalThing,
    ·   ─────┬─────
    ·        ╰── I don't know what this is, do you? other_thing: ExternalThing
 61 │ }
    ╰────
```

new:

```
Error: 
  × parsing failed
    ╭─[resources/codegen_inputs/test.rs:89:1]
 89 │     value: u16,
 90 │     other_thing: ExternalThing,
    ·                ──────┬──────
    ·                      ╰── Error: undeclared type
 91 │ }
    ╰────
```

and as an aside, I do not know why all of the spans are always a few columns out of alignment.. 😒

cmyr:
> We got GATs, do you want to use here or attack that in a subsequent PR?

definitely a follow-up thing, needs a bunch of experimenting I suspect.

font-codegen/src/parsing.rs
rsheeter: > I do not know why all of the spans are always a few columns out of alignment

At one point I was firmly convinced I had caused that and then couldn't figure out how before realizing it happened even in the original code rofl

PR #97, opened 2022-10-28
Add blanket From<T> and From<Option<T>> for offset markers

cmyr:
(based off of #96)

This seems generally useful, and is now possible since we've gotten rid of all the other weird From impls in #96


dfrg (approved):
Simple and useful. Lgtm!

PR #98, opened 2022-10-29
More basic traits for FWord and UfWord

cmyr:
This adds default, as well as From/Into the inner numerical types.


dfrg (approved):
Looks good to me.

PR #99, opened 2022-10-29
Tweak signature of NullableOffsetMarker::new

cmyr:
Longer term I don't expect this to be called directly very often, but if it is this signature is a bit nicer, since you don't need to explicity specify Some(_) during normal use.

This also adds NullableOffsetMarker::null as a convenience method. Does this mean we can make new() just take T? Maybe? But I think allowing None in new makes some sense, since this would now be very annoying to construct in the not-unreasonable case that you have an Option<T> in your compiler and you want to convert it to a NullableOffsetMarker<T>.

In any case I expect that once we land #97, the most common way to make an offset marker will just be via From/Into, and this will be moot.

### DIFF
--- a/write-fonts/src/offsets.rs
+++ b/write-fonts/src/offsets.rs
@@ -110,8 +110,15 @@ impl<const N: usize, T> NullableOffsetMarker<T, N> {
 }
 
 impl<const N: usize, T> NullableOffsetMarker<T, N> {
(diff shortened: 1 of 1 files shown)


dfrg:
I agree it’s worth landing #97 and then seeing if this still makes sense.

cmyr:
I no longer think this is necessary.

PR #101, opened 2022-11-01
compile STAT table

cmyr:
(based off of #96, which should go in first)

This adds write-fonts support for STAT.

notes from the commit:

> This required a surprising ammount of attribute trickery :|
> 
> - #[to_owned(..)] on the various flags because of an annoynace where
>   we have duplicate flag definitions and need to express how to map between them.
>   (this will be resolved when we make read-fonts a hard dep)
> - #[compile_type] and #[to_owned] on the offset_to_axis_value_offsets
>   field because we want to use a shim table in read-fonts, but a vec of
>   offsets in write-fonts
> 
> All in all getting this fairly straightforward table working took a
> surprising amount of elbow grease.

It was work on this PR that motivated #96; I had to write a custom FromObjRef impl and was frustrated that there was not a simple predictable mechanism for writing these conversions.


cmyr:
Update: this is ready to review, after it motivated a bunch of cleanup; it's now much more straightforward.

resources/codegen_inputs/stat.rs
```diff
@@ -19,12 +22,15 @@ table Stat {
...
     /// The number of axis value tables.
+    #[compile(array_len($offset_to_axis_value_offsets))]
     axis_value_count: u16,
     /// Offset in bytes from the beginning of the STAT table to the
     /// start of the design axes value offsets array. If axisValueCount
     /// is zero, set to zero; if axisValueCount is greater than zero,
     /// must be greater than zero.
     #[read_offset_with($axis_value_count)]
+    #[compile_type(OffsetMarker<Vec<OffsetMarker<AxisValue>>, WIDTH_32>)]
+    #[to_owned(convert_axis_value_offsets(obj.offset_to_axis_values()))]
```
rsheeter, line 33: Holy leaking Rust into schema. How might we avoid having to do this, have the codegen be able to figure it out from less Rust-specific information?

(maybe we should just add that to the appropriate issue)

rsheeter (approved):
LGTM, I have some concerns w/how much this is leaking Rust into the codegen input but I think that is better tackled in independent issues

resources/codegen_inputs/stat.rs
```diff
@@ -1,16 +1,19 @@
...
+#![parse_module(read_fonts::tables::stat)]
 
 /// [STAT](https://docs.microsoft.com/en-us/typography/opentype/spec/stat) (Style Attributes Table)
 table Stat {
     /// Major/minor version number. Set to 1.2 for new fonts.
     #[version]
+    #[compile(MajorMinor::VERSION_1_2)]
     version: MajorMinor,
     /// The size in bytes of each axis record.
+    #[compile(8)]
```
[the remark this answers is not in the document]
cmyr: What `compile` indicates is that this value cannot be modified by the user, and is instead computed at compile time. The attribute was previously called `compute`, but that also seemed ambiguous.

Maybe this would be clearer if we defined a constant like `AXIS_RECORD_SIZE`, and used that instead of a literal?

resources/codegen_inputs/stat.rs
cmyr: This is 100% specific to the architecture of this crate, and is also a design choice. To walk through how we get here:

- this field is an *offset* to an *array of offsets*
- this means that there are two levels of indirection
- we have a mechanism for indicating arguments for resolving an offset, but if there are multiple layers of offsets in a given field this mechanism becomes ambiguous: do the arguments apply to the first offset or the second?
- This is easy to solve with a shim table, which provides us a simple way to separate the layers, and specify attributes for each one
- and that generates reasonable code for read-fonts, but in write-fonts the extra table is completely superfluous, since in write-fonts we don't need to resolve offsets
- and so to improve the write-fonts API, we say, "this is actually just a vec, and here is how you convert it".

If this was a common pattern, it would make sense to support it in a general way, but it isn't (this is the only place we have encountered it so far). I don't think this is something we would ever want to infer in a schema, and I wouldn't really want to special-case it in the codegen. I think that if we were using a language agnostic schema we would probably just lose this annotation and accept a worse API.

rsheeter: Problem is this moves us a step further away from a language-agnostic schema, it just got harder to get there in a meaningful way. IMHO it would be worth thinking through what it would take to not do this quite so Rust-ily.

PR #106, opened 2022-11-02
Fix arg order in TableProvider::hmtx

### LINKED ISSUE #105: otexplorer throws `hmtx: Error 'An offset was out of bounds'` with NotoColorEmoji
get the font from https://github.com/googlefonts/noto-emoji/blob/main/fonts/Noto-COLRv1.ttf

then try use otexplorer to dump the hmtx table and observe this unexpected error:

```
$ cargo run --bin otexplorer -- -t hmtx ../noto-emoji/fonts/Noto-COLRv1.ttf
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/otexplorer -t hmtx ../noto-emoji/fonts/Noto-COLRv1.ttf`
hmtx: Error 'An offset was out of bounds'
```

cmyr:
fixes #105

### DIFF
--- a/read-fonts/src/table_provider.rs
+++ b/read-fonts/src/table_provider.rs
@@ -33,7 +33,7 @@ pub trait TableProvider<'a> {
         let number_of_h_metrics = self.hhea().map(|hhea| hhea.number_of_h_metrics())?;
         self.expect_data_for_tag(tables::hmtx::TAG)
             .and_then(|data| {
-                FontReadWithArgs::read_with_args(data, &(num_glyphs, number_of_h_metrics))
+                FontReadWithArgs::read_with_args(data, &(number_of_h_metrics, num_glyphs))
             })
     }
 
--- a/read-fonts/src/table_provider.rs
+++ b/read-fonts/src/table_provider.rs
@@ -101,3 +101,56 @@ pub trait TableProvider<'a> {
             .and_then(FontRead::read)
     }
 }
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    /// https://github.com/googlefonts/fontations/issues/105
+    #[test]
+    fn bug_105() {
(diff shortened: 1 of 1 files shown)


rsheeter:
Typically for a bug fix I like to see a test that exposes the bug and is fixed by the change, should be feasible here I think?

rsheeter (approved):
LGTM subject to adding a test

read-fonts/src/table_provider.rs
```diff
@@ -101,3 +101,56 @@ pub trait TableProvider<'a> {
...
+    /// https://github.com/googlefonts/fontations/issues/105
+    #[test]
+    fn bug_105() {
+        // serve some dummy versions of the tables used to compute hmtx. The only
+        // fields that matter are maxp::num_glyphs and hhea::number_of_h_metrics,
+        // everything else is zero'd out
+        struct DummyProvider;
+        impl TableProvider<'static> for DummyProvider {
+            fn data_for_tag(&self, tag: Tag) -> Option<FontData<'static>> {
+                if tag == Tag::new(b"maxp") {
```
anthrotype, line 119: beginner question: would it not be more "idiomatic" to write this if/else block as a match statement?

cmyr: It would, but unfortunately `match` does not allow us to have a 'method call' in a pattern, even if it is a `const` method, so we can't match against `Tag::new(b"xxxx")`, and would need to have `if` guards anyway, like:

```rust
match tag {
    t if t == Tag::new(b"maxp") => (),
   // etc
}
```

which I wouldn't hate, but it doesn't feel like exactly a match statement? it is a bit more concise though..

anthrotype: oh i see. thanks for the explanation!

PR #112, opened 2022-11-03
Implement Default for write-fonts types

### LINKED ISSUE #107: Implement Default for compile types
This will make these types much easier to work with; for instance we will be able to easily create a new empty table and then populate it, or set some fields of a table and then use `..Default::default()` to fill in the blanks.

Importantly for this, I think the default type for non-nullable offsets should be an empty table, instead of a null table. (I am starting to think that normal `OffsetMarker` should not be an `Option<T>` internally; this detail is a consequence of the fact that we want to allow reading a broken file and then mutating its contents, but i think if an offset in a file is malformed we should just record the issue somewhere and insert a new default table in its place, which we will be able to do once we have implemented `Default` for the compile types)

This will likely also require us to add syntax to the codegen inputs to allow the user to specify a default value for a field.

cmyr:
In general this is as simple as adding `Default` to the `#[derive(_)]` attribute, but we also allow the user to specify specific values via a #[default(..)] attribute, in which case we have to generate a manual implementation.

This patch also includes a number of supporting fixups:
- implement `Default` for the font-types scalars (except for MajorMinor and Version16Dot16, for which the appropriate default value is generally context specific)
- Updating a few codegen input files to specify default or compile values for versions where they were missing
- Add explanation for #[default(..)] attr in codegen README
- Use new Default impls where appropriate when constructing types in tests

closes #107

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -82,8 +82,11 @@ impl Fields {
 
     /// `Ok(true)` if no fields have custom default values, `Ok(false)` otherwise.
     ///
-    /// Returns an error if a field that requires a custom default or compile value
-    /// does not have one.
+    /// This serves double duty as a validation method: if we know that default
+    /// should not be derived on a field (such as with version fields) but there
+    /// is no apprporiate annotation, we will return an error explaining the problem.
+    /// This is more helpful than generating code that does not compile, or compile_write_stmt
+    /// but is likely not desired.
     pub(crate) fn can_derive_default(&self) -> syn::Result<bool> {
         for field in self.iter() {
             if !field.supports_derive_default()? {
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -4,6 +4,7 @@ use std::ops::Deref;
 
 use proc_macro2::TokenStream;
 use quote::{quote, ToTokens};
+use syn::spanned::Spanned;
 
 use crate::parsing::{Attr, FieldValidation, OffsetTarget, Phase};
 
@@ -73,6 +74,23 @@ impl Fields {
 
     pub(crate) fn iter_compile_write_stmts(&self) -> impl Iterator<Item = TokenStream> + '_ {
         self.fields.iter().map(Field::compile_write_stmt)
+    }
+
+    pub(crate) fn iter_compile_default_inits(&self) -> impl Iterator<Item = TokenStream> + '_ {
+        self.fields.iter().filter_map(Field::compile_default_init)
+    }
+
+    /// `Ok(true)` if no fields have custom default values, `Ok(false)` otherwise.
+    ///
+    /// Returns an error if a field that requires a custom default or compile value
+    /// does not have one.
+    pub(crate) fn can_derive_default(&self) -> syn::Result<bool> {
+        for field in self.iter() {
+            if !field.supports_derive_default()? {
+                return Ok(false);
+            }
+        }
+        Ok(true)
     }
 
     // we use this to add disable a clippy lint if needed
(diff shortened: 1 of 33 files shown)


font-codegen/src/fields.rs
```diff
@@ -75,6 +76,23 @@ impl Fields {
         self.fields.iter().map(Field::compile_write_stmt)
     }
 
+    pub(crate) fn iter_compile_default_inits(&self) -> impl Iterator<Item = TokenStream> + '_ {
+        self.fields.iter().filter_map(Field::compile_default_init)
+    }
+
+    /// `Ok(true)` if no fields have custom default values, `Ok(false)` otherwise.
```
rsheeter, line 83: I don't find it tremendously obvious why you don't just return bool from this description?

font-codegen/src/record.rs
```diff
@@ -209,13 +209,31 @@ pub(crate) fn generate_compile_impl(
...
+                }
+            }
+        }
+        }
+    });
+
     Ok(quote! {
         #( #docs )*
-        #[derive(Clone, Debug)]
+        #[derive(Clone, Debug, #maybe_derive_default)]
```
rsheeter, line 230: Optional: I dislike the trailing comma when maybe_derive_default is empty. Imagine we add a few more maybe_whatever's, you could get stuff like `#[derivce(Clone, Debug, , , )]`. That would lead me to put the , into maybe_derive_default.

font-codegen/src/table.rs
```diff
@@ -349,14 +349,17 @@ pub(crate) fn generate_group_compile(
...
     let mut variant_decls = Vec::new();
     let mut write_match_arms = Vec::new();
     let mut validate_match_arms = Vec::new();
     let mut from_obj_match_arms = Vec::new();
     let from_type = quote!(#parse_module :: #name);
+    let mut first_var_name = None;
     for var in &item.variants {
+        if first_var_name.is_none() {
+            first_var_name = Some(&var.name);
+        }
```
rsheeter, line 362: I'm failing to grasp why this needs to be an Option and set in the loop. I imagine a GenericGroup with no variants is meaningless so what would be wrong with just writing something like `let first_var_name = &item.variants.first().unwrap().name;`?

font-codegen/src/table.rs
```diff
@@ -418,6 +427,12 @@ pub(crate) fn generate_format_compile(
         quote! ( #( #docs )* #name(#typ) )
     });
 
+    let default_variant = &item
+        .variants
+        .first()
+        .expect("empty format groups not supported")
+        .name;
```
rsheeter, line 434: This feels misplaced to me, shouldn't having >0 variants be an invariant of generic group, checked when parsing and then safe to trust thereafter? - if so you could unwrap because you *know* it's not empty.

I'm actually somewhat surprised the type system doesn't facilitate expressing a "not empty" vector.

font-codegen/src/fields.rs
cmyr: fair, hope that it's clearer now.

font-codegen/src/table.rs
rsheeter: Also this is duplicative of above, maybe both should be calling the same helper?

font-codegen/src/record.rs
cmyr: the trialing comma will be removed when we run rustfmt on the generated code. Does that change your thinking?

cmyr: maybe we should have a trailing comma in `#maybe_derive_default` as well?

write-fonts/src/font_builder.rs
```diff
@@ -85,3 +85,9 @@ fn checksum_and_padding(table: &[u8]) -> (u32, u32) {
 
     (sum.wrapping_add(rem), padding as u32)
 }
+
+impl TTCHeader {
+    fn compute_version(&self) -> MajorMinor {
+        panic!("TTCHeader writing not supported")
```
rsheeter, line 91: nit: yet?

font-codegen/src/table.rs
cmyr: yes fair point.

font-codegen/src/table.rs
cmyr: This is a case of me using `expect` to document why a given unwrap should not be a problem in practice, but I'm happy either way.

write-fonts/src/tables/name.rs
```diff
@@ -14,6 +14,14 @@ impl Name {
...
         .unwrap()
     }
+
+    fn compute_version(&self) -> u16 {
+        if self.lang_tag_record.is_some() {
+            1
+        } else {
+            0
+        }
+    }
```
rsheeter, line 24: is the `if` preferable to something like 

    fn compute_version(&self) -> u16 {
        self.lang_tag_record.map_or_else(|| 0, |_| 1)
    }

rsheeter (approved):
LGTM, couple of nits and questions noted in comments

write-fonts/src/tables/name.rs
cmyr: clippy also doesn't like me here, and has suggested `u16::from(self.lang_tag_record.is_some())` which seems reasonably clear.

PR #113, opened 2022-11-03
No null offsetmarker

cmyr:
This PR makes `OffsetMarker` not be an `Option` internally. It makes `OffsetMarker` implement `Deref<Target=T>`, (and `NullableOffsetMarker` impls `Deref<Target=Option<T>>`) so they now act basically as transparent wrappers.

Basically, I think this was a design mistake. From the commit message:

>For the record, the original idea was that if we were parsing an
existing font into compile types, and we encountered an offset that was
malformed, we would represent that with the `None` case, and then during
validation we would warn if an offset was unexpectedly null.

>As we now expect all tables and records to implement Default, we can
drop the Option, and just insert a new empty table in this particular
case.

This is a big API simplification, and should remove the need for the user to interact directly with these marker types in most cases (and we can hopefully improve API to fix any remaining cases.)

### DIFF
--- a/font-codegen/src/table.rs
+++ b/font-codegen/src/table.rs
@@ -295,7 +295,7 @@ fn generate_to_owned_impl(item: &Table, parse_module: &syn::Path) -> syn::Result
         quote! {
             where
                 U: FontRead<'a>,
-                #t: FromTableRef<U> + 'static,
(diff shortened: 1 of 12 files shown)


rsheeter (approved):
LGTM

PR #116, opened 2022-11-04
Impl default for generated bitflags

cmyr:
The default will always be empty flags. If this is not acceptable, we can add an attribute for specifying the desired default value, or an attribute indicating we won't generate a Default impl.

Default for enums is more annoying: not all enums have a meaningful default value, and we don't want to guess, so maybe for now the best bet is to just require the user to implement the trait, or otherwise just specify an explicit default value for each field in a table that contains a raw enum.

Another approach altogether (which had mentioned in relation to NameIds) would be to not use an actual rust enum to represent these enums, and make them instead by newtype structs with associated constants. This feels un-rusty, but also an awful lot simpler and closer to representing the underlying truth.

### DIFF
--- a/read-fonts/generated/generated_glyf.rs
+++ b/read-fonts/generated/generated_glyf.rs
@@ -214,6 +214,7 @@ impl<'a> std::fmt::Debug for SimpleGlyph<'a> {
 
 bitflags::bitflags! {
     /// Flags used in [SimpleGlyph]
+    #[derive(Default)]
     pub struct SimpleGlyphFlags: u8 {
         /// Bit 0: If set, the point is on the curve; otherwise, it is off
         /// the curve.
@@ -421,6 +422,7 @@ impl<'a> std::fmt::Debug for CompositeGlyph<'a> {
 
 bitflags::bitflags! {
     /// Flags used in [CompositeGlyph]
+    #[derive(Default)]
     pub struct CompositeGlyphFlags: u16 {
         /// Bit 0: If this is set, the arguments are 16-bit (uint16 or
         /// int16); otherwise, they are bytes (uint8 or int8).
--- a/font-codegen/src/flags_enums.rs
+++ b/font-codegen/src/flags_enums.rs
@@ -22,6 +22,7 @@ pub(crate) fn generate_flags(raw: &BitFlags) -> proc_macro2::TokenStream {
     quote! {
         bitflags::bitflags! {
             #( #docs )*
+            #[derive(Default)]
             pub struct #name: #typ {
                 #( #variants )*
             }
--- a/read-fonts/generated/generated_gpos.rs
+++ b/read-fonts/generated/generated_gpos.rs
@@ -215,6 +215,7 @@ impl<'a> SomeTable<'a> for PositionLookup<'a> {
 
 bitflags::bitflags! {
     /// See [ValueRecord]
+    #[derive(Default)]
     pub struct ValueFormat: u16 {
         /// Includes horizontal adjustment for placement
         const X_PLACEMENT = 0x0001;
--- a/read-fonts/generated/generated_stat.rs
+++ b/read-fonts/generated/generated_stat.rs
@@ -891,6 +891,7 @@ impl<'a> SomeRecord<'a> for AxisValueRecord {
 
 bitflags::bitflags! {
     /// [Axis value table flags](https://docs.microsoft.com/en-us/typography/opentype/spec/stat#flags).
+    #[derive(Default)]
     pub struct AxisValueTableFlags: u16 {
         /// If set, this axis value table provides axis value information
         /// that is applicable to other fonts within the same font family.


read-fonts/generated/generated_glyf.rs
```diff
@@ -212,7 +212,7 @@ impl<'a> std::fmt::Debug for SimpleGlyph<'a> {
...
...ORT_VECTOR = 0x10 ; # [doc = " Bit 5: This flag has two meanings, depending on how the"] # [doc = " Y_SHORT_VECTOR flag is set. If Y_SHORT_VECTOR is set, this bit"] # [doc = " describes the sign of the value, with 1 equalling positive and"] # [doc = " 0 negative. If Y_SHORT_VECTOR is not set and this bit is set,"] # [doc = " then the current y-coordinate is the same as the previous"] # [doc = " y-coordinate. If Y_SHORT_VECTOR is not set and this bit is also"] # [doc = " not set, the current y-coordinate is a signed 16-bit delta"] # [doc = " vector."] const Y_IS_SAME_OR_POSITIVE_Y_SHORT_VECTOR = 0x20 ; # [doc = " Bit 6: If set, contours in the glyph description may overlap."] # [doc = " Use of this flag is not required in OpenType — that is, it is"] # [doc = " valid to have contours overlap without having this flag set. It"] # [doc = " may affect behaviors in some platforms, however. (See the"] # [doc = " discussion of “Overlapping contours” in Apple’s"] # [doc = " specification for details regarding behavior in Apple"] # [doc = " platforms.) When used, it must be set on the first flag byte"] # [doc = " for the glyph. See additional details below."] const OVERLAP_SIMPLE = 0x40 ; } }
```
cmyr, line 215: unfortunately the `bitflags!` macro does not use valid rust syntax, so `rustfmt` can't format it correctly. It would be nice to improve this, though, so maybe there's something I can do manually.

rsheeter: Is bitflags quasi-standard? - my immediate noob thought is to not use it, just generate whatever the code we want is directly. As an added bonus this removes the format nuisance (#117).

rsheeter:
> not use an actual rust enum to represent these enums, and make them instead by newtype structs with associated constants. This feels un-rusty, but also an awful lot simpler and closer to representing the underlying truth.

Simpler and closer to representing the underlying truth sound like things the Rust community would approve of.

My Rust noob first thought would be to enumerate individual bitflag values and have a type to represent an arbitrary set of them. You can query them individually using the enum.

read-fonts/generated/generated_glyf.rs
cmyr: bitflags is definitely the de facto standard way to declare a set of flags in Rust, but I do agree we could consider replacing it at some point with a manual implementation, and then we could get rid of #117.

cmyr:
> > not use an actual rust enum to represent these enums, and make them instead by newtype structs with associated constants. This feels un-rusty, but also an awful lot simpler and closer to representing the underlying truth.
> 
> Simpler and closer to representing the underlying truth sound like things the Rust community would approve of.
> 
> My Rust noob first thought would be to enumerate individual bitflag values and have a type to represent an arbitrary set of them. You can query them individually using the enum.

So it's worth remembering that the opentype spec contains both flags and enums; the distinction being that a flags field can contain multiple values, whereas an enum is discrete, and we handle these differently in fontations.

For *flags*, which we're dealing with here, we use the `bitflags` crate, and the thing that produces is a newtype around an integer, with constants declared for the various flag values, and implementations of the various bitwise operators, plus convenience methods for `contains`, `union`, etc.

For *enums*, currently we define a primitive rust enum, which is an enum where all of the variants are 'empty', and so it is just stored as a discriminant, and the discriminant can be explicit. The problem with this is that it is UB in rust to construct a primitive enum from a value that is not one of its discriminants, so if we encounter an invalid or unknown value in a file when reading one of these enums, we need to either fail the read, or else replace it with some other value; and in the cases of enums, there often isn't really an obvious choice for what the default should be.

rsheeter:
I meant that you could enumerate the individual valid values of a bitfield. You'd store it in a newtype of the appropriate int because the combination of several isn't a valid enum value. You can then query using the enum. That is, the enum merely replaces the consts for flag values. Is it better to have consts on a struct? - they feel like an enum to me.

dfrg:
> Another approach altogether (which had mentioned in relation to NameIds) would be to not use an actual rust enum to represent these enums, and make them instead by newtype structs with associated constants.

I was only considering this for names because they are effectively an open set. I think we should continue to use enums where they make sense.

dfrg:
> I meant that you could enumerate the individual valid values of a bitfield. You'd store it in a newtype of the appropriate int because the combination of several isn't a valid enum value. You can then query using the enum. That is, the enum merely replaces the consts for flag values. Is it better to have consts on a struct? - they feel like an enum to me.

In this case, you'd end up with different type names for the flag enum and the container (Value and ValueFlags?) which would be confusing. The bitflags crate is so pervasive in the Rust community that I think we'd need a really good reason to not use it.

PR #117, opened 2022-11-04
[codegen] Manually reformat generated bitflags! decls

cmyr:
Because the contents of this macro is not value Rust code, rustfmt doesn't know what to do with it, which means we currently end up with monolithic single line blobs in our generated files for any declared bitflags.

This patch adds an additional post-processing step that finds any such declarations and manually reparses and rewrites them.

Maybe this is not a good idea? This was definitely me scratching an itch at the end of the day.

### DIFF
--- a/font-codegen/src/formatting.rs
+++ b/font-codegen/src/formatting.rs
@@ -1,0 +1,225 @@
+//! improve readability of generated code
+
+use std::borrow::Cow;
+use std::fmt::Write;
+
+use regex::Captures;
+
+/// reformats the generated code to improve readability.
+pub(crate) fn format(tables: proc_macro2::TokenStream) -> Result<String, syn::Error> {
+    // if this is not valid code just pass it through directly, and then we
+    // can see the compiler errors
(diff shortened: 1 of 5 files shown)


rsheeter:
Is there a way we could emit valid rust code that would format correctly instead?

rsheeter (approved):

cmyr:
Not for as long as we're using the `bitflags!` macro. The rule with rustfmt is that it will format the contents of a macro, but only if those contents are valid rust code. The syntax of bitflags looks like,

```rust
bitflags! {
    struct Flags: u32 {
        const A = 0b00000001;
        const B = 0b00000010;
        const C = 0b00000100;
        const ABC = Self::A.bits | Self::B.bits | Self::C.bits;
    }
}
```

and this is not valid rust code; the `struct $Ident: $Type` at the top is not a valid struct declaration, you cannot declare consts in a struct body, etcetera.

The only real way to address this would be to somehow change the syntax of the macro so that the contents were valid rust.

PR #124, opened 2022-11-05
Bug fixes and improvements to glyf outline parsing

dfrg:
Fixes some bugs in outline parsing and also adds a new `read_points` method (in addition to the iterator) in `SimpleGlyph` that is much faster when the user can provide preallocated buffers.

Ok to merge.


rsheeter (approved):
LGTM but ... I wonder if we should delete the iterator as it appears to be a footgun.

dfrg:
kept the iterator but renamed `read_points` -> `read_points_fast` and added notes of perf to the doc comments

Issue #82, opened 2022-10-23
When codegen fails a backtrace is nice


rsheeter:
I found it very helpful to have `eprintln!("{}", Backtrace::capture());` before places that return syn::Error's when modifying the codegen. Without it the lovely error told me where in the input the problem was but not so much where in the codegen it might be.

I wonder if it's worth having that as a standard practice?

rsheeter:
Maybe in a debug log?

rsheeter:
https://doc.rust-lang.org/std/backtrace/index.html is experimental; blocked on https://github.com/rust-lang/rust/issues/53487

rsheeter:
Backtrace is no longer experimental in 1.65

cmyr:
Also I realized something about this tension (between showing backtraces versus showing the inline errors) and perhaps i commented on this somewhere earlier? But basically the backtraces are useful if you're working on the actual codegen tool itself, but the inline errors are much more useful if you're working on some other table, and just *using* the tool.

rsheeter:
I think if I add backtraces in a debug log both users are happy?

PR #128, opened 2022-11-07
Capture backtraces when codegen goes awry

rsheeter:
Feel free to merge if acceptable. Helps with #82, though spots where we propagate an error with ? still get missed. Baby steps.

### DIFF
--- a/font-codegen/src/lib.rs
+++ b/font-codegen/src/lib.rs
@@ -1,5 +1,7 @@
 //! Generating types from the opentype spec
 
+use std::backtrace::Backtrace;
+
 use log::debug;
 use proc_macro2::TokenStream;
 use quote::quote;
@@ -30,7 +32,11 @@ pub fn generate_code(code_str: &str, mode: Mode) -> Result<String, syn::Error> {
     // Generation is done in phases (https://github.com/googlefonts/fontations/issues/71):
     // 1. Parse
     debug!("Parse (mode {:?})", mode);
-    let mut items: Items = syn::parse_str(code_str)?;
+    // This is the one step where we can't readily intercept the error with logged_syn_error
+    let mut items: Items = syn::parse_str(code_str).map_err(|e| {
+        debug!("{}", Backtrace::capture());
+        e
+    })?;
     items.sanity_check(Phase::Parse)?;
 
     // 2. Contemplate (semantic analysis)
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -1,6 +1,6 @@
 //! raw parsing code
 
-use std::{collections::HashMap, ops::Deref, str::FromStr};
+use std::{backtrace::Backtrace, collections::HashMap, fmt::Display, ops::Deref, str::FromStr};
 
 use log::{debug, trace};
 use proc_macro2::{Span, TokenStream};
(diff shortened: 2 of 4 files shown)


font-codegen/src/parsing.rs
```diff
@@ -1398,10 +1401,15 @@ fn get_single_lifetime(input: &syn::PathArguments) -> syn::Result<Option<syn::Li
     match get_single_generic_arg(input)? {
         None => Ok(None),
         Some(syn::GenericArgument::Lifetime(arg)) => Ok(Some(arg.clone())),
-        _ => Err(syn::Error::new(input.span(), "expected single lifetime")),
+        _ => Err(logged_syn_error(input.span(), "expected single lifetime")),
     }
 }
 
+pub(crate) fn logged_syn_error<T: Display>(span: Span, message: T) -> syn::Error {
```
cmyr, line 1408: So this is an improvement, but as we're only doing this with errors that we create explicitly (as opposed to errors that occur during another parse method, for instance when we do `let name = input.parse::<syn::Ident>()?`).

And unfortunately, errors of the other type (generated in syn, and merely propagated by us) are probably the majority of our errors?

font-codegen/src/lib.rs
```diff
@@ -30,7 +32,11 @@ pub fn generate_code(code_str: &str, mode: Mode) -> Result<String, syn::Error> {
     // Generation is done in phases (https://github.com/googlefonts/fontations/issues/71):
     // 1. Parse
     debug!("Parse (mode {:?})", mode);
-    let mut items: Items = syn::parse_str(code_str)?;
+    // This is the one step where we can't readily intercept the error with logged_syn_error
+    let mut items: Items = syn::parse_str(code_str).map_err(|e| {
+        debug!("{}", Backtrace::capture());
```
cmyr, line 37: Not sure if this is intentional or not, but this will only capture a bactrace if the `RUST_BACKTRACE` env var is set. If we want to *always* capture a backtrace, we can use [`Backtrace::force_capture`](https://doc.rust-lang.org/1.65.0/src/std/backtrace.rs.html#311).

cmyr (approved):
Happy to merge this as an improvement, but I think longer term we might want to move away from `syn::Error`, and create our own error type that generates a backtrace when it is created; this would have an impl of `From<syn::Error>`, and in that impl we would grab the backtrace, which would then work anywhere there is `?`.

font-codegen/src/lib.rs
```diff
@@ -30,7 +32,11 @@ pub fn generate_code(code_str: &str, mode: Mode) -> Result<String, syn::Error> {
...
+        e
+    })?;
```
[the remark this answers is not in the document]
rsheeter: If we fail parsing entirely, e.g. I go add DUCK to some_input.rs, then we hit this w/o hitting any of our parse stuff. if I deliberately break something, say change u16 to u16z somewhere, we get a trace to the right place in our code.

font-codegen/src/lib.rs
rsheeter: Not capturing when backtrace is off feels correct to me, do you think we should always capture?

rsheeter:
> create our own error type that generates a backtrace when it is created

That does sound much nicer.

font-codegen/src/parsing.rs
rsheeter: When I tried introducing some basic errors, e.g. a type that didn't exist, I did get a trace. In any case, IIUC using our own error that auto-captures traces should entirely resolve this.

PR #130, opened 2022-11-07
Add support for mapping codepoints to glyph ids

dfrg:
Adds basic methods for mapping codepoints to glyph ids for cmap subtable formats 4 and 12.

This should allow me to completely replace the current font parsing code in piet-gpu with read-fonts.


rsheeter (approved):
LGTM IFF tests are added. This seems very straight forward to test with the ttx sample mechanism.

rsheeter:
> This should allow me to completely replace the current font parsing code in piet-gpu with read-fonts.

That's exciting!

PR #131, opened 2022-11-08
Parallelize codegen execution

cmyr:
This uses rayon to execute multiple codegen tasks in parallel, which is a significant speedup (I'm seeing wall-clock time cut in more than 1/3) for multi-file codegen runs.

running `time cargo run --bin=codegen resources/codegen_plan.toml`:

```
before: 4.43s user 0.61s system 95% cpu 5.262 total
after:  6.68s user 0.73s system 462% cpu 1.603 total
```

### DIFF
--- a/font-codegen/Cargo.toml
+++ b/font-codegen/Cargo.toml
@@ -23,3 +23,4 @@ serde = {version = "1.0", features = ["derive"] }
 xflags = "0.2.4"
 log = "0.4"
 env_logger = "0.9.0"
+rayon = "1.5.3"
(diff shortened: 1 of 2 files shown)


rsheeter:
Would rayon also be recommended for ||ism in compiler?

rsheeter:
I really want to Just Merge!

rsheeter (approved):
LGTM

cmyr:
> Would rayon also be recommended for ||ism in compiler?

It would definitely be the first thing I would try!

PR #132, opened 2022-11-08
Constructors

cmyr:
This adds `fn new` constructors to most tables and records, and also adds per-format constructors to multi-format tables.

Getting this to work requires a bit more analysis during codegen, since the format groups need to know the arguments for the constructors of each format variant. My approach to this was to do a minor refactor, where we now store our items as an `IndexMap<Ident, Item>`. This map can then also be reused to generate the 'known idents' map that we used to resolve fields.

Overall I'm quite happy with this; it's probably just a few hundred lines of human code, but it goes and adds new functionality across all of our existing tables and records.


font-codegen/src/fields.rs
```diff
@@ -232,6 +245,16 @@ impl Fields {
     }
 }
 
+/// All the state required to generate a constructor for a table/record
+/// that includes this field.
+pub(crate) struct FieldConstructorInfo<'a> {
+    pub(crate) name: &'a syn::Ident,
```
rsheeter, line 251: We have other structures that contain syn::Ident and those do not have lifetimes. Can we remove the lifetime? - it's ok if it means we copy Ident's imho.

font-codegen/src/fields.rs
```diff
@@ -1216,7 +1257,20 @@ impl FieldType {
...
+    /// they live in the FieldAttrs that are not accessible from FieldType.
+    ///
+    /// The `for_constructor` argument is `true` only we want the type of an
+    /// argument to a constructor method. Specifically, this does not wrap
+    /// offset-types in offset markers (these conversions occur in the constructor).
+    fn compile_type(
+        &self,
+        nullable: bool,
+        version_dependent: bool,
+        for_constructor: bool,
```
rsheeter, line 1272: not loving the bool typing here, it's resulting in code like `self.typ.compile_type(self.is_nullable(), false, true))` where the is_nullable is clear but the literals have non-obvious meaning. Feels like a case where a typed flag (enum?) or builder would result in clearer code.

font-codegen/Cargo.toml
```diff
@@ -24,3 +24,4 @@ xflags = "0.2.4"
 log = "0.4"
 env_logger = "0.9.0"
 rayon = "1.5.3"
+indexmap = "1.9.1"
```
rsheeter, line 27: I suggest that adding a dependency is a sufficiently large change it should be explained in comments. For example, why can't we use the normal HashMap?

font-codegen/src/table.rs
```diff
@@ -738,3 +786,22 @@ impl TableReadArgs {
         })
     }
 }
+
+// An overwrought and likely incorrect way of converting 'Format1' to 'format_1' -_-
```
rsheeter, line 790: Very unit testable :) Would like to see more unit tests in our CLs in general.

This looks like a manual implementation of a regex, e.g. replace letter-number with letter-dash-number, which could make sense for perf in some situations but ... probably not this one?

rsheeter (approved):
LGTM subject to tidying up the bool,bool,bool fn, the calls to that are pretty hard to interpret at a glance

rsheeter:
> it's probably just a few hundred lines of human code, but it goes and adds new functionality across all of our existing tables and records.

# 🦸

font-codegen/Cargo.toml
cmyr: good point: the motivation here is that hashmap does not guarantee ordering, and we want to generate code for our items in the order that they are declared, otherwise any rerun of codegen risks shuffling everything.

font-codegen/src/fields.rs
cmyr: you are absolutely right, pure laziness on my part.

font-codegen/src/fields.rs
cmyr: In general I try to borrow where doing so is A) possible and B) not overly difficult to do; I'm not sure if we have other structures that meet these criteria? Happy to change, since this is unimportant to practice, but wanted to explain my reasoning.

PR #133, opened 2022-11-08
Re-export font-types in read-fonts

dfrg:
Simple change that re-exports font-types from read-fonts so that those types can be accessed from `read_fonts::types` without having to sync dependency versions in your own Cargo.toml.

### DIFF
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -37,6 +37,9 @@ pub use read::{ComputeSize, FontRead, FontReadWithArgs, ReadArgs, ReadError, Var
 pub use table_provider::TableProvider;
 pub use table_ref::TableRef;
 
+/// Public re-export of the font-types crate.
+pub use font_types as types;
+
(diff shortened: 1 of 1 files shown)


read-fonts/src/lib.rs
```diff
@@ -37,6 +37,9 @@ pub use read::{ComputeSize, FontRead, FontReadWithArgs, ReadArgs, ReadError, Var
 pub use table_provider::TableProvider;
 pub use table_ref::TableRef;
 
+/// Public re-export of the font-types crate.
+pub use font_types as types;
```
rsheeter, line 41: Purely out of curiosity, why are we changing the name (as types)?

rsheeter (approved):
LGTM

read-fonts/src/lib.rs
dfrg: `read_fonts::font_types` is a mouthful and a bit redundant. It's fairly common to chop off the common parts when you're re-exporting within a family of crates.

PR #137, opened 2022-11-10
Enhancements to the Fixed type

dfrg:
I've run into some friction when using the fixed point types for computing things like deltas and variation region scalars. This adds `Mul`, `Div` and `Neg` impls for `Fixed` along with exposed associated constants for 0 and 1. Also adds a `to_fixed` conversion for `F2dot14`.

Ok to merge if there are no issues.

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -124,7 +124,7 @@ macro_rules! float_conv {
                 //what matters is that we are rounding *away from zero*.
                 #[cfg(all(not(feature = "std"), not(test)))]
                 Self(
-                    (x * Self::ONE as $ty + (0.5 * (-1.0 * x.is_sign_negative() as u8 as $ty)))
+                    (x * Self::ONE.0 as $ty + (0.5 * (-1.0 * x.is_sign_negative() as u8 as $ty)))
                         as _,
                 )
             }
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -18,9 +18,14 @@ macro_rules! fixed_impl {
             /// This type's smallest representable value
             pub const EPSILON: Self = Self(1);
 
+            /// Representation of 0.0.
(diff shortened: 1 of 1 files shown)


font-types/src/fixed.rs
```diff
@@ -1,6 +1,6 @@
 //! fixed-point numerical types
 
-use std::ops::{Add, AddAssign, Sub, SubAssign};
+use core::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
```
cmyr, line 3: I have `core` aliased as `std` if we're `no_std`. I'm not sure where I got this pattern from, but I don't think I invented it. In any case, I think we use `std` here and if we want to change the overall pattern we can do that in a separate PR?

cmyr (approved):
Looks good. These are definitely still rough around the edges; in particular I don't like the `from_` methods (`from_f64` here, and `from_f32` on `F2Dot14`) and think they should both maybe just have `fn new` constructors? But that has no impact here.


This is also the kind of code that I would generally write unit tests for, but then it's also some of the code I am least confident writing so ymmv.

font-types/src/fixed.rs
dfrg: oh! I actually thought I messed this up by doing automatic imports so this was my attempt to fix it. I have no objections to that pattern so I'll revert this change.

dfrg:
I added a couple unit tests for mul/div. I also find the constructors a bit less than satisfying and I honestly would like access to the inner values without moving through from/to_be_bytes(), but I'm trying to resist and see how far I can get without them. We'll revisit as things progress.

cmyr:
If you want access to the inner values I think a `to_bits` and `from_bits` method would be fine? These exist on `f32` and `f64`, so there's precedent.

cmyr:
But I guess I'd also be curious where you need this, it does feel like an abstraction leak?

PR #138, opened 2022-11-10
Add read support for metrics variations (ItemVariationStore, et al)

dfrg:
Adds the foundations for reading metrics variations. This isn't quite ready yet. It needs tests, additional tables (VVAR, MVAR) and updates to GDEF and COLR.

Just submitting the draft PR now to get some feedback and make sure I haven't done anything too crazy.

In initial testing, HVAR is returning identical results to FreeType for at least one font.


read-fonts/src/lib.rs
```diff
@@ -21,6 +21,7 @@ mod table_ref;
 pub mod tables;
 #[cfg(feature = "traversal")]
 pub mod traversal;
+pub mod variations;
```
cmyr, line 24: so I've been wondering if actually *all* tables (including layout and variations) should live in the *tables* module? It has a certain clarity, e.g. It makes clear that everything that is *not* in the tables module is sort of core implementation stuff.

This is not really related to this PR, but this feels like a good time to discuss it.

resources/codegen_inputs/mvar.rs
```diff
@@ -0,0 +1,31 @@
...
+/// The [MVAR (Metrics Variations)](https://docs.microsoft.com/en-us/typography/opentype/spec/mvar) table
+table Mvar {
+    /// Major version number of the horizontal metrics variations table — set to 1.
+    /// Minor version number of the horizontal metrics variations table — set to 0.
+    version: MajorMinor,
+    /// Not used; set to 0.
+    #[skip_getter]
+    _reserved: u16,
+    /// The size in bytes of each value record — must be greater than zero.
+    value_record_size: u16,
```
cmyr, line 12: so we probably want to support this, although I'm okay if we do it later. I have a pattern that handles this that I'm using in STAT, and it should be portable to here.

resources/codegen_inputs/variations.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// DeltaSetIndexMap format: set to 0.
+    #[format = 0]
+    format: u8,
+    /// A packed field that describes the compressed representation of 
+    /// delta-set indices. See details below.
+    entry_format: u8,
+    /// The number of mapping entries.
+    map_count: u16,
+    /// The delta-set index mapping data. See details below.
+    #[count(((($entry_format & 0x30) >> 4) + 1) as usize * $map_count as usize)]
```
cmyr, line 14: I won't tell rod if you don't tell rod?

more seriously I might prefer if we implement a method on the type that calculates the count, and here we just call that method?

cmyr (commented):
just did a quick skim, looks reasonable to me!

read-fonts/src/lib.rs
dfrg: I was wondering the same thing. I think all should live in tables and the layout/variations modules should be moved there as well. I can't think of any reason to special case them and they'll stand out anyway due to the longer names.

resources/codegen_inputs/mvar.rs
dfrg: I'm really leery of adding complexity for this kind of thing and I suspect (like fvar's VariationAxisRecord) that this size is forever fixed because implementations have hard coded it. Happy to revisit in the future.

resources/codegen_inputs/variations.rs
dfrg: Will do!

read-fonts/src/lib.rs
rsheeter: +1 to all in tables

read-fonts/src/tables/hvar.rs
```diff
@@ -0,0 +1,98 @@
...
+        let ix = match self.advance_width_mapping() {
+            Some(Ok(dsim)) => dsim.get(gid as u32)?,
+            _ => DeltaSetIndex {
+                outer: 0,
+                inner: gid,
+            },
+        };
+        let ivs = self.item_variation_store()?;
+        ivs.compute_delta(ix, coords)
+    }
```
rsheeter, line 29: Hvar::advance_width_delta and Vvar::advance_height_delta differ *only* in what ix is set to and which varstore is used. I feel a common function in your future.

read-fonts/src/tables/hvar.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub fn rsb_delta(&self, glyph_id: GlyphId, coords: &[F2Dot14]) -> Result<Fixed, ReadError> {
+        let gid = glyph_id.to_u16();
+        let ix = match self.rsb_mapping() {
+            Some(Ok(dsim)) => dsim.get(gid as u32)?,
+            _ => return Err(ReadError::NullOffset),
+        };
+        let ivs = self.item_variation_store()?;
+        ivs.compute_delta(ix, coords)
+    }
+}
```
rsheeter, line 54: Hvar lsb_delta, rsb_delta and Vvar txb_delta, bsb_delta appear to be almost identical and in need of a common helper.

They are also very close to advance_(width|height)_delta, differing only in what happens if the mapping is None. Looks like a common function whose parameters include a fn to handle the mapping None case might go quite far?

read-fonts/src/variations.rs
```diff
@@ -0,0 +1,200 @@
...
+            None => return Ok(Fixed::default()),
+        };
+        let regions = self.variation_region_list()?.variation_regions();
+        let region_indices = data.region_indexes();
+        let mut delta = Fixed::ZERO;
+        for (i, region_delta) in data.delta_set(index.inner).enumerate() {
+            let region_index = region_indices
+                .get(i)
+                .ok_or(ReadError::MalformedData(
+                    "invalid delta sets in ItemVariationStore",
```
rsheeter, line 70: It would be nice to include specifics such as the index and the value of i. In Python or Java we'd stick it in the message but perhaps in Rust it would be more normal to have actual fields on MalformedData?

read-fonts/src/tables/hvar.rs
dfrg: Good catch. I'll pull it out. Same with the side bearing/vorg code.

read-fonts/src/variations.rs
rsheeter: Same comment for "invalid entry size in DeltaSetIndexMap" above

dfrg: We can add another variant to ReadError (I added one for missing metric tags in MVAR), but I'm not sure how far we want to go down that path

read-fonts/src/tables/hvar.rs
dfrg: Will do!

resources/test_fonts/ttx/master_iup.ttx
```diff
@@ -0,0 +1,551 @@
...
+
+    <TTGlyph name="q" xMin="38" yMin="-152" xMax="418" yMax="468">
+      <component glyphName="b" x="494" y="456" scale="-0.99994" flags="0x4"/>
+    </TTGlyph>
+
+    <TTGlyph name="space"/><!-- contains no outline data -->
+
+  </glyf>
+
+  <name>
```
rsheeter, line 245: Suggest trimming out parts of the ttx that don't matter. For example, we can probably delete most or all of the `<name>` content.

resources/test_fonts/ttx/master_iup.ttx
```diff
@@ -0,0 +1,551 @@
...
+    </namerecord>
+    <namerecord nameID="257" platformID="3" platEncID="1" langID="0x409">
+      Ascender
+    </namerecord>
+    <namerecord nameID="258" platformID="3" platEncID="1" langID="0x409">
+      Regular
+    </namerecord>
+  </name>
+
+  <post>
```
rsheeter, line 296: If you don't specifically care, `<post>` can likely be deleted

resources/test_fonts/ttx/master_iup.ttx
dfrg: I'll trim. I'd like to keep `<glyf>` and `<gvar>` to use for later testing.

resources/test_fonts/ttx/master_iup.ttx
```diff
@@ -0,0 +1,551 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rsheeter, line 1: IMHO this ttx is complex enough a comment, either in the xml or in something like a sibling master_iup.md, to explain what you are setting up and why it makes for an interesting test would be helpful.

rsheeter (approved):
LGTM subject to:

1. Reduction of duplication (left comments inline)
1. Tests

resources/codegen_inputs/variations.rs
rsheeter: I don't mind the math - would be ok with some sort of "thou shalt implement a fn" too - but the as usize does bug me.

read-fonts/src/variations.rs
rsheeter: A MalformedDataAtIndex error with a field for the offending index seems like it would go quite far. Make the index an enum with variants for at least naked u32 and DeltaSetIndex and you've covered a *lot* of cases.

read-fonts/src/tables/hvar.rs
dfrg: Done

resources/test_fonts/ttx/master_iup.ttx
dfrg: Just removed this ttx and reused the vazirmatn_var_trimmed that already exists since it contains an HVAR table.

dfrg:
Dedup'd common code and reused the existing variable font (vazirmatn) for tests. Also moved the variations module to tables. We can do the same for layout in another PR.

PR #141, opened 2022-11-12
Add read support for avar table

dfrg:
This adds support for the `avar` table with a method to apply the piecewise linear mapping to a normalized coordinate.

One thing of note is that this adds a `mul_div` method to `Fixed`. FreeType's `FT_MulDiv` produces very slightly different results from `FT_MulFix` followed by `FT_DivFix` and this replicates that so we produce identical results.

This also makes use of `VarLenArray` and the traverse method for that field is stubbed out in similar fashion to the one in `post`. Can we solve this with a custom impl of `SomeArray`?


resources/codegen_inputs/avar.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// The number of variation axes for this font. This must be the same number as axisCount in the 'fvar' table.
+    axis_count: u16,
+    /// The segment maps array — one segment map for each axis, in the order of axes specified in the 'fvar' table.
+    #[count(..)]
+    #[traverse_with(traverse_segment_maps)]
+    axis_segment_maps: VarLenArray<SegmentMaps<'a>>,
+}
+
+/// [SegmentMaps](https://learn.microsoft.com/en-us/typography/opentype/spec/avar#table-formats) record
+table SegmentMaps {
```
cmyr, line 20: this is supposed to be a record, not a table (and thank the lord, this was going to break a bunch of things)

dfrg: yeah, if I make it a record, it crashes codegen with 'array tokens never cooked'. Table just worked so I ran with it

dfrg: I get the complication in that tables imply some form of offset resolution, but it seems like anything inside a VarLenArray will require special casing regardless?

cmyr: it's just a really funny case. The basic issue is that we don't bother implementing `FontRead` for records, because in general they always exist in arrays, and we transmute, but `VarLenArray` uses `FontRead` under the covers.

The reason I ran into this is because it should be easy for this to work with traversal, *assuming* that the inner types are records. It doesn't work if they're tables, at least right now, because we always expect tables to be behind offsets.

As I write this, I think the best course of action right now would just be a manual impl of `FontRead` for this particular case; then at least we get traversal working.

The reason traversal is particularly tricky in `post` is because we don't really have any generalized concept of a 'string', so it isn't obvious how to even represent that as a FieldType.

I'm going to add a commit here if that's okay that hopefully gets this working, with traversal?

dfrg: By all means. Hack away!

cmyr (approved):
Okay I think this is all good to go now.

dfrg:
These changes seem beneficial to codegen across the board. Thanks for the assist!

PR #142, opened 2022-11-14
Flatten layout module into tables module

cmyr:
All the different layout tables are now direct submodules of `tables`, as is the layout module itself.


dfrg (approved):
Lots of changes but seems like just module structure stuff. This looks good to me.

cmyr:
yes this has no functional change, just a bunch of path changes in generated code explode the diff.

PR #143, opened 2022-11-14
Add LookupFlag type

cmyr:
This can be shared between the two crates. Was blocked on 'extern scalar', which now exists.

Requires custom traversal logic, because traversal has know way of knowing about this type, and I'd like to avoid special-casing it in that code.

This will conflict with #142, and I don't have strong opinions about which should go in first.


dfrg (approved):
This also looks good. I assume we don't want to use `bitflags` here because of the mark attachment type mask?

PR #144, opened 2022-11-14
Make DEFAULT_GLYPH_NAMES array public

cmyr:
There's an argument that maybe this should live somewhere else? But for now I just want to use it from fea-rs.


dfrg (approved):
This is part of the specification for the `post` table so it’s entirely reasonable for it to live here.

PR #145, opened 2022-11-15
[codegen] Remove :Default bound on constructor generics

cmyr:
This is a minor fixup; this bound is overly constrictive and not relevant in any of the code that we generate so far. We can always get more constrictive later, as needed.


dfrg (approved):
Seems reasonable!

PR #146, opened 2022-11-15
Fix logic around the Mac Roman encoding

cmyr:
This hadn't been used, and the initial impl was a total brainfart.


dfrg (approved):
Lgtm

read-fonts/src/tables/name.rs
```diff
@@ -238,3 +238,54 @@ static MAC_ROMAN: [u16; 128] = [
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn mac_roman() {
+        static INPUT: &str = "Joachim Müller-Lancé";
+        for c in INPUT.chars() {
+            let enc = MacRomanMapping.encode(c).unwrap();
+            assert_eq!(MacRomanMapping.decode(enc), c);
```
rsheeter, line 288: To confirm, this test failed before and now passes?

cmyr: correct.

PR #147, opened 2022-11-15
Add a GPOS compilation test

cmyr:
This was motivated by a compilation error in fea-rs. It looks like we are not the culprit.


dfrg (approved):
+1. Nice to see the new `let else` statement getting some use.

PR #150, opened 2022-11-22
Helpers for compiling shared layout tables

cmyr:
We use a newtype pattern so we can differentiate between the GPOS and GSUB uses of certain common tables, but that is challange during compilation, where we don't really care about this distinction until the very end. This patch lets us work with the common tables while we build them up, and then cast them into the appropriate concrete type when we're done.


dfrg (approved):
Looks good. It’s likely we’ll want helpers for this sort of thing on the read side as well when dealing with lookups. I’m not sure exactly what they should look like yet though.

PR #151, opened 2022-11-22
Reexport read-fonts and font-types from write-fonts

cmyr:
Aliased as `read` and `types`.

@dfrg this takes a slightly different approach from #133; it uses `pub extern crate _ as _;`, instead of `pub use _ as _;`. The main difference with this approach as far as I can tell is that it makes the aliases publicly visible without any qualifiers, e.g. you can do `let x: types::Fixed = ()` whereas with `pub use _` you can only do `let x: crate::types::Fixed: ..`? Is there any other reason to prefer one pattern over the other?

This can be merged if approved.


dfrg (approved):
+1. Should we make a similar change in read-fonts?

cmyr:
yea, I'm happy to make that change as a followup?

PR #153, opened 2022-11-24
A few more compile-motivated tweaks

cmyr:
This is a rollup of three small commits fixing issues I encountered while hacking on feature compilation.

### DIFF
--- a/read-fonts/src/font_data.rs
+++ b/read-fonts/src/font_data.rs
@@ -258,3 +258,11 @@ impl AsRef<[u8]> for FontData<'_> {
         self.bytes
     }
 }
+
+//kind of ugly, but makes FontData work with FontBuilder. If FontBuilder stops using
+//Cow in its API, we can probably get rid of this?
+impl<'a> From<FontData<'a>> for std::borrow::Cow<'a, [u8]> {
+    fn from(src: FontData<'a>) -> Self {
+        src.bytes.into()
+    }
+}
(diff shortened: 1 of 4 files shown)


read-fonts/src/font_data.rs
```diff
@@ -258,3 +258,12 @@ impl AsRef<[u8]> for FontData<'_> {
         self.bytes
     }
 }
+
+//kind of ugly, but makes FontData work with FontBuilder. If FontBuilder stops using
+//Cow in its API, we can probably get rid of this?
```
rsheeter, line 263: We should perhaps create, and cite, an issue to not use Cow in the FontBuilder API?

write-fonts/src/font_builder.rs
```diff
@@ -21,6 +21,11 @@ impl<'a> FontBuilder<'a> {
         self
     }
 
+    /// Returns `true` if the builder contains a table with this tag.
```
rsheeter, line 24: Comment adds nothing to type signature, suggest removal.

rsheeter (approved):
LGTM

read-fonts/src/font_data.rs
cmyr: #155

write-fonts/src/font_builder.rs
cmyr: I think in general under-documentation is a far greater problem than over-documentation, and requiring the documentation of all public items is something that we can (and I think we should) do in CI. 🤷

PR #154, opened 2022-11-25
Compile Device tables, including in ValueRecord

cmyr:
This is a bit annoying because ValueRecord is hand-rolled, and so I had to rework a bunch of it to make it actually contain device tables.

My understanding is that this isn't really used much, but it's in the `feaLib` tests so that's enough for me 😅


rsheeter (approved):
Seems reasonable to me

PR #156, opened 2022-11-28
Bump versions to 0.0.3

cmyr:
This will let me reference a stable version in fea-rs.

Also I apparently haven't been tagging releases, so will  start doing that.


rsheeter (approved):
LGTM. I really want to just merge :D

PR #159, opened 2022-11-29
Tag tweaks

cmyr:
This improves the validation logic around tags to fix a few issues:

- tags cannot begin with spaces, but can contain trailing spaces
- validation logic was different between new and new_checked

In addition it adds an AsRef<[u8]> impl and a to_bytes method, for accessing the raw bytes.

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -25,46 +25,44 @@ impl Tag {
     ///
     /// This method panics if the tag is not valid per the requirements above.
     pub const fn new(src: &[u8]) -> Tag {
-        assert!(
-            !src.is_empty() && src.len() < 5,
-            "input must be 1-4 bytes in length"
-        );
-        let mut raw = [b' '; 4];
-        let mut i = 0;
-        while i < src.len() {
-            if src[i] <= 0x20 || src[i] > 0x7e {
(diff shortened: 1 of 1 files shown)


font-types/src/tag.rs
```diff
@@ -83,10 +81,15 @@ impl Tag {
     pub const fn from_be_bytes(bytes: [u8; 4]) -> Self {
         Self(bytes)
     }
+
+    /// Return the raw byte array representing this tag.
+    pub fn to_bytes(self) -> [u8; 4] {
```
rsheeter, line 86: Maybe this isn't a convention in Rust but to_bytes sounds to me like it does a conversion whereas this feels more like as_bytes

rsheeter (approved):

font-types/src/tag.rs
cmyr: There is a handy chart for deciding when to use `to_`, `as_`, and `into_`: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

..which is slightly confusing, because it feels like we should use `into`, but I don't generally see `into` on Copy types? But I'll go with that anyway.

`as_bytes` would make sense if the signature was `(&self) -> &[u8]`

PR #161, opened 2022-12-02
Fix include_bytes! breakage

cmyr:
Instead of directly including these files, we now have a build script
that will copy the files into a build directory *if they exist*, and
will write an empty file in that location if they do not.

This ensures that include_bytes! always points to a file that exists,
which means that we will compile as a package.


rsheeter (approved):
Feels bad man. LGTM since working beats not working.

PR #162, opened 2022-12-04
STAT compilation improvements

cmyr:
- fixed a bug where we were calculating design_axis_count incorrectly
- add a custom constructor, to ensure elided_fallback_name_id is always present
- add a basic test case


dfrg (approved):
LGTM

PR #164, opened 2022-12-06
TopLevelTable trait + codegen attribute

cmyr:
This adds the `#[tag = "sfnt"]` attribute in codegen, and uses that to implement the `TopLevelTable` trait, which just associates a const with an item.

The big annoyance here is with hmtx/vmtx and hhea/vhea; because these share an implementation, we need to create new types to hang this trait off of, and we can't use the attribute.

I'm personally coming back to the opinion that our lives would be simpler if we just let there be duplicate codegen for these types; i'm not convinced that we will actually ever want to share code between them, and it would reduce the amount of boilerplate we need to implement on our own.

There are a few things missing from this:
- we still have the TAG constants in each table module
- I haven't dealt with hmtx/vmtx etc in `write-fonts`, since I want to see what the consensus is on this before hand.

### DIFF
--- a/font-codegen/Cargo.toml
+++ b/font-codegen/Cargo.toml
@@ -12,6 +12,7 @@ name = "codegen"
 path = "src/main.rs"
 
 [dependencies]
+font-types = { version = "0.0.3", path = "../font-types" }
 rustfmt-wrapper = "0.1"
 regex = "1.5"
 miette = { version =  "4.6", features = ["fancy"] }
--- a/read-fonts/src/table_provider.rs
+++ b/read-fonts/src/table_provider.rs
@@ -4,6 +4,14 @@ use types::Tag;
 
 use crate::{tables, FontData, FontRead, FontReadWithArgs, ReadError};
 
+/// A table that has an associated tag.
+///
+/// This is true of top-level tables, but not their various subtables.
+pub trait TopLevelTable {
+    /// The table's tag.
+    const TAG: Tag;
+}
+
 /// An interface for accessing tables from a font (or font-like object)
 pub trait TableProvider<'a> {
     fn data_for_tag(&self, tag: Tag) -> Option<FontData<'a>>;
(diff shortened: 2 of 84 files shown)


font-codegen/Cargo.toml
```diff
@@ -12,6 +12,7 @@ name = "codegen"
 path = "src/main.rs"
 
 [dependencies]
+font-types = { version = "0.0.3", path = "../font-types" }
```
rsheeter, line 15: QQ, why do we need a version on a path dependency?

read-fonts/src/table_provider.rs
```diff
@@ -12,133 +20,115 @@ pub trait TableProvider<'a> {
...
+
+    fn expect_table_args<T: TopLevelTable + FontReadWithArgs<'a>>(
+        &self,
+        args: &T::Args,
+    ) -> Result<T, ReadError> {
+        self.expect_data_for_tag(T::TAG)
+            .and_then(|data| T::read_with_args(data, args))
+    }
+
     fn head(&self) -> Result<tables::head::Head<'a>, ReadError> {
```
rsheeter, line 35: At this point do we even need per-table functions? - to me it would feel idiomatic to have something like `fn table` that could return whichever one you want so you can write things like:

```rust
let head: Head = font.table()?
```

rsheeter: You'd have to do a little bit of gymnastics to figure out read args but ... I think it might be achievable?

rsheeter (approved):
LGTM

font-codegen/Cargo.toml
cmyr: all dependencies must be on crates.io, and so must have versions. Including the path dependency is just useful during development, so you can make changes across crates and you'll always be using whatever you have checked out.

read-fonts/src/table_provider.rs
cmyr: I personally think it is a nice convenience to just have a simple method that returns a table, and I definitely think it is good to hide away the `FontReadWithArgs` business.

PR #165, opened 2022-12-06
Add builder for post 2.0

cmyr:
This lets us write out a valid post table, which is useful for fea-rs.

### DIFF
--- a/write-fonts/src/tables/post.rs
+++ b/write-fonts/src/tables/post.rs
@@ -1,5 +1,7 @@
 //! The post table
 
+use std::collections::HashMap;
+
 include!("../../generated/generated_post.rs");
 
 //TODO: I imagine we're going to need a builder for this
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/post.rs
```diff
@@ -8,6 +10,38 @@ include!("../../generated/generated_post.rs");
...
+            .map(|(i, name)| (*name, i as u16))
+            .collect::<HashMap<_, _>>();
+        let mut name_index = Vec::new();
+        let mut storage = Vec::new();
+
+        for name in order {
+            match known_glyphs.get(name) {
+                Some(i) => name_index.push(*i),
+                None => {
+                    let idx = (known_glyphs.len() + storage.len()).try_into().unwrap();
```
dfrg, line 28: Not particularly germane to this until we get `> 16 bit` glyph ids, but I do think we should consider propagating errors rather than panicking wherever possible so that these can be reported in a friendly manner to the user. Nobody likes internal compiler errors :)

dfrg (approved):
One small comment that is not blocking, but more food for thought. Otherwise, looks good!

write-fonts/src/tables/post.rs
cmyr: Good observation... the pattern I've taken here is that there are separate validation and compilation passes; the validation pass ensures invariants are all upheld (and reports issues in aggregate) and then the compilation pass barrels through, assuming that the data is good.

I'm not totally sure this is the correct approach (for one thing, it encourages writing '//TODO: validate this' everywhere, but if we want to change that it will involve a general rearchitecting.

PR #167, opened 2022-12-09
Split hv tables

### LINKED ISSUE #166: rethink vhea/hhea & vmtx/hmtx
I have run into a few issues with using a common type to represent two distinct tables here.

First off, using a common type doesn't work with the `TopLevelTable` trait (added in #164). Since a single type represents multiple top-level tables, there is no way to implement this trait correctly.

The second issue is that `hhea` is always version 1.0, and `vhea` is at version 1.1. This means that we need to *somehow* track which table we're in when we compile, since we need to write out different versions for the different tables.

Currently, I can see two ways to address this issue: manually, or via codegen. 

Addressing it manually means manually adding different wrapper structs that contain a common inner table that is codegen'd. This is very possible, but it will involve a lot of boilerplate; basically we will need to manually implement each of the traits we expect to be present. This isn't too bad, but it is annoying, and it also means that if we make any future changes to our traits we will need to manually maintain this set of implementations. On the plus side, we can probably get away with using a macro to handle this implementation, so at least we only write it once, and not twice; I've taken that approach in #164, where [it looks like this](https://github.com/googlefonts/fontations/pull/164/files#diff-e548efd21754e3d0fb83797c32ce32ca03ee50a8359c9c193485798590be6fb4).

Addressing it via codegen means having each of these tables get its own codegen file, which will generate a distinct type. The downside of this is that it involves duplication. The upside of this is that we can delete a bunch of hand-written code, and avoid the future maintenance costs that this entails, and it also means that these tables will now 'look and feel' more consistently like all the other top-level tables in the crate.

We've already had this discussion (in #104) but in light of actually working with the existing code, my feelings have changed, and I think that just having each of these tables be generated separately is probably the better compromise.

Also: One of the arguments for having shared implementations was that it would be possible to reuse these tables in both vertical & horizontal contexts. it's really not clear to me if this is a practical or theoretical concern, but if it is a practical concern I think that we could address it with a trait, when it arises?

Does anyone else have any strong feelings? cc @anthrotype

cmyr:
This moves each of hhea/vhea and hmtx/vmtx into its own codegen file.

Closes #166.


read-fonts/src/tables.rs
```diff
@@ -13,8 +13,7 @@ pub mod head;
 pub mod hhea;
 pub mod hmtx;
 pub mod hvar;
-pub mod hvhea;
-pub mod hvmtx;
+//pub mod hvmtx;
```
anthrotype, line 16: leftover?

dfrg (approved):
This seems good to go now.

PR #168, opened 2022-12-09
Variations: make more use of EntryFormat type

cmyr:
This can let our logic be more easily shared between read- and write-fonts.

@dfrg you were casting these to u32, but I don't think there's any overflow risk... was there some other rationale I might have missed?


dfrg:
I think the cast carried over from my earlier implementation before I was aware that the rhs of a shift wasn’t required to be of the same type as lhs. I also don’t see any potential for overflow here.

dfrg (approved):
Looks good!

PR #169, opened 2022-12-09
Basic compilation of variation tables

cmyr:
This isn't super useful as-is; we will need builders for these types that have various packed representations, but this at least gets the types in place so that they can be referenced by other tables.


dfrg (approved):
Great! I assume with this in place that the `item_var_store_offset` field in `GDEF` can be properly typed now?

cmyr:
@dfrg yes, good call, I'll add that here.

PR #170, opened 2022-12-09
BASE table

cmyr:
I'm just going to come out and admit it: I'm watching the football, and just catching up on a bunch of brainless busywork TODOs from my catch.

To whit: support for the BASE table. 💁


font-codegen/src/fields.rs
```diff
@@ -1090,6 +1090,11 @@ impl Field {
         } else if let Some(computed) = &self.attrs.compile {
             match &computed.attr {
                 CustomCompile::Expr(inline_expr) => {
+                    // this is slightly iffy: it isn't the compile type, so it
+                    // may cause problems later, but this whole "compute a value and cast"
+                    // approach is underpowered because sometimes we would like to just
```
rsheeter, line 1095: "sometimes we would like" is perhaps less helpful than specific examples

font-codegen/src/fields.rs
```diff
@@ -1090,6 +1090,11 @@ impl Field {
         } else if let Some(computed) = &self.attrs.compile {
             match &computed.attr {
                 CustomCompile::Expr(inline_expr) => {
+                    // this is slightly iffy: it isn't the compile type, so it
```
rsheeter, line 1093: Maybe say what the code would say if it *was* the compile type?

rsheeter (approved):
LGTM, just a few very minor nits

PR #172, opened 2022-12-12
Compile OS/2 table

cmyr:
Straightforward, but the last missing core (non variable) table.

While I was in here I also added a few more doc links for OS/2 fields.


resources/codegen_inputs/os2.rs
```diff
@@ -2,19 +2,29 @@
...
     /// The Average Character Width parameter specifies the arithmetic average
     /// of the escapement (width) of all non-zero width glyphs in the font.
     x_avg_char_width: i16,
+    /// [Weight class](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass).
+    ///
     /// Indicates the visual weight (degree of blackness or thickness of
     /// strokes) of the characters in the font. Values from 1 to 1000 are valid.
+    #[default(400)]
     us_weight_class: u16,
-    /// Indicates a relative change from the normal aspect ratio (width to
```
dfrg, line 15: Looks like some of this comment was cut off?

dfrg (approved):
Just a couple issues with comments. Otherwise looks good.

In addition to the inline note, s/extend/extent/ in the comment for `us_win_descent`.

PR #174, opened 2022-12-13
[codegen] Make #[count] not accept expressions

cmyr:
This introduces a new, more constrained grammar for this attribute, supporting two forms.

In the simple form, it accepts a single argument, which may be either a `$field`, a literal integer (`52`) or the special token `..`, which means "all remaining data", and is only valid for the last field in a table.

In the non-simple form, it accepts a function like syntax, where the 'function' is one of a small set of predefined options. These functions expect some set number of pre-defined arguments, which may be either `$field`s or literals.

supported 'functions' are currently `subtract`, `half`, `delta_value_count` and `map_data_size`.

----

This actually wasn't too bad? It's a lot more code to do what we want, but it does have a nice conceptual clarity, and it makes it much more possible to imagine things like 'computing the inverse of a function', when the range of possible functions is discrete and known ahead of time. I also think the syntax ends up being quite reasonable.

I'm going to have to perform similar surgery in a few other locations, but I hope that this was one of the larger ones.

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -235,13 +235,6 @@ pub(crate) struct FieldReadArgs {
     pub(crate) inputs: Vec<syn::Ident>,
 }
 
-#[derive(Clone, Debug)]
-pub(crate) enum CountArg {
-    All(syn::token::Dot2),
-    Field(syn::Ident),
-    Literal(syn::LitInt),
-}
-
 /// Annotations for how to calculate the count of an array.
 ///
 /// ```no_compile
@@ -257,9 +250,15 @@ pub(crate) enum Count {
     },
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum CountArg {
+    All(syn::token::Dot2),
+    Field(syn::Ident),
+    Literal(syn::LitInt),
+}
+
 #[derive(Clone, Copy, Debug)]
 pub(crate) enum CountTransform {
-    //None,
     /// requires exactly two args, defined as $arg1 - $arg2
     Sub,
     /// requires exactly one arg. defined as $arg1 / 2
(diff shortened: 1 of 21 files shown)


font-codegen/src/parsing.rs
```diff
@@ -235,12 +235,37 @@ pub(crate) struct FieldReadArgs {
     pub(crate) inputs: Vec<syn::Ident>,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum CountArg {
```
rsheeter, line 239: Nit: suggest examples in comment

rsheeter: Also ".." for all is perhaps a touch Rust'ish?

Count(*)  // I can guess what this means
Count(..)  // Python hat on, whats this?

rsheeter: (belatedly realizes the examples are right below) ... maybe just contemplate the .. then :)

font-codegen/src/parsing.rs
```diff
@@ -235,12 +235,37 @@ pub(crate) struct FieldReadArgs {
...
+    Simple(CountArg),
+    NotSimple {
+        args: Vec<CountArg>,
+        xform: CountTransform,
+    },
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum CountTransform {
+    //None,
```
rsheeter, line 262: Remove commented out none?

font-codegen/src/parsing.rs
```diff
@@ -1184,22 +1209,85 @@ impl Item {
...
         } else {
-            input.parse().map(Self::Expr)
+            input.parse().map(Self::Simple)
+        }
+    }
+}
+
+impl Parse for CountTransform {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let ident = input.parse::<syn::Ident>()?;
```
rsheeter, line 1259: Why isn't this a match?

font-codegen/src/parsing.rs
```diff
@@ -1184,22 +1209,85 @@ impl Item {
...
+        if ident == "subtract" {
+            Ok(CountTransform::Sub)
+        } else if ident == "half" {
+            Ok(CountTransform::Half)
+        } else if ident == "delta_value_count" {
+            Ok(CountTransform::DeltaValueCount)
+        } else if ident == "map_data_size" {
+            Ok(CountTransform::DeltaSetIndexData)
```
rsheeter, line 1267: Why the name discrepency?

font-codegen/src/parsing.rs
```diff
@@ -1184,22 +1209,85 @@ impl Item {
...
+        } else {
+            Err(syn::Error::new(ident.span(), "unknown named transform, expected one of 'subtract', 'half' 'delta_value_count', or 'map_data_size'"))
```
rsheeter, line 1269: Can one generate the valid values list? - once you've got 10 of these lists at least one will be wrong

font-codegen/src/parsing.rs
```diff
@@ -1184,22 +1209,85 @@ impl Item {
...
+            CountTransform::DeltaValueCount => 3,
+            CountTransform::DeltaSetIndexData => 2,
+        }
+    }
+}
+
+impl ToTokens for CountArg {
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        match self {
+            CountArg::All(_) => panic!("Count::All is a special case"),
```
rsheeter, line 1288: This message will not help me fix the problem, can you offer advice?

rsheeter (approved):
I quite like it, lgtm

font-codegen/src/parsing.rs
cmyr: the attribute syntax that we use here doesn't need to be universal, it just needs to map to the schema; this is all going to get turned into yaml, where we will just use 'all' or something.

PR #175, opened 2022-12-14
[codegen] Minor refactor of Count attr

cmyr:
I realized there was a cleaner way of handling this, by moving the All variant out of CountArg and into the top-level Count struct; this reduces the number of special cases we need to think about.


rsheeter (approved):
Nit: it might be nice if the comment was slightly more specific about the change, e.g. use blah instead of blah

cmyr:
> Nit: it might be nice if the comment was slightly more specific about the change, e.g. use blah instead of blah

I hope I did this in the extended commit message?

PR #176, opened 2022-12-14
Fix incorrect format value for CharacterVariantParams

cmyr:


rsheeter (approved):
If only we had a test!

cmyr:
we do, just downstream 😅

PR #177, opened 2022-12-14
Fix version type in vhea table

cmyr:
We had specified this as MajorMinor (essentially a packed pair of u16s) but the type in the spec is Version16Dot16, which has its own special representation.


dfrg (approved):
And yet I still find myself surprised by all the inconsistencies 🤷‍♂️

cmyr:
Yea, this one is *particularly* weird. It seems pretty obvious from the spec that Version16Dot16 is soft-deprecated, so really no clue how it snuck into this table.

PR #179, opened 2022-12-15
Set default value for LangSys::required_feature_index

cmyr:
It would be nice to make this be Option<u16>, but that is blocked on us having some slightly better way of specifying custom compilation behaviour; this is an improvement in the meantime.


dfrg (approved):
I believe this is a relic from the days when OpenType layout might have been simple. It seems to be ignored in general so defaulting it looks like a good idea.

PR #180, opened 2022-12-15
[codegen] Make #[available] not accept expressions

cmyr:
Further work on making us schema-ready. Previously this attribute accepted any expression that evaluated to something which implemented the 'Comparable' trait, for the appropriate version type. In practice, this was so that we could provide both literal integers as well as named constants, such as `MajorMinor::VERSION_1_1`.

The schema will not support the concept of an arbitrary expression, however, and so this needs a more constrained representation.

With this patch, we no longer support the use of named constants in this attribute; instead the argument must be either a single integer literal, or a comma-separated major/minor pair of literal integers, which we will convert to the appropriate type when comparing.

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -174,7 +174,7 @@ pub(crate) struct Field {
 pub(crate) struct FieldAttrs {
     pub(crate) docs: Vec<syn::Attribute>,
     pub(crate) nullable: Option<syn::Path>,
-    pub(crate) available: Option<Attr<syn::Expr>>,
+    pub(crate) available: Option<Attr<Available>>,
     pub(crate) skip_getter: Option<syn::Path>,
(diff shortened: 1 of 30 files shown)


font-codegen/src/parsing.rs
```diff
@@ -235,6 +235,12 @@ pub(crate) struct FieldReadArgs {
     pub(crate) inputs: Vec<syn::Ident>,
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct Available {
```
rsheeter, line 239: Nit: I didn't find the meaning of available intuitive. Maybe it's worth the cost of a long name? - for example available_from_version(2) I would immediately get

cmyr: happy to bikeshed that name, I agree it isn't great. If we agree on a change will do a separate PR. :)

PR #183, opened 2022-12-16
Add read support for fvar table

dfrg:
This adds the `fvar` table. Support for the variable size `InstanceRecord` is handled similarly to `ValueRecord` using `ComputedArray` and a custom implementation `FontReadWithArgs` along with traversal stuff.

Also adds otexplorer support for fvar and avar.

Contains tests for reading axes and instances as well as computing normalized coordinates.


read-fonts/src/tables/instance_record.rs
```diff
@@ -0,0 +1,70 @@
+//! An fvar InstanceRecord
+
+use super::InstanceRecord;
```
cmyr, line 3: any reason not to define the type in this file?

resources/codegen_inputs/fvar.rs
```diff
@@ -0,0 +1,56 @@
...
+    axis_instance_arrays_offset: Offset16<AxisInstanceArrays>,
+    /// This field is permanently reserved. Set to 2.
+    #[skip_getter]
+    #[compile(2)]
+    _reserved: u16,
+    /// The number of variation axes in the font (the number of records in the axes array).
+    axis_count: u16,
+    /// The size in bytes of each VariationAxisRecord — set to 20 (0x0014) for this version.
+    #[compile(20)]
+    axis_size: u16,
```
cmyr, line 23: in theory, I imagine we want to be using this value to determine the stride length during parsing, or else we risk incorrect results with some future spec bump. There's a similar issue in STAT, where it is also not resolved; I'm okay with leaving it as-is for now, but we should definitely open an issue.

I have a half-written patch for STAT somewhere, so hopefully the pattern I use there will also be appropriate here.

aside: the idea of a record that may change size in the future is nasty, I feel like this should *really* just be a table with a version number. :/

cmyr (approved):
Looks good, two little notes inline.

read-fonts/src/tables/instance_record.rs
dfrg: Oh, I think I had done everything in `fvar.rs` before I noticed the pattern you used with `ValueRecord` and just didn't copy the base type over. I'll update it.

resources/codegen_inputs/fvar.rs
dfrg: I agree, and I think the spec is completely broken here. I honestly don't see how they can change either of the sizes in this table without breaking just about every implementation in the wild. 

I'll keep this as-is and let you carry over whatever pattern you're considering for `STAT` when compilation comes up for this table.

read-fonts/src/tables/instance_record.rs
dfrg: Turns out the fields on that type weren't pub either, so I'm glad you called this out :)

PR #185, opened 2022-12-17
[write-fonts] Small ValueRecord additions

cmyr:
This makes ValueRecord hashable, and also adds a method to ValueFormat for computing the byte size of the record.

This has again highlighted one annoyance of this tool, which is there's no way to change or augment the set of traits that we derive for a given object; it would be nice if we could add support in codegen for this. (I am mentioning it here so that I can reference this in a new issue.)


dfrg (approved):
lgtm!

PR #187, opened 2022-12-17
Change class sequence rules to use u16

dfrg:
These are currently defined as sequences of `GlyphId` but should actually be arbitrary `u16` class values as stored in a `ClassDef`.


cmyr (approved):
Thanks!

PR #188, opened 2022-12-20
Coverage table & class def fixups

cmyr:
- There was a bug where we didn't dedupe entries in the coverage table before building
- There was an inefficiency where we would allow explicit class 0 entries when building classdefs
- I wanted a method on the write-fonts ClassDef to get the class for an item

### DIFF
--- a/write-fonts/src/tables/layout.rs
+++ b/write-fonts/src/tables/layout.rs
@@ -321,6 +321,7 @@ impl CoverageTableBuilder {
     /// Create a new builder from a vec of `GlyphId`.
     pub fn from_glyphs(mut glyphs: Vec<GlyphId>) -> Self {
         glyphs.sort_unstable();
+        glyphs.dedup();
         CoverageTableBuilder { glyphs }
     }
 
@@ -623,5 +624,17 @@ mod tests {
         let result = encode_delta(DeltaFormat::Local2BitDeltas, &inp);
         assert_eq!(result.len(), 1);
         assert_eq!(result[0], 0x5540_u16);
+    }
+
+    fn make_glyph_vec<const N: usize>(gids: [u16; N]) -> Vec<GlyphId> {
+        gids.into_iter().map(GlyphId::new).collect()
+    }
+
+    #[test]
+    fn coverage_builder() {
+        let coverage = make_glyph_vec([1u16, 2, 9, 3, 6, 9])
+            .into_iter()
+            .collect::<CoverageTableBuilder>();
+        assert_eq!(coverage.glyphs, make_glyph_vec([1, 2, 3, 6, 9]));
     }
 }
--- a/write-fonts/src/tables/layout.rs
+++ b/write-fonts/src/tables/layout.rs
@@ -229,6 +229,27 @@ impl ClassDef {
         one.into_iter().flatten().chain(two.into_iter().flatten())
     }
 
+    /// Return the glyph class for the provided glyph.
+    ///
+    /// Glyphs which have not been assigned a class are given class 0
+    pub fn get(&self, glyph: GlyphId) -> u16 {
+        self.get_raw(glyph).unwrap_or(0)
+    }
+
+    // exposed for testing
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/layout.rs
```diff
@@ -229,6 +229,27 @@ impl ClassDef {
         one.into_iter().flatten().chain(two.into_iter().flatten())
     }
 
+    /// Return the glyph class for the provided glyph.
+    ///
+    /// Glyphs which have not been assigned a class are given class 0
+    pub fn get(&self, glyph: GlyphId) -> u16 {
+        self.get_raw(glyph).unwrap_or(0)
```
rsheeter, line 236: unwrap_or_default?

write-fonts/src/tables/layout.rs
```diff
@@ -229,6 +229,27 @@ impl ClassDef {
...
+    }
+
+    // used for testing
```
rsheeter, line 239: used by pub fn get so this is perhaps misleading?

write-fonts/src/tables/layout.rs
```diff
@@ -229,6 +229,27 @@ impl ClassDef {
...
+    fn get_raw(&self, glyph: GlyphId) -> Option<u16> {
+        match self {
+            ClassDef::Format1(table) => glyph
+                .to_u16()
+                .checked_sub(table.start_glyph_id.to_u16())
+                .and_then(|idx| table.class_value_array.get(idx as usize))
+                .copied(),
+            ClassDef::Format2(table) => table.class_range_records.iter().find_map(|rec| {
+                (rec.start_glyph_id <= glyph && rec.end_glyph_id <= glyph).then_some(rec.class)
```
rsheeter, line 248: class range records are sorted by start glyph so bsearch? Or are you concerned the ones modified for write aren't? - if so plz comment

write-fonts/src/tables/layout.rs
```diff
@@ -229,6 +229,27 @@ impl ClassDef {
...
+            }),
+        }
+    }
+
     pub fn class_count(&self) -> u16 {
         //TODO: implement a good integer set!!
```
rsheeter, line 254: This wants to be an issue, presumably one that says "make me a memory safe hb_set"?

write-fonts/src/tables/layout.rs
```diff
@@ -624,4 +646,33 @@ mod tests {
...
+                .iter()
+                .map(|(gid, cls)| (GlyphId::new(*gid), *cls))
+                .collect::<ClassDefBuilder>()
+                .build()
+        }
+
+        let class = make_class([(4, 0), (5, 1)]);
+        assert!(class.get_raw(GlyphId::new(4)).is_none());
+        assert_eq!(class.get_raw(GlyphId::new(5)), Some(1));
+        assert!(class.get_raw(GlyphId::new(100)).is_none());
```
rsheeter, line 676: why not test the public get fn?

rsheeter (approved):
LGTM w/nits

write-fonts/src/tables/layout.rs
cmyr: The issue is I actually want to know if there is a zero value in the record, instead of it being used as a placeholder; the public fn does unwrap_or(0).

write-fonts/src/tables/layout.rs
cmyr: they're sorted, but this collection can never be that huge, so I would expect the average cost of linear and binary search to come out about equal.

So mostly just "that'll do" 🤷

write-fonts/src/tables/layout.rs
cmyr: totally a personal thing, I just sometimes prefer the explicitness.

write-fonts/src/tables/layout.rs
cmyr: #192

PR #191, opened 2022-12-20
Initial glyph loading/scaling crate

dfrg:
This adds a new crate called `punchcut` containing the initial TrueType glyph scaling code based on `read-fonts`. The caching and hinting code has been extracted to hopefully make this more reviewable.

A few notes:
* There are fixed point math functions that have some overlap with code in `font-types`. FreeType provides outline points as either integral font units or fixed point (26.6) values depending on whether a scale has been applied. We copy that here and that makes some of the math and types less than ideal. I'm open to options for making this nicer (without loss in perf/memory) but I'd like to not block the initial cut on it.
* There is a tiered architecture that exposes the individual sources (just `glyf` for now) as well as a higher level interface at the root of the crate. The layering feels pretty good to me right now, but comments for improvement are welcome.
* I've tried to comment some of the bits in `source/glyf/scaler.rs` that might be tricky. Feel free to point out any that require further explanation.

This has nice integration with `kurbo`: the `Outline` type implements `kurbo::Shape` so these can be fed directly into `vello` (née piet-gpu). I plan on using this for glyph loading support in that crate soon™.


punchcut/docs/capi.md
```diff
@@ -0,0 +1,154 @@
...
+    PC_HINTING_NONE = 0,
+    // Full gridfitting.
+    PC_HINTING_LEGACY = 1,
+    // Grayscale antialiasing (FreeType light + grayscale)
+    PC_HINTING_GRAYSCALE_SUBPIXEL = 2,
+    // RGB subpixel hinting (FreeType light + LCD)
+    PC_HINTING_SUBPIXEL = 3,
+    // Same as RGB subpixel, but prevents all movement
+    // in the horizontal direction.
+    PC_HINTING_MODERN = 4,
```
cmyr, line 28: is this name in general use? I have a slight allergy to the use of 'modern' in general, since it doesn't really communicate much more than "new when I named this".

punchcut/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+}
+
+use font::Tag;
+use source::*;
+
+use core::str::FromStr;
+
+pub use error::*;
+pub use outline::*;
+pub use scaler::*;
```
cmyr, line 30: I don't love globs but 🤷

punchcut/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl Context {
+    /// Creates a new glyph loading context.
+    pub fn new() -> Self {
+        Self {
+            glyf: glyf::Context::new(),
+            glyf_outline: glyf::Outline::new(),
+            coords: vec![],
+            variations: vec![],
+        }
```
cmyr, line 119: this looks like it should be able to derive default?

punchcut/src/outline.rs
```diff
@@ -0,0 +1,293 @@
...
+    Shape,
+};
+
+use crate::source::*;
+
+/// Simple outline represented as a sequence of path commands.
+#[derive(Clone, Default, Debug)]
+pub struct Outline {
+    verbs: Vec<Verb>,
+    points: Vec<(f32, f32)>,
```
cmyr, line 16: Is this another case where we wish we had 32-bit kurbo? I definitely think some of the code below would be easier to follow if the math was implemented in terms of a `Point` type...

punchcut/src/outline.rs
```diff
@@ -0,0 +1,293 @@
...
+            verb_pos: 0,
+            point_pos: 0,
+        }
+    }
+
+    /// Appends a glyf outline, converting it into a sequence of path elements.
+    /// Returns false if the source outline is malformed.
+    pub fn append_glyf(&mut self, outline: &glyf::Outline) -> bool {
+        #[inline(always)]
+        fn conv(p: glyf::Point, s: f32) -> (f32, f32) {
```
cmyr, line 61: maybe worth just adding an impl in `read-fonts`.

punchcut/src/outline.rs
cmyr, line 16: Just for my edification, what's the rationale for using this representation, instead of an enum ala `kurbo::PathEl`? Is this about better memory efficiency?

punchcut/src/outline.rs
```diff
@@ -0,0 +1,293 @@
...
+                PathEl::CurveTo(pt(p0), pt(p1), pt(p2))
+            }
+            Verb::Close => PathEl::ClosePath,
+        })
+    }
+}
+
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(u8)]
+enum Verb {
```
cmyr, line 287: silly nit, but maybe move this up to be near the `Outline` struct? would improve the experience of reading this file I think

punchcut/src/scaler.rs
```diff
@@ -0,0 +1,177 @@
...
+        Self {
+            context,
+            font_id: None,
+            size: 0.0,
+            hint: None,
+        }
+    }
+
+    /// Sets a unique font identifier for hint state caching.
+    pub fn font_id(mut self, font_id: Option<u64>) -> Self {
```
cmyr, line 32: do we expect the user to call this with `None`? Who is responsible for generating these ids?

punchcut/src/scaler.rs
```diff
@@ -0,0 +1,177 @@
...
+        self.font_id = font_id;
+        self
+    }
+
+    /// Sets the font size in pixels per em units. A size of 0.0 will disable
+    /// scaling and result in glyphs defined in font units.
```
cmyr, line 38:
```suggestion
    /// Sets the font size in pixels per em units.
    ///
    /// A size of 0.0 will disable scaling and result in glyphs defined in font units.
```

punchcut/src/scaler.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /// Sets the hinting mode.
+    pub fn hint(mut self, hint: Option<Hinting>) -> Self {
+        self.hint = hint;
+        self
+    }
+
+    /// Specifies a variation with a set of normalized coordinates. This will
+    /// clear any variations specified with the variations method.
```
cmyr, line 51:
```suggestion
    /// Specifies a variation with a set of normalized coordinates.
    ///
    /// This will clear any variations specified with the variations method.
```

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,621 @@
...
+            let pp0x = self.phantom[0].x;
+            if pp0x != 0 {
+                for p in &mut outline.points {
+                    p.x -= pp0x;
+                }
+            }
+            outline.is_scaled = self.scaler.is_scaled;
+        }
+        Ok(())
+    }
```
cmyr, line 396: it would  be nice if we could figure out a way to split this function up a bit, although I appreciate all the comments.

cmyr (approved):
A bunch of comments inline, but overall, (and skipping over the fact that I don't have a particularly firm grasp on all of the business logic) this seems quite clean!

I think there's likely some room for improvement around types, as you say, but I don't have a ton of concrete proposals, outside of 'more things should be types, and more things like conversion logic and operations should be methods on those types'.

so lgtm, but I am not a domain expert.

punchcut/docs/capi.md
madig: "Linear" might describe it better. Also, what's up with vertical text hinting, anyway?

madig: Also, the options here seem a bit arbitrary... Gray-scale/Subpixel and amount and direction of gridfitting are orthogonal concepts, where for the latter we have accidental control in TTFs (hinting done by code in font) and full control in CFFs (hinting done by renderer).

madig:
So, this crate is for loading glyphs from various OT formats and turn them into kurbo datatypes, potentially applying hinting?

dfrg:
Yes, that’s the intent. There are working hinters for both TrueType and CFF but I left the former out to keep the PR size more manageable and the latter is waiting on CFF parsing in fontations.

madig:
Interesting. Did you port FreeType's TTF hinting engine?

punchcut/docs/capi.md
```diff
@@ -0,0 +1,154 @@
+# C API for punchcut
```
drott, line 1: Would it make sense to split this out from the pull request and discuss separately?

punchcut/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+/// Modes for hinting.
+///
+/// Only the `glyf` source supports all hinting modes.
+#[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
+#[allow(dead_code)]
+pub enum Hinting {
+    /// "Full" hinting mode. May generate rough outlines and poor horizontal
+    /// spacing.
+    Legacy,
```
drott, line 40: Agree with @cmyr's related feedback on the naming in the C-API draft. If these are taken from FreeType - I'd be very happy if we could make at least "Legacy" and "Modern" more descriptive.

punchcut/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+license = "MIT/Apache-2.0"
+description = "Loading glyphs from OpenType font files."
+repository = "https://github.com/googlefonts/fontations"
+readme = "README.md"
+categories = ["text-processing", "parsing", "graphics"]
+
+[dependencies]
+read-fonts = { version = "0.0.4", path = "../read-fonts" }
+peniko = { git = "https://github.com/linebender/peniko" }
```
drott, line 13: If I understand this correctly, this dependency is used for the `Shape` impl on top of `Outline`. Could this dependency be made optional, and/or the `Shape` implementation be a separate pull request? If/when punchcut is used in a Skia backend to extract outlines, I think we could avoid the peniko/kurbo dependency.

punchcut/src/source/glyf/math.rs
```diff
@@ -0,0 +1,135 @@
+//! Various fixed point math functions used in TrueType scaling and hinting.
```
drott, line 1: If I am not mistaken, a couple of functions in this file are not used yet and could be landed separately, haven't found references for `muldiv_no_round` or `hypot` for example.

punchcut/src/lib.rs
```diff
@@ -0,0 +1,200 @@
+/*!
+Glyph loading.
+*/
+
+#![forbid(unsafe_code)]
+// TODO: this is temporary-- remove when hinting is added.
+#![allow(dead_code, unused_imports, unused_variables)]
+
+/// Re-export of peniko crate.
+pub use peniko;
```
drott, line 10: Related to previous comment in `Cargo.toml`, could we make the peniko dependency optional?

drott (commented):
Hope that's okay if I leave a couple of drive-by comments after I was familiarizing myself with the punchcut API and the suggestions for the C FFI.

punchcut/docs/capi.md
dfrg: The grayscale/subpixel options are not necessarily arbitrary. These are returned as flags in the GETINFO instruction and do actually affect hinting, specifically for fonts in the ClearType collection. The names are unfortunate though.

I suggest replacing them with Full, Light, LightSubpixel and VerticalSubpixel. Do those make more sense?

punchcut/src/lib.rs
dfrg: The number of symbols is relatively small, so happy to remove the globs :)

punchcut/src/lib.rs
dfrg: I believe so. I'll add it.

punchcut/src/outline.rs
dfrg: Yes, `PathEl` is 56 bytes while the typical quadratic segment can be represented in 17 bytes in this representation. This is a fairly significant difference. From the most recent discussion, we may remove this entirely though and just add a path sink style trait that would avoid the issue.

punchcut/src/outline.rs
dfrg: It might make sense to change the `Point` struct in `read-fonts` to take a type parameter and then potentially add a "convert to f32 with scale method"

punchcut/src/outline.rs
dfrg: Right. I think I made the rationale in the above comment about f32 in kurbo. This is probably moot now though.

punchcut/src/scaler.rs
dfrg: The choice of allowing `None` here is to allow builder methods to be called with the "default" values to avoid breaking the method chain.

The client would be responsible for generating the ids. I tried with a typed auto-generated unique id (CacheKey) in swash but the friction was real when integrating with other code so I wanted to make this as generic as possible.

punchcut/src/source/glyf/scaler.rs
dfrg: I can definitely split this up into functions for simple and composite glyphs and that should make each more tractable.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,621 @@
...
+                        .flags
+                        .contains(CompositeGlyphFlags::USE_MY_METRICS)
+                    {
+                        // The USE_MY_METRICS flag indicates that this component's phantom
+                        // points should override those of the composite glyph.
+                        self.phantom = phantom;
+                    }
+                    // Scaling does an internal conversion to 26.6 so we don't use the
+                    // fixed types in read-fonts. Maybe a better solution here? We want to match
+                    // FreeType semantics.
```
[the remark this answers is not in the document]
dfrg: We can add some convenience methods to make this a bit nicer, but ultimately there's some inherent complexity around how both scaling and hinting handle a mix of integers and fixed 26.6 values. It might be worth just taking the time to identify and implement these things.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,621 @@
...
+        self.phantom[1].x = self.phantom[0].x + advance;
+        self.phantom[1].y = 0;
+        self.phantom[2].x = advance / 2;
+        self.phantom[2].y = bounds[3] as i32 + tsb;
+        self.phantom[3].x = advance / 2;
+        self.phantom[3].y = self.phantom[2].y - vadvance;
+    }
+}
+
+// Phantom point management.
```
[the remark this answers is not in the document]
dfrg: Good point :) I'll add a comment. The gist is that phantom points are added to the outline to allow the interpreter to compute hinted metrics-- specifically advance width.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,621 @@
...
+    }
+}
+
+/// Slot for the hinting cache.
+#[derive(Copy, Clone)]
+enum CacheSlot {
+    /// Uncached font.
+    Uncached,
+    /// Font and size cache indices.
+    Cached(usize, usize),
```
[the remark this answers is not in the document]
dfrg: Will do.

punchcut/docs/capi.md
dfrg: I don't think I meant to push that to the branch for this PR. Per our discussion, we need something more sophisticated so I'm happy to remove the file and punt on this.

punchcut/src/lib.rs
dfrg: I made suggestions on Colin's comment above. I wonder if those feel better to you?

punchcut/Cargo.toml
dfrg: We could remove it at this time, but the dependency exists for future COLRv1 support where we'll need to represent affine transforms and the various brush types. We don't absolutely require it... we can define our own types in this crate, but I'd prefer to keep it so that those types are consistent across the ecosystem. Perhaps we can do some measurement on binary size to see what the actual effect is?

punchcut/src/source/glyf/math.rs
dfrg: `hypot` is used for scaled component offsets but `muldiv_no_round` and a few others are only used in the hinting code. I can remove these and restore them in a subsequent PR.

punchcut/src/lib.rs
dfrg: As mentioned above, we can remove it, but future COLRv1 support will end up duplicating types from these libraries. FWIW, we own both peniko and kurbo.

dfrg:
> Interesting. Did you port FreeType's TTF hinting engine?

Yes, I ported it a while back for the [swash](https://github.com/dfrg/swash) crate.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,621 @@
...
+    pub fn get_into(&mut self, glyph_id: GlyphId, outline: &mut Outline) -> Result<()> {
+        outline.clear();
+        self.context.unscaled.clear();
+        self.context.original.clear();
+        self.context.deltas.clear();
+        if glyph_id.to_u16() >= self.font.glyph_count {
+            return Err(Error::GlyphNotFound(glyph_id));
+        }
+        GlyphScaler::new(self).load(glyph_id, outline, 0)?;
+        Ok(())
```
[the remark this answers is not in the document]
dfrg: Yep!

punchcut/docs/capi.md
dfrg: Removed from this PR and stored at https://gist.github.com/dfrg/f1b9022541e66b881b6e82fd2f9969b1 for future reference.

punchcut/Cargo.toml
drott: If I am doing this right (see below) looks like `kurbo` is 1.8M and `peniko` 459k. IMO this is quite substantial. In Chrome we already get warnings for Android binary size (which we need to keep small for being able to install Chrome effectively through play store) for CLs that increase the binary size by 20k. 

As for the initial prototype we probably won't have COLRv1 support through punchcut immediately, I'd still prefer to make this dependency optional and include it, ideally optionally too, later for COLRv1 support in punchcut.

> represent affine transforms and the various brush types.

By "various brush types" do you mean the gradient definitions? By brush I tend to imagine something like a path-effect or shape to a line or curve stroke. If we're only interested in the type definitions for the interface, it seems quite heavy to depend on all of peniko and kurbo if the size impact is the whole library - but I am not sure to what extent the linking steps can prune unused parts.

`$ cargo build -r --lib`

then 

```bash
~/dev/peniko/target/release/deps$ ls -lha *rlib
-rw-r--r-- 1 drott primarygroup 502K Jan 12 10:37 libarrayvec-b8ff81c131b08d72.rlib
-rw-r--r-- 1 drott primarygroup 1.8M Jan 12 10:37 libkurbo-e1fffba60c3a2368.rlib
-rw-r--r-- 2 drott primarygroup 459K Jan 12 10:37 libpeniko-4b15967201e87c2a.rlib
-rw-r--r-- 1 drott primarygroup 444K Jan 12 10:37 libsmallvec-cd2b8e69191b620b.rlib
```

punchcut/docs/capi.md
drott: These seem like improvements over "legacy" and "modern". As we can't find the full technical description into the enum value, these aliases or shorthands work for me.

punchcut/Cargo.toml
dfrg: Thanks for the analysis. We’ve settled on limiting dependencies to crates in this repo so I’ll remove these and perhaps restore them later behind a feature gate.

I wanted to note that rlibs contain full binary code along with a significant chunk of metadata and IR to facilitate incremental builds and generic instantiation in downstream crates. They’re not particularly representative of final artifact size.

rsheeter: +1 to gate this on removal of those deps. Also ... how does one correctly measure binary size?

rsheeter: Filed https://github.com/googlefonts/fontations/issues/200 to establish an accepted and easily repeated mechanism to measure size impact.

punchcut/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Features
+
+Current (✔️), near term (🔜) and planned (⌛) feature matrices:
+
+#### Vector sources:
+
+| Source | Loading | Variations | Hinting |
+|--------|---------|------------|---------|
+| glyf   | ✔️     |  🔜        | ✔️*    |
```
rsheeter, line 17: IMHO hinting should be planned with a * since it doesn't work at time of writing.

punchcut/README.md
```diff
@@ -0,0 +1,43 @@
...
+| CFF    | ⌛     | ⌛         | ⌛     |
+| CFF2   | ⌛     | ⌛         | ⌛     |
+| COLR   | 🔜     | 🔜         | **      |
```
rsheeter, line 20: Suggest distinguishing COLRv1, COLRv2 akin to CFF and CFF2

punchcut/README.md
```diff
@@ -0,0 +1,43 @@
...
+\*\* This will be supported but is probably not desirable due the general affine transforms
+present in the paint graph.
+
+#### Bitmap sources:
+
+| Source | Loading |
+|--------|---------|
+| EBDT   | ⌛     |
+| CBDT   | ⌛     |
+| sbix   | ⌛     |
```
rsheeter, line 33: Optional, I don't think breaking out bitmaps from vectors is buying us much; could be a single matrix.

punchcut/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## The name?
+
+Wikipedia says "[punchcutting](https://en.wikipedia.org/wiki/Punchcutting) is a craft used in traditional
+typography to cut letter punches in steel as the first stage of making metal type." The punches carry the
+outline of the desired letter which can be used to create a mold to transfer the design to various
+surfaces.
+
+The primary purpose of this crate is the generation of outlines from font data, so the name seemed
+appropriate.
```
rsheeter, line 43: Provide (a link could suffice) a basic example of use, say loading a font from a file and getting the outline of something.

punchcut/src/error.rs
```diff
@@ -0,0 +1,52 @@
+use read_fonts::{types::GlyphId, ReadError};
```
rsheeter, line 1: We should have a common glyphname too! https://github.com/googlefonts/fontations/issues/201

punchcut/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self::Read(e)
+    }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::NoSources => write!(f, "No glyph sources are available for the given font"),
+            Self::GlyphNotFound(gid) => write!(f, "Glyph {} was not found in the given font", gid),
+            Self::RecursionLimitExceeded(gid) => write!(
```
rsheeter, line 33: Suggest printing what the limit is

punchcut/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+                f,
+                "Recursion limit exceeded when loading composite component {}",
+                gid
+            ),
+            Self::HintingFailed(gid) => write!(f, "Bad hinting bytecode for glyph {}", gid),
```
rsheeter, line 38: Feature gate this one.

Also, we should provide any additional information we have. Perhaps we know _something_ about what went wrong?

punchcut/src/error.rs
```diff
@@ -0,0 +1,52 @@
...
+            Self::InvalidAnchorPoint(gid) => write!(
+                f,
+                "Invalid anchor point indices for composite glyph {}",
```
rsheeter, line 41: If possible add detail, e.g. what indices

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+pub use error::{Error, Result};
+pub use scaler::{Scaler, ScalerBuilder};
+pub use sink::PathSink;
+
+/// Modes for hinting.
+///
+/// Only the `glyf` source supports all hinting modes.
+#[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
+#[allow(dead_code)]
+pub enum Hinting {
```
rsheeter, line 37: Disappear behind feature gate? ... I'm going to stop writing that for *every* hinting item from here on

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    VerticalSubpixel,
+}
+
+/// Type for a normalized variation coordinate.
+pub type NormalizedCoord = read_fonts::types::F2Dot14;
+
+/// Type for a glyph identifier.
+pub type GlyphId = read_fonts::types::GlyphId;
+
+/// Setting for specifying a variation by tag and value.
```
rsheeter, line 58: Useless comment, tell me something about it. What's it used for, what coordinate system is the value in, what thing(s) in the otspec should I use it to represent.

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+#[derive(Copy, Clone, Debug)]
+pub struct Variation {
+    /// Tag for the variation.
+    pub tag: Tag,
+    /// Value for the variation.
```
rsheeter, line 63: Commenting the field Value on the Variation type "Value for the variation" is just noise. Either find something interesting to say or don't comment.

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    fn from(s: ([u8; 4], f32)) -> Self {
+        Self {
+            tag: Tag::new_checked(&s.0[..]).unwrap_or_default(),
+            value: s.1,
+        }
+    }
+}
+
+/// Context for loading glyphs from any available source.
+#[derive(Default)]
```
rsheeter, line 95: I could see wanting to log this, perhaps add Debug>. And ... I don't see why to disallow Clone.

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+pub struct Context {
+    /// Inner context for loading TrueType outlines.
+    glyf: glyf::Context,
+    /// Internal storage for TrueType outlines.
+    glyf_outline: glyf::Outline,
```
rsheeter, line 100: It says above it's for any source but it appears to only support glyf. If "any source" is aspirational comment the type so and explain how you imagine extending it. For example, would the glyf context and glyf become an enum?

punchcut/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    #[test]
+    fn scaled_16_ppem() {
+        let font = FontRef::new(read_fonts::test_data::test_fonts::VAZIRMATN_VAR).unwrap();
+        test_glyph(
+            &font,
+            GlyphId::new(3),
+            16.0,
+            // Path elements scaled to 16ppem as computed by FreeType
```
rsheeter, line 163: From the information here I don't know how to add additional tests, how exactly do I produce the FreeType #'s for an arbitrary glyph? - could go into an md if you like, in which case "see readme" here might make sense

punchcut/src/scaler.rs
```diff
@@ -0,0 +1,172 @@
+use super::{source::*, Context, Error, Hinting, NormalizedCoord, PathSink, Result, Variation};
```
rsheeter, line 1: Suggest a comment that explains, or links to an explanation, what glyph scaling is, why you do it, etc. Imagine a Noogler is to work with this, would they "get it"?

punchcut/src/scaler.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.context
+            .variations
+            .extend(variations.into_iter().map(|v| v.into()));
+        self
+    }
+
+    /// Builds a scaler using the currently configured settings
+    /// for the specified font.
+    pub fn build(self, font: &impl TableProvider<'a>) -> Scaler<'a> {
+        if !self.context.variations.is_empty() {
```
rsheeter, line 84: Easier to read with less nesting; suggest a helper fn that early-exits with guard clauses for some of the nop cases such as no variation settings, no fvar, or no axes.

punchcut/src/scaler.rs
rsheeter, line 1: I thought import * was bad practice in Rust?

punchcut/src/sink.rs
```diff
@@ -0,0 +1,74 @@
+use peniko::kurbo::{BezPath, PathEl};
```
rsheeter, line 1: We are going to remove the peniko and kurbo deps iiuc, so presumably this needs to significantly change?

punchcut/README.md
dfrg: I'll add some example code. Inline is fairly common in Rust READMEs

punchcut/src/error.rs
dfrg: This would likely need to be feature gated as `font-types` can't depend on external libraries for `smol_str` and also doesn't depend on `std`, so `Arc` is out. I was thinking a `fontc-types` crate might be useful for the compiler.

punchcut/src/lib.rs
dfrg: This type will actually go away once the `feta` crate lands as it has a generic `Setting<T>` designed to support both variations and features. I'll make sure that one is better documented.

punchcut/src/lib.rs
dfrg: Same as above

punchcut/src/source/glyf.rs
```diff
@@ -0,0 +1,109 @@
...
+            original: vec![],
+            deltas: vec![],
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{Context, Outline, Scaler};
+    use crate::{font::*, GlyphId};
```
rsheeter, line 45: The code handles many scenarios we do not test, perhaps we should add more ttx's to read-fonts and make tests? - for example, variations seems to be entirely untested.

punchcut/src/lib.rs
dfrg: The context is font independent so it would grow support for other formats (CFF, COLR) with additional fields. I'll add a note.

punchcut/src/source/glyf.rs
```diff
@@ -0,0 +1,109 @@
...
+mod math;
+mod outline;
+mod scaler;
+
+pub use outline::Outline;
+pub use scaler::Scaler;
+
+pub use read_fonts::tables::glyf::Point;
+
+/// Context for loading for TrueType glyphs.
```
rsheeter, line 15: The term context throws me here, this is a fairly raw representation of an actual glyph isn't it?

punchcut/src/source/glyf.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl Default for Context {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Context {
+    /// Creates a new context.
```
rsheeter, line 32: Unhelpful comment. Wait maybe I said I'd stop complaining about those... NOW I'll stop :)

punchcut/src/source/glyf.rs
rsheeter: I'd be OK deferring that to an issue that enumerates scenarios but I do think we want to test this a little more.

punchcut/src/lib.rs
dfrg: I'd very much like to have a separate repo that is built for large scale testing of all fontations code with crates designed to do that and where we can properly document methodology. I believe @cmyr would like the same for the compiler side. 

In the meantime, I'll add some instructions to reproduce.

punchcut/src/scaler.rs
dfrg: I'll add some more context. I can probably just copy from the swash docs (https://docs.rs/swash/0.1.6/swash/scale/index.html)

punchcut/src/scaler.rs
dfrg: I'll break it up a bit.

punchcut/src/scaler.rs
dfrg: It's really not necessary here as the only item in there is `glyf`. I'll fix it.

On the broader subject, I think it's generally fine for *private* imports when you're dealing with large sets of symbols. It's also common when libraries offer `prelude` modules which are designed to be glob imported.

punchcut/src/source/glyf/math.rs
```diff
@@ -0,0 +1,83 @@
...
```
rsheeter, line 1: Needs comments and tests. What is it for, why is it here, why is our transform missing dx, dy, etc.

Also ... is this really only useful to punchcut? - it feels slightly misplaced, it's a fundamental buried in a task-specific crate. Should we have a font-math?

punchcut/src/source/glyf/outline.rs
```diff
@@ -0,0 +1,165 @@
+use super::Point;
+use crate::PathSink;
+
+/// TrueType outline.
+#[derive(Default, PartialEq, Eq, Debug)]
+pub struct Outline {
+    /// Set of points that define the shape of the outline.
+    pub points: Vec<Point>,
+    /// Set of tags (one per point).
+    pub tags: Vec<u8>,
```
rsheeter, line 10: Why not Vec(Point, PointTag) then? - would make the tag-per-point relation explicit and unscrewupable.

punchcut/src/sink.rs
dfrg: I'll just drop the trait impls from that module. We can potentially add them back behind a feature gate in the future, but it's not strictly important.

punchcut/src/source/glyf/outline.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Index of the end points for each contour in the outline.
+    pub contours: Vec<u16>,
+    /// True if the scaler applied a scale, in which case the points are in
+    /// 26.6 fixed point format. Otherwise, they are in integral font units.
+    pub is_scaled: bool,
```
rsheeter, line 15: Can wemodel this explicitly instead? Spitballing, make two types, make Point<T>, and then hold an enum here that is either `ScaledPoints(Vec<Point<26.6 guage fixed point>>)` or `ScaledPoints(Vec<Point<FontUnit>>)`

punchcut/src/source/glyf.rs
dfrg: It represents the intermediate state that is necessary for loading a glyph. This is where the hinting cache will live so it appears sparse at the moment and may seem less "context-like" at a glance.

punchcut/src/source/glyf/outline.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn clear(&mut self) {
+        self.points.clear();
+        self.tags.clear();
+        self.contours.clear();
+        self.is_scaled = false;
+    }
+
+    /// Converts the outline to a sequence of path commands and invokes the callback for
+    /// each on the given sink.
+    pub fn to_path(&self, sink: &mut impl PathSink) -> bool {
```
rsheeter, line 34: more comments, imagine a coder new to fonts is trying to follow.

punchcut/src/source/glyf/outline.rs
```diff
@@ -0,0 +1,165 @@
...
+        const TAG_MASK: u8 = 0x3;
+        const CONIC: u8 = 0x0;
+        const ON: u8 = 0x1;
+        const CUBIC: u8 = 0x2;
+        let s = if self.is_scaled { 1. / 64. } else { 1. };
+        let points = &self.points;
+        let tags = &self.tags;
+        let mut count = 0usize;
+        let mut last_was_close = false;
+        for c in 0..self.contours.len() {
```
rsheeter, line 48: the single letter names are not making this easier to read

punchcut/src/source/glyf/outline.rs
```diff
@@ -0,0 +1,165 @@
...
+                return false;
+            }
+            let mut v_start = points[cur];
+            let v_last = points[last];
+            let mut tag = tags[cur] & TAG_MASK;
+            if tag == CUBIC {
+                return false;
+            }
+            let mut step_point = true;
+            if tag == CONIC {
```
rsheeter, line 65: When do we encounter conics?

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,640 @@
...
+/// Limit for recursion when loading composite glyphs.
+const RECURSION_LIMIT: usize = 32;
+
+/// TrueType glyph scaler for a specific font and configuration.
+pub struct Scaler<'a> {
+    /// Backing context.
+    context: &'a mut Context,
+    /// Current font data.
+    font: Font<'a>,
+    /// Font identifier for the hinting cache.
```
rsheeter, line 23: help the reader understand how such ids are created, preferably by having a type, at the very least a newtype not u64, and some lovely comments on said type

punchcut/src/source/glyf/math.rs
dfrg: Most of these are very specific to TrueType scaling and hinting. The transform function, for example, only deals with a 2x2 matrix (because that is the only thing present in a composite glyph) and has an additional scale factor to handle the mixture of 16.16 and 26.6 values.

I'll add comments and tests.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,640 @@
...
+    /// Loads an outline for the specified glyph identifier.
+    pub fn get(&mut self, glyph_id: GlyphId) -> Result<Outline> {
+        let mut outline = Outline::new();
+        self.get_into(glyph_id, &mut outline)?;
+        Ok(outline)
+    }
+}
+
+/// State for loading a glyph.
+struct GlyphScaler<'a, 'b> {
```
rsheeter, line 95: just 'a I can handle but if you are going to have more than 1 I feel strongly names and/or commentary makes sense.

punchcut/src/source/glyf/scaler.rs
```diff
@@ -0,0 +1,640 @@
...
+            ppem,
+            scale,
+            hint,
+            coords,
+        })
+    }
+
+    /// Loads an outline for the specified glyph identifier to the preallocated
+    /// target.
+    pub fn get_into(&mut self, glyph_id: GlyphId, outline: &mut Outline) -> Result<()> {
```
rsheeter, line 74: Does this get deleted as we make our primary interface for initial experiments callback based?

punchcut/src/source/glyf/outline.rs
dfrg: A lot of context is lost with hinting and variations being missing. The tags array is also used during delta computation and hinting and is temporarily associated with other point arrays (original, unscaled, deltas) during those processes.

punchcut/src/source/glyf/outline.rs
dfrg: I'd rather not have dynamic allocations embedded in an enum. One change I had considered here is to always return F26Dot6 values since the precision is more than enough to handle FWord and loading glyphs without a scale is likely fairly rare.

rsheeter (approved):
LGTM overall.

I think three key things need to be addressed, either here or in follow-ons:

1. Testing is very light, perhaps it could become data driven with instructions on how to regenerate or extend the data?
1. Commenting is very light; imagine you want a competent coder who isn't a font expert to be able to have a clue
   * Also please remove pointless comments. For example, `hints: bool` doesn't benefit from a comment saying True if hints enabled.
   * That is, I'm not looking for pointless comment every field stuff, I want the core logic and references to be captured where not obvious
1. Everything related to hinting needs to be either feature gated or deleted

If you will do follow-on PRs please file issues so we don't forget.

Edit: also apologies for taking so long to get to reviewing this.

punchcut/src/source/glyf/outline.rs
dfrg: This is actually a quadratic. The terminology was taken from FT_Outline_Decompose. I'll rename it.

punchcut/src/source/glyf/scaler.rs
dfrg: Happy to newtype, but it needs to be flexible. We don't want to be responsible for generation or management of unique font ids. I mentioned above that I tried that with swash and it ended up being a huge source of friction for just about everybody trying to work with it (including me :)

punchcut/src/source/glyf/scaler.rs
dfrg: Will make this more clear.

punchcut/src/source/glyf/scaler.rs
dfrg: We can hide the `glyf` module but this won't go away. TrueType can't be fully implemented without such data structures. The top level interface is now callback based and can be used for both TrueType and CFF.

punchcut/src/lib.rs
dfrg: Don't mind adding `Debug` but it'll likely print a massive amount of nested data, particularly with the hinting stuff. `Clone` is also fine but I'm having a hard time thinking of a reasonable use case for it.

punchcut/src/error.rs
dfrg: It doesn't currently expose any details about what went wrong. That's been a todo of mine for a while.

dfrg:
Thanks for the review. 

I'd rather address (at least the bulk of) the issues here and ping for follow ups rather than merging and letting them linger.

punchcut/src/scaler.rs
dfrg: Thanks for doing the work for me. Done :)

dfrg:
Addressed most of the low-hanging fruit, plus removed the peniko/kurbo dependencies and feature gated all of the hinting code. I'll look at the rest tomorrow (today?)

dfrg:
Opened issue #203 to discuss better testing wrt comparison with FreeType.

punchcut/src/source/glyf/math.rs
dfrg: The internal code now uses a new `F26Dot6` type in font-types and this module has been removed entirely. The small bits of single use code have been moved inline and commented.

punchcut/src/source/glyf/outline.rs
dfrg: The points field is now `Vec<F26Dot6>` regardless of scaling mode and the `is_scaled` field has been removed.

punchcut/src/lib.rs
dfrg: I added a python script for this to `resources/test_fonts/extract_glyphs.py` and updated `rebuild.sh` to call it. This will generate `extracted/<font-name>-glyphs.text` for each font. I documented the format in the [readme](https://github.com/dfrg/fontations/blob/load-glyphs/resources/test_fonts/README.md).

This should make it fairly simple to add new fonts for testing.

punchcut/src/source/glyf.rs
dfrg: I added some infra for automatically extracting the test data as mentioned above. Opened #204 to track adding ttx files to cover more/interesting cases.

dfrg:
> 1. Testing is very light, perhaps it could become data driven with instructions on how to regenerate or extend the data?
> 2. Commenting is very light; imagine you want a competent coder who isn't a font expert to be able to have a clue
> 3. Everything related to hinting needs to be either feature gated or deleted

Status now: 

1. Partially done in that new testing infra is in place including automatic extraction of the outline data. Test cases are still sparse and I've decided to defer that with #204 opened to track.
2. Still pending. I'll spend some time on more complete documentation in the next few days.
3. Done

dfrg:
We've agreed that now is a good time to merge this. Thanks to everyone for both detailed reviews and patience. The code has been substantially improved as a result.

There are some minor remaining adjustments needed which I've written up in #209 to track.

PR #193, opened 2023-01-05
Bump versions to 0.0.5

cmyr:


dfrg (approved):
Are we skipping 0.0.4 for font-types in an effort to keep all versions in sync moving forward?

cmyr:
Yea, I think if a new version actually requires a release of all the crates it makes sense to sync up the versions?

PR #198, opened 2023-01-12
Remove bitflags dependency

### LINKED ISSUE #197: Remove bitflags dependency
In https://github.com/googlefonts/fontations/pull/116#issuecomment-1304644307 use it until we have a good reason not to. We now have a good reason not to: it's not present in Chromium at time of writing.

cmyr:
With this patch, we no longer rely on bitflags, and instead manually implement flag-like behaviour for our flag-like types.

This removes our own external dependency, and also lets us delete a bunch of special-cased formatting code.

closes #197


dfrg (approved):
Looks good! Given that this removes some unfortunate formatting code and make us free of all external dependencies on the read side, I'm inclined to suggest we merge this.

rsheeter (approved):
LGTM, +1 to merge

PR #199, opened 2023-01-12
Small changes to fixed point types and CharIter

dfrg:
This adds `to_bits` and `from_bits` conversions to the fixed point types. We can work toward better typing in punchcut, but these feel reasonable and will alleviate some of the pain of roundtripping through bytes in meantime. This also adds `Fixed::to_f2dot14`.

Additionally, derives `Clone` for `name::CharIter` in the spirit of making iterators clone where possible. I ran into this when working on localized string support in the new font metadata library.

Really small PR, so feel free to merge.

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -174,13 +174,13 @@ crate::newtype_scalar!(Fixed, [u8; 4]);
 
 impl Fixed {
     /// Converts a 16.16 to 2.14 fixed point value.
+    ///
+    /// This specific conversion is defined by the spec:
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#coordinate-scales-and-normalization>
+    ///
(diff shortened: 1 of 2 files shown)


font-types/src/fixed.rs
```diff
@@ -162,9 +172,22 @@ float_conv!(Fixed, to_f64, from_f64, f64);
...
 
+impl Fixed {
+    /// Converts a 16.16 to 2.14 fixed point value.
+    pub fn to_f2dot14(self) -> F2Dot14 {
+        // This specific conversion is defined by the spec:
+        //
+        // "5. Convert the final, normalized 16.16 coordinate value to 2.14 by this method: add 0x00000002,
+        // and sign-extend shift to the right by 2."
+        //
+        // https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#coordinate-scales-and-normalization
```
cmyr, line 183: I would put this in the documentation

cmyr (approved):
One nit inline, looks good!

PR #214, opened 2023-01-30
Tuple variation store (gvar)

cmyr:
This patch adds support for the gvar table, including parsing tuple variations.

There's a bunch of hand-written code in here, and I suspect there will be some refactoring involved when we add cvar support. It's also possible that more of this can be codegened, but in the end I went with the expedient approach and just tried to get something working.

I also think this API will need some tweaks and additions. Currently I just expose raw iterators over each set of tuples for a given glyph, and then each delta in a given tuple; the work of determining what tuples to apply for a given position in d-space, (as well as scaling deltas) is not implemented here (although I did port chad's code that computes the scalars, at least).


write-fonts/generated/generated_variations.rs
```diff
@@ -7,6 +7,135 @@ use crate::codegen_prelude::*;
...
+        self.values.write_into(writer);
+    }
+}
+
+impl Validate for Tuple {
+    fn validate_impl(&self, ctx: &mut ValidationCtx) {
+        ctx.in_table("Tuple", |ctx| {
+            ctx.in_field("values", |ctx| {
+                if self.values.len() > (u16::MAX as usize) {
+                    ctx.report("array excedes max length");
```
behdad, line 124: Typo

read-fonts/src/tables/variations.rs
```diff
@@ -11,6 +11,479 @@ pub struct DeltaSetIndex {
...
+    //0x1000	Reserved	Reserved for future use — set to 0.
+    //
+    /// Mask for the low 12 bits to give the shared tuple records index.
+    const TUPLE_INDEX_MASK: u16 = 0x0FFF;
+
+    fn tuple_len(self, axis_count: u16, flag: usize) -> usize {
+        match flag {
+            0 => self.embedded_peak_tuple(),
+            1 => self.intermediate_region(),
+            _ => panic!("only 0 or 1 allowed here"),
```
behdad, line 48: I'm not sure if the flag comes from the font or user. If from the font, then panic is not desired. If from the user, maybe an enum is better? @rsheeter

cmyr: This flag comes from the codegen input, and the codegen tries to use a language-agnostic schema. In this particular corner of codegen, (when calculating a count) the only inputs that can be provided are unsigned integers, so I'm overloading that in this particular instance to represent an enum. 😒

write-fonts/generated/generated_variations.rs
cmyr: #216

read-fonts/src/tables/variations.rs
```diff
@@ -11,6 +11,479 @@ pub struct DeltaSetIndex {
...
+        let tuple_byte_len = F2Dot14::RAW_BYTE_LEN * axis_count as usize;
+        let index = self.tuple_index();
+        FIXED_LEN
+            + index
+                .embedded_peak_tuple()
+                .then_some(tuple_byte_len)
+                .unwrap_or_default()
+            + index
+                .intermediate_region()
+                .then_some(tuple_byte_len)
```
dfrg, line 186:
```suggestion
                .then_some(tuple_byte_len * 2)
```
Intermediate regions are defined by two tuples. This took me a bit of time to find :)

read-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,489 @@
...
+#[derive(Clone)]
+pub struct TupleVariation<'a> {
+    axis_count: u16,
+    header: TupleVariationHeader<'a>,
+    shared_tuples: SharedTuples<'a>,
+    packed_deltas: PackedDeltas<'a>,
+    point_numbers: PackedPointNumbers<'a>,
+}
+
+impl<'a> TupleVariation<'a> {
```
dfrg, line 188:
```suggestion
impl<'a> TupleVariation<'a> {
    /// Returns true if this tuple provides deltas for all points in a glyph.
    pub fn all_points(&self) -> bool {
        self.point_numbers.count() == 0
    }
    
```
Add a method to detect if a tuple provides deltas for all points. This allows us to skip the interpolation code path for these tuples which is a pretty big performance win.

read-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,489 @@
...
+                }
+            } else {
+                if coord < peak.min(ZERO) || coord > peak.max(ZERO) {
+                    return None;
+                }
+                scalar = scalar * coord / peak;
+            }
+        }
+        Some(scalar)
+    }
```
dfrg, line 242:
```suggestion
    pub fn compute_scalar(&self, coords: &[F2Dot14]) -> Option<Fixed> {
        const ZERO: Fixed = Fixed::ZERO;
        let mut scalar = Fixed::ONE;
        let peak = self.peak();
        let inter_start = self.header.intermediate_start_tuple();
        let inter_end = self.header.intermediate_end_tuple();
        if peak.len() != self.axis_count as usize {
            return None;
        }

        for i in 0..self.axis_count {
            let i = i as usize;
            let coord = coords.get(i).copied().unwrap_or_default().to_fixed();
            let peak = peak.get(i).unwrap_or_default().to_fixed();
            if peak == ZERO || peak == coord {
                continue;
            }

            if coord == ZERO {
                return None;
            }

            if let (Some(inter_start), Some(inter_end)) = (&inter_start, &inter_end) {
                let start = inter_start.get(i).unwrap_or_default().to_fixed();
                let end = inter_end.get(i).unwrap_or_default().to_fixed();
                if coord <= start || coord >= end {
                    return None;
                }
                if coord < peak {
                    scalar = scalar.mul_div(coord - start, peak - start);
                } else {
                    scalar = scalar.mul_div(end - coord, end - peak);
                }
            } else {
                if coord < peak.min(ZERO) || coord > peak.max(ZERO) {
                    return None;
                }
                scalar = scalar.mul_div(coord, peak);
            }
        }
        Some(scalar)
    }
```
Sorry for the length. This does the following:

* changes the `coords` parameter to `&[F2Dot14]`  since that is the expected type for normalized coordinates
* modifies the loop to iterate over all axes regardless of the length of coords
* replaces multiply+divide with use of the `mul_div` function which gives slightly different results but is required to match FreeType

dfrg (approved):
This is a huge chunk of work and it's all beautifully done. I have a small diff that fixes one tiny bug and makes some API changes. I left the entirety of it in a few inline suggestions that you can feel free to commit. Otherwise, I'll send a PR after this lands.

Either way, this looks good to me!

read-fonts/src/tables/gvar.rs
cmyr: I have a question about this bit:

```rust
        for i in 0..self.axis_count {
            let i = i as usize;
            let coord = coords.get(i).copied().unwrap_or_default().to_fixed();
```

is this about ensuring we do something sane if `axis_count` and `coords.len()` differ? Should they ever differ? If not I would prefer just checking that they match before we enter the loop, and returning `None` if they dont?

dfrg: The reasoning here is that you can pass in a coord array that is smaller than the axis count and it just "does the right thing" in assuming all missing coords are at their default positions in design space.

read-fonts/src/tables/variations.rs
```diff
@@ -11,6 +11,479 @@ pub struct DeltaSetIndex {
...
+            Tuple {
+                values: self.data.read_array(range).unwrap(),
+            }
+        })
+    }
+
+    /// Compute the actual length of this table in bytes
+    fn byte_len(&self, axis_count: u16) -> usize {
+        const FIXED_LEN: usize = u16::RAW_BYTE_LEN + TupleIndex::RAW_BYTE_LEN;
+        let tuple_byte_len = F2Dot14::RAW_BYTE_LEN * axis_count as usize;
```
cmyr, line 177:
```suggestion
        let tuple_byte_len = F2Dot14::RAW_BYTE_LEN * 2 * axis_count as usize;
```

read-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,489 @@
...
+            .tuple_index()
+            .tuple_records_index()
+            .and_then(|idx| self.shared_tuples.tuples().get(idx as usize).ok())
+            .or_else(|| self.header.peak_tuple())
+            .unwrap_or_default()
+    }
+
+    // transcribed from pinot/moscato
+    /// Compute the scalar for a this tuple at a given point in design space.
+    ///
```
cmyr, line 201:
```suggestion
    ///
    /// The `coords` slice must be of lesser or equal length to the number of axes.
    /// If it is less, missing (trailing) axes will be assumed to have zero values.
    ///
```

PR #215, opened 2023-01-30
[codegen] Add #[derive_traits(..)] attribute

### LINKED ISSUE #186: Allow specifying additional derived traits in codegen
It would be nice to be able to say, for instance, that you would like to derive `Hash` or `PartialOrd` for some record. We could then be somewhat judicious about this, since not all traits *can* be implemented for all types, and not all traits may be desirable for all types.

cmyr:
This was an old todo that I hit again when working on gvar; sometimes we want some particular table/record to implement a specific trait, on a case-by-case basis. This adds support for a new table-level attribute where you can provide a list of traits that will be derived for the type.

While getting this to work I ran into a few existing cases where we were not deriving traits that we could be, so I added those as well (for instance, we will now derive Ord/PartialOrd for generated flags as well as BigEndian<T>).

closes #186

### DIFF
--- a/write-fonts/generated/generated_name.rs
+++ b/write-fonts/generated/generated_name.rs
@@ -134,7 +134,7 @@ impl FromObjRef<read_fonts::tables::name::LangTagRecord> for LangTagRecord {
 }
 
 ///[Name Records](https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-records)
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
 pub struct NameRecord {
     /// Platform ID.
     pub platform_id: u16,
--- a/write-fonts/generated/generated_name.rs
+++ b/write-fonts/generated/generated_name.rs
@@ -134,7 +134,7 @@ impl FromObjRef<read_fonts::tables::name::LangTagRecord> for LangTagRecord {
 }
 
 ///[Name Records](https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-records)
-#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Debug, Default, PartialOrd, Ord, PartialEq, Eq)]
 pub struct NameRecord {
     /// Platform ID.
     pub platform_id: u16,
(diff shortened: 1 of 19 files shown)


write-fonts/generated/generated_name.rs
```diff
@@ -127,7 +127,7 @@ impl FromObjRef<read_fonts::tables::name::LangTagRecord> for LangTagRecord {
 }
 
 ///[Name Records](https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-records)
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
```
rsheeter, line 130: The terminology here is exceptionally and arguably unecessarily Rust-focused while the underlying concepts are not. Feels like an unnecessary step backwards from a general schema. For example, why not `#[capabilities(hash, compare, sort)]`

rsheeter (approved):
LGTM other than mild concern over unnecessarily Rust-specific terminology

write-fonts/generated/generated_name.rs
rsheeter: ^ refers to the `#[derive_traits(...)]` in the codegen inputs, apparently I clicked the wrong file

cmyr: Hmm. I think this is another example of a situation where the annotations in codegen are not always necessarily some language-agnostic thing, but are sometimes very specifically instructions for what code we want to generate for this (rust) project. In this case, the idea that these types need to be (e.g.) hashable isn't really a matter of the spec, but more a consequence of some consuming code somewhere, where I have decided to store something in a hashmap; that may not be the case in another implementation.

Similarly, an attempt to generalize this idea beyond rust is also tricky. Does `compare` mean `PartialEq + Eq`, or will `PartialEq` suffice?

There's a reasonable argument that we should just derive these traits for every type; and the reason I would want to avoid that is just that we have a ton of types, and so deriving a bunch of traits for all of them causes us to generate a bunch of extra code. Maybe this is fine?

rsheeter: My naive thought would be capabilities:compare means PartialEq + Eq except where infeasible because the type contains floats.

cmyr:
okay several months later I have updated this to use generic language.

PR #217, opened 2023-01-31
Initial simple glyph compilation

cmyr:
This can naively compile non-component, non-variable glyphs.

Very much a checkpoint, but I think it makes sense to iterate in-repo for this work?

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -133,6 +133,9 @@ impl<'a> SimpleGlyph<'a> {
 }
 
 /// Point in a simple glyph.
+///
+/// This is the representation used by punchcut, where we use higher precision
+/// and use both integer and fixed-point representations
 #[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
 pub struct Point<T> {
     /// X cooordinate.
@@ -149,12 +152,33 @@ impl<T> Point<T> {
 }
 
 /// Point with an associated on-curve flag in a simple glyph.
-#[derive(Clone, Copy, Debug)]
+///
+/// This type is a simpler representation of the data in the blob.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CurvePoint {
-    /// The point.
-    pub point: Point<i32>,
+    /// X cooordinate.
+    pub x: i16,
+    /// Y cooordinate.
+    pub y: i16,
     /// True if this is an on-curve point.
     pub on_curve: bool,
+}
+
+impl CurvePoint {
+    /// Construct a new `CurvePoint`
+    pub fn new(x: i16, y: i16, on_curve: bool) -> Self {
+        Self { x, y, on_curve }
+    }
+
+    /// Convenience method to construct an on-curve point
+    pub fn on_curve(x: i16, y: i16) -> Self {
+        Self::new(x, y, true)
+    }
+
+    /// Convenience method to construct an off-curve point
+    pub fn off_curve(x: i16, y: i16) -> Self {
+        Self::new(x, y, false)
+    }
 }
 
 #[derive(Clone)]
@@ -173,13 +197,8 @@ impl<'a> Iterator for PointIter<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         self.advance_flags()?;
         self.advance_points();
-        Some(CurvePoint {
-            point: Point {
-                x: self.cur_x as i32,
-                y: self.cur_y as i32,
-            },
-            on_curve: self.cur_flags.contains(SimpleGlyphFlags::ON_CURVE_POINT),
-        })
+        let is_on_curve = self.cur_flags.contains(SimpleGlyphFlags::ON_CURVE_POINT);
+        Some(CurvePoint::new(self.cur_x, self.cur_y, is_on_curve))
     }
 }
 
@@ -518,7 +537,7 @@ mod tests {
         assert_eq!(
             simple_glyph
                 .points()
-                .map(|pt| (pt.point.x, pt.point.y, pt.on_curve))
+                .map(|pt| (pt.x, pt.y, pt.on_curve))
                 .collect::<Vec<_>>(),
             &[
                 (5, 0, true),
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -1,0 +1,298 @@
+//! The [glyf (Glyph Data)](https://docs.microsoft.com/en-us/typography/opentype/spec/glyf) table
+
+use kurbo::{BezPath, Shape};
+
+use read_fonts::tables::glyf::{CurvePoint, SimpleGlyphFlags};
+
+use crate::FontWrite;
+
+/// A single quadratic bezier contour
+#[derive(Clone, Debug)]
+pub struct Contour(Vec<CurvePoint>);
+
+/// A simple (without components) glyph
+pub struct SimpleGlyf {
+    x_max: i16,
+    y_max: i16,
+    x_min: i16,
+    y_min: i16,
+    contours: Vec<Contour>,
+    _instructions: Vec<u8>,
+}
+
+/// An error if an input curve is malformed
+#[derive(Clone, Debug)]
+pub enum BadKurbo {
+    HasCubic,
+    TooSmall,
+    MissingMove,
+}
+
+/// A helper trait for converting other point types to open-type compatible reprs
+pub trait OtPoint {
+    fn get(self) -> (i16, i16);
+}
+
+impl OtPoint for kurbo::Point {
+    fn get(self) -> (i16, i16) {
+        (ot_round(self.x as f32), ot_round(self.y as f32))
+    }
+}
+
+impl OtPoint for (i16, i16) {
+    fn get(self) -> (i16, i16) {
+        self
+    }
+}
+
+// adapted from simon:
+// https://github.com/simoncozens/rust-font-tools/blob/105436d3a617ddbebd25f790b041ff506bd90d44/otmath/src/lib.rs#L17
+fn ot_round(val: f32) -> i16 {
+    (val + 0.5).floor() as i16
+}
+
+impl Contour {
+    /// Create a new contour begining at the provided point
+    pub fn new(pt: impl OtPoint) -> Self {
+        let (x, y) = pt.get();
+        Self(vec![CurvePoint::on_curve(x, y)])
+    }
+
+    /// The total number of points in this contour
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// `true` if this contour is empty
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Add a line segment
+    pub fn line_to(&mut self, pt: impl OtPoint) {
+        let (x, y) = pt.get();
+        self.0.push(CurvePoint::on_curve(x, y));
+    }
+
+    /// Add a quadratic curve segment
+    pub fn quad_to(&mut self, p0: impl OtPoint, p1: impl OtPoint) {
+        let (x0, y0) = p0.get();
+        let (x1, y1) = p1.get();
+        self.0.push(CurvePoint::off_curve(x0, y0));
+        self.0.push(CurvePoint::on_curve(x1, y1));
+    }
+}
+
+impl SimpleGlyf {
+    /// Attempt to create a simple glyph from a kurbo `BezPath`
+    ///
+    /// The path may contain only line and quadratic bezier segments. The caller
+    /// is responsible for converting any cubic segments to quadratics before
+    /// calling.
+    ///
+    /// Returns an error if the input path is malformed; that is, if it is empty,
+    /// contains cubic segments, or does not begin with a 'move' instruction.
+    //TODO: figure out a more general API? maybe a nested builder thing, where you
+    //build contours, and from those contours build a glyph? idk?
+    pub fn from_kurbo(path: &BezPath) -> Result<Self, BadKurbo> {
+        let mut contours = Vec::new();
+        let mut current = None;
+
+        for el in path.elements() {
+            match el {
+                kurbo::PathEl::MoveTo(pt) => {
+                    if let Some(prev) = current.take() {
+                        contours.push(prev);
+                    }
+                    current = Some(Contour::new(*pt));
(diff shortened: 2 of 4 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// Add a line segment
+    pub fn line_to(&mut self, pt: impl OtPoint) {
+        let (x, y) = pt.get();
+        self.0.push(CurvePoint::on_curve(x, y));
+    }
+
+    /// Add a quadratic curve segment
+    pub fn curve_to(&mut self, p0: impl OtPoint, p1: impl OtPoint) {
```
behdad, line 78: Maybe rename? Typically `curve_to` means a cubic.

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+    ///
+    /// The path may contain only line and quadratic bezier segments. The caller
+    /// is responsible for converting any cubic segments to quadratics before
+    /// calling.
+    ///
+    /// Returns an error if the input path is malformed; that is, if it is empty,
+    /// contains cubic segments, or does not begin with a 'move' instruction.
+    //TODO: figure out a more general API? maybe a nested builder thing, where you
+    //build contours, and from those contours build a glyph? idk?
+    pub fn from_kurbo(path: &BezPath) -> Result<Self, BadKurbo> {
```
dfrg, line 97: Should we take `&impl Shape` here instead? I suppose it doesn’t matter for our immediate use case but makes the API more general and doesn’t have any additional cost.

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+                    current = Some(Contour::new(*pt));
+                }
+                kurbo::PathEl::LineTo(pt) => {
+                    current.as_mut().ok_or(BadKurbo::MissingMove)?.line_to(*pt)
+                }
+                kurbo::PathEl::QuadTo(p0, p1) => current
+                    .as_mut()
+                    .ok_or(BadKurbo::MissingMove)?
+                    .curve_to(*p0, *p1),
+                kurbo::PathEl::CurveTo(_, _, _) => return Err(BadKurbo::HasCubic),
```
dfrg, line 116: We could lower to quads here. I’m not sure what the implications might be.

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+                // I think we can just ignore this, and remove duplicate points
```
dfrg, line 117: All contours are implicitly closed so this seems reasonable.

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+                // at the end?
+                kurbo::PathEl::ClosePath => (),
+            }
+        }
+
+        contours.extend(current);
+
+        for contour in &mut contours {
+            //TODO: verify that single-point contours are actually meaningless?
```
dfrg, line 126: I recall something about fonts with `sbix` tables using single point contours to represent image offsets. @behdad might have more info on this.

dfrg (approved):
This looks good to me! Might make sense for @anthrotype to glance at the delta/packing code

write-fonts/src/tables/glyf.rs
cmyr: fair enough :)

write-fonts/src/tables/glyf.rs
cmyr: yea, I wanted to leave the conversion out of it since it might get complicated with variable fonts; the premise here is that you've already done your conversion and verified that it's sane etc.

write-fonts/src/tables/glyf.rs
cmyr: I think we should take *something*, but i'm really not sure what? I think a bunch of these decisions will depend on how we're representing variable glyphs.

dfrg: Makes sense.

write-fonts/src/tables/glyf.rs
behdad: They use two points, to represent the glyph bounding box.

write-fonts/src/tables/glyf.rs
rsheeter: +1 to leave it to higher levels. For example I would expect the compiler to  look at every instance of a contour and safely lower it while retaining interpolation compatibility.

read-fonts/src/tables/glyf.rs
```diff
@@ -149,14 +152,35 @@ impl<T> Point<T> {
...
-#[derive(Clone, Copy, Debug)]
+///
+/// This type is a simpler representation of the data in the blob.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct CurvePoint {
-    /// The point.
-    pub point: Point<i32>,
+    /// X cooordinate.
+    pub x: i16,
+    /// Y cooordinate.
```
rsheeter, line 161: I'm not sure commenting Point.y as "y coordinate" is anything more than noise

rsheeter: I'm seeing similar noise-level comments quite a bit.

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn get(self) -> (i16, i16) {
+        self
+    }
+}
+
+// adapted from simon:
+// https://github.com/simoncozens/rust-font-tools/blob/105436d3a617ddbebd25f790b041ff506bd90d44/otmath/src/lib.rs#L17
+fn ot_round(val: f32) -> i16 {
+    (val + 0.5).floor() as i16
+}
```
rsheeter, line 52: Is glyf the only place that performs ot-round's? If not perhaps shift to a helper library/common location?

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
+//! The [glyf (Glyph Data)](https://docs.microsoft.com/en-us/typography/opentype/spec/glyf) table
+
+use kurbo::{BezPath, Shape};
+
+use read_fonts::tables::glyf::{CurvePoint, SimpleGlyphFlags};
+
+use crate::FontWrite;
+
+/// A single quadratic bezier contour
```
rsheeter, line 9: Looking at the definition I think you mean it won't contain any curve of order 3 or higher. That is, there will be no sequence of more than 2 off-curve points between on-curves. The descritpion made me think it would exclusively contain quads but you appear to allow lines as well.

Perhaps the comment could be made more specific?

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+impl SimpleGlyf {
+    /// Attempt to create a simple glyph from a kurbo `BezPath`
+    ///
+    /// The path may contain only line and quadratic bezier segments. The caller
+    /// is responsible for converting any cubic segments to quadratics before
+    /// calling.
+    ///
+    /// Returns an error if the input path is malformed; that is, if it is empty,
+    /// contains cubic segments, or does not begin with a 'move' instruction.
+    //TODO: figure out a more general API? maybe a nested builder thing, where you
```
rsheeter, line 95: I suggest you move your TODO into an issue

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+                        contours.push(prev);
+                    }
+                    current = Some(Contour::new(*pt));
+                }
+                kurbo::PathEl::LineTo(pt) => {
+                    current.as_mut().ok_or(BadKurbo::MissingMove)?.line_to(*pt)
+                }
+                kurbo::PathEl::QuadTo(p0, p1) => current
+                    .as_mut()
+                    .ok_or(BadKurbo::MissingMove)?
```
rsheeter, line 114: Can you hoist the repetitive current.as_mut().ok_or(missing move)? to a binding outside the match? - you'd be checking it for ClosePath too but close doesn't make a whole lot of sense without ever starting

write-fonts/src/tables/glyf.rs
rsheeter: Would that duplicate point removal ever make something meant to be interpolation friendly not be?

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+        let mut cur = 0;
+        for contour in &self.contours {
+            cur += contour.len();
+            (cur as u16 - 1).write_into(writer);
+        }
+        (self._instructions.len() as u16).write_into(writer);
+        self._instructions.write_into(writer);
+
+        let deltas = self.compute_point_deltas().collect::<Vec<_>>();
+        //TODO: calculate flag repeats here
```
rsheeter, line 253: Suggest filing an issue unless you are going to more or less immediately address

write-fonts/src/tables/glyf.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use read::{FontData, FontRead};
+
+    use super::*;
+
+    #[test]
```
rsheeter, line 274: Feels light on testing. 

Thinking "aloud," can we read a glyph from one of our binary test fonts, write it out, and get the same bytes?

Also, you have various error checks in the code, and no tests with of bad input.

rsheeter:
> I think it makes sense to iterate in-repo for this work?

+1. What do you see as next steps?

write-fonts/src/tables/glyf.rs
cmyr: It shouldn't; it is addressing the fact that in ot all contours have the same start/end point, but in kurbo a bezpath is composed of segments, and if you want a 'closed' contour you generally finish with a segment that has an end point equal to your starting point.

write-fonts/src/tables/glyf.rs
cmyr: yea maybe this moves to font-types eventually?

write-fonts/src/tables/glyf.rs
cmyr: yes, there is a follow-up PR coming.

write-fonts/src/tables/glyf.rs
cmyr: I'm not sure exactly what you mean? I can't make it a method, since `current` is an `Option`, and I can't add methods to types i don't define, so this is the cleanest approach I can think of.

about `ClosePath`, I think we just don't really worry about that; it isn't semantically meaningful for us, although we might want to verify when we sanitize our input that if there is a `ClosePath` element, the last path segment ends at the first point in the path.

write-fonts/src/tables/glyf.rs
anthrotype: Rod actually raises a valid point here. I haven't checked this code too deeply, but I recall in fonttools we had a similar issue in the converter pen that translates from point-wise to segment-wise representation of a path, and we needed to special case the situation when the last segment of a closed contour is a "line" and its point overlaps with the contour's "move" point: https://github.com/fonttools/fonttools/pull/1720/

In the sources, which contain list of points, as well as in the output glyf table, which also contains only points, it's clear and non ambiguous whether there's or not a duplicate point at the end/start of the contour; when translating points to segment and back to points it's important to not lose this duplicate points, because they may be intentional to support clever interpolation of variable contours (e.g. where a line segment "disappears" in some masters)

write-fonts/src/tables/glyf.rs
cmyr: yes good points, see #219 :)

read-fonts/src/tables/glyf.rs
dfrg: Agree with Colin that missing public docs look weird and inconsistent. I also think it gives an impression that the library is not well maintained. FWIW, the standard library also has many cases of “noise” doc comments: https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.new

write-fonts/src/tables/glyf.rs
rsheeter: Move it up a level, e.g. to outside the match.

```rust
for el in path.elements() {
  let current = current.as_mut().ok_or(BadKurbo::MissingMove)?;
  match el { ... use current here }
```

write-fonts/src/tables/glyf.rs
rsheeter: Being meaningless and being an error condition aren't quite the same thing. Are we sure this should be an *error*?

write-fonts/src/tables/glyf.rs
cmyr: that would error immediately, since we don't populate `current` until we actually see a match statement.

rsheeter: Sorry, missed that when reading in gh diff view. Still feels repetitive/noisy to me. What if ... current was a wee struct that managed current/prev and exposed things like move_to/line_to/etc?

A ContourBuilder is what I'm imagining I think.

```rust
let mut builder = ContourBuilder::new();

for el in path.elements() {
  kurbo::PathEl::MoveTo => {
    // move_to initiates a a new contour and returns Option<Contour> for any accumulated content
    if let Some(contour) = builder.move_to(pt) {
      contours.push(contour)
    }
  },
  kurbo::PathEl::LineTo => builder.line_to(pt)?,
  // etc
}
```

write-fonts/src/tables/glyf.rs
rsheeter: I suggest we follow the lead of https://github.com/googlefonts/fontations/pull/221 and move it to font-types nowish?

write-fonts/src/tables/glyf.rs
rsheeter: I think we'll find a builder fits fontmake-rs quite well too. When I have some free time I'll try to hack together a poc of fontmake-rs using such an API that we can look at.

write-fonts/src/tables/glyf.rs
rsheeter: To give a simple example, consider a box that becomes a triangle by moving two corners to a point in between. Those points - if at the end - are now duplicate for the triangle case but not the square. If you drop them you'd make paths that used to interpolate not do so anymore.

PR #218, opened 2023-02-02
Compute repeating flags when compiling glyphs

cmyr:
I'd left this as a TODO in the first patch.

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -222,6 +222,57 @@ impl FontWrite for CoordDelta {
     }
 }
 
+/// A little helper for writing flags that may have a 'repeat' byte
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct RepeatableFlag {
+    flag: SimpleGlyphFlags,
+    repeat: u8,
+}
+
+impl FontWrite for RepeatableFlag {
+    fn write_into(&self, writer: &mut crate::TableWriter) {
+        debug_assert_eq!(
+            self.flag.contains(SimpleGlyphFlags::REPEAT_FLAG),
+            self.repeat > 0
+        );
+
+        self.flag.bits().write_into(writer);
+        if self.flag.contains(SimpleGlyphFlags::REPEAT_FLAG) {
+            self.repeat.write_into(writer);
+        }
+    }
+}
+
+impl RepeatableFlag {
+    /// given an iterator over raw flags, return an iterator over flags + repeat values
+    // writing this as an iterator instead of just returning a vec is very marginal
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -222,6 +222,57 @@ impl FontWrite for CoordDelta {
...
+        flags: impl IntoIterator<Item = SimpleGlyphFlags>,
+    ) -> impl Iterator<Item = RepeatableFlag> {
+        let mut iter = flags.into_iter();
+        let mut prev = None;
+
+        std::iter::from_fn(move || loop {
+            match (iter.next(), prev.take()) {
+                (None, prev) => return prev,
+                (Some(flag), None) => prev = Some(RepeatableFlag { flag, repeat: 0 }),
+                (Some(flag), Some(mut last)) => {
```
dfrg, line 260: The inherent logic for this is pretty nasty but this solution captures it nicely!

dfrg (approved):
Lgtm!

write-fonts/src/tables/glyf.rs
```diff
@@ -295,4 +344,81 @@ mod tests {
...
+        assert_eq!(read.end_pts_of_contours(), &[3]);
+        let points = read.points().collect::<Vec<_>>();
+        assert_eq!(points[0].x, 20);
+        assert_eq!(points[0].y, -100);
+        assert_eq!(points[1].x, 25);
+        assert_eq!(points[1].y, -90);
+        assert_eq!(points[2].x, 50);
+        assert_eq!(points[2].y, -69);
+        assert_eq!(points[3].x, 80);
+        assert_eq!(points[3].y, -20);
```
rsheeter, line 380: I'm curious why we don't assert `read.points().collect::<Vec<_>>()` equals `vec![Point(..), etc]`?

write-fonts/src/tables/glyf.rs
rsheeter: Ew. Also +1.

write-fonts/src/tables/glyf.rs
```diff
@@ -295,4 +344,81 @@ mod tests {
...
         assert!(points[4].on_curve);
     }
+
+    #[test]
+    fn compile_repeatable_flags() {
+        let mut path = BezPath::new();
+        path.move_to((20., -100.));
+        path.line_to((80., -20.));
+        path.line_to((50., -69.));
+        path.line_to((25., -90.));
```
rsheeter, line 354: Optional, BezPath::from_svg("M20,-100 L80,-20 etc") might be a more concise way to write this sort of test input?

PR #221, opened 2023-02-02
Move Point to font-types and add math ops

dfrg:
This moves the `Point<T>` type from `read_fonts::tables::glyf` to `font-types` and adds the basic per component math operators as well as multiplication and division by scalars.

Also adds a few additional conversions for the fixed point types.


font-types/src/point.rs
```diff
@@ -0,0 +1,244 @@
+use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+
+/// Two dimensional point with a generic coordinate type.
+#[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
+pub struct Point<T> {
+    /// X cooordinate.
```
cmyr, line 6:
```suggestion
    /// X coordinate.
```

cmyr (approved):
Looks good, modulo one typo (which I think I probably copy-pasted somewhere else a while ago)

PR #223, opened 2023-02-02
compiling composite glyphs

cmyr:
This is really just the basic implementation code; as with #217, we have to figure out exactly how we're going to feed our input into this.

This also has one major limitation: it does not have access to the paths of the component glyphs, which means it cannot compute the bounding boxes for those glyphs. This will be something else to resolve when we figure out what our API is actually going to look like.

Based off of #219, which should go in first.


write-fonts/src/tables/glyf.rs
```diff
@@ -25,6 +28,48 @@ pub struct SimpleGlyph {
...
+}
+
+/// A single component glyph (part of a [`CompositeGlyph`]).
+#[derive(Clone, Debug)]
+pub struct Component {
+    glyph: GlyphId,
+    anchor: Anchor,
+    pub flags: ComponentFlags,
+    transform: Transform,
+}
```
rsheeter, line 46: For reference, at time of writing fontmake-rs knows the glyph name - and could doubtless come up with a gid - and the transform. https://github.com/googlefonts/fontmake-rs/blob/main/fontir/src/ir.rs.

rsheeter:
> cannot compute the bounding boxes for those glyphs

I suggest we accept this limitation for the moment and ask the client to provide, e.g. `add_component(component, bbox)`. fontmake-rs would then set components of the glyph G as dependencies of G so at the time we compute G we have at least the IR (or even the binary?) and from that can obtain bboxes

 We might evolve away from that but I believe that would be sufficient to get started. @anthrotype does that sound remotely sane?

write-fonts/src/write.rs
```diff
@@ -145,6 +145,16 @@ impl TableWriter {
         data.add_offset(obj_id, width, self.offset_adjustment);
     }
 
+    /// Add a padding byte of necessary to ensure the table length is word-aligned.
+    ///
+    /// This is necessary for things like the glyph table, which require offsets
+    /// to be aligned on 2-byte boundaries.
+    pub fn ensure_word_aligned(&mut self) {
```
rsheeter, line 152: "word" feels a little ambiguous to me here. wdyt about something like

```
/// Adds 0 bytes until the table length is a multiple of 2
pub fn pad2(&mut self) { ... }
```

write-fonts/src/tables/glyf.rs
```diff
@@ -487,4 +689,39 @@ mod tests {
...
+        let glyf = font.glyf().unwrap();
+        let read_glyf::Glyph::Composite(orig) = loca.get_glyf(GlyphId::new(2), &glyf).unwrap().unwrap() else { panic!("not a composite glyph") };
+
+        let mut iter = orig
+            .components()
+            .map(|comp| Component::new(comp.glyph, comp.anchor, comp.transform, comp.flags));
+        let mut composite = CompositeGlyph::new(iter.next().unwrap());
+        composite.add_component(iter.next().unwrap());
+        composite._instructions = orig.instructions().unwrap_or_default().to_vec();
+        //FIXME: figure out how we're actually computing bboxes here
```
rsheeter, line 706: As noted elsewhere, perhaps we just pass bbox to add_component for now?

write-fonts/src/tables/glyf.rs
```diff
@@ -487,4 +689,39 @@ mod tests {
...
+        };
+        assert!(iter.next().is_none());
+        let bytes = crate::dump_table(&composite).unwrap();
+        let ours = read::tables::glyf::CompositeGlyph::read(FontData::new(&bytes)).unwrap();
+
+        let our_comps = ours.components().collect::<Vec<_>>();
+        let orig_comps = orig.components().collect::<Vec<_>>();
+        assert_eq!(our_comps.len(), orig_comps.len());
+        assert_eq!(&our_comps[0], &orig_comps[0]);
+        assert_eq!(&our_comps[1], &orig_comps[1]);
```
rsheeter, line 721: I'm failing to spot where we assured ourselves there would be 2; perhaps we should additionally assert len is 2?

write-fonts/src/tables/glyf.rs
```diff
@@ -487,4 +689,39 @@ mod tests {
...
+        assert_eq!(ours.instructions(), orig.instructions());
+        assert_eq!(orig.offset_data().len(), bytes.len());
+
+        assert_eq!(orig.offset_data().as_ref(), bytes);
```
rsheeter, line 725: It's hard to tell what the transform actually *is*, maybe worth asserting each component has the expected transform

rsheeter: It's also hard to tell what flags are covered and thus what other tests we might want. I will venture to guess this one glyph from vazir probably doesn't cover all the cases we want to validate.

anthrotype:
>> cannot compute the bounding boxes for those glyphs

> I suggest we accept this limitation for the moment and ask the client to provide, e.g. add_component(component, bbox). ... @anthrotype does that sound remotely sane?

yes, sounds good to me

read-fonts/src/tables/glyf.rs
```diff
@@ -788,4 +828,68 @@ mod tests {
...
+        };
+        assert_eq!(
+            anchor.compute_flags(),
+            CompositeGlyphFlags::ARG_1_AND_2_ARE_WORDS
+        )
+    }
+
+    #[test]
+    fn compute_transform_flags() {
+        fn make_xform(xx: f32, yy: f32, xy: f32, yx: f32) -> Transform {
```
anthrotype, line 869: the order of parameter to this `make_xform` helper looks odd to me, might as well follow the same order (xx, yx, xy, yy) in which these are defined in the Transform struct (which in turn follows the canonical column-major ordering of these)

write-fonts/src/tables/glyf.rs
```diff
@@ -327,12 +372,169 @@ impl FontWrite for SimpleGlyph {
...
+        self.glyph.write_into(writer);
+        self.anchor.write_into(writer);
+        self.transform.write_into(writer);
+    }
+}
+
+impl CompositeGlyph {
+    /// Create a new composite glyph, with the provided component.
+    ///
+    /// Additional components can be added with [`add_component`][Self::add_component]
```
anthrotype, line 422: why special case one component per composite glyph? I'd have expected CompositeGlyph::new to take one or more components (a slice of len > 1 probably)

cmyr: It's possible I'm being overly clever here, but the basic idea is to enforce the invariant that a composite glyph always has at least one component. There's no obvious way how else I would express this in rust without getting very fancy. I don't really want to use a slice since I don't know how the caller is storing components (it is likely to be in something like a hashmap, though) and it's also hard to express the idea of having the constructor take a non-empty iterator. I definitely expect to see this API get tweaked as we start to actually try to use it for things, though.

rsheeter: `new(vec<Component>...) -> Result<Self, SomeErrorT>` would be much less annoying to work with.

rsheeter: Actually ... as with pens this would arguably be nice as an iterator adapter. Collecting into a Composite would be quite nice as long as the error handling is tidy.

write-fonts/src/tables/glyf.rs
```diff
@@ -327,12 +377,187 @@ impl FontWrite for SimpleGlyph {
...
+        }
+    }
+}
+
+impl Component {
+    /// Create a new component.
+    pub fn new(
+        glyph: GlyphId,
+        anchor: Anchor,
+        transform: Transform,
```
rsheeter, line 397: I want to give you a glyph name and a kurbo::Affine

write-fonts/src/tables/glyf.rs
rsheeter: #251 shows a sketch of suggestion above.

write-fonts/src/write.rs
```diff
@@ -145,6 +145,16 @@ impl TableWriter {
         data.add_offset(obj_id, width, self.offset_adjustment);
     }
 
+    /// Add a padding byte of necessary to ensure the table length is an even number.
+    ///
+    /// This is necessary for things like the glyph table, which require offsets
+    /// to be aligned on 2-byte boundaries.
+    pub fn ensure_2byte_aligned(&mut self) {
```
rsheeter, line 152: nit: I find ensure less informative than pad wrt what is actually going to happen, e.g. pad_to_2byte_aligned

PR #224, opened 2023-02-03
Don't rotate/reverse paths

cmyr:
This can be handled somewhere higher up the stack, as needed.


dfrg (approved):
Lgtm. Makes sense for code at this level to deal with the actual outline encoding.

PR #226, opened 2023-02-03
Move glyf2path code to read-fonts

dfrg:
This moves the glyf outline to path converter code from `punchcut` to `read-fonts` to enable roundtrip testing of `write-fonts` code and in the hope that it might be generally useful there.

Also adds a single roundtrip test to demonstrate usage.


font-types/src/path.rs
```diff
@@ -0,0 +1,24 @@
+/// Interface for accepting a sequence of path commands.
+///
+/// This is a general abstraction to unify ouput for processes that decode and/or
+/// transform outlines.
+///
+/// The type parameter `T` represents the coordinate type.
+pub trait PathSink<T> {
```
rsheeter, line 7: noob question: is Into<T> a bad idea here and if so why?

rsheeter (approved):

font-types/src/path.rs
cmyr: Do you mean `T` as the coordinate type, or do you mean `Into<T>` as an alternative to this whole interface?

font-types/src/path.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// Emit a line segment from the current point to (x, y).
+    fn line_to(&mut self, x: T, y: T);
+
+    /// Emit a quadratic bezier segment from the current point with a control
+    /// point at (cx0, cy0) and ending at (x, y).
+    fn quad_to(&mut self, cx0: T, cy0: T, x: T, y: T);
+
+    /// Emit a cubic bezier segment from the current point with control
+    /// points at (cx0, cy0) and (cx1, cy1) and ending at (x, y).
+    fn curve_to(&mut self, cx0: T, cy0: T, cx1: T, cy1: T, x: T, y: T);
```
cmyr, line 20: My question here becomes: what exactly is the scope of this interface? do we anticipate supporting cubics? Are we converting multiple contours into a single path, by adding move/close?

font-types/src/path.rs
cmyr: Also: is there a strong argument for having distinct coordinate types? How all do we expect this to be used?

read-fonts/src/tables/glyf.rs
```diff
@@ -477,6 +478,175 @@ impl<'a> SomeTable<'a> for Component {
...
+}
+
+/// Converts a `glyf` outline described by points, flags and contour end points to a sequence of
+/// path elements and and invokes the appropriate callback on the given sink for each.
+///
+/// The input points are expected in `F26Dot6` format as that is the standard result of scaling
+/// a TrueType glyph. Output points are generated in `f32`.
+pub fn to_path(
+    points: &[Point<F26Dot6>],
+    flags: &[u8],
```
cmyr, line 518: are these `SimpleGlyphFlags`? some other flag type? Would this code benefit from having that type be defined?

cmyr (commented):
seems reasonable, some notes inline :)

font-types/src/path.rs
dfrg: I'm honestly not sure. This was initially non-generic and used f32. When I was copying it over to read-fonts, I attempted to make both PathSink and to_path generic, but ended up stalling on the necessary division by 2 for computing mid points. I left PathSink as a generic because it didn't seem to hurt. Happy to go back to unconditional f32s if this seems like premature abstraction.

I also need more context for `Into<T>`

font-types/src/path.rs
dfrg: We'll need to support cubics when we add the CFF table. I know there is also a desire to add cubics to glyf, so I don't see any reason to leave this out.

It's worth mentioning that this is the trait that is used to extract glyphs in punchcut as part of our new callback-only API, so it needs to be general.

read-fonts/src/tables/glyf.rs
dfrg: This is a bit complicated. For the purposes of `to_path`, they would only have the on-curve (or a prospective cubic off-curve) bit set. But this array is overloaded and is used by both variation delta computation and hinting to store temporary flags, specifically for IUP processing. It could potentially be typed, but I'm not sure exactly what that might look like or if it would be helpful.

font-types/src/path.rs
dfrg: And yes, it combines multiple contours with move/close. This is essentially the same as `FT_Outline_Decompose`.

rsheeter: I suggest a comment re FT_O_D

rsheeter: We should support cubic, needed for both CFF and hopefully in future glyf as well (https://github.com/harfbuzz/boring-expansion-spec, 2nd point)

font-types/src/path.rs
```diff
@@ -0,0 +1,24 @@
+/// Interface for accepting a sequence of path commands.
```
anthrotype, line 1: This trait looks to me a lot like the "pen" protocol in fonttools and font-related python libraries. I wonder where the "sink" metaphor comes from? I'm sure both have their own "traditions". FWIW the "pen" is familiar to type designers and font engineers. Not immediately suggesting that you rename, just curious.

dfrg: The metaphor comes from the concept of sources and sinks-- the former being something that generates output and the latter something that receives input. If the name is not appropriately descriptive, a reasonable alternative in Rust land would be to call it `PathBuilder` but I usually like to reserve that for concrete types.

Btw, just checked out the pen module in fonttools. That's pretty nifty! Composable path adapters are a great idea and much easier to implement efficiently now that we have GATs.

font-types/src/path.rs
cmyr: Yea I'd vote for just leaving it as f32 🤷

read-fonts/src/tables/glyf.rs
cmyr: the main way I would imagine it being helpful would be as types-as-documentation; currently I look at this and I see a byte array and it could be anything. If it were a type you would also be able to add some simple convenience methods for it for testing various bits, which might also let you make this whole method slightly more readable? I'm fine either way, but if you want to keep it like this it might be nice to document somewhere what these flags are, exactly.

font-types/src/path.rs
dfrg: Seems reasonable. I'll change it back.

font-types/src/path.rs
rsheeter: https://en.wikipedia.org/wiki/Sink_(computing). To me this is more commonly associated with data pipelines and while not _wrong_ here I'm not sure it's a perfect fit either.

I don't think `trait PathBuilder` is all that unreasonable? - being a trait is a very clear message about the concrete'ness of it.

`trait Pen` is also entirely reasonable IMHO and the consistency with prior art is a non-trivial plus.

read-fonts/src/tables/glyf.rs
dfrg: Sold. I’ll add a `PointFlags` type with methods for the various bits.

font-types/src/path.rs
dfrg: Thinking on this a bit, Pen reads nicely to me. I’ll make the change.

PR #227, opened 2023-02-04
Glyph variations

dfrg:
This patch adds the computation and application of variation deltas to the TrueType glyph scaler based on the recently completed work on the gvar table.

Also updates the extract_glyphs.py script to sample glyphs at various design space positions when generating test data.


rsheeter:
Per f2f lets try to add links to FT where we are mirroring FT

punchcut/src/lib.rs
```diff
@@ -132,15 +132,20 @@ mod tests {
...
+                .coords(&expected_outline.coords)
+                .build(&font);
             scaler
                 .outline(expected_outline.glyph_id, &mut path)
                 .unwrap();
             if path.0 != expected_outline.path {
                 panic!(
-                    "mismatch in glyph path for id {} with size {}: path: {:?} expected_path: {:?}",
+                    "mismatch in glyph path for id {} (size: {}, coords: {:?}): path: {:?} expected_path: {:?}",
                     expected_outline.glyph_id.to_u16(),
```
rsheeter, line 146: Nit: do we really need to_u16? We can't make our glyph id type at least Debug if not Display?

punchcut/src/scaler.rs
```diff
@@ -122,35 +123,43 @@ impl<'a> ScalerBuilder<'a> {
...
-                        .and_then(|mappings| mappings.get(i).transpose().ok())
-                        .flatten()
-                        .map(|mapping| mapping.apply(coord))
-                        .unwrap_or(coord);
-                    let coord = coord.to_f64() as f32;
-                    *dest_coord = NormalizedCoord::from_f32(coord);
-                }
+        for variation in &self.context.variations {
+            // Make sure we iterate over all axes to support the case with multiple axes
+            // with the same name.
```
rsheeter, line 132: Suggest expanding to note this permits non-linear interpolation. Could cite https://github.com/PeterConstable/OT_Drafts/blob/master/NLI/UnderstandingNLI.md.

punchcut/src/scaler.rs
```diff
@@ -122,35 +123,43 @@ impl<'a> ScalerBuilder<'a> {
...
+            for (i, axis) in axes
+                .iter()
+                .enumerate()
+                .filter(|(_, axis)| axis.axis_tag() == variation.tag)
+            {
```
rsheeter, line 137: I assume we search axes over and over to avoid stack allocating a map tag:axis and it's performant because we assume a relatively small # of axes. That might be worth noting in a comment?

punchcut/src/scaler.rs
```diff
@@ -122,35 +123,43 @@ impl<'a> ScalerBuilder<'a> {
...
+                let coord = axis.normalize(Fixed::from_f64(variation.value as f64));
+                let coord = avar_mappings
+                    .as_ref()
+                    .and_then(|mappings| mappings.get(i).transpose().ok())
+                    .flatten()
+                    .map(|mapping| mapping.apply(coord))
+                    .unwrap_or(coord);
+                self.context.coords[i] = coord.to_f2dot14();
             }
         }
```
rsheeter, line 147: Do we use logging at all? - if we do, logging the case where a variation matches no axis might be worth considering.

punchcut/src/scaler.rs
```diff
@@ -122,35 +123,43 @@ impl<'a> ScalerBuilder<'a> {
...
-                }
+        for variation in &self.context.variations {
+            // Make sure we iterate over all axes to support the case with multiple axes
+            // with the same name.
+            for (i, axis) in axes
+                .iter()
+                .enumerate()
+                .filter(|(_, axis)| axis.axis_tag() == variation.tag)
+            {
+                let coord = axis.normalize(Fixed::from_f64(variation.value as f64));
```
rsheeter, line 138: It might be nicer to support `normalize(f32)` or to implement `From<f32> for Fixed`?

punchcut/src/scaler.rs
```diff
@@ -122,35 +123,43 @@ impl<'a> ScalerBuilder<'a> {
...
+                let coord = avar_mappings
+                    .as_ref()
+                    .and_then(|mappings| mappings.get(i).transpose().ok())
+                    .flatten()
+                    .map(|mapping| mapping.apply(coord))
+                    .unwrap_or(coord);
+                self.context.coords[i] = coord.to_f2dot14();
```
rsheeter, line 145: Similar to ^, I think I'd have expected a `.into()` here; is there a reason we aren't using `From`?

punchcut/src/source/glyf.rs
```diff
@@ -11,33 +12,27 @@ pub use scaler::Scaler;
...
-impl Default for Context {
-    fn default() -> Self {
-        Self::new()
-    }
+    /// Storage for simple glyph deltas.
+    deltas: Vec<Point<Fixed>>,
+    /// Storage for composite glyph deltas.
+    composite_deltas: Vec<Point<Fixed>>,
+    /// Temporary storage for computing interpolated deltas.
+    interp_points: Vec<Point<Fixed>>,
```
rsheeter, line 29: nits:

1. all your other `Vec<Point<Fixed>>` are named deltas, why is this one points while commented to be deltas?
1. You've used `Vec<Point<Fixed>>` enough it might be worth considering a type alias, e.g. `simple_glyph_deltas: Deltas`

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+    for delta in deltas.iter_mut() {
+        *delta = Default::default();
+    }
+    let Ok(var_data) = gvar.glyph_variation_data(glyph_id) else {
+        // Empty variation data for a glyph is not an error.
+        return Ok(());
+    };
+    for tuple in var_data.tuples() {
+        let Some(scalar) = tuple.compute_scalar(coords) else {
+            continue;
```
rsheeter, line 27: It would be nice if the documentation for compute_scalar clarified it will never return 0, you get None or Some(non-zero value)

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+        };
+        if tuple.all_points() {
```
rsheeter, line 29: I wasn't sure what `all_points` meant and had to check the docs. I suggest a more informative name. For example, `tuple.has_deltas_for_all_glyph_points` would have been immediately obvious.

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+        points.len() - 4
+    } else {
+        points.len() - 3
+    };
+    for tuple in var_data.tuples() {
+        let Some(scalar) = tuple.compute_scalar(coords) else {
+            continue;
+        };
+        if tuple.all_points() {
+            // When a tuple contains all points, we can simply accumulate the deltas directly.
```
rsheeter, line 90: This comment applies to composite as well.

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+    // shift of the outline.
+    let actual_len = if has_var_lsb {
+        points.len() - 4
+    } else {
+        points.len() - 3
+    };
+    for tuple in var_data.tuples() {
+        let Some(scalar) = tuple.compute_scalar(coords) else {
+            continue;
+        };
```
rsheeter, line 88: You duplicate the notion of looping over all the tuples that are applicatable. At a glance I believe you could switch from:

```rust
    for tuple in var_data.tuples() {
        let Some(scalar) = tuple.compute_scalar(coords) else {
            continue;
        };
```

To something like `for (tuple, scalar) in var_data.active_tuples_at(coord)` in both simple and composite.

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+        if tuple.all_points() {
+            // When a tuple contains all points, we can simply accumulate the deltas directly.
+            for (delta, tuple_delta) in deltas[0..actual_len].iter_mut().zip(tuple.deltas()) {
+                *delta += tuple_delta.apply_scalar(scalar);
+            }
```
rsheeter, line 93: The fast path is duplicated - other than slicing the deltas - as well. Could make it a function. Or... at risk of complexity for the sheer joy of it, make a function that does the filtering, handles the fast path, and takes a callback to handle the case where we can't fast-path.

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+/// Modeled after the FreeType implementation: <https://github.com/freetype/freetype/blob/bbfcd79eacb4985d4b68783565f4b494aa64516b/src/truetype/ttgxvar.c#L3881>
+fn interpolate_deltas(
+    points: &[Point<i32>],
+    flags: &[PointFlags],
+    contours: &[u16],
+    out_points: &mut [Point<Fixed>],
+) -> Option<()> {
+    if contours.is_empty() {
+        return Some(());
+    }
```
rsheeter, line 137: since the entirety of your code other than setting point_ix to 0 is in the loop over contours perhaps the guard clause is unnecessary?

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+    let mut point_ix = 0usize;
+    for &end_point_ix in contours {
+        let end_point_ix = end_point_ix as usize;
+        let first_point_ix = point_ix;
+        // Search for first point that has a delta.
```
rsheeter, line 142: can you clarify the meaing of not finding any point with a delta?

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+
+/// Shift the coordinates of all points between `start` and `end` inclusive
+/// using the difference given by the point at `ref_`.
+///
+/// Modeled after the FreeType implementation: <https://github.com/freetype/freetype/blob/bbfcd79eacb4985d4b68783565f4b494aa64516b/src/truetype/ttgxvar.c#L3776>
+fn shift_range(
+    points: &[Point<i32>],
+    start: usize,
+    end: usize,
+    ref_: usize,
```
rsheeter, line 209: Why the trailing _?

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+    Ok(())
+}
+
+/// Interpolate points without delta values, similar to the IUP hinting
+/// instruction.
+///
+/// Modeled after the FreeType implementation: <https://github.com/freetype/freetype/blob/bbfcd79eacb4985d4b68783565f4b494aa64516b/src/truetype/ttgxvar.c#L3881>
+fn interpolate_deltas(
+    points: &[Point<i32>],
+    flags: &[PointFlags],
```
rsheeter, line 131: are you supposed to have the same # of points, flags, out_points? Perhaps a debug_assert or two, or a comment clarifying the expected sizes if not?

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+    let delta = *ref_out - ref_in;
+    if delta.x == Fixed::ZERO && delta.y == Fixed::ZERO {
+        return Some(());
+    }
+    for out_point in out_points.get_mut(start..ref_)? {
+        *out_point += delta;
+    }
+    for out_point in out_points.get_mut(ref_ + 1..=end)? {
+        *out_point += delta;
+    }
```
rsheeter, line 223: the two loops is weird, you specifically don't want to shift ref by itself?

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+            }
+            // If we only have a single delta, shift the contour.
+            if cur_delta_ix == first_delta_ix {
+                shift_range(
+                    points,
+                    first_point_ix,
+                    end_point_ix,
+                    cur_delta_ix,
+                    out_points,
+                )?;
```
rsheeter, line 174: shift and interpolate with numerous numeric args where some are offset is not the easiest thing to read. I wonder how we could address that? 

Speculatively,

```rust
// early on
let points = PointJiggler { source: points, dest: out_points };

// later
points.delta(cur_delta_idx).apply_to(first_point_ix..end_point_idx)
```

punchcut/src/source/glyf/deltas.rs
rsheeter: https://github.com/dfrg/fontations/pull/1 shows exploratory hackery in this direction, feel free to use or ignore as you see fit.

punchcut/src/source/glyf/deltas.rs
```diff
@@ -0,0 +1,289 @@
...
+                    }
+                    out_point.$coord = out;
+                }
+            }
+        };
+    }
+    interp_coord!(x);
+    interp_coord!(y);
+    Some(())
+}
```
rsheeter, line 289: Computation of deltas, interpolating ranges, shifting ranges, etc all seem remarkably testable?

rsheeter: Passing a function to get variation data instead of an entire Gvar might make testing easier?

rsheeter (approved):
Overall LGTM, left a few comments. I know we have data driven tests but IMHO it wouldn't hurt to have narrowly focused unit tests for individual computations as well.

punchcut/src/lib.rs
dfrg: Looks like it already implements both. Must have been an oversight on my part. Fixed!

punchcut/src/scaler.rs
dfrg: Thanks for the link. Added.

punchcut/src/scaler.rs
dfrg: Correct. I added a note

punchcut/src/scaler.rs
dfrg: We don't and probably shouldn't take on a logging dependency given our current set of constraints. I'd expect axis validation to happen at a higher level in the layout stack. I wouldn't mind adding a collection of `unmatched_variations` that can be queried from constructed `Scaler` object if you think that might be useful?

punchcut/src/scaler.rs
dfrg: The higher level interface we'll expose for metadata (including axis/instance enumeration) will predominantly use f32 and we can move this over to make use of that when it's added. I'd like to keep the low level read-fonts API close to the spec.

As for `From`, the docs don't specifically require it but general convention is that the trait should only be implemented for lossless conversions. Same applies to the suggested Fixed -> F2Dot14 conversion below.

punchcut/src/source/glyf.rs
dfrg: `interp_points` gets passed as `working_points` to the internal delta computation code and it does actually store points that are used for interpolating deltas. I can change the name to match that and add a note to the comment about actual usage.

For the deltas, I'd feel better with a `type Delta = Point<Fixed>` to express the intent without throwing away the information that it's a `Vec`

punchcut/src/source/glyf.rs
```diff
@@ -57,16 +52,30 @@ mod tests {
...
-                Scaler::new(&mut cx, &font, None, expected_outline.size, None, &[]).unwrap();
+            let mut scaler = Scaler::new(
+                &mut cx,
+                &font,
+                None,
+                expected_outline.size,
+                None,
+                &expected_outline.coords,
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
dfrg: The top level `Context` uses a builder specifically to avoid this sort of confusion. Since this particular code will soon be removed from the public API anyway, I'd like to keep it as is to avoid complicating the higher level code which would have to branch on this rather than passing the fields from the builder directly.

I also haven't added the new `FontId` wrapper type that was previously discussed as a replacement for the raw u64. When that gets done, this could potentially be supplied as `FontId::none()` or `FontId::uncached()` which would remove the ambiguity.

punchcut/src/source/glyf/deltas.rs
dfrg: Done

punchcut/src/source/glyf/deltas.rs
dfrg: Fixed

punchcut/src/source/glyf/deltas.rs
dfrg: Slightly beyond my underscore threshold but I kinda like it, so done :)

punchcut/src/source/glyf/deltas.rs
dfrg: pinot actually has a method that does this but that didn't make it over to the implementation here in the first cut. Done!

punchcut/src/source/glyf/deltas.rs
dfrg: Good catch. Removed the worthless early exit.

punchcut/src/source/glyf/deltas.rs
dfrg: If there are no deltas within a contour then the tuple simply doesn't apply and the contour is left unmodified. Added a note to explain this.

punchcut/src/source/glyf/deltas.rs
dfrg: `ref` is a keyword. I changed these to `ref_ix` to copy the style in other parts of the code and to avoid the dangling underscore.

punchcut/src/source/glyf/deltas.rs
dfrg: Correct. Added the debug asserts.

punchcut/src/source/glyf/deltas.rs
dfrg: Yes, we only call shift on a contour that contains a single point with an applied delta. In this case, we apply that same delta to the remaining points. The loops could be combined with a `split_at_mut` and a `chain`, but I think that is less clear than the double loop. I'll add a comment to explain what's going on.

punchcut/src/source/glyf/deltas.rs
dfrg: Replaced shift and interpolate with these:
```
jiggler.shift(first_point_ix..=end_point_ix, cur_delta_ix)?;

jiggler.interpolate(
    cur_delta_ix + 1..=end_point_ix,
    RefPoints(cur_delta_ix, first_delta_ix),
)?;
```
So they now take range + ref point(s)

punchcut/src/source/glyf/deltas.rs
dfrg: Nice, ty! Copied over manually because I had already created a few conflicts by this point.

punchcut/src/source/glyf/deltas.rs
dfrg: Testing full delta computation in isolation of glyf is annoying because the algorithm depends on the outline to determine interpolation factors for deltas. I did, however, add unit tests for the underlying delta interpolation code which is designed to trigger both interpolations and shifts.

I think this, combined with the local tests in gvar plus the data driven tests gives us good coverage. Happy to extend this in another PR if you feel it's insufficient.

PR #230, opened 2023-02-07
Add test for glyph with multiple contours

cmyr:
This uncovers a difference in how we encode SimpleGlyphFlags in the case where there is a single repeat, and changes our implementation to match fontmake's.

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -251,10 +251,11 @@ impl RepeatableFlag {
         let mut iter = flags.into_iter();
         let mut prev = None;
         // if a flag repeats exactly once, then there is no (space) cost difference
-        // between using a repeat flag followed by a value of '1' and just repeating
-        // the flag (without setting the repeat bit). Our approach naturally suggests
-        // the first approach, but fontmake takes the second, so we add an extra
-        // step where if we see a case where there's a single repeat, we split it into
+        // between 1) using a repeat flag followed by a value of '1' and 2) just
+        // repeating the flag (without setting the repeat bit).
+        // It would be simplest for us to go with option 1), but fontmake goes
+        // with 2). We like doing what fontmake does, so we add an extra step
+        // where if we see a case where there's a single repeat, we split it into
         // two separate non-repeating flags.
         let mut decompose_single_repeat = None;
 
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -250,8 +250,19 @@ impl RepeatableFlag {
     ) -> impl Iterator<Item = RepeatableFlag> {
         let mut iter = flags.into_iter();
         let mut prev = None;
+        // if a flag repeats exactly once, then there is no (space) cost difference
+        // between using a repeat flag followed by a value of '1' and just repeating
+        // the flag (without setting the repeat bit). Our approach naturally suggests
+        // the first approach, but fontmake takes the second, so we add an extra
+        // step where if we see a case where there's a single repeat, we split it into
+        // two separate non-repeating flags.
```
rsheeter, line 258: I would view matching existing as the default choice absent a reason to do otherwise. We know fontmake emits fonts that works, we know we want our compiler built on this code to produce the same output as fontmake other than well-understood differences. So ... what is our reason to do other than what fontmake does?

rsheeter: Hm. I misread I think. So you *do* match fontmake and the key point is we're choosing repetition specifically to match fontmake. Perhaps we could say that more directly?

```
        // if there are two consecutive identical flags, then there is no (space) cost difference
        // between 1) using a repeat flag followed by a value of '1' and 2) repeating
        // the flag without setting the repeat bit. fontmake takes option 2) so we match
        // that here.
```

Feel free to ignore if you feel this no better than the original explanation.

write-fonts/src/tables/glyf.rs
```diff
@@ -427,6 +446,32 @@ mod tests {
...
+        dbg!(
+            SimpleGlyphFlags::from_bits(1),
+            SimpleGlyphFlags::from_bits(9)
+        );
+        assert_eq!(our_points, their_points);
+        assert_eq!(orig.glyph_data(), ours.glyph_data());
+        assert_eq!(orig_bytes.len(), bytes.len());
+        assert_eq!(orig_bytes.as_ref(), bytes);
+    }
+
```
rsheeter, line 474: Shouldn't we have a test for the flag repeat change too? - the test matches the PR title but the PR includes that additional change.

rsheeter (approved):
LGTM other than a nit about changing flag behavior w/o a test

PR #231, opened 2023-02-08
Compute proper midpoint in glyf::to_path

dfrg:
The last updates to `to_path` incorrectly changed this to be a fixed point division. Divide the raw bits by 2 instead.

(just merge me)

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -610,7 +610,9 @@ pub fn to_path(
     fn to_f32(x: F26Dot6) -> f32 {
         x.to_f64() as f32
     }
-    const TWO: F26Dot6 = F26Dot6::from_i32(2);
+    fn midpoint(a: Point<F26Dot6>, b: Point<F26Dot6>) -> Point<F26Dot6> {
+        ((a + b).map(F26Dot6::to_bits) / 2).map(F26Dot6::from_bits)
+    }
     let mut count = 0usize;
     let mut last_was_close = false;
     for contour_ix in 0..contours.len() {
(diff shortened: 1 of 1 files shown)


rsheeter:
What harm does this cause? Is it possible to write a test that exposes giving the wrong answer that becomes fixed with this change?

rsheeter (approved):
LGTM

read-fonts/src/tables/glyf.rs
```diff
@@ -610,7 +610,9 @@ pub fn to_path(
     fn to_f32(x: F26Dot6) -> f32 {
         x.to_f64() as f32
     }
-    const TWO: F26Dot6 = F26Dot6::from_i32(2);
+    fn midpoint(a: Point<F26Dot6>, b: Point<F26Dot6>) -> Point<F26Dot6> {
```
rsheeter, line 613: this is obtuse enough I suggest a comment. bonus points for a citation.

read-fonts/src/tables/glyf.rs
```diff
@@ -610,7 +610,9 @@ pub fn to_path(
...
+        ((a + b).map(F26Dot6::to_bits) / 2).map(F26Dot6::from_bits)
```
cmyr, line 614: Is the `Div` impl for `F26Dot6` broken, or is it right "sometimes"? If it exists but does not produce the correct result in this case, that feels like a heck of a footgun.

read-fonts/src/tables/glyf.rs
dfrg: The relevant code is here: https://github.com/freetype/freetype/blob/de8b92dd7ec634e9e2b25ef534c54a3537555c11/src/base/ftoutln.c#L123

I’ll add a note and include the link.

dfrg:
> What harm does this cause? Is it possible to write a test that exposes giving the wrong answer that becomes fixed with this change?

As Colin points out, our `Div` impl for `F26Dot6` is incorrect and that's where proper tests should be applied should we choose to keep that method. Opened issue #232 to address this.

PR #234, opened 2023-02-13
Fix missing Ord impl for flag types

### LINKED ISSUE #233: fea-rs doesn't build against fontations head
ValueFormat seems to no longer be Ord:

```
error[E0277]: the trait bound `ValueFormat: Ord` is not satisfied
    --> fea-rs/src/compile/lookups/gpos.rs:223:22
     |
223  |                     .entry((v1.format(), v2.format()))
     |                      ^^^^^ the trait `Ord` is not implemented for `ValueFormat`
```

cmyr:
This was overlooked when we stopped using bitflags.

fixes #233

(just merge me)


rsheeter (approved):
LGTM, will merge as soon as CI turns green

PR #236, opened 2023-02-14
Convert read SimpleGlyph to write SimpleGlyph

dfrg:
Implements `FromObjRef` and `FromTableRef` for `write_fonts::tables::glyf::SimpleGlyph` to allow conversion from the read type to the write type using the `from_table_ref` method.

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -460,16 +460,25 @@ mod tests {
         path.0
     }
 
+    // For `indexToLocFormat == 0` (short version), offset divided by 2 is stored, so add a padding
+    // byte if the length is not even to ensure our computed bytes match those of our test glyphs.
+    fn pad_for_loca_format(loca: &read::tables::loca::Loca, mut bytes: Vec<u8>) -> Vec<u8> {
+        if matches!(loca, read::tables::loca::Loca::Short(_)) && bytes.len() & 1 != 0 {
+            bytes.push(0);
+        }
+        bytes
+    }
+
     #[test]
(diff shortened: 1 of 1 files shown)


rsheeter:
Merging because this unblocks me, worst case we can tweak later.

write-fonts/src/tables/glyf.rs
```diff
@@ -430,6 +460,35 @@ mod tests {
...
     }
 
+    // For `indexToLocFormat == 0` (short version), offset divided by 2 is stored, so add a padding
+    // byte if the length is not even to ensure our computed bytes match those of our test glyphs.
+    fn pad_for_loca_format(loca: &read::tables::loca::Loca, mut bytes: Vec<u8>) -> Vec<u8> {
+        if matches!(loca, read::tables::loca::Loca::Short(_)) && bytes.len() & 1 != 0 {
+            bytes.push(0);
+        }
+        bytes
+    }
```
cmyr, line 470: @dfrg this shouldn't be necessary? The `FontWrite` impl for `SimpleGlyph` should handle it already.

dfrg: I may have missed that but when I ran tests they were failing with a missing padding byte when the length of the glyph data was odd.

cmyr: interesting, that would indicate a bug on my part I think!

dfrg: I assumed the adjustment wasn’t made because it ultimately depends on whether short or long offsets are required and we need the full set of compiled glyphs to make that determination

PR #237, opened 2023-02-16
Implement a subset of FontTools Pens for write-fonts, leaning on kurbo for things like transform

rsheeter:
Having this makes porting algorithms implemented against pens in FontTools and nanoemoji easier.

My immediate need is to match https://github.com/googlefonts/ufo2ft/blob/dd738cdcddf61cce2a744d1cafab5c9b33e92dd4/Lib/ufo2ft/util.py#L205 for https://github.com/googlefonts/fontmake-rs/pull/123 so I need transform and reverse. I will presumably need a glyf pen too but I haven't done that yet.

### DIFF
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -26,7 +26,7 @@ pub trait Pen {
 }
 
 /// Captures commands to [Pen] to facilitate implementations that buffer commands.
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum PenCommand {
     MoveTo {
         x: f32,
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -4,7 +4,8 @@
 /// transform outlines.
 ///
 /// /// AbstractPen in Python terms.
-/// <https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/basePen.py#L54>
+/// <https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/basePen.py#L54>.
+/// Implementations
 pub trait Pen {
     /// Emit a command to begin a new subpath at (x, y).
     fn move_to(&mut self, x: f32, y: f32);
@@ -22,4 +23,51 @@ pub trait Pen {
 
     /// Emit a command to close the current subpath.
     fn close(&mut self);
+}
+
+/// Captures commands to [Pen] to facilitate implementations that buffer commands.
+#[derive(Debug, Copy, Clone)]
+pub enum PenCommand {
+    MoveTo {
+        x: f32,
+        y: f32,
+    },
+    LineTo {
+        x: f32,
+        y: f32,
+    },
+    QuadTo {
+        cx0: f32,
+        cy0: f32,
+        x: f32,
+        y: f32,
+    },
+    CurveTo {
+        cx0: f32,
+        cy0: f32,
+        cx1: f32,
+        cy1: f32,
+        x: f32,
+        y: f32,
+    },
+    Close,
+}
+
+impl PenCommand {
+    pub fn apply_to<T: Pen>(&self, pen: &mut T) {
+        match *self {
+            PenCommand::MoveTo { x, y } => pen.move_to(x, y),
+            PenCommand::LineTo { x, y } => pen.line_to(x, y),
+            PenCommand::QuadTo { cx0, cy0, x, y } => pen.quad_to(cx0, cy0, x, y),
+            PenCommand::CurveTo {
+                cx0,
+                cy0,
+                cx1,
+                cy1,
+                x,
+                y,
+            } => pen.curve_to(cx0, cy0, cx1, cy1, x, y),
+            PenCommand::Close => pen.close(),
+        }
+    }
 }
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -70,4 +70,17 @@ impl PenCommand {
             PenCommand::Close => pen.close(),
         }
     }
+
+    /// The directly stated - not implied - end point of the command.
+    ///
+    /// Notably, Close does have an end point but it is not directly stated so it returns None.
+    pub fn end_point(&self) -> Option<(f32, f32)> {
+        match *self {
+            PenCommand::MoveTo { x, y } => Some((x, y)),
+            PenCommand::LineTo { x, y } => Some((x, y)),
+            PenCommand::QuadTo { x, y, .. } => Some((x, y)),
+            PenCommand::CurveTo { x, y, .. } => Some((x, y)),
+            PenCommand::Close => None,
+        }
+    }
 }
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -76,10 +76,10 @@ impl PenCommand {
     /// Notably, Close does have an end point but it is not directly stated so it returns None.
     pub fn end_point(&self) -> Option<(f32, f32)> {
         match *self {
-            PenCommand::MoveTo { x, y } => Some((x, y)),
-            PenCommand::LineTo { x, y } => Some((x, y)),
-            PenCommand::QuadTo { x, y, .. } => Some((x, y)),
-            PenCommand::CurveTo { x, y, .. } => Some((x, y)),
+            PenCommand::MoveTo { x, y }
+            | PenCommand::LineTo { x, y }
+            | PenCommand::QuadTo { x, y, .. }
+            | PenCommand::CurveTo { x, y, .. } => Some((x, y)),
             PenCommand::Close => None,
         }
     }
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -2,6 +2,9 @@
 ///
 /// This is a general abstraction to unify ouput for processes that decode and/or
 /// transform outlines.
+///
+/// /// AbstractPen in Python terms.
+/// <https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/basePen.py#L54>
 pub trait Pen {
     /// Emit a command to begin a new subpath at (x, y).
     fn move_to(&mut self, x: f32, y: f32);
--- a/write-fonts/src/pens.rs
+++ b/write-fonts/src/pens.rs
@@ -1,7 +1,5 @@
 //! Pen implementations based on <https://github.com/fonttools/fonttools/tree/main/Lib/fontTools/pens>
 
-use std::mem;
-
 use font_types::Pen;
 use kurbo::{Affine, BezPath, Point};
 
@@ -30,34 +28,26 @@ impl<'a, T: Pen> TransformPen<'a, T> {
 }
 
 impl<'a, T: Pen> Pen for TransformPen<'a, T> {
-    fn move_to(&mut self, mut x: f32, mut y: f32) {
-        (x, y) = self.map_point(x, y);
+    fn move_to(&mut self, x: f32, y: f32) {
+        let (x, y) = self.map_point(x, y);
         self.inner_pen.move_to(x, y);
     }
 
-    fn line_to(&mut self, mut x: f32, mut y: f32) {
-        (x, y) = self.map_point(x, y);
+    fn line_to(&mut self, x: f32, y: f32) {
+        let (x, y) = self.map_point(x, y);
         self.inner_pen.line_to(x, y);
     }
 
-    fn quad_to(&mut self, mut cx0: f32, mut cy0: f32, mut x: f32, mut y: f32) {
-        (cx0, cy0) = self.map_point(cx0, cy0);
-        (x, y) = self.map_point(x, y);
+    fn quad_to(&mut self, cx0: f32, cy0: f32, x: f32, y: f32) {
+        let (cx0, cy0) = self.map_point(cx0, cy0);
+        let (x, y) = self.map_point(x, y);
         self.inner_pen.quad_to(cx0, cy0, x, y);
     }
 
-    fn curve_to(
-        &mut self,
-        mut cx0: f32,
-        mut cy0: f32,
-        mut cx1: f32,
-        mut cy1: f32,
-        mut x: f32,
-        mut y: f32,
-    ) {
-        (cx0, cy0) = self.map_point(cx0, cy0);
-        (cx1, cy1) = self.map_point(cx1, cy1);
-        (x, y) = self.map_point(x, y);
+    fn curve_to(&mut self, cx0: f32, cy0: f32, cx1: f32, cy1: f32, x: f32, y: f32) {
+        let (cx0, cy0) = self.map_point(cx0, cy0);
+        let (cx1, cy1) = self.map_point(cx1, cy1);
+        let (x, y) = self.map_point(x, y);
         self.inner_pen.curve_to(cx0, cy0, cx1, cy1, x, y);
     }
 
@@ -84,8 +74,8 @@ impl BezPathPen {
         }
     }
 
-    pub fn take_path(&mut self) -> BezPath {
-        mem::replace(&mut self.path, BezPath::new())
+    pub fn into_inner(self) -> BezPath {
+        self.path
     }
 }
 
@@ -134,6 +124,7 @@ mod tests {
         double.move_to(1.0, 1.0);
         double.line_to(2.0, 2.0);
 
-        assert_eq!("M1 1L2 2M2 2L4 4", bez.take_path().to_svg());
+        // We should see the move/line passed through double is doubled
+        assert_eq!("M1 1L2 2M2 2L4 4", bez.into_inner().to_svg());
     }
 }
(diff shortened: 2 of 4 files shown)


write-fonts/src/pens.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.inner_pen.move_to(x, y);
+    }
+
+    fn line_to(&mut self, mut x: f32, mut y: f32) {
+        (x, y) = self.map_point(x, y);
+        self.inner_pen.line_to(x, y);
+    }
+
+    fn quad_to(&mut self, mut cx0: f32, mut cy0: f32, mut x: f32, mut y: f32) {
+        (cx0, cy0) = self.map_point(cx0, cy0);
```
dfrg, line 44: This actually caught me by surprise as it’s fairly unusual in Rust. I’d recommend keeping the parameters immutable and adding a `let` here.

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+impl BezPathPen {
+    pub fn new() -> BezPathPen {
+        BezPathPen {
+            path: BezPath::new(),
+        }
+    }
+
+    pub fn take_path(&mut self) -> BezPath {
```
dfrg, line 87: Unless you really need to hold on to the pen, the common name for this is `into_inner` which takes self by value and just returns the path. It might also make sense to add a `From` impl.

dfrg (approved):
Looks good with some suggestions for more idiomatic code.

write-fonts/src/pens.rs
rsheeter: Good catch, so updated.

write-fonts/src/pens.rs
rsheeter: Fixed, ty

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+    fn close(&mut self) {
+        self.inner_pen.close();
+    }
+}
+
+pub struct BezPathPen {
+    path: BezPath,
+}
+
+fn pt(x: f32, y: f32) -> Point {
```
anthrotype, line 63: I find the name of this `pt` helper function too short, also it's not immediately clear from the name only that a cast is taking place. How about `pt_f64`?

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+        }
+    }
+
+    fn map_point(&self, x: f32, y: f32) -> (f32, f32) {
+        let pt = self.transform
+            * Point {
+                x: x as f64,
+                y: y as f64,
+            };
+        (pt.x as f32, pt.y as f32)
```
anthrotype, line 26: all this back and forth between f32 and f64 made me wonder.. are 32 bits enough for the pen protocol? Kurbo itself uses f64 for BezPath and Point. I guess it depends on the circumstances how much precision one needs. Perhaps the font-types Pen trait could be made generic over the coodinate float type. Or maybe I'm fine like that.

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
+//! Pen implementations based on <https://github.com/fonttools/fonttools/tree/main/Lib/fontTools/pens>
+
+use font_types::{Pen, PenCommand};
+use kurbo::{Affine, BezPath, Point};
+
+/// A pen that transforms params using [kurbo::Affine].
+pub struct TransformPen<'a, T: Pen> {
```
anthrotype, line 7: maybe the `pens` module could be turned in to a directory containing several modules for each pen? It matches the approach taken in fontTools, and gives each pen greater visibility I believe. Currently in here there's three new pens defined in here, TransformPen, BezPathPen and ReverseContourPen. As the number of pens increase this can get crowded

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+                cy1: cy0,
+                x: end_x,
+                y: end_y,
+            },
+            // Close is a line from (end_x,end_y)=>(start_x, start_y) so reversed it's a line to end x/y
+            PenCommand::Close => PenCommand::LineTo { x: end_x, y: end_y },
+        });
+    }
+
+    // a closing line to start is a Z
```
anthrotype, line 194: what you're doing here (replacing the last LineTo of the reversed path with a ClosePath when it overlaps the starting point) is problematic because you may inadvertently drop a duplicate lineTo point that immediately followed moveTo pre-reversal and which deliberately has the same (x,y) coordinates, thus potentially making VF sources incompatible after reversing.

In fontTools ReverseContourPen (as well as in the PointToSegmentPen that converts between points and segment-oriented pens) we have an option outputImpliedClosingLine that when True it always output the implied closing line regardless of whether the last point overlaps moveTo or not. Setting this option is preferable when working with VF masters.
eg. see this font where some contour points collaps onto one another but only in some masters https://github.com/googlefonts/fontmake/issues/572

But even when outputImpliedClosingLine=False, fonttools ReverseContourPen will _not_ omit the last lineTo (which was the first one, before reversing) if it overlaps with the starting point, otherwise the number of points and segments would change after reversing
See https://github.com/googlefonts/cu2qu/issues/51 (fixed by https://github.com/fonttools/fonttools/pull/1080)

see this test case: 
https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Tests/pens/reverseContourPen_test.py#L227-L250

and also this one: https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Tests/pens/reverseContourPen_test.py#L38-L53

actually _all_ the tests in reverseContourPen_test.py would be nice to port over...

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+    // subpath by definition contains 0 or 1 move commands, and the move is at [0] if it exists
+    let (start_x, start_y) = if let PenCommand::MoveTo { x, y } = commands[0] {
+        // When reversed, the move is to the end point of the last command
+        // in a typical [move, ..., close] structure, end point == start point
+        let (end_x, end_y) = end(x, y, *commands.last().unwrap());
+        reversed.push(PenCommand::MoveTo { x: end_x, y: end_y });
+        commands = &commands[1..];
+        (x, y)
+    } else {
+        (0.0, 0.0)
```
anthrotype, line 156: in fontTools Pen protocol, a moveTo is required all the time, it's impossible/forbidden to start a new subpath without one. Where's this (0,0) fallback coming from? We should return an an error here.

anthrotype: actually, in the fonttools pen protocol, there is only one exception to this rule that all subpaths must begin with a moveTo: i.e. a contour that has no on-curve points and is only made up of off-curves from quadratic curves, it's a TrueType glyf thing. In this case the whole contour is comprised of one big qCurveTo with n>1 off-curves and the last on-curve point is set to `None` (and a closePath() to mark its end).
see https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Lib/fontTools/pens/basePen.py#L332

However, I think in our Pen trait quad_to means a single quadratic bezier, not a fancy TrueType-style quadratic B-spline with implied on-curves, so at the moment we can't sypport that

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+        pen.close();
+    }
+
+    fn draw_closed_test_shape(pen: &mut impl Pen) {
+        pen.move_to(125.0, 100.0);
+        pen.quad_to(200.0, 150.0, 175.0, 300.0);
+        pen.curve_to(150.0, 150.0, 50.0, 150.0, 25.0, 300.0);
+        pen.quad_to(0.0, 150.0, 75.0, 100.0);
+        pen.line_to(100.0, 50.0);
+        pen.close();
```
anthrotype, line 302: I think I found an issue. All your closed test shapes end with a line_to so this does not show up in your tests.
Basically, whenever the last pen command before the close is _not_ a line_to (e.g. a curve_to or quad_to or move_to even), then after reversing you end up with an _additional_ line segment that overlaps the move.
E.g. given this path (notice it ends with a curve_to before the closing 'Z'):

"M100,100 L150,200 L50,200 C75,200 75,100 100,100 Z" 

If I reverse it using your current code, I get this (I added commas for legibility):

"M100,100 L100,100 C75,100 75,200 50,200 L150,200 Z"

Note that first 'L' segment following the 'M' is a duplicate and was not in the original closed path.

Compare this with the output from fontTools ReverseContourPen:

```python
from fontTools.pens.recordingPen import RecordingPen
from fontTools.pens.reverseContourPen import ReverseContourPen
rec = RecordingPen()
rev = ReverseContourPen(rec)
rev.moveTo((100.0, 100.0))
rev.lineTo((150.0, 200.0))
rev.lineTo((50.0, 200.0))
rev.curveTo((75.0, 200.0), (75.0, 100.0), (100.0, 100.0))
rev.closePath()
print(rec.value)
"""
[('moveTo', ((100.0, 100.0),)),
 ('curveTo', ((75.0, 100.0), (75.0, 200.0), (50.0, 200.0))),
 ('lineTo', ((150.0, 200.0),)),
 ('closePath', ())]
"""
```

write-fonts/src/pens.rs
dfrg: These types evolved separately and the primary purpose for the `Pen` trait (originally called `PathSink`) was to support callback based output from the glyph loader. Incidentally, the original trait was generic over coordinate type.

write-fonts/src/pens.rs
anthrotype: All this talk about implied closing line lead me to discover what I believe is a bug in fonttools' implementation as well LOL

https://github.com/fonttools/fonttools/pull/2995

The ReverseContourPen's `outputImpliedClosingLine` parameter is a very recent addition by behdad, we haven't released it yet (a parameter with that name previously existed only in the PointToSegmentPen). It turns out it has a very similar bug whereby a duplicate lineTo, overlapping moveTo, gets inserted after reversing a closed path.
The bug is not present in the default case, when outputImpliedClosingLine=False.

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+///
+/// Reverses the winding direction of the contour. Keeps the first point unchanged.
+///
+/// <https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/reverseContourPen.py#L8>
+pub struct ReverseContourPen<'a, T: Pen> {
+    inner_pen: &'a mut T,
+    pending: Vec<PenCommand>,
+}
+
+fn end(subpath_start_x: f32, subpath_start_y: f32, subpath_cmd: PenCommand) -> (f32, f32) {
```
dfrg, line 125: This seems like it might be useful as a public `end_point` method on `PenCommand` that returns an `Option<(f32, f32)>` yielding `None` in the case of the close command.

anthrotype: 👍  I thought exactly the same

write-fonts/src/pens.rs
```diff
@@ -0,0 +1,369 @@
...
+        (x, y)
+    } else {
+        (0.0, 0.0)
+    };
+
+    // Reverse the commands between move (if any) and final close (if any)
+    for (idx, cmd) in commands.iter().enumerate().rev() {
+        let (end_x, end_y) = if idx > 0 {
+            end(start_x, start_y, commands[idx - 1])
+        // } else if ends_with_close {
```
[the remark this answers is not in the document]
rsheeter: Err no, that should be gone.

write-fonts/src/pens.rs
rsheeter: Assuming the pen starts at 0,0 was just an intuition really, e.g. if you just said "line to x,y" it would mean "from current, which is 0,0 by default." I'm ok with making that an error.

rsheeter: Fixed.

write-fonts/src/pens.rs
rsheeter: I would like to tackle any such modification to the Pen trait outside this PR

write-fonts/src/pens.rs
rsheeter: Renamed as_kurbo_point which I think captures the intent somewhat better than "pt" :)

write-fonts/src/pens.rs
rsheeter: Good call, will update so.

write-fonts/src/pens.rs
rsheeter: > I added commas for legibility

@dfrg are we allowed to change the to_svg format for Kurbo? - it is quite illegible right now

rsheeter: Forgive my density but I don't quite follow.

```
E.g. given this path (notice it ends with a curve_to before the closing 'Z'):

"M100,100 L150,200 L50,200 C75,200 75,100 100,100 Z" => MLLCZ

If I reverse it using your current code, I get this (I added commas for legibility):

"M100,100 L100,100 C75,100 75,200 50,200 L150,200 Z" => MLCLZ
```

So before and after we have 1 M, 2 L's, 1 C, and a Z. In the original the Z is pointless - we already ended the C at 100,100 - so when reversed the first command is pointless. I feel like I'm missing something.

write-fonts/src/pens.rs
rsheeter: Good call re porting tests, will do.

write-fonts/src/pens.rs
dfrg: I don’t think kurbo makes any particular guarantees about the svg output. I assume you’d prefer a format like the comment above?

rsheeter: > I assume you’d prefer a format like the comment above?

Yes. The , helps spot which values are a point and the space *before* the letter makes it *far* easier to spot the commands.

write-fonts/src/pens.rs
rsheeter: I don't have strong feelings about it. I think in Rust the documentation system would split the types out independent of what files they live in so perhaps this sort of organization is less important?

write-fonts/src/pens.rs
rsheeter: I'm dubious about the FontTools practice of eliminating the line from a close during reversal. Imagine you have several versions of move/line/line/close and in _some_ of them the last line ends at the start ... if you eliminate the "unnecessary" line from close as suggested by https://github.com/fonttools/fonttools/blob/bf265ce49e0cae6f032420a4c80c31d8e16285b8/Tests/pens/reverseContourPen_test.py#L23 - note that an input MLLLC turns into an output MLLC - then these shapes will no longer interpolate.

rsheeter: I suppose ^ is to say I like the outputImpliedClosingLine=True behavior and do not currently implement the FontTools default mode of dropping the closing lines.

font-types/src/pen.rs
```diff
@@ -70,4 +70,17 @@ impl PenCommand {
...
+    ///
+    /// Notably, Close does have an end point but it is not directly stated so it returns None.
+    pub fn end_point(&self) -> Option<(f32, f32)> {
+        match *self {
+            PenCommand::MoveTo { x, y } => Some((x, y)),
+            PenCommand::LineTo { x, y } => Some((x, y)),
+            PenCommand::QuadTo { x, y, .. } => Some((x, y)),
+            PenCommand::CurveTo { x, y, .. } => Some((x, y)),
+            PenCommand::Close => None,
+        }
```
dfrg, line 84: If you're binding the same set of values in a sequence of match arms, you can combine them. This could be written a bit more succinctly:
```rust
        use PenCommand::*;
        match *self {
            MoveTo { x, y } | LineTo { x, y } | QuadTo { x, y, .. } | CurveTo { x, y, .. } => {
                Some((x, y))
            }
            Close => None,
        }
```
But the current code is also fine.

write-fonts/src/pens.rs
```diff
@@ -270,12 +272,64 @@ impl<'a, T: Pen> Pen for ReverseContourPen<'a, T> {
...
+}
+
+impl RecordingPen {
+    pub fn new() -> RecordingPen {
+        RecordingPen {
+            commands: Vec::new(),
+        }
+    }
+
+    pub fn commands(&self) -> &Vec<PenCommand> {
```
dfrg, line 287: Common practice would be for this to return a slice instead.

font-types/src/pen.rs
rsheeter: Fancy!

rsheeter: Done, ty

write-fonts/src/pens.rs
rsheeter: Fixed, ty

write-fonts/src/pens.rs
rsheeter: I confirmed theory above with Rust people who know more than me so I am inclined to leave this for now

cmyr: yea my personal thinking here is that it makes sense to split things out when a single file would start to get unwieldy, and there is a nice clear way things could be separated.

rsheeter: When would you typically feel a single file is unwieldy? Is it lines, number of types, ...?

For me it would probably be as you got into thousands of lines or the file started to contain increasingly unrelated ideas. Concretely, to me this file is ... wieldy.

PR #238, opened 2023-03-01
Add a script to help set version

rsheeter:
Setting version by hand seems tiresome, first swing at automation.


rsheeter:
#239 does it better

PR #248, opened 2023-03-01
Update version to 0.1.0

### LINKED ISSUE #247: Move to 0.1.0
https://doc.rust-lang.org/cargo/reference/semver.html#change-categories "Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible." That inconveniently means every change to 0.0.x is incompatible so we should move to 0.1.0.

This causes problems like:

1. Release fontations 0.0.6 with a new feature for fontmake-rs
1. Update fontmake-rs to 0.0.6

_Be sad, fontmake-rs doesn't compile because fea-rs points to 0.0.5 which is definitely incompatible_

rsheeter:
Ran `python3 resources/scripts/version.py --ver 0.1.0`

Fixes #247.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,5 +11,5 @@ members = [
 ]
 
 [workspace.package]
-version = "0.0.6"
+version = "0.1.0"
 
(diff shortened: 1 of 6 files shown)


cmyr (approved):
🚢

PR #249, opened 2023-03-01
skrifa metadata round 1: metrics

dfrg:
First pass at adding high level metadata to the skrifa crate.

This patch adds support for reading metrics both at the global font level (with the `Metrics` type which is roughly equivalent to `SkFontMetrics`) and the `GlyphMetrics` type that exposes advance widths, left side bearings and bounding boxes for each glyph.

This also introduces the `MetadataProvider` trait that effectively extends any `read_fonts::TableProvider` to support access to the new metadata-- in this case, with the `metrics()` and `glyph_metrics()` methods. This set will grow with future patches.


skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+/// Extents of a rectangular region.
+#[derive(Copy, Clone, PartialEq, Default, Debug)]
+pub struct BoundingBox {
+    /// Minimum x-coordinate.
+    pub x_min: f32,
+    /// Minimum y-coordinate.
+    pub y_min: f32,
+    /// Maximum x-coordinate.
+    pub x_max: f32,
+    /// Maximum y-coordinate.
```
rsheeter, line 29: The comments here approach being noise.

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+
+use read_fonts::{
+    tables::{glyf::Glyf, hmtx::LongMetric, hvar::Hvar, loca::Loca},
+    types::{BigEndian, GlyphId},
+    TableProvider,
+};
+
+use crate::NormalizedCoord;
+
+/// Metrics for a text decoration.
```
rsheeter, line 11: This would be a magnificent place to explain what a decoration *is*

rsheeter: Also ... are there any invariants? What would a negative or 0 thickness mean?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+    pub x_min: f32,
+    /// Minimum y-coordinate.
+    pub y_min: f32,
+    /// Maximum x-coordinate.
+    pub x_max: f32,
+    /// Maximum y-coordinate.
+    pub y_max: f32,
+}
+
+/// Metrics that apply to all glyphs in a font.
```
rsheeter, line 33: Suggest clarifying if this is for the entire design space or for a specific position in design space?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+/// Metrics for a text decoration.
+#[derive(Copy, Clone, PartialEq, Default, Debug)]
+pub struct Decoration {
+    /// Offset to the top of the decoration from the baseline.
+    pub offset: f32,
+    /// Thickness of the decoration.
+    pub thickness: f32,
+}
+
+/// Extents of a rectangular region.
```
rsheeter, line 20: Do we not have the concept of a box in font-types? If not ... should we?

skrifa/src/meta/metrics.rs
rsheeter: Sugest clarifying with links that this data is drawn from https://learn.microsoft.com/en-us/typography/opentype/spec/head, etc.

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+}
+
+impl Metrics {
+    /// Creates new metrics for the given font, size in pixels per em, and
+    /// normalized variation coordinates.
+    ///
+    /// If `size` is `None` or `Some(0.0)`, resulting metric values will be in font units.
+    pub fn from_font<'a>(
+        font: &impl TableProvider<'a>,
+        size: Option<f32>,
```
rsheeter, line 75: A new user is likely to be mystified as to what size value to pass. What guidance for typical use can we provide?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+            metrics.is_monospace = post.is_fixed_pitch() != 0;
+            metrics.italic_angle = post.italic_angle().to_f64() as f32;
+            metrics.underline = Some(Decoration {
+                offset: post.underline_position().to_i16() as f32 * scale,
+                thickness: post.underline_thickness().to_i16() as f32 * scale,
+            });
+        }
+        let hhea = font.hhea();
+        if let Ok(hhea) = &hhea {
+            metrics.max_width = Some(hhea.advance_width_max().to_u16() as f32 * scale);
```
rsheeter, line 105: Noob question, if you wrote scale * hhea.advance_width_max().to_u16() would you get to drop the cast?

skrifa/src/meta/metrics.rs
rsheeter: Looking ahead I think a "decoration" is always a line, presumably a line drawn on top of the normal content?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+        }
+        // Choosing proper line metrics is a challenge due to the changing
+        // spec, backward compatibility and broken fonts.
+        //
+        // We use the same strategy as FreeType:
```
rsheeter, line 110: This seems like where you link to the FreeType equivalent?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+                }
+            }
+        }
+        if let (Ok(mvar), true) = (font.mvar(), !coords.is_empty()) {
+            use read_fonts::tables::mvar::tags::*;
+            macro_rules! metric_delta {
+                ($tag: ident) => {
+                    mvar.metric_delta($tag, coords).unwrap_or_default().to_f64() as f32 * scale
+                };
+            }
```
rsheeter, line 162: I appreciate the power macros afford but here normal code would work fine and my bias would be to prefer normal code when there isn't a good reason why not. Specifically:

```rust
// Define a free fn to lookup a delta
fn lookup_delta(mvar: &TableRef<MvarMarker>, tag: Tag, coords: &[F2Dot14], scale: f32) -> f32 {
    mvar.metric_delta(tag, coords).unwrap_or_default().to_f64() as f32 * scale
}

// OK, but I don't want to have noise from passing mvar, coords, scale!!
// So, instead of defining a macro, use a closure to capture the values
let metric_delta = |tag| lookup_delta(&mvar, tag, coords, scale);

// Now call metric_delta(HASC) instead of metric_delta!(HASC)

```

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+    loca_glyf: Option<(Loca<'a>, Glyf<'a>)>,
+    coords: &'a [NormalizedCoord],
+}
+
+impl<'a> GlyphMetrics<'a> {
+    /// Creates new glyph metrics from the given font, size in pixels per em units, and normalized
+    /// variation coordinates.
+    ///
+    /// If `size` is `None` or `Some(0.0)`, resulting metric values will be in font units.
+    pub fn from_font(
```
rsheeter, line 203: this might just be me but ... from_font makes me think there is some other option, like from_bytes vs from_file or some such. Here I don't know that you'd ever be able to create them for anything but a font so I wonder why this isn't `new`?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+        }
+        let mut advance = self
+            .h_metrics
+            .get(glyph_id.to_u16() as usize)
+            .map(|metric| metric.advance())
+            .unwrap_or(self.default_advance_width) as i32;
+        if let Some(hvar) = &self.hvar {
+            advance += hvar
+                .advance_width_delta(glyph_id, self.coords)
+                // FreeType truncates metric deltas...
```
rsheeter, line 260: he said, linking to wher eit happens

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+#[derive(Clone)]
+pub struct GlyphMetrics<'a> {
+    glyph_count: u16,
+    scale: f32,
+    h_metrics: &'a [LongMetric],
+    default_advance_width: u16,
+    lsbs: &'a [BigEndian<i16>],
+    hvar: Option<Hvar<'a>>,
+    loca_glyf: Option<(Loca<'a>, Glyf<'a>)>,
+    coords: &'a [NormalizedCoord],
```
rsheeter, line 195: I feel like we're conflating concepts here. Glyph metrics without coordinates is a valid concept, and glyph metrics with the coords specified is an instance. Naively I wonder if we should model it so more explicitly, e.g. construct GlyphMetrics *without* coords and then call something like instance or at to pin them down?

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+                thickness: 51.0,
+            }),
+        };
+        assert_eq!(metrics, expected);
+    }
+
+    #[test]
+    fn metrics_missing_os2() {
+        let font = FontRef::new(VAZIRMATN_VAR).unwrap();
+        let metrics = font.metrics(None, &[]);
```
rsheeter, line 365: Not crazy about this, I feel like I'd have to lookup the metrics fn to know what this means.

skrifa/src/meta/metrics.rs
```diff
@@ -0,0 +1,434 @@
...
+                (advance_width, lsb)
+            })
+            .collect::<Vec<_>>();
+        assert_eq!(expected, &result[..]);
+    }
+
+    #[test]
+    fn glyph_metrics_var() {
+        let font = FontRef::new(VAZIRMATN_VAR).unwrap();
+        let coords = &[NormalizedCoord::from_f32(-0.8)];
```
rsheeter, line 415: Could this exhibit the process for *user* coords? - I want to see something bulk up with weight. Something is at -0.8 is a bit impenetrable.

skrifa/src/meta/provider.rs
```diff
@@ -0,0 +1,25 @@
+use super::metrics::{GlyphMetrics, Metrics};
+use crate::NormalizedCoord;
+
+/// Interface for types that can provide font metadata.
+pub trait MetadataProvider<'a>: raw::TableProvider<'a> + Sized {
```
rsheeter, line 5: This is currently just metrics, not any broader metadata; I wonder if it should be narrowed? If not perhaps we could explain what we anticipate the scope growing to be.

This is just personal preference but ... I kind of like the narrowness and would personally be a fan of it becoming more specific, e.g. `MetricReader` (MetricFactory in Java terms, unless we can somehow get to MetricFactoryFactory which would earn +20 pts)

rsheeter (approved):
Couple of nits inline, generally looks good

skrifa/src/meta/metrics.rs
dfrg: Yes, decoration is an underline or strike through. I borrowed the terminology from css which I think would be familiar to most users? We could call this Line but to me that implies things like ascent, descent and leading in this context. 

There are no invariants— it simply returns what the font specifies. Both thickness values are signed in the spec so it seems reasonable to return absolute value here. Otherwise it should be up to the rendering implementation to decide what values make sense.

skrifa/src/meta/metrics.rs
dfrg: It’s for a specific position in design space. I’ll add a note about that and the relevant links.

skrifa/src/meta/metrics.rs
dfrg: Good call. I’ll move this there and make it generic over coordinate type.

skrifa/src/meta/metrics.rs
dfrg: Since all “size” values in this crate should be in ppem, I can add a section on sizing to the crate root doc and link to it from each method that makes use.

skrifa/src/meta/metrics.rs
dfrg: Rust doesn’t support implicit conversions regardless of operand order. This happens to be one of my favorite features of the language.

skrifa/src/meta/metrics.rs
dfrg: Will do!

skrifa/src/meta/metrics.rs
dfrg: Yep! The closure works just fine even without the intermediate function. Will change.

skrifa/src/meta/metrics.rs
dfrg: Not sure if I anticipated other constructors. Having a single one named new makes sense.

skrifa/src/meta/metrics.rs
dfrg: 🔗

skrifa/src/meta/metrics.rs
dfrg: An empty coordinates array implies the default position in design space. I’ll think of a way to make this more explicit.

skrifa/src/meta/metrics.rs
dfrg: As above, I’ll try to improve the signature. I think removing the option around size would help.

skrifa/src/meta/metrics.rs
dfrg: I’ll add a test here with user coords in the PR that adds variation metadata so I have the infra in place.

skrifa/src/meta/provider.rs
dfrg: The code I have staged for full metadata has many more methods here for attributes, charmaps, strings, axes and instances (so far). Rust requires you to import traits to make use of the methods so I think breaking them up adds friction for the user for no appreciable gain. I’d prefer to keep it as is for now and revisit after the full set lands.

skrifa/src/meta/metrics.rs
dfrg: Looking a bit deeper into this, FreeType does [weird stuff](https://github.com/freetype/freetype/blob/5c37b6406258ec0d7ab64b8619c5ea2c19e3c69a/src/sfnt/sfobjs.c#L1390) by shifting the position to be in the _middle_ of the stroke which doesn't seem to be useful.

Skia summarily [undoes this](https://github.com/google/skia/blob/04e32a9e56411e7fcdd894255323629ee6e7b103/src/ports/SkFontHost_FreeType.cpp#L1599) and also negates the offsets (converting from Y-up to Y-down).

The thickness values aren't touched at all.

I'd suggest we keep the current code as is, returning the raw values contained in the font.

skrifa/src/meta/metrics.rs
rsheeter: I'm imagining unpositioned vs positioned I think. At default is positioned. Some APIs, e.g. give me a specific metric, would require use of metrics at a specific position.

skrifa/src/meta/metrics.rs
dfrg: Moved this to `font_types::BoundingBox<T>` and made the comments less noisy. Hopefully this will be generally useful.

skrifa/src/meta/metrics.rs
dfrg: Done!

skrifa/src/meta/metrics.rs
dfrg: Added a new `Size` type that is better documented.

skrifa/src/meta/metrics.rs
dfrg: Per f2f, this code now looks like:
```rust
let metrics = font.metrics(Size::unscaled(), NormalizedCoords::default());
```

PR #251, opened 2023-03-07
[DISCUSS] Compile composite glyphs tweaks

rsheeter:
@cmyr's #223 tweaked to better accomodate what IR looks like in https://github.com/googlefonts/fontmake-rs/pull/120. See in particular commits by only me for what changed. Notably, I found this interface much easier to work with than creating a mutable composite with a single comonent then calling add a bunch of times:

```rust
impl CompositeGlyph {
    pub fn try_from(
        source: impl IntoIterator<Item = (Component, Bbox)>,
    ) -> Result<Self, CompositeGlyphError> { ... }
}
```

I rebased on main to unblock myself. This makes setting https://github.com/googlefonts/fontations/tree/composite-glyf as the branch a mess until it too is rebased on main. 

I imagine we'll want to update and merge #223 then close this.

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -423,7 +423,7 @@ struct FieldLengths {
 }
 
 /// Transform for a composite component.
-#[derive(Clone, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Transform {
     /// X scale factor.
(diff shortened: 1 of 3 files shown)


cmyr:
Okay I have taken your commit here and brought it into #223 with some minor tweaks (struct for the error, from_iter instead of try_from, which implies the `TryFrom` trait)

PR #254, opened 2023-03-08
[ci] Migrate from actions-rs, update checkout

cmyr:
There were a bunch of deprecation warnings in the CI logs, which raised two issues:

- we need to update any actions that we call from CI to [versions using node 16](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)
- the actions-rs actions we've been using to install the toolchain and run cargo [are unmaintained](https://github.com/actions-rs/toolchain/issues/216), so we move over to another action from dtolonay

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,7 +23,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: cargo doc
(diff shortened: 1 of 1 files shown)


rsheeter:
More deletes than add/updates ftw!

rsheeter (approved):

PR #255, opened 2023-03-08
Bump version to 0.1.1

cmyr:
It does feel weird to me that we're going to end up in situations where we have, say, versions [0.1, 0.1.1, 0.2, 0.3] for font-types, where all of those versions point to the same commit? Especially if anyone else is using font-types?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,5 +11,5 @@ members = [
 ]
 
 [workspace.package]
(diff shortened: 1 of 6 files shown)


rsheeter (approved):

cmyr:
closing, we need to figure out a better approach.

PR #257, opened 2023-03-09
[clippy]: Push the rock back up the hill

cmyr:
Just Merge Me™️

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -101,6 +101,7 @@ impl RangeRecord {
     }
 }
 
+#[allow(clippy::derivable_impls)]
 impl Default for DeltaFormat {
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/layout.rs
```diff
@@ -101,6 +101,7 @@ impl RangeRecord {
     }
 }
 
+#[allow(clippy::derivable_impls)]
```
rsheeter, line 104: If we can derive ti ... shouldn't we?

rsheeter (approved):

read-fonts/src/tables/layout.rs
cmyr: the type is declared in generated code, so we would need syntax in codegen for specifying default values, which we don't have.

PR #260, opened 2023-03-10
Bump write-fonts version

rsheeter:
So I can release #258


rsheeter:
Closing to attempot it using fancier tooling

PR #266, opened 2023-03-11
Let codegen input specify enum defaults

### LINKED ISSUE #265: write-fonts doesn't support enum because they do not impl Default
If you add something with an enum, cmap in my case, to write-fonts the result doesn't compile. Reproduced on a test file:

```shell
error[E0277]: the trait bound `MyEnum: Default` is not satisfied
  --> write-fonts/src/../generated/generated_test_enum.rs:19:5
   |
17 | #[derive(Clone, Debug, Default)]
   |                        ------- in this derive macro expansion
18 | pub struct MyRecord {
19 |     pub my_enum: MyEnum,
   |     ^^^^^^^^^^^^^^^^^^^ the trait `Default` is not implemented for `MyEnum`
```

Will post a PR with demonstration of bug shortly.

Occurs because the compile type for a record is Default and enums, emitted from https://github.com/googlefonts/fontations/blob/main/font-codegen/src/flags_enums.rs#L317, are not. Slapping `#[default]` on the Unknown is sufficient to compile - shown in #266 as this unblocks me - but perhaps it would be best to support default in the codegen input?

rsheeter:
Codegen input can now specify `#[default]` on enum variants. If none is specified `Unknown` will be the default.

Context: Currently an enum without a handwritten default cannot compile a write-fonts type. See #265. Based on `layout.rs` you can work around this by implementing default but I'd rather state it in the codegen input.

Fixes #265 .


font-codegen/src/parsing.rs
```diff
@@ -567,9 +573,27 @@ impl Parse for RawEnum {
...
+                    None
+                }
+            })
+            .collect();
+        if defaults.len() > 1 {
+            return Err(logged_syn_error(
+                name.span(),
+                format!("multiple defaults {defaults:?}"),
+            ));
+        }
```
cmyr, line 595: we have a pattern where we have a separate pass, after parsing, where we check that attributes make sense; this is the `sanity_check` method.

So what I would do is to just parse this naively here, and then add a `sanity_check` method to `RawEnum`, which checks that there's at most one item marked default.

(also you'll need to update `Item::sanity_check` to forward the call to the type; currently it doesn't, since we don't currently implement this for raw enums)

font-codegen/src/flags_enums.rs
```diff
@@ -321,25 +321,30 @@ pub(crate) fn generate_raw_enum(raw: &RawEnum) -> TokenStream {
...
+    let maybe_default = (!raw.variants.iter().any(|v| v.attrs.default)).then(|| quote!(#[default]));
     quote! {
         #( #docs )*
-        #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+        #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
         #[repr(#typ)]
         pub enum #name {
             #( #variants )*
             #[doc(hidden)]
+            #maybe_default
```
cmyr, line 347: I don't think we ever want to have `Unknown` be the default. The existence of `Unknown` is sort of a hack; it is working around the fact that it's possible for an enum value to be invalid in an existing font, and we need some way to represent that, but we cannot compile the `Unknown` variant, and we don't want the user to be constructing them.

I think that if the user does not specify a `#[default]` variant, we should not implement the trait, and the user can figure out what they want to do.

write-fonts/src/tables.rs
```diff
@@ -1,6 +1,8 @@
 //! A directory of all the font tables.
 
 pub mod base;
+// TODO fix bug with enum defaults
+//pub mod cmap;
```
cmyr, line 5: I doubt we want to commit this?

cmyr (approved):
looks good, some notes inline!

font-codegen/src/flags_enums.rs
rsheeter: That's fair, but then we are left with the issue that you get an errors whose solution is less than obvious when you generate a broken compile type. Maybe I'll file an issue to have a nice error in that case.

font-codegen/src/parsing.rs
```diff
@@ -394,11 +394,17 @@ pub(crate) struct RawEnum {
...
-    pub(crate) docs: Vec<syn::Attribute>,
+    pub(crate) attrs: EnumVariantAttrs,
     pub(crate) name: syn::Ident,
     pub(crate) value: syn::LitInt,
 }
 
+#[derive(Debug, Default, Clone)]
+pub(crate) struct EnumVariantAttrs {
+    pub(crate) docs: Vec<syn::Attribute>,
+    pub(crate) default: bool,
```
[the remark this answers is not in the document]
rsheeter: Oops, good catch. Will fix.

font-codegen/src/parsing.rs
rsheeter: I've been away from codegen too long, forgot ol' sanity check. Good catch, will do.

write-fonts/src/tables.rs
rsheeter: No, but we should definitely fix that bug!

font-codegen/src/parsing.rs
rsheeter: Fixed.

PR #267, opened 2023-03-11
Add the compile type for cmap

rsheeter:
Built on #266 because otherwise it doesn't compile. Just the codegen default, no handwritten additions yet. Baby steps.


rsheeter:
Ugh, please hold. I just realized cmap doesn't compile for additional reasons beyond the enum issue. Fooled myself it did by commenting out the tables.rs entry :D

PR #271, opened 2023-03-11
Unblock cmap compile type

### LINKED ISSUE #269: write-fonts generates FromObjRef that doesn't compile for cmap
After fixing enum woes in #265 I still have cmap compile type problems:

1. Uint24 isn't Default
1. Various reserved methods lack the suffix `_byte_range`

### LINKED ISSUE #270: Offset to T generates duplicate definitions of access fn
Adding `padded: Offset16<PadLikeCmap>,` to `table BasicTable` in `resources/codegen_inputs/test_records.rs` produces multiple definitions of padded:

```rust
    pub fn padded(&self) -> Offset32 {
        let range = self.shape.padded_byte_range();
        self.data.read_at(range.start).unwrap()
    }

    /// Attempt to resolve [`padded`][Self::padded].
    pub fn padded(&self) -> Result<PadLikeCmap<'a>, ReadError> {
        let data = self.data;
        self.padded().resolve(data)
    }
```

### LINKED ISSUE #273: Too many args is suppressed for records but not tables
The codegenerator generates a cilppy allow for too many args for a record ctor but not for a table ctor

rsheeter:
Fix issues blocking cmap compile type. Fixes #269. Fixes #270. Fixes #273.

Names that violate _offset(s) naming convention now get nice errors instead of compile issues in generated code:

```shell
Error: 
  × parsing failed
    ╭─[resources/codegen_inputs/test_records.rs:9:1]
  9 │ table BasicTable {
 10 │     padded: Offset32<PadLikeCmap>,
    ·     ───┬──
    ·        ╰── name must end in _offset
 11 │     #[compile(array_len($simple_records))]
    ╰────
```

Skip getter will complain if no way to acquire a value is defined, such as when applied to a reserved field:

```shell
 10 │     #[skip_getter]
 11 │     _reserved: u16,
    ·     ────┬────
    ·         ╰── There is no getter and no defined way to get a value. Perhaps you need ...?
```

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -199,6 +199,8 @@ pub(crate) struct FieldAttrs {
     pub(crate) to_owned: Option<Attr<InlineExpr>>,
     /// Custom validation behaviour
     pub(crate) validation: Option<Attr<FieldValidation>>,
+    /// Whether the value is computed by the user
+    pub(crate) user_computed: Option<syn::Path>,
 }
 
 #[derive(Debug, Clone)]
@@ -958,6 +960,7 @@ static READ_OFFSET_WITH: &str = "read_offset_with";
 static TRAVERSE_WITH: &str = "traverse_with";
 static TO_OWNED: &str = "to_owned";
 static VALIDATE: &str = "validate";
+static USER_COMPUTED: &str = "user_computed";
 
 impl Parse for FieldAttrs {
     fn parse(input: ParseStream) -> syn::Result<Self> {
@@ -1005,6 +1008,8 @@ impl Parse for FieldAttrs {
                 this.traverse_with = Some(Attr::new(ident.clone(), attr.parse_args()?));
             } else if ident == FORMAT {
                 this.format = Some(Attr::new(ident.clone(), parse_attr_eq_value(attr.tokens)?))
+            } else if ident == USER_COMPUTED {
+                this.user_computed = Some(attr.path);
             } else {
                 return Err(logged_syn_error(
                     ident.span(),
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -1237,7 +1237,7 @@ impl Items {
 }
 
 impl Item {
-    fn name(&self) -> &syn::Ident {
+    pub(crate) fn name(&self) -> &syn::Ident {
         match self {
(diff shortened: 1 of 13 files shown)


resources/codegen_inputs/test_records.rs
```diff
@@ -7,10 +7,11 @@
 #![parse_module(read_fonts::codegen_test::records)]
 
 table BasicTable {
+    padded_offset: Offset32<PadLikeCmap>,
```
cmyr, line 10: can we remove these changes now? It looks like they were added to explore what wasn't working, but I think this now just duplicates stuff that exists in `test_offsets_arrays.rs`.

font-codegen/src/parsing.rs
```diff
@@ -199,6 +199,8 @@ pub(crate) struct FieldAttrs {
     pub(crate) to_owned: Option<Attr<InlineExpr>>,
     /// Custom validation behaviour
     pub(crate) validation: Option<Attr<FieldValidation>>,
+    /// Whether the value is computed by the user
+    pub(crate) user_computed: Option<syn::Path>,
```
cmyr, line 203: Can you provide a bit more detail about problem this is solving?

cmyr (commented):
Would it be possible to split this up a bit? There's a bunch of stuff going on here, and I think some of it (like adding a new attribute) would benefit from focused discussion, whereas other chunks of it (`Default` for u24, warning if an offset is not `_offset` or `_offsets`) are obviously improvements that stand on their own.

(I'd also drop the commits that touch the test files, since they look like noise, unless I'm missing something)

rsheeter:
> Would it be possible to split this up a bit? 

Will do. Hopefully we can merge those quickly, the chain of dependencies for some downstream PRs is growing a little long.

resources/codegen_inputs/test_records.rs
rsheeter: This is not adequately tested by test_offsets_arrays. I could make it so I suppose.

rsheeter:
Closing this; will open smaller narrower PRs for it's parts.

resources/codegen_inputs/test_records.rs
cmyr: oh I see, I didn't understand what made this special when I first looked.

PR #274, opened 2023-03-13
Fix broken call to extract_glyphs

### LINKED ISSUE #268: Extract glyphs doesn't work on M1 Mac
On an M1 Mac:

```shell
Installing fonttools and freetype-py
Requirement already satisfied: pip in ./.venv/lib/python3.10/site-packages (23.0.1)
Requirement already satisfied: fonttools[unicode]==4.38 in ./.venv/lib/python3.10/site-packages (from -r /usr/local/google/home/rsheeter/oss/fontations/resources/test_fonts/requirements.txt (line 1)) (4.38.0)
Requirement already satisfied: freetype-py==2.3.0 in ./.venv/lib/python3.10/site-packages (from -r /usr/local/google/home/rsheeter/oss/fontations/resources/test_fonts/requirements.txt (line 2)) (2.3.0)
Requirement already satisfied: unicodedata2>=14.0.0 in ./.venv/lib/python3.10/site-packages (from fonttools[unicode]==4.38->-r /usr/local/google/home/rsheeter/oss/fontations/resources/test_fonts/requirements.txt (line 1)) (15.0.0)
Traceback (most recent call last):
  File "/usr/local/google/home/rsheeter/oss/fontations/resources/test_fonts/extract_glyphs.py", line 6, in <module>
    import freetype
ModuleNotFoundError: No module named 'freetype'
```

dfrg:
Changes `rebuild.sh` to activate the virtual environment before attempting to actually use it.

Fixes #268 

Just merge if it looks good.

### DIFF
--- a/resources/test_fonts/rebuild.sh
+++ b/resources/test_fonts/rebuild.sh
@@ -13,7 +13,7 @@ VENV_DIR=$SCRIPT_DIR/../../.venv
 REQUIREMENTS=$SCRIPT_DIR/requirements.txt
 PIP=$VENV_DIR/bin/pip
 TTX=$VENV_DIR/bin/ttx
-EXTRACT_GLYPHS = $SCRIPT_DIR/extract_glyphs.py
+EXTRACT_GLYPHS=$SCRIPT_DIR/extract_glyphs.py
 
 # check that we have python3 + virtualenv installed:
 if ! python3 -m venv -h  >/dev/null 2>&1; then
@@ -30,7 +30,11 @@ echo "Installing fonttools and freetype-py"
 $PIP install --upgrade pip
 $PIP install -r $REQUIREMENTS
 
+source $VENV_DIR/bin/activate
+
 for f in $(ls $SRC_DIR/*.ttx); do
     $TTX -o $OUT_DIR/$(basename "$f" .ttx).ttf --no-recalc-timestamp -b $f
     python3 $EXTRACT_GLYPHS $OUT_DIR/$(basename "$f" .ttx).ttf
 done
+
+deactivate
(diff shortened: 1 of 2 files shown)


cmyr:
fwiw the decision to not activate the venv (and instead to just call the venv-installed binaries directly) was intentional, and suggested by @anthrotype.

resources/test_fonts/rebuild.sh
```diff
@@ -30,7 +30,11 @@ echo "Installing fonttools and freetype-py"
 $PIP install --upgrade pip
 $PIP install -r $REQUIREMENTS
 
+source $VENV_DIR/bin/activate
+
 for f in $(ls $SRC_DIR/*.ttx); do
     $TTX -o $OUT_DIR/$(basename "$f" .ttx).ttf --no-recalc-timestamp -b $f
     python3 $EXTRACT_GLYPHS $OUT_DIR/$(basename "$f" .ttx).ttf
```
anthrotype, line 37: @cmyr @dfrg  if instead of calling `python3` you had called `$VENV_DIR/bin/python` then you would not need to first source the activate script

dfrg:
I’m a bit out of my element and having CLASS_PATH flashbacks so happy to adjust to whatever the python pros suggest. So if I use the python3 in the venv bin dir, I’ll have access to the associated libs installed by pip?

anthrotype:
> if I use the python3 in the venv bin dir, I’ll have access to the associated libs installed by pip?

yes. As you prefer, I don't mind if the shell script activate the venv, there's no risk that the venv environment leak into the calling shell (unless one were to explicitly do `source rebuild.sh` which I don't think it's the case for this).

dfrg:
I agree it's much cleaner to not activate the venv explicitly. #277 should do this properly. Thanks for educating me :)

PR #275, opened 2023-03-13
Helper for constructing a simple cmap4

rsheeter:
Add an api to write a cmap from a high level definition. Builds on #272.

I intend the internals of `from_mappings` to change over time, but the entrypoint to remain stable.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -39,7 +39,7 @@ impl CmapSubtable {
         let entry_selector = entry_selector as u16;
         let range_shift = seg_count * 2 - search_range;
 
-        let id_range_offsets = id_deltas.iter().map(|_| 0).collect();
+        let id_range_offsets = vec![0; id_deltas.len()];
         CmapSubtable::Format4(Cmap4::new(
             size_of_cmap4(seg_count, 0),
             lang,
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -2,62 +2,142 @@
 //!
 //! [cmap]: https://docs.microsoft.com/en-us/typography/opentype/spec/cmap
 
+use std::collections::HashSet;
+
 include!("../../generated/generated_cmap.rs");
+
+// https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#unicode-platform-platform-id--0
+const UNICODE_BMP_ENCODING: u16 = 3;
+
+fn size_of_cmap4(seg_count: u16, gid_count: u16) -> u16 {
+    8 * 2  // 8 uint16's
+    + 2 * seg_count * 4  // 4 parallel arrays of len seg_count, 2 bytes per entry
+    + 2 * gid_count // 2 bytes per gid in glyphIdArray
+}
+
+impl CmapSubtable {
+    fn create_format_4(
+        lang: u16,
+        end_code: Vec<u16>,
+        start_code: Vec<u16>,
+        id_deltas: Vec<i16>,
+    ) -> Self {
+        if HashSet::from([start_code.len(), end_code.len(), id_deltas.len()]).len() != 1 {
+            panic!("uneven parallel arrays, very bad. Very very bad.");
+        }
+
+        let seg_count: u16 = start_code.len().try_into().unwrap();
+        // Spec: Log2 of the maximum power of 2 less than or equal to segCount (log2(searchRange/2),
+        // which is equal to floor(log2(segCount)))
+        let entry_selector = (seg_count as f32).log2().floor();
+
+        // Spec: Maximum power of 2 less than or equal to segCount, times 2
+        // ((2**floor(log2(segCount))) * 2, where “**” is an exponentiation operator)
+        let search_range = 2u16.pow(entry_selector as u32).checked_mul(2).unwrap();
+
+        // if 2^entry_selector*2 is a u16 then so is entry_selector
+        let entry_selector: u16 = entry_selector as u16;
+        let range_shift: u16 = seg_count * 2 - search_range;
+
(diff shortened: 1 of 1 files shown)


rsheeter:
Per f2f this morning I am going to migrate some of the code from https://github.com/googlefonts/fontmake-rs/pull/134 to here.

write-fonts/src/tables/cmap.rs
```diff
@@ -74,9 +154,29 @@ mod tests {
             panic!("Expected a cmap4 in {encoding_record:?}");
         };
 
-        assert_eq!(start_code, to_vec(cmap4.start_code()));
-        assert_eq!(end_code, to_vec(cmap4.end_code()));
-        assert_eq!(id_delta, to_vec(cmap4.id_delta()));
-        assert_eq!(id_range_offsets, to_vec(cmap4.id_range_offsets()));
+        // The spec example says entry_selector 4 but the calculation it gives seems to yield 2 (?)
```
anthrotype, line 157: > The spec example says entry_selector 4 but the calculation it gives seems to yield 2 (?)

FWIW, also fontTools.ttLib.ttFont::getSearchRange agrees with you

write-fonts/src/tables/cmap.rs
```diff
@@ -2,62 +2,142 @@
...
+        for (cp, gid) in mappings.into_iter() {
+            let cp = (cp as u32).try_into().unwrap();
+            let next_in_run = (
+                prev.0.checked_add(1).unwrap(),
+                prev.1.checked_add(1).unwrap(),
+            );
+            let current = (cp, gid.to_u16());
+            // Codepoint and gid need to be continuous
+            if current != next_in_run {
+                // Start a new run
```
anthrotype, line 81: fonttools's `_c_m_a_p` module contain a `splitRange` function that attempts 

> to split a range of character codes into subranges with consecutive glyph IDs in such a way that the cmap4 subtable can be stored "most" efficiently

https://github.com/fonttools/fonttools/blob/444a349ef4dd1bae2d43f6ea0e06d73fb6466b9a/Lib/fontTools/ttLib/tables/_c_m_a_p.py#L783

I wonder if we should do something like that here as well

write-fonts/src/tables/cmap.rs
```diff
@@ -2,62 +2,142 @@
...
+impl CmapSubtable {
+    fn create_format_4(
+        lang: u16,
+        end_code: Vec<u16>,
+        start_code: Vec<u16>,
+        id_deltas: Vec<i16>,
+    ) -> Self {
+        if HashSet::from([start_code.len(), end_code.len(), id_deltas.len()]).len() != 1 {
+            panic!("uneven parallel arrays, very bad. Very very bad.");
+        }
```
cmyr, line 27: I would probably just write this as,

```rust
assert!(end_code.len() == start_code.len() && end_code.len() == id_deltas.len(), "msg");
```

this doesn't really matter, but importing HashSet and doing an allocation here just seems unnecessary.

write-fonts/src/tables/cmap.rs
```diff
@@ -2,62 +2,142 @@
...
+        // which is equal to floor(log2(segCount)))
+        let entry_selector = (seg_count as f32).log2().floor();
+
+        // Spec: Maximum power of 2 less than or equal to segCount, times 2
+        // ((2**floor(log2(segCount))) * 2, where “**” is an exponentiation operator)
+        let search_range = 2u16.pow(entry_selector as u32).checked_mul(2).unwrap();
+
+        // if 2^entry_selector*2 is a u16 then so is entry_selector
+        let entry_selector: u16 = entry_selector as u16;
+        let range_shift: u16 = seg_count * 2 - search_range;
```
cmyr, line 40:
```suggestion
        let entry_selector = entry_selector as u16;
        let range_shift = seg_count * 2 - search_range;
```

The type here is trivially inferred, so I don't see much advantage in declaring it explicitly?

cmyr (approved):
looks good!

write-fonts/src/tables/cmap.rs
rsheeter: I think we should but perhaps as a followon

write-fonts/src/tables/cmap.rs
```diff
@@ -2,62 +2,142 @@
...
+
+        let id_range_offsets = id_deltas.iter().map(|_| 0).collect();
```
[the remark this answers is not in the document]
rsheeter: Oh interesting, I haven't seen that before. I like it.

write-fonts/src/tables/cmap.rs
rsheeter: Deferred to https://github.com/googlefonts/fontations/issues/282

write-fonts/src/tables/cmap.rs
```diff
@@ -2,62 +2,142 @@
...
+    /// Generates a [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap) that is expected to work in most modern environments.
+    ///
+    /// For the time being just emits [format 4](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#format-4-segment-mapping-to-delta-values)
+    /// so we can drive towards compiling working fonts. In time we may wish to additionally emit format 12 to support
+    /// novel codepoints.
+    pub fn from_mappings(mappings: impl IntoIterator<Item = (char, GlyphId)>) -> Cmap {
+        let mut end_code: Vec<u16> = Vec::new();
+        let mut start_code: Vec<u16> = Vec::new();
+        let mut id_deltas: Vec<i16> = Vec::new();
+        let mut id_range_offsets: Vec<u16> = Vec::new();
```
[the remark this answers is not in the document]
rsheeter: Fixed. Also realized I no longer use id_range_offsets at all.

PR #280, opened 2023-03-13
Require fields have some means of acquiring a value

rsheeter:
When they don't, such as when a reserved field is marked `#[skip_getter]` the write types get confused.

Most of the time the source of definition for a getter-less field should be a const, typically a  `#[compile(0]`:

```rust
    /// Set to 0.
    #[skip_getter]
    #[compile(0)]
    reserved_pad: u16,
```

Sometimes not, such as font.rs:

```rust
    #[skip_getter]
    offset: u32,
```

Hoisted out of #271. In #271 attribute `user_computed` was added to allow codegen to report errors when this is done by accident while still letting special butterfly fields exist. Here I have chosen to remove the skip getter from offset instead. That required tweaks to lib.rs.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -59,7 +59,7 @@ impl Fields {
             if !fld.has_defined_value() {
                 return Err(logged_syn_error(
                     fld.name.span(),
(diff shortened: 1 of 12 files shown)


cmyr:
yep just for the record I think this is the right approach, happy to see us not need to add another attribute.

PR #284, opened 2023-03-15
make structs for types that impl FontReadWithArgs

cmyr:
Previously we passed around untyped tuples, which has caused some confusion.

Honestly: this was a bit of a hassle, and I'm not sure I love it? The main downside is that it adds a bunch of code to the codegen crate. It also complicates the syntax in our codgen inputs for specifying arguments, as you need to be able to specify the name of the argument for which you are providing a value, unless the value and the argument already share a name.

If we keep it, I need to go clean up some docs.

(I do have an idea for how this might be improved, by using the `fn pick_speed(Opts { go_fast, how_fast }: Opts) -> ()` syntax, but this involves changing the API so we always pass arguments by value. This is probably fine, and I'll explore it separately, probably tomorrow)

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -2,7 +2,7 @@
 
 use std::{borrow::Cow, ops::Deref};
 
-use proc_macro2::TokenStream;
+use proc_macro2::{Span, TokenStream};
 use quote::{quote, ToTokens};
 use syn::spanned::Spanned;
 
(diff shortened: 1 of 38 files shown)


rsheeter:
> I'm not sure I love it?

I agree ... but I think the upside over passing (u16, u16) to say hmtx is real. I do immediately wonder about other options. For example, what if we used named args until you hit N and then switched to an arg struct? - in other languages that would be a very normal pattern.

cmyr:
To reiterate what I'd said in code review: the reason that we don't just generate custom constructors is that we want to express this idea in a trait, so that we have a common pattern we can use throughout the generated code. I'm happy to generate custom constructors *as well*, though? I'll do that in a separate PR, if it's good enough than maybe we won't actually need to add these types.

cmyr:
superseded by #288

PR #285, opened 2023-03-15
Make ot_round more widely accessible and comment it

rsheeter:
Put it in write-fonts because `floor()` is not available in font-types. Modelled on std::ops, e.g. https://doc.rust-lang.org/std/ops/trait.Sub.html.


cmyr (approved):
I wouldn't change a thing :)

PR #288, opened 2023-03-15
Better read arg constructors

cmyr:
I wanted to write this separately from #284, since maybe it is actually just a better solution overall? With this, the user should never need to interact with `FontReadWithArgs` at all, and can use the generated methods, where the arguments get names.

I used `read` instead of `new` for the method name because that's the name we generally use for constructors in `read-fonts`, otherwise I would prefer `new`.


rsheeter (approved):
As a user this makes my life better, lgtm.

anthrotype (approved):

anthrotype:
> I used read instead of new

I think `read` fits better in this case than `new`, because there is also the raw "data" parameter and a 'reading' of this data is involved; it's not merely a constructor where the parameters passed in are more or less set to the instance.

cmyr:
Great, I'm going to close #284, I don't think the added complexity is justified if we have these alternative constructors.

PR #292, opened 2023-03-19
[skrifa] cleanup and better encapsulate hinting

dfrg:
This patch doesn't have any functional changes. Just a bit of code shuffling:
* Removes the unnecessary intermediate `GlyphScaler` type (and the double lifetime!)
* Moves font "instance" parameters (size, scale and coords) to the `ScalerFont` type
* Encapsulates hinting state in a new module

This is mostly so I can maintain the bytecode interpreter in a fork (for now!) without requiring changes to the main scaler code. It's also just better organized all around.


cmyr (approved):
take that, entropy!

PR #294, opened 2023-03-21
[read-fonts] add UVS mapping for cmap14

dfrg:
Adds a new `map_variant` method to `Cmap14` to support Unicode variation sequences.


read-fonts/build.rs
```diff
@@ -9,6 +9,7 @@
 use std::path::Path;
 
 static FILES_TO_MOVE: &[&str] = &[
+    "../resources/test_fonts/ttf/cmap14_font1.ttf",
```
rsheeter, line 12: This reminds me, it's fairly easy to forget this and find out when the file isn't available and you try to use it. Should we just copy ttf/*.ttf?

read-fonts/src/tables/cmap.rs
```diff
@@ -91,6 +102,57 @@ impl<'a> Cmap12<'a> {
...
+                let range = &ranges[i];
+                let start = range.start_unicode_value().into();
+                if codepoint < start {
+                    hi = i;
+                } else if codepoint > (start + range.additional_count() as u32) {
+                    lo = i + 1;
+                } else {
+                    return Some(MapVariant::UseDefault);
+                }
+            }
```
rsheeter, line 142: Why are you using binary_search_by below but not above? - when I tried it, convinced there was some problem with the idea I hadn't spotted, it passed the tests and the code felt better:

```rust
if default_uvs.ranges().binary_search_by(|range| {
    let start = range.start_unicode_value().into();
    if codepoint < start {
        Ordering::Greater
    } else if codepoint > (start + range.additional_count() as u32) {
        Ordering::Less
    } else {
        Ordering::Equal
    }
}).is_ok() {
    return Some(MapVariant::UseDefault);
}
```

read-fonts/src/tables/cmap.rs
```diff
@@ -91,6 +102,57 @@ impl<'a> Cmap12<'a> {
...
+        }
+        // Binary search the non-default UVS table if present. This maps codepoint+selector to a variant glyph.
+        let non_default_uvs = selector_record.non_default_uvs(self.offset_data())?.ok()?;
+        let mapping = non_default_uvs.uvs_mapping();
+        let ix = mapping
+            .binary_search_by(|rec| <_ as Into<u32>>::into(rec.unicode_value()).cmp(&codepoint))
```
rsheeter, line 148: This line is outstandingly illegible. Perhaps we can afford one more line to make it legible? - IIUC it's saying the same thing as:

```rust
.binary_search_by(|rec| {
    let rec_codepoint: u32 = rec.unicode_value().into();
    rec_codepoint.cmp(&codepoint)
})
```

rsheeter (approved):
LGTM w/a question and a nit

read-fonts/build.rs
dfrg: We should and it would also be nice to generate the file that includes the font data as well. I'd say this belongs in another PR though.

read-fonts/src/tables/cmap.rs
dfrg: Made it so. Also did the same for the similar code above.

PR #297, opened 2023-03-21
Fix Cmap::from_mappings

rsheeter:
Naively you'd think Unicode would be but you would be wrong; per @anthrotype fontmake emits Windows cmaps because they are the most widely supported.

Also, it helps a lot with binary search if things are in order.


cmyr (approved):
👍

PR #302, opened 2023-03-22
Add all off curve ⏰ unit test for to_path

dfrg:
To prevent the most recent bug from happening again.

Just merge?

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -813,6 +813,49 @@ mod tests {
     use crate::{FontRef, GlyphId, TableProvider};
 
     #[test]
+    fn all_off_curve_to_path() {
+        fn pt(x: i32, y: i32) -> Point<F26Dot6> {
+            Point::new(x, y).map(F26Dot6::from_bits)
+        }
+        let points = [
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/glyf.rs
```diff
@@ -812,6 +812,49 @@ mod tests {
...
+                self.0.push_str(&format!(
+                    "C{cx0:.3},{cy0:.3} {cx1:.3},{cy1:.3} {x:.3},{y:.3} "
+                ));
+            }
+            fn close(&mut self) {
+                self.0.push_str("z ");
+            }
+        }
+        let mut pen = SvgPen(String::default());
+        to_path(&points, &flags, &contours, &mut pen).unwrap();
```
rsheeter, line 854: I think you need to either choose more distinctive points or comment the test; it's quite hard to tell what  it's meant to detect/prevent if you don't know to begin with.

Fewer points and int points would make it all more readable.

dfrg: I tried to be cheeky by using a contour from the actual outline that caused the failure, but the test was a bit dense.

Reduced the number of points to 4 and made them "rounder" numbers.

PR #304, opened 2023-03-26
Generate a write type for avar

rsheeter:
Required for https://github.com/googlefonts/fontmake-rs/issues/208. JMM.

Almost a pure codegen addition, just had to make MajorMinor Default so avar would compile.


dfrg (approved):
in `resources/codgen_inputs/avar.rs` we should add `compile` attributes for `Avar::axis_count` and `SegmentMaps::position_map_count`. I believe these should just be `#[compile(array_len($field))]`. 

Otherwise, looks good.

font-types/src/version.rs
```diff
@@ -13,7 +13,7 @@ pub struct Version16Dot16(u32);
 /// represented as a `major_version`, `minor_version` pair. This type encodes
 /// those as a single type, which is useful for some of the generated code that
 /// parses out a version.
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
cmyr, line 16: So I'm way late on this, but the pattern we use elsewhere is to explicitly compute the version, using the `#[compile]` annotation. As written, it would be easy to accidentally write out an `avar` table with the version (0, 0), which is invalid.

PR #306, opened 2023-03-26
Add fvar to write-fonts

rsheeter:
Required for https://github.com/googlefonts/fontmake-rs/issues/209.


write-fonts/src/tables/instance_record.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Optional. The name ID for entries in the 'name' table that provide PostScript names for this instance.
+    pub post_script_name_id: Option<NameId>,
+}
+
+impl FontWrite for InstanceRecord {
+    fn write_into(&self, writer: &mut TableWriter) {
+        self.subfamily_name_id.write_into(writer);
+        self.flags.write_into(writer);
+        self.coordinates.write_into(writer);
+        self.post_script_name_id
```
dfrg, line 27: This is generating the full instance record unconditionally even though the instance size is computed based on the presence of at least one non-None PostScript name. I think we’ll generate an invalid table if all are None.

dfrg (approved):
Generally looks good modulo one logic bug.

I’m not sure if you intended avar to be included here as well but it seems fine to me.

write-fonts/src/tables/instance_record.rs
rsheeter: Note to self: trust None means do not emit and instead update this to be all Some / all none in set_instancesize.

rsheeter: Fixed to only write if present. Having trouble getting validation wired; filed #308

PR #310, opened 2023-03-27
Add the ability to attach validation to a table

rsheeter:
And use it on fvar and prove it works.

### DIFF
--- a/write-fonts/src/tables/fvar.rs
+++ b/write-fonts/src/tables/fvar.rs
@@ -8,6 +8,19 @@ pub use instance_record::InstanceRecord;
 include!("../../generated/generated_fvar.rs");
 
 impl Fvar {
+    /// We need everyone to have, or not have, post_script name so we can have a single record size.
+    pub fn check_instance_size(&self, ctx: &mut ValidationCtx) {
+        let sum: i32 = self
+            .axis_instance_arrays
+            .instances
+            .iter()
+            .map(|ir| ir.post_script_name_id.map(|_| 1).unwrap_or(-1))
+            .sum();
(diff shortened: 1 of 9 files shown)


write-fonts/src/tables/fvar.rs
```diff
@@ -8,6 +8,19 @@ pub use instance_record::InstanceRecord;
 include!("../../generated/generated_fvar.rs");
 
 impl Fvar {
+    /// We need everyone to have, or not have, post_script name so we can have a single record size.
+    pub fn check_instance_size(&self, ctx: &mut ValidationCtx) {
```
dfrg, line 12: @cmyr do these need to be public?

write-fonts/src/tables/fvar.rs
```diff
@@ -8,6 +8,19 @@ pub use instance_record::InstanceRecord;
...
+        let sum: i32 = self
+            .axis_instance_arrays
+            .instances
+            .iter()
+            .map(|ir| ir.post_script_name_id.map(|_| 1).unwrap_or(-1))
```
dfrg, line 17: Cleverness noted :)

While we’re here, we should probably also check that `instance.coordinates.len() == axis_count` for all instances.

dfrg (approved):
Looks good with a suggestion for an additional check.

write-fonts/src/tables/fvar.rs
rsheeter: That's a good point, let me see if anything kerplodes if not.

rsheeter: It appears they do not, unless there's some reason someone would call them from The Outside. For now I will update to not be.

PR #312, opened 2023-03-28
[skrifa] localized strings

dfrg:
This adds a high level interface for reading localized strings (otherwise known as names in the OpenType spec). The API is broadly modeled after both Skia and DirectWrite.

Re-exports `font_types::NameId` as `StringId` for naming consistency.

Adds a `StringIds` iterator (accessible from `MetadataProvider::string_ids()`) that yields all available identifiers in the name table.

Also adds a `LocalizedStrings` iterator (accessible from `MetadataProvider::localized_strings(id: StringId)`) that yields all name entries for a given identifier.

Language ids and tags are converted to BCP-47 language identifiers.

Additionally, changes the read-fonts build script to copy all fonts in ttf/ and all files in extracted/ to the output directory. Figured I'd make this change now since I added another test font.


skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+    #[test]
+    fn localized() {
+        let font = FontRef::new(read_fonts::test_data::test_fonts::NAMES_ONLY).unwrap();
+        let mut subfamily_names = font
+            .localized_strings(StringId::SUBFAMILY_NAME)
+            .map(|s| (s.language().unwrap().to_string(), s.to_string()))
+            .collect::<Vec<_>>();
+        subfamily_names.sort_by(|a, b| a.0.cmp(&b.0));
+        let expected = [
+            (String::from("el-GR"), String::from("Κανονικά")),
```
drott, line 652: Could we test a non-European language for a result here as well, Arabic and Chinese in particular? (the intent being to cover the UTF-8 conversion handling.)

resources/test_fonts/ttx/names_only.ttx
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.37">
+  <name>
+    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
```
drott, line 4: The Skia internal implementation seems to support at least MacRoman for platformID=Macintosh. What happens if that occurs in a name table with this impl?

drott (commented):
Great to see this taking shape - I am not really in a position yet to review Rust codes, I think others are better at that, just some curious comments.

resources/test_fonts/ttx/names_only.ttx
drott: I found https://github.com/googlefonts/fontations/blob/main/read-fonts/src/tables/name.rs#L185 - so do I understand it right that the name table support in read_fonts handles the decoding for retrieving `name_string()` and does support `MacRoman`?

skrifa/src/meta/strings.rs
dfrg: Absolutely. I’ll add subfamily names in those languages.

resources/test_fonts/ttx/names_only.ttx
dfrg: Yes, we do support Mac Roman and the actual decoding occurs in the `read-fonts` name table implementation. I’ll change the platform id for the English subfamily name in the names font so that it’s represented in the test.

skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+};
+
+use core::fmt;
+
+#[doc(inline)]
+pub use read_fonts::types::NameId as StringId;
+
+#[derive(Clone)]
+/// Iterator over the characters of a string.
+pub struct Chars<'a> {
```
cmyr, line 34: total nit:
```suggestion
/// Iterator over the characters of a string.
#[derive(Clone)]
pub struct Chars<'a> {
```

skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+        let name = font.name().ok();
+        let records = name
+            .as_ref()
+            .map(|name| name.name_record().iter())
+            .unwrap_or([].iter());
+        Self {
+            records,
+            last_id: None,
+        }
+    }
```
cmyr, line 64: This feels like a minor mixing of concerns to me, although I'm not sure it's that important?

Basically: if we don't have a name table, it would make sense to me to handle that by returning an `Option<StringIds<'a>>` from the method on metadata provider, and I would maybe make this constructor take a name table directly? To put it another way, this method is silently handling an error condition in a way that makes it difficult to identify. There might be a good reason for this choice, but it does jump out at me.

skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+                return id;
+            }
+        }
+    }
+}
+
+/// Iterator over a collection of localized strings for a specific identifier.
+#[derive(Clone)]
+pub struct LocalizedStrings<'a> {
+    name: Option<Name<'a>>,
```
cmyr, line 85: again I find it slightly odd that we will create this iterator even if there is no `Name` table?

skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+}
+
+/// This value is chosen arbitrarily to accommodate common language tags that
+/// are almost always <= 11 bytes (LLL-SSSS-RR where L is primary language, S
+/// is script and R is region) and to keep the Language enum at a reasonable
+/// 32 bytes in size.
+const MAX_INLINE_LANGUAGE_LEN: usize = 30;
+
+#[derive(Copy, Clone, Debug)]
+#[repr(u8)]
```
cmyr, line 191: just curious what the motivation is for `repr(u8)` here?

skrifa/src/meta/strings.rs
```diff
@@ -0,0 +1,683 @@
...
+    },
+    Static(&'static str),
+}
+
+impl Language {
+    fn new(name: &Name, record: &NameRecord) -> Option<Self> {
+        let language_id = record.language_id();
+        // For version 1 name tables, prefer language tags:
+        // https://learn.microsoft.com/en-us/typography/opentype/spec/name#naming-table-version-1
+        if name.version() == 1 && language_id >= 0x8000 {
```
cmyr, line 205: would love a named constant in place of `0x8000`

cmyr (approved):
A few little things inline, but looks good!

read-fonts/build.rs
```diff
@@ -8,23 +8,26 @@
...
-static FILES_TO_MOVE: &[&str] = &[
-    "../resources/test_fonts/ttf/cmap14_font1.ttf",
-    "../resources/test_fonts/ttf/linear_gradient_rect_colr_1.ttf",
-    "../resources/test_fonts/ttf/simple_glyf.ttf",
-    "../resources/test_fonts/ttf/vazirmatn_var_trimmed.ttf",
-    "../resources/test_fonts/extracted/vazirmatn_var_trimmed-glyphs.txt",
-];
-
 fn main() {
+    println!("cargo:rerun-if-changed=../resources/test_fonts/");
```
anthrotype, line 12: you may need to add `read-fonts/build.rs` to the `resources/scripts/println_ignore.txt` allowlist, otherwise after #313 was merged the prepush check for println! would trigger at this line (merge or rebase `main` to verify)

dfrg: Thanks for catching this! I didn't realize that PR had been merged. Fixed.

skrifa/src/meta/strings.rs
dfrg: This is mostly about ergonomics. The intent, specifically with metadata, was to not surface these types of things. My expectation is that for anyone consuming this library, the logic for dealing with `Option<StringIds<'a>>` and an empty iterator would be exactly the same so I'd like to not impose useless boilerplate.

Generally, I'd expect implementations that require a minimum set of (valid?) tables to perform that validation at font loading time using the more fine-grained error reporting provided by `read-fonts`.

dfrg:
Turns out that `string_ids()` cannot be implemented to return a unique set in O(n) without allocating due to the sort order in the table. Since this functionality is not provided by any other API, I’m going to remove it.

In lieu of this, I’ll add a `NameId::well_known()` method in `font-types` that returns an iterator over the standard name set. It’s not useful to list the names beyond those without context anyway and users can dig into the `read-fonts` crate for deeper introspection of the `name` table if necessary.

skrifa/src/meta/strings.rs
dfrg: I think what I'd like to do here is document this behavior and add a method that reports the set of supported metadata that is available in the font based on the presence of required tables. This lets client code determine if a font is considered usable for its purposes at load time rather than having to deal with errors at lower levels in the stack. This feels better to me for integration in full scale text rendering systems but I'd be happy to discuss further in the forthcoming PR.

resources/test_fonts/ttx/names_only.ttx
```diff
@@ -0,0 +1,26 @@
...
+      Κανονικά
+    </namerecord>
+    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x415">
+      Normalny
+    </namerecord>
+    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x42d">
+      Arrunta
+    </namerecord>
+    <namerecord nameID="2" platformID="3" platEncID="1" langID="0x804">
+      正常
```
drott, line 20: Thanks for adding that, one thing that came to my mind after the review yesterday: IIRC we've had bugs in Chromium where we didn't handle the encoding correctly, or the font had the wrong encoding. I tried finding the bug, but what this raised: To what extent do we want to test how we handle broken fonts here? What happens if there's invalid UTF-8 in the font? (I assume the records are stored in UTF-8?)

dfrg: Unicode names are encoded in UTF-16BE. The decoding is handled here: https://github.com/googlefonts/fontations/blob/e3dc30dca577cee306c4256d0ef8c1e31fc88dfe/read-fonts/src/tables/name.rs#L147

Invalid characters are emitted as U+FFFD (Unicode replacement character) so it’s a potentially lossy conversion.

dfrg: It looks like if the buffer ends with the first half of a surrogate pair, it is simply dropped. We probably want to emit a replacement character in that case as well.

PR #313, opened 2023-03-29
add pre-push git hook to check for accidental println! statements

anthrotype:
similar to https://github.com/googlefonts/fontmake-rs/pull/204, with the only difference that in the case of fontations there are a couple of files where the use of `println!` is intentional and legitimate (font-codegen::preprocessor or otexplorer) so I added an allow-list to be able to exclude those from the check.


cmyr (approved):
Looks good, thanks @anthrotype !

PR #315, opened 2023-03-30
Fix cmap4 construction for large deltas

rsheeter:
Test exposes and fixes what I believe to be the issue underlying https://github.com/googlefonts/fontmake-rs/issues/223.

I have also tested by building Texturina with this fontmake-rs with this fix, fontmake-rs no longer kerplodes and produces a cmap that ttx can read.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -71,21 +71,29 @@ impl Cmap {
 
         let mut prev = (u16::MAX - 1, u16::MAX - 1);
         for (cp, gid) in mappings.into_iter() {
+            let gid = gid.to_u16();
             let cp = (cp as u32).try_into().unwrap();
             let next_in_run = (
                 prev.0.checked_add(1).unwrap(),
                 prev.1.checked_add(1).unwrap(),
             );
-            let current = (cp, gid.to_u16());
+            let current = (cp, gid);
             // Codepoint and gid need to be continuous
             if current != next_in_run {
                 // Start a new run
                 start_code.push(cp);
                 end_code.push(cp);
 
-                // we might need to reach further than an i16 can take us
-                // using idRangeOffset ... but we're saving that for another day
-                id_deltas.push((gid.to_u16() as i32 - cp as i32).try_into().unwrap());
+                // TIL Python % 0x10000 and Rust % 0x10000 do not mean the same thing.
+                // rem_euclid is almost what we want, except as applied to small values
+                // ex -10 rem_euclid 0x10000 = 65526
+                let delta: i32 = gid as i32 - cp as i32;
+                let delta = if let Ok(delta) = TryInto::<i16>::try_into(delta) {
+                    delta
+                } else {
+                    delta.rem_euclid(0x10000).try_into().unwrap()
+                };
+                id_deltas.push(delta);
             } else {
                 // Continue the prior run
                 let last = end_code.last_mut().unwrap();
@@ -201,5 +209,21 @@ mod tests {
         }
         assert_ne!(ordered, disordered);
         assert_generates_simple_cmap(disordered);
+    }
+
+    #[test]
+    fn generate_cmap4_large_values() {
+        let mut mappings = simple_cmap_mappings();
+        // Example from Texturina.
+        let codepoint = char::from_u32(0xa78b).unwrap();
+        let gid = GlyphId::new(153);
+        mappings.push((codepoint, gid));
+
+        let cmap = write::Cmap::from_mappings(mappings);
+
+        let bytes = dump_table(&cmap).unwrap();
+        let font_data = FontData::new(&bytes);
+        let cmap = Cmap::read(font_data).unwrap();
+        assert_eq!(cmap.map_codepoint(codepoint), Some(gid));
     }
 }


anthrotype:
```python
from fontTools.fontBuilder import FontBuilder
fb = FontBuilder(1000)
fb.setupGlyphOrder([".notdef"] + [f"glyph{i}" for i in range(1, 154)])
fb.setupCharacterMap({0xA78B: "glyph153"})
font = fb.font
cmap = font["cmap"]
del cmap.tables[1]  # fonttools also builds a platformID=0 subtable, we don't so I just delete this to be able to compare
data = cmap.compile(font)
print(data)
# b"\x00\x00\x00\x01\x00\x03\x00\x01\x00\x00\x00\x0c\x00\x04\x00 \x00\x00\x00\x04\x00\x04\x00\x01\x00\x00\xa7\x8b\xff\xff\x00\x00\xa7\x8b\xff\xffY\x0e\x00\x01\x00\x00\x00\x00"
```

```rust
use std::collections::HashMap;
use font_types::GlyphId;
use read_fonts::{FontData, FontRead};
use read_fonts::tables::cmap as read;
use write_fonts::tables::cmap as write;

fn main() {
    let bytes = b"\x00\x00\x00\x01\x00\x03\x00\x01\x00\x00\x00\x0c\x00\x04\x00 \x00\x00\x00\x04\x00\x04\x00\x01\x00\x00\xa7\x8b\xff\xff\x00\x00\xa7\x8b\xff\xffY\x0e\x00\x01\x00\x00\x00\x00";
    let font_data = FontData::new(bytes);
    let cmap = read::Cmap::read(font_data).unwrap();
    println!("fonttools: {:#?}", cmap);
    let mut mappings: HashMap<char, GlyphId> = HashMap::new();
    mappings.insert('\u{A78B}', GlyphId::new(153));
    let cmap2 = write::Cmap::from_mappings(mappings);
    println!("fontations/cmap_giddy: {:#?}", cmap2);
}
```

this prints:

```
fonttools: Cmap {
    version: 0,
    num_tables: 1,
    encoding_records: [
        EncodingRecord {
            platform_id: 3,
            encoding_id: 1,
            subtable_offset: Cmap4 {
                format: 4,
                length: 32,
                language: 0,
                seg_count_x2: 4,
                search_range: 4,
                entry_selector: 1,
                range_shift: 0,
                end_code: [
                    42891,
                    65535,
                ],
                start_code: [
                    42891,
                    65535,
                ],
                id_delta: [
                    22798,
                    1,
                ],
                id_range_offsets: [
                    0,
                    0,
                ],
                glyph_id_array: [],
            },
        },
    ],
}
fontations/cmap_giddy: Cmap {
    encoding_records: [
        EncodingRecord {
            platform_id: Windows,
            encoding_id: 1,
            subtable: OffsetMarker {
                obj: Format4(
                    Cmap4 {
                        length: 34,
                        language: 0,
                        seg_count_x2: 4,
                        search_range: 4,
                        entry_selector: 1,
                        range_shift: 0,
                        end_code: [
                            42891,
                            65535,
                        ],
                        start_code: [
                            42891,
                            65535,
                        ],
                        id_delta: [
                            0,
                            1,
                        ],
                        id_range_offsets: [
                            4,
                            0,
                        ],
                        glyph_id_array: [
                            153,
                        ],
                    },
                ),
            },
        },
    ],
}
```

Note how in the fonttools' cmap subtable, the id_range_offsets are all 0 and the glyph_id_array is empty, and the id_delta for the first segment is 22798; whereas in fontations' cmap, the id_delta is 0 and the id_range_offsets/glyph_id_array are used.

If you take start_code=42891 and id_delta=22798 and sum them up using modulo 65536 like the spec says, then you get 153, which is the expected glyph index. No need for the extra id_range_offsets and glyph_id_array in this case.

rsheeter:
Oh if I just missed the mod 65536 line that's easier which would be nice.

write-fonts/src/tables/cmap.rs
```diff
@@ -65,46 +107,84 @@ impl Cmap {
...
-            let cp = (cp as u32).try_into().unwrap();
-            let next_in_run = (
-                prev.0.checked_add(1).unwrap(),
-                prev.1.checked_add(1).unwrap(),
-            );
-            let current = (cp, gid.to_u16());
-            // Codepoint and gid need to be continuous
-            if current != next_in_run {
+            let cp = cp as u32;
+            let delta = gid.to_u16() as i64 - cp as i64;
```
anthrotype, line 120: > Oh if I just missed the mod 65536 line that's easier which would be nice.

yes, here you should do % 65536 just like fonttools does, then it fill fit:

```python
idDelta.append((indices[0] - startCode[i]) % 0x10000)
```

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_c_m_a_p.py#L1003

rsheeter: Done

rsheeter: Hm. I still get overflows.

anthrotype: (153 - 0xa78b) % 0x10000) gives me 22798 in python. no idea

rsheeter: TIL Rust % means remainder and Python % means modulus.

rsheeter: OK, so the simple solution here is to do what Python means by %. Good catch.

dfrg (approved):

write-fonts/src/tables/cmap.rs
```diff
@@ -71,21 +71,29 @@ impl Cmap {
...
                 // Start a new run
                 start_code.push(cp);
                 end_code.push(cp);
 
-                // we might need to reach further than an i16 can take us
-                // using idRangeOffset ... but we're saving that for another day
-                id_deltas.push((gid.to_u16() as i32 - cp as i32).try_into().unwrap());
+                // TIL Python % 0x10000 and Rust % 0x10000 do not mean the same thing.
+                // rem_euclid is almost what we want, except as applied to small values
+                // ex -10 rem_euclid 0x10000 = 65526
```
anthrotype, line 89: Hm fonttools uses _unsigned_ uint16 for idDeltas, so just doing `(indices[i] - startCode[i]) % 0x10000` always works even for very small indices as in your example above -10 % 0x10000 == 65526:

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_c_m_a_p.py#L778
note the use of "H" for initializing array.array, which means unsigned short integer (lowercase "h" would mean signed):
https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_c_m_a_p.py#L1018
there is how it obtains back a gid by summing up (charCode + idDelta) % 0xFFFF, in the subtable4 decompile method:  
https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/tables/_c_m_a_p.py#L913

However, you're right that the spec uses int16:
https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#format-4-segment-mapping-to-delta-values

I wonder if it used to be uint16 at some point in the past, then the spec was updated later on..
In any case, I don't think it matters the difference, since those idDeltas are not exposed to the public API in fonttools and only used in the compile/decompile steps. We seem to be getting the same numbers in the end, right?

anthrotype: look what I found: https://github.com/fonttools/fonttools/issues/2575

anthrotype: interestingly, in the linked discussion, https://github.com/MicrosoftDocs/typography-issues/issues/980, Peter Constable notes

> this would have been made easier if the type for idDelta were kept as unsigned as it was in the original TrueType spec...

anthrotype: I see when read-fonts decompiles the glyph index it performs a cast to u16 immediately after summing codepoint and the (signed i16) delta, like this `(codepoint as i32 + delta) as u16`
https://github.com/googlefonts/fontations/blob/1f85b76a4697c03221388cfd430fff9b4664a3b5/read-fonts/src/tables/cmap.rs#L61

So in the end it doesn't matter whether we interpet idDeltas as signed (as OT spec and fontations do) or as unsigned (as fonttools and old TrueType spec do), the end result is the same.

anthrotype (approved):

write-fonts/src/tables/cmap.rs
rsheeter: I didn't realize it had changed signedness, thats amazing.

PR #316, opened 2023-03-30
[skrifa] character mapping

dfrg:
This adds a high level interface for mapping characters (and Unicode variation sequences) to glyph identifiers. Implements a search strategy for the best mapping subtables based on a combination the methods used by FreeType and HarfBuzz (which prioritizes symbol mapping subtables).


skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+        assert_eq!(charmap.map(' '), None);
+        assert_eq!(charmap.map(0x101723_u32), Some(GlyphId::new(23)));
+        assert_eq!(charmap.map(0x101725_u32), Some(GlyphId::new(25)));
+        assert_eq!(charmap.map(0x102523_u32), Some(GlyphId::new(53)));
+        assert_eq!(charmap.map(0x102526_u32), Some(GlyphId::new(56)));
+        assert_eq!(charmap.map(0x102527_u32), Some(GlyphId::new(57)));
+    }
+
+    #[test]
+    fn map_symbol_pua() {
```
drott, line 341: You're describing above that symbol encoding will get selected over a regular encoding - can we cover that in tests as well?

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    fn map_impl(&self, codepoint: u32) -> Option<GlyphId> {
+        match &self.subtable {
+            UnicodeSubtable::Format4(subtable) => subtable.map_codepoint(codepoint),
+            UnicodeSubtable::Format12(subtable) => subtable.map_codepoint(codepoint),
+        }
+    }
+
+    fn adjust_symbol_pua(codepoint: u32) -> Option<u32> {
```
drott, line 172: Nit, naming:

```suggestion
    fn shift_symbol_to_pua(codepoint: u32) -> Option<u32> {
```

drott (commented):
Great to see this, looking forward to integrating that with Skia. A couple of comments below.

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+pub struct Charmap<'a> {
+    map: Option<Map<'a>>,
+    variant_map: Option<Cmap14<'a>>,
+}
+
+impl<'a> Charmap<'a> {
+    /// Creates a new character map from the given font.
+    pub fn new(font: &impl TableProvider<'a>) -> Self {
+        if let Ok(cmap) = font.cmap() {
+            let (maps, map, variant_map) = find_best_mappings(&cmap);
```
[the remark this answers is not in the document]
dfrg: Yes, this code is a bit dense. I’ll rename/reorganize a bit to add some clarity.

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+};
+
+pub use read_fonts::tables::cmap::MapVariant;
+
+/// Mapping of characters to nominal glyph identifiers.
+///
+/// The mappings are derived from the [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap)
+/// table.
+///
+/// ## Supported formats
```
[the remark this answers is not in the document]
dfrg: Great question. My understanding is that the non-Unicode tables are simply unsupported. While both FreeType and HarfBuzz (and read-fonts) are capable of parsing them, they are never selected unless specifically requested because all inputs are expected to be Unicode and none of these libraries contain internal converters for legacy encodings. I’ll investigate this some more.

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+    UnicodeBmp = 1,
+    UnicodeFull = 2,
+    Symbol = 3,
+}
+
+/// Searches the `cmap` table for the best Unicode subtables.
+///
+/// Returns the `MappingIndex` accelerator along with character and UVS
+/// subtables.
+fn find_best_mappings<'a>(
```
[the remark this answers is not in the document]
dfrg: Will make this more descriptive.

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+    MappingIndex,
+    Option<UnicodeSubtable<'a>>,
+    Option<Cmap14<'a>>,
+) {
+    const ENCODING_MS_SYMBOL: u16 = 0;
+    const ENCODING_MS_UNICODE_CS: u16 = 1;
+    const ENCODING_APPLE_ID_UNICODE_32: u16 = 4;
+    const ENCODING_APPLE_ID_VARIANT_SELECTOR: u16 = 5;
+    const ENCODING_MS_ID_UCS_4: u16 = 10;
+    let mut maps = MappingIndex::default();
```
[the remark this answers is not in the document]
dfrg: Will update this.

skrifa/src/meta/charmap.rs
dfrg: Good catch! I’ll add a non symbol subtable to the test font and an additional test to ensure the correct subtable is selected.

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,374 @@
...
+/// selection (greater is better).
+#[derive(Copy, Clone, PartialEq, PartialOrd)]
+enum MapKind {
+    None = 0,
+    UnicodeBmp = 1,
+    UnicodeFull = 2,
+    Symbol = 3,
+}
+
+/// Searches the `cmap` table for the best Unicode subtables.
```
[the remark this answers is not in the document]
dfrg: Indeed. I’ll make this more clear.

skrifa/src/meta/charmap.rs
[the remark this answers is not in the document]
dfrg: Based on this: https://skia.googlesource.com/skia/+/c7b7b8fff26925debf40d30ae79b79b959fc91c3/src/ports/SkFontHost_FreeType.cpp#403, it looks like Skia accepts the default Unicode mapping or falls back to symbol if no Unicode mapping exists. I don't see code for selecting any other encodings which leads me to believe that we won't suffer regressions with this PR.

The fallback order does imply that we are doing the opposite here with regard to the preference for a symbol encoding. Since HarfBuzz does prioritize symbol, and per our chat last week, we discovered that the HB code is used for character mapping in Blink, I'm inclined to keep the strategy here as is.

Thoughts?

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    /// Returns true if a Unicode variation sequence mapping is available.
+    pub fn has_variant_map(&self) -> bool {
+        self.variant_map.is_some()
+    }
+
+    /// Maps a character to a nominal glyph identifier. Returns `None` if a mapping does
+    /// not exist.
+    pub fn map(&self, ch: impl Into<u32>) -> Option<GlyphId> {
```
drott, line 82: More references:
* https://bugs.chromium.org/p/chromium/issues/detail?id=589340 discussion on differences when moving to HarfBuzz glyph lookup functions, and at the time found that Firefox does not seem to support symbol cmaps.
* But also: https://github.com/harfbuzz/harfbuzz/issues/1918#issuecomment-693630196 - where @behdad explains that CoreText does not prioritise symbol.

I had two things in mind: One would be: 
1) We leave it up the application control this and provide a `map_symbol()` method in addition where mapping through the symbol table is performed when one is found, and always select a non-symbol cmap for `fn map()`. 
2) Or we leave as is with the reasoning that if someone puts a symbol table in their font, they probably want it to work for looking up symbols primarily - if there are other glyphs in that font, it could be a regular cmap lookup with PUA codepoint lookup results.

Option 2. has less magic, but might not be 100% required at the moment, given that we did not encounter any issues with that in Chrome to my knowledge for the past couple years.

skrifa/src/meta/charmap.rs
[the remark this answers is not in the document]
drott: Sorry for breaking threading, commented here: https://github.com/googlefonts/fontations/pull/316/files#r1157276977

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,415 @@
...
+    /// subtables.
+    mapping_index: MappingIndex,
+    /// Either a symbol subtable or the Unicode subtable with the
+    /// greatest coverage.
+    map: Option<UnicodeSubtable<'a>>,
+    /// Subtable that supports mapping Unicode variation sequences.
+    variant_map: Option<Cmap14<'a>>,
+}
+
+impl<'a> MappingSelection<'a> {
```
drott, line 251: Thanks for moving the algorithm here, that makes it indeed easier to read.

skrifa/src/meta/charmap.rs
dfrg: Thank you for the additional context! Shipping for several years in both Chrome and HarfBuzz feels like really strong precedent so my preference would be to go with option 2 and not diverge from current behavior unless we find some reason to do otherwise.

skrifa/src/meta/charmap.rs
dfrg: I agree. Thanks for the suggestion

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,415 @@
...
+/// or a Unicode subtable with the greatest coverage.
+///
+/// For `variant_map`, best means a format 14 subtable.
+struct MappingSelection<'a> {
+    /// The mapping index accelerator that holds indices of the following
+    /// subtables.
+    mapping_index: MappingIndex,
+    /// Either a symbol subtable or the Unicode subtable with the
+    /// greatest coverage.
+    map: Option<UnicodeSubtable<'a>>,
```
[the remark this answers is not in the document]
dfrg: No worries! Would `default_map` read better to you here? I’d like something that aligns nicely with `variant_` for symmetry between the two.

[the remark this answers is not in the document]
drott: 🤔 `default_map` does not describe it more IMO. How about `char_map` or `codepoint_map` - with that we'd have the symmetry?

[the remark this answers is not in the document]
dfrg: I like `codepoint_map`. Will make it so.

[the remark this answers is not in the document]
dfrg: I renamed this to `codepoint_subtable` to better capture what it actually holds. I also replaced `map`/`Map` accordingly for other private types and fields.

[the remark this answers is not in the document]
drott: Sounds good, thanks.

skrifa/src/meta/charmap.rs
drott: Yes, ok with me. I can't foresee that precisely yet, but due to requirements of another client, we may need more control, similar to `FT_Select_Charmap` here in the future, but we can revisit this then.

dfrg: Happy to revisit if a client requests more fine grained control.

drott (approved):
Thanks, LGTM!

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,420 @@
...
+            Default::default()
+        }
+    }
+
+    /// Creates a new character map for the given font using the tables referenced by
+    /// the precomputed indices.
+    ///
+    /// The font should be the same as the one used to construct this object.
+    pub fn charmap<'a>(&self, font: &impl TableProvider<'a>) -> Charmap<'a> {
+        if let Ok(cmap) = font.cmap() {
```
rsheeter, line 130: guard clause?

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,420 @@
...
+) -> Option<CmapSubtable<'a>> {
+    records
+        .get(index as usize)
+        .and_then(|record| record.subtable(data).ok())
+}
+
+#[derive(Clone)]
+struct CodepointSubtable<'a> {
+    subtable: SupportedSubtable<'a>,
+    is_symbol: bool,
```
rsheeter, line 169: is_symbol seems like it might warrant a comment

skrifa/src/meta/charmap.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    fn shift_symbol_to_pua(codepoint: u32) -> Option<u32> {
+        // From HarfBuzz:
+        // For symbol-encoded OpenType fonts, we duplicate the
+        // U+F000..F0FF range at U+0000..U+00FF.  That's what
+        // Windows seems to do, and that's hinted about at:
+        // https://docs.microsoft.com/en-us/typography/opentype/spec/recom
+        // under "Non-Standard (Symbol) Fonts".
+        // See <https://github.com/harfbuzz/harfbuzz/blob/453ded05392af38bba9f89587edce465e86ffa6b/src/hb-ot-cmap-table.hh#L1595>
+        if codepoint <= 0x00FF {
```
rsheeter, line 198: Suggestion: take &self here and then make the if statement self.is_symbol && codepoint...

That lets you change map to something like self.map_impl(codepoint).or_else(Self::shift_symbol_to_pua)

dfrg: We can't quite get the clean syntax you suggest since `shift_symbol_to_pua` will require two arguments. Inlining the code would simplify things though.

PR #319, opened 2023-03-31
Write gvar

cmyr:
~This is a WIP; in particular it isn't tested yet, and I'm also going to break off some commits into separate PRs.~

This is a high quality PR that is ready for review.


write-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,451 @@
...
+    gid: GlyphId,
+    variations: Vec<GlyphDeltas>,
+}
+
+/// Glyph deltas for one point in the design space.
+pub struct GlyphDeltas {
+    peak_tuple: Tuple,
+    intermediate_region: Option<(Tuple, Tuple)>,
+    deltas: Vec<(i16, i16)>,
+}
```
rsheeter, line 24: I think the fields here might warrant comment?

write-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,451 @@
...
+
+impl Gvar {
+    /// Construct a gvar table from a vector of per-glyph variations.
+    ///
+    /// Variations must be present for each glyph, but may be empty.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the axis count is not consistent between variations (excluding
```
rsheeter, line 33: why not an Error?

write-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn write_into(&self, writer: &mut TableWriter) {
+        self.bits().write_into(writer)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
rsheeter, line 391: for such an important part that involves so many hand-coded bits the testing feels ... a little light?

rsheeter (approved):
I'd prefer error to panic but tis broadly lgtm.

write-fonts/src/tables/gvar.rs
```diff
@@ -0,0 +1,451 @@
...
+    fn smoke_test() {
```
rsheeter, line 392: it would be nice to test a well commented simple scenario, say a box that varies. Maybe some of it's points are coincident sometimes, e.g. a box that turns into a triangle by narrowing the top line.

Could become an integration test, build a font with font builder, yank it's outline with skrifa, get the right thing.

write-fonts/src/tables/gvar.rs
cmyr: It was an error initially, but I couldn't see any way that would be used by the caller other than unwrapping. My basic thinking, here, is that it is the caller's responsibility to ensure that the invariants are met, and if they aren't then there is a bug in the caller's implementation; there is no normal circumstance where one of these error conditions might be triggered.

write-fonts/src/tables/gvar.rs
cmyr: Definitely. My intention was to get a working version of this done, so that we could start playing around with it. I'm happy to try and come up with some more thorough integration test as followup.

Issue #321, opened 2023-03-31
Font-scope validation


rsheeter:
Such as things like fvar.axis_count must == gvar.axis_count

PR #322, opened 2023-04-02
Set the binary search assist fields of the table directory

### LINKED ISSUE #314: FontBuilder should set searchRange, entrySelector, and rangeShift
Code like `let cmap_table = tables::cmap::Cmap::from_mappings([]);` seems to produce a `subtable_offset: +12` when fontTools produces `subtable_offset: +20` (according to otexplorer). Ots then complains as thus:

```
WARNING: bad table directory searchRange
WARNING: bad table directory entrySelector
WARNING: bad table directory rangeShift
File sanitized successfully!
```

rsheeter:
Avoids having every output of write-fonts produce OTS warnings. Fixes #314.

JMM if this looks ok.

### DIFF
--- a/write-fonts/src/font_builder.rs
+++ b/write-fonts/src/font_builder.rs
@@ -15,6 +15,35 @@ pub struct FontBuilder<'a> {
     tables: BTreeMap<Tag, Cow<'a, [u8]>>,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum BuildFontError {
+    TooManyTables,
+}
+
+impl TableDirectory {
+    pub fn from_table_records(
+        table_records: Vec<TableRecord>,
+    ) -> Result<TableDirectory, BuildFontError> {
+        if table_records.len() > u16::MAX as usize {
+            return Err(BuildFontError::TooManyTables);
+        }
+
+        // See https://learn.microsoft.com/en-us/typography/opentype/spec/otff#table-directory
+        // Computation works at the largest allowable num tables so don't stress the as u16's
+        let entry_selector = (table_records.len() as f64).log2().floor() as u16;
+        let search_range = (2.0_f64.powi(entry_selector as i32) * 16.0) as u16;
+        let range_shift = (table_records.len() * 16 - search_range as usize) as u16;
+
+        Ok(TableDirectory::new(
+            TT_SFNT_VERSION,
+            search_range,
+            entry_selector,
+            range_shift,
+            table_records,
+        ))
+    }
+}
+
 impl<'a> FontBuilder<'a> {
     pub fn add_table(&mut self, tag: Tag, data: impl Into<Cow<'a, [u8]>>) -> &mut Self {
         self.tables.insert(tag, data.into());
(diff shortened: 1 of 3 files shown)


write-fonts/src/font_builder.rs
```diff
@@ -15,6 +15,36 @@ pub struct FontBuilder<'a> {
...
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum BuildFontError {
+    TooManyTables,
+}
+
+impl TableDirectory {
+    pub fn from_table_records(
+        table_records: Vec<TableRecord>,
+    ) -> Result<TableDirectory, BuildFontError> {
+        if table_records.len() > u16::MAX as usize {
```
cmyr, line 27: this seems so impossible to do in real use that I would be content with an assert; then we could forget about returning a `Result` for now, avoiding needing to do a minor version bump?
My medium-term vision is to get rid of this current `FontBuilder` entirely, and replace it with something where can just add any `TopLevelTable` object (which knows its tag) and then we would store these uncompiled, and compile them in parallel on build; this would also give us somewhere to do a font-wide validation pass (#321)

rsheeter: I'm ok with dropping result here. I think the new API will need a Result then.

rsheeter: Indeed that's very close to what FontBuilder *is* at this point; if build can have errors it can then do validation?

rsheeter: No more result.

cmyr (approved):
great, thanks!

write-fonts/src/font_builder.rs
```diff
@@ -15,6 +15,27 @@ pub struct FontBuilder<'a> {
...
+impl TableDirectory {
+    pub fn from_table_records(table_records: Vec<TableRecord>) -> TableDirectory {
+        assert!(table_records.len() <= u16::MAX as usize);
+
+        // See https://learn.microsoft.com/en-us/typography/opentype/spec/otff#table-directory
+        // Computation works at the largest allowable num tables so don't stress the as u16's
+        let entry_selector = (table_records.len() as f64).log2().floor() as u16;
+        let search_range = (2.0_f64.powi(entry_selector as i32) * 16.0) as u16;
+        // The result doesn't really make sense with 0 tables but ... let's at least not fail
+        let range_shift = (table_records.len() * 16).saturating_sub(search_range as usize) as u16;
```
anthrotype, line 27: i wonder, since we are doing this at least twice, here and in cmap table, maybe we should have a common method to compute these three fields?

anthrotype: the other one is here:

https://github.com/googlefonts/fontations/blob/47cf723d23d1de419800e33c65eca918a2d75ccc/write-fonts/src/tables/cmap.rs#L29-L39

anthrotype: they are basically the same, except for the item size, 16 in the case of the table directory, 2 in the case of cmap. FontTools uses the same `getSearchRange` function for both, parametrizing the itemSize.
https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/ttLib/ttFont.py#L1138-L1146

PR #325, opened 2023-04-02
Tilting at contours of windmills

rsheeter:
Maybe this is reflective of being a noob more than anything else but the combination of the nested function and std::iter::from_fn feels hard to follow. I tried to make it more idiomatic.

Better? - if so please approve.
Worse? - if so please explain why so I can learn.

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -202,59 +202,87 @@ impl SimpleGlyph {
     ///
     /// This does not do the final binary encoding, and it also does not handle
     /// repeating flags, which doesn't really work when we're an iterator.
-    ///
-    // this is adapted from simon's implementation at
-    // https://github.com/simoncozens/rust-font-tools/blob/105436d3a617ddbebd25f790b041ff506bd90d44/fonttools-rs/src/tables/glyf/glyph.rs#L268
     fn compute_point_deltas(
         &self,
     ) -> impl Iterator<Item = (SimpleGlyphFlags, CoordDelta, CoordDelta)> + '_ {
-        // reused for x & y by passing in the flags
-        fn flag_and_delta(
-            value: i16,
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -202,62 +202,90 @@ impl SimpleGlyph {
...
     pub fn contours(&self) -> &[Contour] {
         &self.contours
     }
 }
 
+// this is adapted from simon's implementation at
+// https://github.com/simoncozens/rust-font-tools/blob/105436d3a617ddbebd25f790b041ff506bd90d44/fonttools-rs/src/tables/glyf/glyph.rs#L268
+struct PointDeltaIter<'a, I>
+where
+    I: Iterator<Item = &'a CurvePoint>,
```
dfrg, line 220: The bound on `I` here should be necessary.

write-fonts/src/tables/glyf.rs
```diff
@@ -202,62 +202,90 @@ impl SimpleGlyph {
...
+{
+    last_x: i16,
+    last_y: i16,
+    points: I,
+}
+
+impl<'a, I> PointDeltaIter<'a, I>
+where
+    I: Iterator<Item = &'a CurvePoint>,
```
dfrg, line 229: ditto

dfrg (approved):
Approved with some small suggestions. I don’t have a strong feeling on this one way or the other so merge at your discretion.

write-fonts/src/tables/glyf.rs
rsheeter: Huh. I'd have sworn it complained when I didn't put that.

rsheeter:
I don't think I'm convinced this is clearly better. It's certainly the way I personally prefer it, with less nesting and lots of small separate bits, but that doesn't seem enough to justify actually changing it.

PR #328, opened 2023-04-03
Expose constructors & consts on variation flag types

cmyr:
JMM


rsheeter (approved):
LGTM, I'm curious why not `From<u16>` as I think that's what I naively would have expected.

PR #330, opened 2023-04-03
Write packed things

cmyr:
This implements the logic for writing out packed deltas and packed point numbers, used in gvar & cvar.

Split out from #319, since this stands alone reasonably well.

### DIFF
--- a/write-fonts/src/tables/variations.rs
+++ b/write-fonts/src/tables/variations.rs
@@ -27,6 +27,121 @@ pub struct PackedPointNumbers {
     numbers: Vec<u16>,
 }
 
+/// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#packed-deltas>
+pub struct PackedDeltas {
+    deltas: Vec<i16>,
+}
+
+impl Validate for PackedDeltas {
+    fn validate_impl(&self, _ctx: &mut ValidationCtx) {}
+}
+
+impl FontWrite for PackedDeltas {
+    fn write_into(&self, writer: &mut TableWriter) {
+        for run in self.iter_runs() {
+            run.write_into(writer)
+        }
+    }
+}
+
+impl PackedDeltas {
+    /// Construct a `PackedDeltas` from a vector of raw delta values.
+    pub fn new(deltas: Vec<i16>) -> Self {
+        Self { deltas }
+    }
+
+    fn iter_runs(&self) -> impl Iterator<Item = PackedDeltaRun> {
+        const MAX_POINTS_PER_RUN: usize = 64;
+
+        fn in_i8_range(val: i16) -> bool {
+            const MIN: i16 = i8::MIN as i16;
+            const MAX: i16 = i8::MAX as i16;
+            (MIN..=MAX).contains(&val)
+        }
+
+        fn count_leading_zeros(slice: &[i16]) -> usize {
+            slice.iter().position(|x| *x != 0).unwrap_or(slice.len())
+        }
+
+        /// compute the number of deltas in the next run, and whether they are i8s or not
+        fn next_run_len(slice: &[i16]) -> (usize, bool) {
+            let (first, rest) = slice.split_first().expect("bounds checked before here");
+            debug_assert!(*first != 0);
+            let is_1_byte = in_i8_range(*first);
+
+            for (i, raw) in rest.iter().copied().enumerate().take(MAX_POINTS_PER_RUN) {
+                let two_zeros = raw == 0 && rest.get(i + 1) == Some(&0);
+                let different_enc_len = in_i8_range(raw) != is_1_byte;
+                if two_zeros || different_enc_len {
+                    return (i + 1, is_1_byte);
+                }
+            }
+
+            (slice.len(), is_1_byte)
+        }
+
+        let mut deltas = self.deltas.as_slice();
+
+        std::iter::from_fn(move || {
+            if deltas.is_empty() {
+                return None;
+            }
+            // first handle the zero case
+            let n_zeros = count_leading_zeros(deltas);
+            if n_zeros > 0 {
+                deltas = &deltas[n_zeros..];
+                return Some(PackedDeltaRun::Zeros(n_zeros as u8));
+            }
+
+            let (len, is_i8) = next_run_len(deltas);
+            let (head, tail) = deltas.split_at(len);
+            deltas = tail;
+            let result = if is_i8 {
+                PackedDeltaRun::OneByte(head)
+            } else {
+                PackedDeltaRun::TwoBytes(head)
+            };
+            Some(result)
+        })
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum PackedDeltaRun<'a> {
+    Zeros(u8),
+    OneByte(&'a [i16]),
+    TwoBytes(&'a [i16]),
+}
+
+impl PackedDeltaRun<'_> {
+    fn compute_flag(&self) -> u8 {
+        /// Flag indicating that this run contains no data,
+        /// and that the deltas for this run are all zero.
+        const DELTAS_ARE_ZERO: u8 = 0x80;
+        /// Flag indicating the data type for delta values in the run.
+        const DELTAS_ARE_WORDS: u8 = 0x40;
+
+        match self {
+            PackedDeltaRun::Zeros(count) => (count - 1) | DELTAS_ARE_ZERO,
+            PackedDeltaRun::OneByte(deltas) => deltas.len() as u8 - 1,
+            PackedDeltaRun::TwoBytes(deltas) => (deltas.len() as u8 - 1) | DELTAS_ARE_WORDS,
+        }
+    }
+}
+
+impl FontWrite for PackedDeltaRun<'_> {
+    fn write_into(&self, writer: &mut TableWriter) {
+        self.compute_flag().write_into(writer);
+        match self {
+            PackedDeltaRun::Zeros(_) => (),
+            PackedDeltaRun::OneByte(deltas) => {
+                deltas.iter().for_each(|v| (*v as i8).write_into(writer))
+            }
+            PackedDeltaRun::TwoBytes(deltas) => deltas.iter().for_each(|v| v.write_into(writer)),
+        }
+    }
+}
+
 impl crate::validate::Validate for PackedPointNumbers {
     fn validate_impl(&self, ctx: &mut ValidationCtx) {
         if self.numbers.len() > 0x7FFF {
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/variations.rs
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
     }
 }
 
+/// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#packed-point-numbers>
+pub struct PackedPointNumbers {
+    is_all: bool,
```
rsheeter, line 26: smells like an enum?

write-fonts/src/tables/variations.rs
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
...
+    }
+
+    fn iter_runs(&self) -> impl Iterator<Item = PackedDeltaRun> {
+        const MAX_POINTS_PER_RUN: usize = 64;
+
+        fn in_i8_range(val: i16) -> bool {
+            const MIN: i16 = i8::MIN as i16;
+            const MAX: i16 = i8::MAX as i16;
+            (MIN..=MAX).contains(&val)
+        }
```
rsheeter, line 60: this might be just me but ... I really dislike nesting pure functions. It makes it much harder to read the code.

write-fonts/src/tables/variations.rs
cmyr: hahaha I turn this into an enum in a commit on #319 💁

write-fonts/src/tables/variations.rs
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
...
+                    return (i + 1, is_1_byte);
+                }
+            }
+
+            (slice.len(), is_1_byte)
+        }
+
+        let mut deltas = self.deltas.as_slice();
+
+        std::iter::from_fn(move || {
```
rsheeter, line 85: I still find this pattern harder to read than a struct plus impl iter.

write-fonts/src/tables/variations.rs
cmyr: funny I actually prefer this style when it isn't too long because I find it easier to read, since the functions are declared right where they're used, instead of being mysterious helpers drifting to the bottom of the file.

fwiw this pattern is used quite a bit in std:
- simple example https://doc.rust-lang.org/1.68.2/src/alloc/vec/mod.rs.html#1406
- more extreme example, declaring a helper struct & impl block inside a fn: https://doc.rust-lang.org/1.68.2/src/alloc/vec/mod.rs.html#1609

I guess how I'd think about this is that it makes the particular function a bit harder to read, at the cost of making the file overall easier to read.

write-fonts/src/tables/variations.rs
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
...
+
+        fn in_i8_range(val: i16) -> bool {
+            const MIN: i16 = i8::MIN as i16;
+            const MAX: i16 = i8::MAX as i16;
+            (MIN..=MAX).contains(&val)
+        }
+
+        fn count_leading_zeros(slice: &[i16]) -> usize {
+            slice.iter().position(|x| *x != 0).unwrap_or(slice.len())
+        }
```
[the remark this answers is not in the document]
cmyr: I prefer your version too. To justify the trait I think this would need to be functionality that was more widely used.

write-fonts/src/tables/variations.rs
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
...
+            let (len, is_i8) = next_run_len(deltas);
+            let (head, tail) = deltas.split_at(len);
+            deltas = tail;
+            let result = if is_i8 {
+                PackedDeltaRun::OneByte(head)
+            } else {
+                PackedDeltaRun::TwoBytes(head)
+            };
+            Some(result)
+        })
```
rsheeter, line 105: IMHO this fn might be clearer if you matched on deltas.get(0), e.g.

```rust
return match deltas.get(0) {
  None => None; // the is_empty case above
  Some(0) => // a run of at least 1 zero!
  Some(..) => // a run of interesting values!
}

File: write-fonts/src/tables/variations.rs (hunk 3 above)
Comment (cmyr) [reply write-fonts/src/tables/variations.rs:85]: I agree, and I prefer a manual impl when the logic is more complicated, but in these simple cases I really appreciate the concision of `from_fn`.

File: write-fonts/src/tables/variations.rs (hunk 5 above)
Comment (cmyr) [reply write-fonts/src/tables/variations.rs:105]: that does seem nice, let me try it out. :)

File: write-fonts/src/tables/variations.rs (hunk 2 above)
Comment (rsheeter) [reply write-fonts/src/tables/variations.rs:60]: Also we're kind of reinventing TryFrom so unless there is a perf issue I wonder if we should just use that?

File: write-fonts/src/tables/variations.rs [hunk 6]
```diff
@@ -21,6 +21,237 @@ impl VariationRegionList {
...
+                    .iter()
+                    .take(MAX_POINTS_PER_RUN)
+                    .scan(prev_point, |prev, point| {
+                        let result = (point - *prev <= U8_MAX).then_some(point);
+                        *prev = *point;
+                        result
+                    })
+                    .count();
+                (count, false)
+            };
```
rsheeter, line 206: I don't think you need to duplicate the iteration if you compute are_words and use it to pick waht function to scan for?

```rust
let are_words = next - prev_point > U8_MAX;
let mut scan_fn = if are_words {
  |v| v > U8_MAX
} else {
  |v| v <= U8_MAX
}
let count = points
   .iter()
   .take(MAX_POINTS_PER_RUN)
   .scan(prev_point, |prev, point| {
   let result = scan_fn(point - *prev).then_some(point);
      *prev = *point;
      result
   })
   .count();
```

write-fonts/src/tables/variations.rs
```diff
@@ -39,3 +270,77 @@ impl<'a> FromObjRef<Option<read_fonts::tables::variations::Tuple<'a>>> for Vec<F
...
+    fn point_pack_words() {
+        let thing = PackedPointNumbers {
+            is_all: false,
+            numbers: vec![1002, 2002, 8408, 12228],
+        };
+
+        let runs = thing.iter_runs().collect::<Vec<_>>();
+        assert!(runs[0].are_words);
+        assert_eq!(runs[0].last_point, 0);
+        assert_eq!(runs[0].points, &[1002, 2002, 8408, 12228]);
```
rsheeter, line 283: Suggest:

```rust
assert_eq!(PackedPointRun{ are_words: true, last_point: 0, etc }, runs);
```

Gives a nicer diff and has it built in to confirm you didn't get 2+ runs.

write-fonts/src/tables/variations.rs
```diff
@@ -39,3 +270,77 @@ impl<'a> FromObjRef<Option<read_fonts::tables::variations::Tuple<'a>>> for Vec<F
...
+
+        let bytes = crate::dump_table(&thing).unwrap();
```
rsheeter, line 285: This appears to be a different test whose only reason to be in the same function is using the same testdata. Suggest making a const or fn to get testdata and then having two actually different tests.

write-fonts/src/tables/variations.rs
```diff
@@ -39,3 +270,77 @@ impl<'a> FromObjRef<Option<read_fonts::tables::variations::Tuple<'a>>> for Vec<F
...
+
+        assert!(runs[1].are_words);
+        assert_eq!(runs[1].last_point, 225);
+        assert_eq!(runs[1].points, &[1002, 2002]);
+
+        assert!(!runs[2].are_words);
+        assert_eq!(runs[2].last_point, 2002);
+        assert_eq!(runs[2].points, &[2008, 2228]);
+
+        assert_eq!(runs.len(), 3);
```
rsheeter, line 312: here too I'd rather you assert the full contents of runs for the nicer diff and to have the expected value grouped up instead of mixed with the accessors to get thigns to compare to.

write-fonts/src/tables/variations.rs
```diff
@@ -39,3 +270,77 @@ impl<'a> FromObjRef<Option<read_fonts::tables::variations::Tuple<'a>>> for Vec<F
...
+        static EXPECTED: &[u8] = &[
+            0x03, 0x0A, 0x97, 0x00, 0xC6, 0x87, 0x41, 0x10, 0x22, 0xFB, 0x34,
+        ];
+
+        let deltas = PackedDeltas::new(vec![10, -105, 0, -58, 0, 0, 0, 0, 0, 0, 0, 0, 4130, -1228]);
+        let runs = deltas.iter_runs().collect::<Vec<_>>();
+        assert_eq!(runs[0], PackedDeltaRun::OneByte(&[10, -105, 0, -58]));
+        assert_eq!(runs[1], PackedDeltaRun::Zeros(8));
+        assert_eq!(runs[2], PackedDeltaRun::TwoBytes(&[4130, -1228]));
+
```
rsheeter, line 332: Same feedback: assert the whole runs above and break the dump table/read into it's own test.

write-fonts/src/tables/variations.rs
```diff
@@ -39,3 +270,77 @@ impl<'a> FromObjRef<Option<read_fonts::tables::variations::Tuple<'a>>> for Vec<F
...
+        let bytes = crate::dump_table(&deltas).unwrap();
+        assert_eq!(bytes, EXPECTED);
+        let read = read_fonts::tables::variations::PackedDeltas::new(FontData::new(&bytes));
+        let decoded = read.iter().collect::<Vec<_>>();
+        assert_eq!(deltas.deltas.len(), decoded.len());
+        assert_eq!(deltas.deltas, decoded);
+        assert_eq!(bytes, EXPECTED);
+    }
+}
```
rsheeter, line 341: In general I beleive you are missing tests for no items and for more items than the max per run?

write-fonts/src/tables/variations.rs
cmyr: I don't think that `TryFrom` should be used *anytime* you have a fallible constructor that takes a single argument. [Per the docs](http://doc.rust-lang.org/1.68.2/std/convert/trait.TryFrom.html), `TryFrom` is for "Simple and safe type conversions that may fail in a controlled way under certain circumstances". If you look at the impls in `std` for `TryFrom`, you will see that they are all very simple conversions, with clear failure conditions.

write-fonts/src/tables/variations.rs
cmyr: I tried it like this but it ends up being more verbose and also harder to debug, since it's harder to track down the actual value contributing to a failure.

write-fonts/src/tables/variations.rs
cmyr: I prefer this style because when a failure occurs, you know immediately what field of which run you need to look at, instead of trying to visually read a big diff to find the one incorrect field.

PR #336, opened 2023-04-05
[skrifa] inline and simplify symbol lookup

dfrg:
Moves the symbol check directly into the `or_else` closure in `CodepointSubtable::map_impl`. Removes the extra indirection.

If this is desirable, JMM.

### DIFF
--- a/skrifa/src/meta/charmap.rs
+++ b/skrifa/src/meta/charmap.rs
@@ -157,14 +157,22 @@ fn get_subtable<'a>(
 #[derive(Clone)]
 struct CodepointSubtable<'a> {
     subtable: SupportedSubtable<'a>,
+    /// True if the subtable is a symbol mapping.
     is_symbol: bool,
 }
 
 impl<'a> CodepointSubtable<'a> {
     fn map(&self, codepoint: u32) -> Option<GlyphId> {
         self.map_impl(codepoint).or_else(|| {
-            if self.is_symbol {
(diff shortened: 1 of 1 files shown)


skrifa/src/meta/charmap.rs
```diff
@@ -157,14 +157,22 @@ fn get_subtable<'a>(
...
+                // For symbol-encoded OpenType fonts, we duplicate the
+                // U+F000..F0FF range at U+0000..U+00FF.  That's what
+                // Windows seems to do, and that's hinted about at:
+                // https://docs.microsoft.com/en-us/typography/opentype/spec/recom
+                // under "Non-Standard (Symbol) Fonts".
+                // See <https://github.com/harfbuzz/harfbuzz/blob/453ded05392af38bba9f89587edce465e86ffa6b/src/hb-ot-cmap-table.hh#L1595>
+                self.map_impl(codepoint + 0xF000)
             } else {
                 None
             }
```
rsheeter, line 178: Optional, this looks like it could use `.then`, e.g. `(self.is_symbol && codepoint <= 0xFF).then(|| self.map_impl(codepoint + 0xF000)`

rsheeter (approved):
LGTM

skrifa/src/meta/charmap.rs
dfrg: This is _almost_ really nice but would require a `flatten()` at the end which IMO kills any benefit of the increased brevity/clarity.

Issue #303, opened 2023-03-23
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()


simoncozens:
I'm trying to use `fea-rs` but `write-fonts` is dying:

```
$ RUST_BACKTRACE=1 cargo run ./features.fea --glyph-order glyphorder -o Gulzar-Regular.ttf
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/fea-rs ./features.fea --glyph-order glyphorder -o Gulzar-Regular.ttf`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', /Users/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/write-fonts-0.1.5/src/write.rs:72:59
```

Relevant bit of backtrace seems to be

```
   4: write_fonts::write::dump_impl
             at /Users/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/write-fonts-0.1.5/src/write.rs:72:13
   5: write_fonts::write::dump_table
             at /Users/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/write-fonts-0.1.5/src/write.rs:47:8
   6: core::ops::function::FnOnce::call_once
             at /rustc/470e518c4b43265020c882bcf3c86632f5494910/library/core/src/ops/function.rs:251:5
   7: core::option::Option<T>::map
             at /rustc/470e518c4b43265020c882bcf3c86632f5494910/library/core/src/option.rs:922:29
   8: fea_rs::compile::lookups::contextual::compute_size
             at ./fea-rs/src/compile/lookups/contextual.rs:648:5
```

rsheeter:
Can you add more specific steps to reproduce? -  e.g. where to obtain the input files, etc. IIRC Gulzar has notably complex features so bonus points for a simplified repro case.

simoncozens:
Here's a minimal reproducer:

[features.fea.gz](https://github.com/googlefonts/fontations/files/11059441/features.fea.gz)
[glyphorder.txt](https://github.com/googlefonts/fontations/files/11059442/glyphorder.txt)
[Gulzar-Regular.ttf.gz](https://github.com/googlefonts/fontations/files/11059444/Gulzar-Regular.ttf.gz)

Looks like it's having problems with overflowing GSUB6 lookups.

cmyr:
Just seeing this now, will investigate! sorry for the delay :)

[...]

PR #339, opened 2023-04-05
Improve diagnostics when table packing fails

cmyr:
This would previously just hit an unwrap and provide very little information. We still provide very little information, but we now do it earlier and in a controlled way.

This should at least explain what happens in cases like #303.

### DIFF
--- a/write-fonts/src/write.rs
+++ b/write-fonts/src/write.rs
@@ -43,7 +43,10 @@ pub fn dump_table<T: FontWrite + Validate>(table: &T) -> Result<Vec<u8>, Validat
     let mut writer = TableWriter::default();
     table.write_into(&mut writer);
     let mut graph = writer.finish();
-    graph.topological_sort();
+    if !graph.topological_sort() {
+        //TODO: implement extension promotion and table splitting
+        panic!("could not find a graph packing that handles all offsets, aborting");
+    }
     Ok(dump_impl(&graph.order, &graph.objects))
 }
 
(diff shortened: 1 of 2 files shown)


write-fonts/src/write.rs
```diff
@@ -43,7 +43,10 @@ pub fn dump_table<T: FontWrite + Validate>(table: &T) -> Result<Vec<u8>, Validat
     let mut writer = TableWriter::default();
     table.write_into(&mut writer);
     let mut graph = writer.finish();
-    graph.topological_sort();
+    if !graph.topological_sort() {
+        //TODO: implement extension promotion and table splitting
+        panic!("could not find a graph packing that handles all offsets, aborting");
```
rsheeter, line 48: feels a touch weird to panic in a function that returns a Result?

rsheeter (approved):
LGTM for now, but longer term I would prefer dump_table not panic.

cmyr:
that is definitely the longer-term goal :)

write-fonts/src/write.rs
cmyr: The error returned by this function contains information about any invalid data in the table that would prevent us from being able to compile it (for instance a vec with more than u16::max items) but a failure to pack the table is something else entirely.

It's possible that this should be rethought? Even if we implement all the fanciness, we probably can't *guarantee* that table packing will always succeed, so maybe this should be fallible..

Issue #209, opened 2023-01-24
Additional improvements to skrifa


dfrg:
The initial punchcut PR (#191) has been lingering and we'd like to land it. This issue lists additional improvements that were requested during review but will be deferred to future PRs.

* Additional high level documentation-- help users understand what the library does and how to use it
* Typing adjustments:
    * ~~The `Variation` type should be better documented. This will be moved to `feta` as a generic type that supports both variations and feature settings.~~
    * ~~Newtype and rationale/description for unique font identifiers (currently raw u64). This also moves to feta as it will be necessary for shaping too.~~

rsheeter:
Also coming up with a norse-themed name? :D

dfrg:
Not sure how I missed that important requirement :)

PR #340, opened 2023-04-06
[skrifa] cleanup: replace Variation with generic Setting<T>

dfrg:
No functional changes here. This is the first in a series of cleanup PRs designed to prepare skrifa for initial release.

Removes the `scale::Variation` type and replaces it with a generic `Setting<T>` type along with a type alias `VariationSetting = Setting<f32>`. Setting is generic because the same structure will also be useful for OpenType feature settings at some point in the future.

Addresses one item on the list in #209.

Also moves a few types from lib.rs to modules.

### DIFF
--- a/skrifa/src/setting.rs
+++ b/skrifa/src/setting.rs
@@ -1,0 +1,80 @@
+//! Generic type for selecting variations and features.
+
+use super::Tag;
+use core::str::FromStr;
+
+/// Setting defined by a selector tag and an associated value.
+///
+/// This type is a generic container for properties that can be activated
+/// or defined by a `(Tag, T)` pair where the tag selects the target
+/// setting and the generic value of type `T` specifies the value for that
+/// setting.
+///
+/// ## Usage
+/// Current usage is for specifying variation axis settings (similar to the
+/// CSS property [font-variation-settings](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variation-settings)).
+/// See [`VariationSetting`].
+///
+/// In the future, this will likely also be used for specifying feature settings
+/// (analogous to the CSS property [font-feature-settings](https://developer.mozilla.org/en-US/docs/Web/CSS/font-feature-settings))
+/// for selecting OpenType [features](https://learn.microsoft.com/en-us/typography/opentype/spec/featuretags).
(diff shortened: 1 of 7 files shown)


skrifa/src/setting.rs
```diff
@@ -0,0 +1,80 @@
...
+            selector: s.0,
+            value: s.1,
+        }
+    }
+}
+
+impl<T> From<(&str, T)> for Setting<T> {
+    fn from(s: (&str, T)) -> Self {
+        Self {
+            selector: Tag::from_str(s.0).unwrap_or_default(),
```
cmyr, line 49: do we really want this? If the string is not a valid tag I feel we're better off failing than just having a meaningless tag?

My preference would be to just bite the bullet and make this slightly less ergonomic, forcing the user to construct tags and ensure they are valid?

cmyr (approved):
One concern about the handling of `Tag`s, otherwise lgtm

skrifa/src/setting.rs
rsheeter: I think a _compiler_ shouldn't emit bad tags but I'm less convinced that should be true of skrifa, we don't add value to the user by rejecting a font that is entirely valid except it violates the rules for what can go in a tag. That would lead me to believe we should accept any <= 4 byte sequence as a tag and offer a way to check validity for users who want that.

dfrg: I understand the concern, but my general feeling here is that specifying an axis tag as, say, "weight axis" is no different than something like "WGHT". Both are incorrect and will result in a failure from the user's perspective.

We can't perform any sort of real validation without having the axis list and even then, I'd argue against doing so because there are very real scenarios in text layout where you have a style like `font-family: A, B; font-variation-settings: 'wght' 100;` and only the `A` family has a weight axis. In this case, we want to try to apply the setting when falling back to `B` but not fail if it doesn't exist.

The position I've adopted for skrifa is that it is generally the wrong layer for validation. User inputs should be checked by the layout engine and font validation (e.g. checking for expected tables or consistent values in those tables) should be done by the enumeration/fallback system. The motto could be "stay safe and do the best we can with the given font and input".

PR #341, opened 2023-04-06
SimpleGlyph::from_kurbo: only pop the last_pt==move_pt for closed paths

anthrotype:
As discussed in chat, we want to match fonttools behavior whereby the removal of the last point of a contour when it sits on top of the first point is only done for explicitly _closed_ contours (that end with a close_path), and not open contours, even if the points' coordinates do overlap.

https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/pens/pointPen.py#L321-L323

This makes sure we produce the same number of simple glyf points as fonttools, whether the path is open or closed. 

And also keep single point contours since fonttools does not prune them, and they are harmless.

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -139,6 +139,18 @@ impl Contour {
     pub fn iter(&self) -> impl Iterator<Item = &CurvePoint> {
         self.0.iter()
     }
+
+    pub fn first(&self) -> Option<&CurvePoint> {
+        self.0.first()
+    }
+
+    pub fn last(&self) -> Option<&CurvePoint> {
+        self.0.last()
+    }
+
+    pub fn pop(&mut self) -> Option<CurvePoint> {
+        self.0.pop()
+    }
 }
 
 impl SimpleGlyph {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -172,24 +184,18 @@ impl SimpleGlyph {
                     .ok_or(BadKurbo::MissingMove)?
                     .quad_to(*p0, *p1),
                 kurbo::PathEl::CurveTo(_, _, _) => return Err(BadKurbo::HasCubic),
-                // I think we can just ignore this, and remove duplicate points
-                // at the end?
-                kurbo::PathEl::ClosePath => (),
+                kurbo::PathEl::ClosePath => {
+                    // remove last point in closed path if has same coords as the move point
+                    let contour = current.as_mut().ok_or(BadKurbo::MissingMove)?;
+                    if contour.len() > 1 && contour.last().unwrap() == contour.first().unwrap() {
```
cmyr, line 190: this should work without needing the unwraps, since `Option<T>` impls `PartialEq` if `T` does, and we know they must exist:
```suggestion
                    if contour.len() > 1 && contour.last() == contour.first() {
```

cmyr (approved):
Looks good, I really appreciate your eye for detail and your familiarity with the prior art for these things!

PR #343, opened 2023-04-06
[skrifa] replace raw u64 with UniqueId

dfrg:
Again, following up on #209, replaces the raw u64 used as a font identifier/cache key with a new `UniqueId` type.

### DIFF
--- a/skrifa/src/unique_id.rs
+++ b/skrifa/src/unique_id.rs
@@ -1,0 +1,45 @@
+//! Unique identifier to support internal caching.
+
+/// Identifier used as a key for internal caches.
+///
+/// The representation of a font in this crate is designed to be flexible for
+/// clients (any type that implements [`TableProvider`](read_fonts::TableProvider)
+/// is supported) without imposing any strict constraints on ownership or data layout.
+///
+/// The price for this flexibility is that an additional mechanism is necessary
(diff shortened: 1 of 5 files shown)


skrifa/src/unique_id.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Users are responsible for ensuring that a unique identifier is not used for
+/// different fonts within a single context. Violating this invariant is guaranteed
+/// to be safe, but will very likely lead to incorrect results.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub struct UniqueId {
+    /// Unique identifier for the data blob containing the content of
+    /// a font file.
+    pub data_id: u64,
+    /// Index of a font in a font collection file.
+    pub index: u32,
```
cmyr, line 36: do we have a design in mind that requires these fields to be accessed directly? If not I'd keep them internal for now, and we can make them public or add getters when we see a need?

cmyr (approved):
lgtm modulo one question :)

skrifa/src/unique_id.rs
dfrg: I waffled on that a bit but can't think of any uses. Making them private seems like a good call.

Issue #345, opened 2023-04-06
We cannot release and we didn't notice until we tried


rsheeter:
Apparently https://github.com/googlefonts/fontations/commit/1e76c0#diff-665af3a887b4e7e7129cccec9f9fef7f9cd0b24b45d0fb0b0a01e16ea0e43d0e broke release. It's not awesome this is true and it's not awesome we didnt' find out until we tried to release.

cmyr:
okay, so here is the issue:

- `read-fonts` includes test data that is stored in files on disk.
- this data is bundled into the binary using the `include_bytes!` macro, for instance here:  https://github.com/googlefonts/fontations/blob/main/read-fonts/src/tests/test_data.rs#L385
- there is no way to ignore `include_bytes!`, e.g. you cannot use `#[cfg(some_flag)]` to conditionally prevent the macro; it is expanded very early in the compilation process.
- but the data lives in the root `resources` directory, which is a *sibling* of the `read-fonts` directory.
- when a crate is distributed, it is distributed as source; basically bundling everything in the crate directory (in this case, `fontations/read-fonts`).
- this means you can't just write `include_bytes!(../resources/something/file)`; this will work at development time, but when you publish the parent directory will be different, and compilation will fail.

And here is the current solution:

- instead of using `include_bytes` with the actual paths, we have a build script,
    - this script copies checks if the files exist, and if it does copies them to a location accessible at compile time
    - and if the files do *not* exist, it writes dummy files
    - this means that the tests would fail if we were to run them from the published sources, but this is fine


I banged my head against this for a while, and don't really have a better solution, but I'm open to ideas!

rsheeter:
Capturing so I can refer to it later, current ideas (non-exclusive choices):

1. Make test data a crate, that crate can use include_bytes and would be referenced as a dev-dependency
1. Make `cargo package` a CI step so we find such issues early
1. Make a test helper function that finds and loads the bytes, perhaps doing horrible things to achieve a return of `&'static` data to enable test code to do things like directly return a table type (normally this is complicated by the need for someone to own the data)

cmyr:
option 2 is done as of #347, so we at least have an interim solution.

rsheeter:
To keep the notes up to date, https://github.com/googlefonts/fontations/pull/362 undoes #347

cmyr:
This is resolved in #365, which moves to using a separate test data crate.

PR #347, opened 2023-04-06
[ci] Run 'cargo package' in ci

cmyr:
would catch things like #345.


rsheeter (approved):

cmyr:
I would like to test that this fails appropriately, let me add a commit that should break it

PR #348, opened 2023-04-07
Additional docs for skrifa

### LINKED ISSUE #209: Additional improvements to skrifa

dfrg:
This provides some more detailed documentation with a focus on the scale module. Also updates the readme to describe supported metadata.

[Readme rendered](https://github.com/dfrg/fontations/blob/skrifa-docs/skrifa/README.md)

No functional changes, but additional `From` impls for `Setting` were added because writing docs exposed that the shape of the `ScalerBuilder::variation_settings` API wasn't quite what I intended.

Closes #209

### DIFF
--- a/skrifa/src/scale/mod.rs
+++ b/skrifa/src/scale/mod.rs
@@ -1,24 +1,49 @@
 //! Extraction of glyph outlines.
 //!
 //! Scaling is the process of decoding an outline, applying variation deltas,
-//! and executing [hinting](https://en.wikipedia.org/wiki/Font_hinting) instructions
-//! for a glyph of a particular size.
+//! and executing [hinting](https://en.wikipedia.org/wiki/Font_hinting)
+//! instructions for a glyph of a particular size.
 //!
-//! The process starts with construction of a [`Context`] that can be used to build
-//! and configure a [`Scaler`] which can generate scaled paths for glyphs.
+//! ## It all starts with a context
 //!
-//! Read on for more detail.
+//! The scaling process generally requires dynamic memory allocations to hold
+//! intermediate results. In addition, TrueType hinting requires execution
+//! of a set of programs to generate state for any instance of a font before
+//! applying glyph instructions.
+//!
+//! To amortize the cost of memory allocations and support caching of hinting
+//! state, we use the [`Context`] type. This type is opaque and contains
+//! internal buffers and caches that can be reused by subsequent scaling
+//! operations.
+//!
+//! Contexts exist purely as a performance optimization and management of them
+//! is up to the user. There are several reasonable strategies of varying
+//! complexity:
+//!
+//! * If performance and heap traffic are not significant concerns, creating
+//! a context per glyph (or glyph run) works in a pinch.
+//! * When making use of a single shared glyph cache, this is an ideal place to
+//! store a context.
+//! * Multithreaded code can use thread locals or a global pool of contexts.
+//!
+//! Regardless of how you manage them, creating a context is trivial:
+//! ```
+//! use skrifa::scale::Context;
+//!
+//! let mut context = Context::new();
+//! ```
+//!
+//! For simplicity, the examples below will use a local context.
 //!
 //! ## Building a scaler
 //!
-//! Scaling requires temporary memory allocations and, in the case of hinting,
-//! also benefits from caching the state of the hinting setup programs. To amortize the
-//! cost of heap allocations and support caching, the [`Context`] type is provided.
-//! This type is opaque and offers a single method called [`new_scaler`](Context::new_scaler)
-//! that produces a [`ScalerBuilder`] to configure and build a [`Scaler`].
+//! Now that we have a [`Context`], we can use the
+//! [`new_scaler`](Context::new_scaler) method to generate an instance of the
+//! [`ScalerBuilder`] type that allows us to configure and build a [`Scaler`].
 //!
-//! Assuming you have some `font` (any type that implements [`TableProvider`](read_fonts::TableProvider)),
-//! this will build a scaler for a size of 16px:
+//! Assuming you have some `font` (any type that implements
+//! [`TableProvider`](read_fonts::TableProvider)), this will build a scaler for
+//! a size of 16px:
 //!
 //! ```
 //! # use skrifa::{scale::*, Size};
@@ -30,7 +55,8 @@
 //! # }
 //! ```
 //!
-//! For variable fonts, the [`variation_settings`](ScalerBuilder::variation_settings) method can
+//! For variable fonts, the
+//! [`variation_settings`](ScalerBuilder::variation_settings) method can
 //! be used to specify user coordinates for selecting an instance:
 //!
 //! ```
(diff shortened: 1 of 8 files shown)


skrifa/src/scale/mod.rs
```diff
@@ -1,13 +1,128 @@
...
+//!
+//! Read on for more detail.
+//!
+//! ## Building a scaler
+//!
+//! Scaling requires temporary memory allocations and, in the case of hinting,
+//! also benefits from caching the state of the hinting setup programs. To amortize the
+//! cost of heap allocations and support caching, the [`Context`] type is provided.
+//! This type is opaque and offers a single method called [`new_scaler`](Context::new_scaler)
+//! that produces a [`ScalerBuilder`] to configure and builder a [`Scaler`].
```
cmyr, line 18:
```suggestion
//! that produces a [`ScalerBuilder`] to configure and build a [`Scaler`].
```

skrifa/src/scale/mod.rs
```diff
@@ -1,13 +1,128 @@
...
+//! For variable fonts, the [`variation_settings`](ScalerBuilder::variation_settings) method can
+//! be used to specify user coordinates for selecting an instance:
+//!
+//! ```
+//! # use skrifa::{scale::*, Size};
+//! # fn build_scaler(font: read_fonts::FontRef) {
+//! let mut context = Context::new();
+//! let mut scaler = context.new_scaler()
+//!     .size(Size::new(16.0))
+//!     .variation_settings(&[("wght", 720.0), ("wdth", 75.0)])
```
cmyr, line 42: not related to the PR but the question that pops into my head, here is, "what if the axis doesn't exist? is there an error condition?

skrifa/src/scale/mod.rs
```diff
@@ -1,13 +1,128 @@
...
+//!
+//! If you already have coordinates in normalized design space, you can specify those directly
+//! with the [`normalized_coords`](ScalerBuilder::normalized_coords) method.
+//!
+//! See the [`ScalerBuilder`] type for all available configuration options.
+//!
+//! ## Getting an outline
+//!
+//! Once we have a configured scaler, extracting an outline is fairly simple. The
+//! [`outline`](Scaler::outline) method on scaler uses a callback approach where the user
```
cmyr, line 55:
```suggestion
//! [`Scaler::outline`] method uses a callback approach where the user
```

(not sure if this breaks the link or not?)

skrifa/src/scale/mod.rs
```diff
@@ -1,13 +1,128 @@
...
+//! // to build an SVG path.
+//! scaler.outline(GlyphId::new(20), &mut path);
+//!
+//! // Print our pretty new path.
+//! println!("{}", path.0);
+//! # }
+//! ```
+//!
+//! The pen based interface is designed to be flexible. Output can be sent directly to a
+//! software rasterizer for scan conversion, converted to an owned path representation (such as
```
cmyr, line 111: just a casual note, but I find the line breaking in this file makes it hard to read the actual diff, since many lines are longer and get soft wrapped. I have my editor set to wrap at 80-90 columns and I find that's a reasonable compromise, maybe something like that would be helpful?

skrifa/src/scale/scaler.rs
```diff
@@ -74,8 +74,9 @@ impl<'a> ScalerBuilder<'a> {
         self
     }
 
-    /// Adds the sequence of variation settings. This will clear any variations
-    /// specified as normalized coordinates.
+    /// Adds the sequence of variation settings.
+    ///
+    /// This will clear any variations specified as normalized coordinates.
```
cmyr, line 79: I think the type signature of this method is confusing enough I would include a few examples, or otherwise describe what types it is possible to use?

skrifa/src/setting.rs
```diff
@@ -61,6 +79,15 @@ impl<T> From<([u8; 4], T)> for Setting<T> {
...
 
+impl<T: Copy> From<&([u8; 4], T)> for Setting<T> {
+    fn from(s: &([u8; 4], T)) -> Self {
+        Self {
+            selector: Tag::new_checked(&s.0[..]).unwrap_or_default(),
+            value: s.1,
+        }
+    }
+}
+
```
cmyr, line 90: looking at this, I personally think I would consider being more opinionated, and support fewer options for how a setting gets represented?

This API definitely sticks out to me (and I think to you, judging by the iteration?) as one of the current rough spots.

Just out of curiosity, how are you calling this API currently? What is your source type (your sequence of variation settings)?

cmyr (approved):
A (surprisingly) few typos and nitty suggestions, and one meta-question about the API, this did a good job of helping me better understand what's going on, though!

rsheeter:
Looks great. One further request: as a potential user I would like the readme to speak to me about panics: do you commit to not have them? Btw will appear only on GH or does it find it's way to crate docs?

rsheeter:
Also ... where's the hello world of how to use it? - readme doesn't have to have it but it should at least point me to it

skrifa/src/scale/mod.rs
dfrg: I think it should be fine

skrifa/src/scale/mod.rs
dfrg: As mentioned in the second paragraph here: https://github.com/googlefonts/fontations/pull/340#discussion_r1160117009, doing this sort of validation provides very little value. Every production text layout engine that I'm aware of will simply drop unsupported axes (and clamp out of range values) so I chose to support that common case here by default. For clients that want to be more strict about their inputs, they can use the forthcoming variations metadata API (or the associated underlying read-fonts code) to do so. I can provide a `validate_settings` method when I do the variations PR to make that simpler.

skrifa/src/scale/mod.rs
dfrg: Apologies! Aside from lengthy URLs, I do try to keep comment widths to roughly 80 characters but it seems to have gotten away from me here. I'll reflow these.

skrifa/src/scale/scaler.rs
dfrg: Fair enough. It is a bit unwieldy. In these cases, I try to provide the most general API possible, hence the `IntoIterator` and `From` but that does make it hard to decipher what concrete types to actually pass to the method. I'll add examples to the docs.

skrifa/src/setting.rs
dfrg: In some sense, I see the `Setting` type and its `VariationSetting` alias as being both a place to attach documentation _and_ a target for supporting various useful conversions. I think the `From` impls cost us very little but allow for a lot of flexibility for clients. For example, in the documentation I updated for the method at your request above, I note that these are all equivalent:
```rust
// slice of VariationSetting
builder.variation_settings(&[
     VariationSetting::new(Tag::new(b"wgth"), 720.0),
     VariationSetting::new(Tag::new(b"wdth"), 50.0),
])
// slice of (Tag, f32)
builder.variation_settings(&[(Tag::new(b"wght"), 720.0), (Tag::new(b"wdth"), 50.0)])
// slice of (&str, f32)
builder.variation_settings(&[("wght", 720.0), ("wdth", 50.0)])
```
I did remove the `([u8; 4], f32)` conversion as it seemed overly redundant. `(Tag, f32)` may also not be useful. I do find the `(&str, f32)` conversion particularly helpful when playing with the code which I do think is important. And fwiw, we do actually use it in vello.

dfrg:
> Looks great. One further request: as a potential user I would like the readme to speak to me about panics: do you commit to not have them? Btw will appear only on GH or does it find it's way to crate docs?

I'll add a note about panics (which we do intend to avoid regardless of any misuse of the API or corrupted font data). The readme won't appear in the docs, but will be on the landing page for the library on crates.io.

dfrg:
> Also ... where's the hello world of how to use it? - readme doesn't have to have it but it should at least point me to it

This PR contains a series of examples in the docs for the scale module. I can link to them from the readme after we publish and the docs are available on docs.rs.

> Context has comments but they are not helpful. "Context for loading glyphs" and "Creates a new glyph loading context" aren't very enlightening. What does it do? WHy does it exist? The test in mod.rs makes it look like it's just noise, you create it purely to get your hands on a scalar and from then on interact only with the scalar. It's possible to discover how they interact but you have to dig.

You may be looking at old code? I tried to flesh that out in this PR. Specific notes are in the following places:

In module level docs: https://github.com/dfrg/fontations/blob/09284aa0c6b4b01351275a11c84192c79dd40960/skrifa/src/scale/mod.rs#L15

On the Context type: https://github.com/dfrg/fontations/blob/09284aa0c6b4b01351275a11c84192c79dd40960/skrifa/src/scale/mod.rs#L162

I'll add a bit more in a new section of the module level docs to give more... context?

dfrg:
This may not be perfect but I think it's in pretty good shape now. Going to go ahead and merge this and then publish.

PR #349, opened 2023-04-08
Try to guide noobs to a complete example of loading with skrifa

rsheeter:
Make https://docs.rs/skrifa/0.0.1/skrifa/ and https://crates.io/crates/skrifa guide a new user to an example sufficient to Do Something. It would be nice if we had an example pen we could reference, it's quite noisy to put one directly in the example  but without one it doesn't do anything.


dfrg:
An end to end example is a great idea, but I think it'd be more useful to have one (or more) in an examples directory under the skrifa crate. We'd have more space for useful comments and the code would actually be runnable so that users can play with it. 

Per IM, we should link directly to the examples dir from the docs.

PR #353, opened 2023-04-10
Try to round trip deltas from a real source

rsheeter:
Demonstrates either misunderstanding or a bug wrt round tripping deltas.


rsheeter:
@cmyr does the test here seem like it _should_ work?

write-fonts/src/tables/variations.rs
```diff
@@ -129,25 +129,30 @@ impl PackedDeltas {
...
+                let different_enc_len = in_i8_range(cur) != is_1_byte;
                 if two_zeros || different_enc_len {
-                    return (i + 1, is_1_byte);
+                    break;
                 }
-            }
 
-            (slice.len(), is_1_byte)
+                idx += 1;
+            }
```
dfrg, line 148: No issues with your adjustment, but it seems like we could keep the original iterator based loop by changing the final return value to `(slice.len().min(MAX_POINTS_PER_RUN), is_1_byte)`. Or, alternatively: `let limit = slice.len().min(MAX_POINTS_PER_RUN);` and using that for both the argument to `take` and the return value.

write-fonts/src/tables/variations.rs
```diff
@@ -129,25 +129,30 @@ impl PackedDeltas {
...
+            (idx, is_1_byte)
         }
 
         let mut deltas = self.deltas.as_slice();
 
         std::iter::from_fn(move || {
-            if *deltas.first()? == 0 {
+            let run = if *deltas.first()? == 0 {
```
dfrg, line 155: Just curious: did clippy complain here or do you just find it more clear to create a local?

dfrg (approved):
LGTM with a couple comments.

write-fonts/src/tables/variations.rs
rsheeter: Having to track for two consecutive zeros makes the take/take_while a mess. Then I need windows, which I believe will lead me to having to add a custom "and then check the last one if len is odd" rig. IMHO it's worse so for now I'm sticking with this version.

write-fonts/src/tables/variations.rs
```diff
@@ -126,19 +126,24 @@ impl PackedDeltas {
...
+            let first = *slice.first().expect("bounds checked before here");
+            debug_assert!(first != 0);
+            let is_1_byte = in_i8_range(first);
 
-            for (i, raw) in rest.iter().copied().enumerate().take(MAX_POINTS_PER_RUN) {
-                let two_zeros = raw == 0 && rest.get(i + 1) == Some(&0);
-                let different_enc_len = in_i8_range(raw) != is_1_byte;
+            let mut idx = 1;
+            while idx < MAX_POINTS_PER_RUN && idx < slice.len() {
+                let cur = slice[idx];
```
cmyr, line 135: Just a general note: I prefer to write these in terms of iteration wherever possible, because it often ends up letting the compiler generate better code. In the past I have seen code like this compiled into something where there is a bounds check on each array access, which is happening in a loop, whereas the iterating version generates code with a single bounds check up front.

PR #354, opened 2023-04-10
Expose and fix issue with packing delta runs of 0

rsheeter:
Split out from #353 

JMM if looks ok

### DIFF
--- a/write-fonts/src/tables/variations.rs
+++ b/write-fonts/src/tables/variations.rs
@@ -117,13 +117,10 @@ impl PackedDeltas {
         }
 
         fn count_leading_zeros(slice: &[i16]) -> u8 {
-            let mut limit = u8::MAX as i32;
             slice
                 .iter()
-                .take_while(|v| {
-                    limit -= 1;
-                    **v == 0 && limit >= 0
-                })
+                .take(u8::MAX.into())
+                .take_while(|v| **v == 0)
                 .count() as u8
         }
 
--- a/write-fonts/src/tables/variations.rs
+++ b/write-fonts/src/tables/variations.rs
@@ -116,8 +116,15 @@ impl PackedDeltas {
             (MIN..=MAX).contains(&val)
         }
 
-        fn count_leading_zeros(slice: &[i16]) -> usize {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/variations.rs
```diff
@@ -116,8 +116,15 @@ impl PackedDeltas {
             (MIN..=MAX).contains(&val)
         }
 
-        fn count_leading_zeros(slice: &[i16]) -> usize {
-            slice.iter().take_while(|v| **v == 0).count()
+        fn count_leading_zeros(slice: &[i16]) -> u8 {
+            let mut limit = u8::MAX as i32;
+            slice
+                .iter()
+                .take_while(|v| {
```
dfrg, line 123: This could just be
```take(u8::MAX as usize).take_while(…)```

**Review by dfrg — APPROVED** [review]
LGTM with one suggestion.

File: write-fonts/src/tables/variations.rs (hunk 1 above)
Comment (rsheeter) [reply write-fonts/src/tables/variations.rs:123]: That's the missing limit() I couldn't find!!

Comment (rsheeter) [reply write-fonts/src/tables/variations.rs]: Fixed, ty

File: write-fonts/src/tables/variations.rs [hunk 2]
```diff
@@ -116,8 +116,12 @@ impl PackedDeltas {
             (MIN..=MAX).contains(&val)
         }
 
-        fn count_leading_zeros(slice: &[i16]) -> usize {
-            slice.iter().take_while(|v| **v == 0).count()
+        fn count_leading_zeros(slice: &[i16]) -> u8 {
+            slice
+                .iter()
+                .take(u8::MAX.into())
```
cmyr, line 122: iirc, the max run length doesn't change here, and should be 64; the run length is stored in the low 6 bits of the control byte; [see the docs here](https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#packed-deltas).

rsheeter: https://github.com/googlefonts/fontations/pull/356

PR #356, opened 2023-04-10
No more than 64 zeros please, we use 6 bits to encode run length

rsheeter:
JMM


cmyr (approved):
Hopefully this was the last little issue? 🤞

rsheeter:
There's still _something_ wrong but it might be on the fontmake-rs side rather than fontations.

PR #364, opened 2023-04-13
[ci] test packages individually

### LINKED ISSUE #363: Make CI run each package individually
Avoid issues like https://github.com/googlefonts/fontmake-rs/issues/255

We should do this for fontations and fontmake-rs

dfrg:
Runs `cargo test -p` for each package rather than a single `cargo test` for the whole workspace.

Closes #363

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,8 +53,17 @@ jobs:
       - name: install stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: cargo test
-        run: cargo test --all-targets --all-features
+      - name: cargo test font-types
+        run: cargo test -p font-types --all-targest --all-features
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -53,8 +53,17 @@ jobs:
       - name: install stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
-      - name: cargo test
-        run: cargo test --all-targets --all-features
+      - name: cargo test font-types
```
rsheeter, line 56: Suggest adding a comment to explain why

rsheeter (approved):

.github/workflows/rust.yml
dfrg: Done

PR #365, opened 2023-04-13
Add font-test-data crate

cmyr:
This adds a new crate, which hosts the raw test data that is shared between the other crates.

This... is hopefully an improvement? We may need to iterate on this somewhat, in particular I suspect it should only be a dev-dependency, which may mean we need to split up the `test_data` feature on read-fonts.


read-fonts/Cargo.toml
```diff
@@ -10,9 +10,13 @@ categories = ["text-processing", "parsing", "graphics"]
...
 [features]
 std = ["font-types/std"]
-test_data = []
+test_data = ["font-test-data"]
 traversal = ["std"]
 default = ["traversal"]
 
 [dependencies]
 font-types = { version = "0.1.4", path = "../font-types" }
+font-test-data = { path = "../font-test-data", optional = true }
```
cmyr, line 19: I'm concerned that this may also break packaging if we aren't publishing this crate, so hold on merging until I can play around a bit more and confirm that.

read-fonts/src/font_data.rs
```diff
@@ -278,6 +278,12 @@ impl AsRef<[u8]> for FontData<'_> {
     }
 }
 
+impl<'a> From<&'a [u8]> for FontData<'a> {
```
cmyr, line 281: This is a useful addition generally, and hopefully fixes some of the ergonomics issues around needing to manually create `FontData`?

dfrg: +1. We could additionally change `FontRead::read` to take `impl Into<FontData>` for further convenience but that doesn’t really seem to be necessary.

dfrg (approved):
This looks great to me and seriously cleans up our test data situation. Thanks!

PR #366, opened 2023-04-14
Use #[capabilites] on Tuple type

cmyr:
This also adds a missing impl of Hash for BigEndian<T>

~based on #215 which should be merged first.~


dfrg (approved):
Not sure why this one went unnoticed for so long. LGTM!

PR #367, opened 2023-04-15
Demonstrate a hang

rsheeter:
Thought I was losing it when fontmake-rs started to hang reading a 10 byte avar. Run `cargo test -p write-fonts hangs` to reproduce.

### DIFF
--- a/write-fonts/src/tables/avar.rs
+++ b/write-fonts/src/tables/avar.rs
@@ -1,3 +1,22 @@
 //! The [avar](https://learn.microsoft.com/en-us/typography/opentype/spec/avar) table
 
 include!("../../generated/generated_avar.rs");
+
+
+#[cfg(test)]
+mod tests {
+    use font_types::MajorMinor;
+    use read_fonts::{FontRead, FontData};
+    use crate::tables::avar::{Avar, SegmentMaps};
+
+    use crate::dump_table;
+
+    #[test]
+    fn hangs() {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/avar.rs
```diff
@@ -1,3 +1,22 @@
...
+    use read_fonts::{FontRead, FontData};
+    use crate::tables::avar::{Avar, SegmentMaps};
+
+    use crate::dump_table;
+
+    #[test]
+    fn hangs() {
+        let avar = Avar::new(
+            MajorMinor::VERSION_1_0,
+            vec![SegmentMaps::new(Vec::new())]);
```
anthrotype, line 18: is this somehow related to the fact that this SegmentMaps are empty? The spec says they must always include at least {-1: -1, 0: 0, 1: 1}

> If the segment map for a given axis has any value maps, then it must include at least three value maps: -1 to -1, 0 to 0, and 1 to 1. These value mappings are essential to the design of the variation mechanisms and are required even if no additional maps are specified for a given axis.

rsheeter: Almost assuredly, as I interpreted "If the segment map for a given axis has any value maps" to allow for having none. IIRC the fix in fontmake-rs was to plug in the -1:-1, 0:0, 1:1 map. However, it still shouldn't hang.

PR #368, opened 2023-04-21
Filter implied on curves for interpolatable glyphs

anthrotype:
~~this is still quite messy and has no unit tests, but seems to work when I use this to build Oswald~~

EDIT: I've updated the PR adding tests, so this is now ready to review.

The PR adds a new public function to write_fonts::tables::glyf module called `simple_glyphs_from_kurbo`, which takes a slice of BezPaths and returns a vector of SimpleGlyphs (or an error if the input paths are not interpolation compatible).

Oncurve points that are impliable (i.e. at equal distance from preceding and following off-curve points) in _all_ the interpolatable paths are dropped from the SimpleGlyph contours.

The same function is also used in the existing SimpleGlyph::from_kurbo constructor to create one SimpleGlyph only instead of a Vec of them.

Linked draft PR in fontmake-rs to make use of this: https://github.com/googlefonts/fontmake-rs/pull/277

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -172,13 +172,13 @@ impl InterpolatableContourBuilder {
     }
 
     /// The first point in each contour
-    fn first(&self) -> Vec<&ContourPoint> {
-        self.0.iter().map(|v| v.first().unwrap()).collect()
+    fn first(&self) -> impl Iterator<Item = &ContourPoint> {
+        self.0.iter().map(|v| v.first().unwrap())
     }
 
     /// The last point in each contour
-    fn last(&self) -> Vec<&ContourPoint> {
-        self.0.iter().map(|v| v.last().unwrap()).collect()
+    fn last(&self) -> impl Iterator<Item = &ContourPoint> {
+        self.0.iter().map(|v| v.last().unwrap())
     }
 
     /// Remove the last point from each contour
@@ -287,6 +287,8 @@ pub fn simple_glyphs_from_kurbo(paths: &[BezPath]) -> Result<Vec<SimpleGlyph>, B
     let mut contours: Vec<InterpolatableContourBuilder> = Vec::new();
     let mut current: Option<InterpolatableContourBuilder> = None;
     let num_glyphs = paths.len();
+    let mut pts = Vec::with_capacity(num_glyphs);
+    let mut quad_pts = Vec::with_capacity(num_glyphs);
     for (i, elements) in path_iters.enumerate() {
         // All i-th path elements are expected to have the same types.
         // elements is never empty (if it were, MultiZip would have stopped), hence the unwrap
@@ -297,42 +299,49 @@ pub fn simple_glyphs_from_kurbo(paths: &[BezPath]) -> Result<Vec<SimpleGlyph>, B
                 if let Some(prev) = current.take() {
                     contours.push(prev);
                 }
-                let pts = elements
-                    .iter()
-                    .map(|el| match el {
-                        &kurbo::PathEl::MoveTo(pt) => Ok(pt),
-                        _ => {
-                            return Err(BadKurbo::InconsistentPathElements(i, el_types(&elements)))
-                        }
-                    })
-                    .collect::<Result<Vec<_>, _>>()?;
-                current = Some(InterpolatableContourBuilder::new(&pts));
-            }
-            kurbo::PathEl::LineTo(_) => {
-                let pts = elements
-                    .iter()
-                    .map(|el| match el {
-                        &kurbo::PathEl::LineTo(pt) => Ok(pt),
-                        _ => {
-                            return Err(BadKurbo::InconsistentPathElements(i, el_types(&elements)))
-                        }
-                    })
-                    .collect::<Result<Vec<_>, _>>()?;
-                current.as_mut().ok_or(BadKurbo::MissingMove)?.line_to(&pts)
-            }
-            kurbo::PathEl::QuadTo(_, _) => {
-                let mut pts = Vec::with_capacity(num_glyphs);
+                pts.clear();
                 for el in &elements {
                     match el {
-                        &kurbo::PathEl::QuadTo(p0, p1) => {
-                            pts.push((p0, p1));
+                        &kurbo::PathEl::MoveTo(pt) => {
+                            pts.push(pt);
                         }
                         _ => {
                             return Err(BadKurbo::InconsistentPathElements(i, el_types(&elements)))
                         }
                     }
                 }
-                current.as_mut().ok_or(BadKurbo::MissingMove)?.quad_to(&pts)
+                current = Some(InterpolatableContourBuilder::new(&pts));
+            }
+            kurbo::PathEl::LineTo(_) => {
+                pts.clear();
+                for el in &elements {
+                    match el {
+                        &kurbo::PathEl::LineTo(pt) => {
+                            pts.push(pt);
+                        }
+                        _ => {
+                            return Err(BadKurbo::InconsistentPathElements(i, el_types(&elements)))
+                        }
+                    }
+                }
+                current.as_mut().ok_or(BadKurbo::MissingMove)?.line_to(&pts)
+            }
+            kurbo::PathEl::QuadTo(_, _) => {
+                quad_pts.clear();
+                for el in &elements {
+                    match el {
+                        &kurbo::PathEl::QuadTo(p0, p1) => {
+                            quad_pts.push((p0, p1));
+                        }
+                        _ => {
+                            return Err(BadKurbo::InconsistentPathElements(i, el_types(&elements)))
+                        }
+                    }
+                }
+                current
+                    .as_mut()
+                    .ok_or(BadKurbo::MissingMove)?
+                    .quad_to(&quad_pts)
             }
             kurbo::PathEl::CurveTo(_, _, _) => return Err(BadKurbo::HasCubic),
             kurbo::PathEl::ClosePath => {
@@ -340,7 +349,7 @@ pub fn simple_glyphs_from_kurbo(paths: &[BezPath]) -> Result<Vec<SimpleGlyph>, B
                 // remove last point in closed path if has same coords as the move point
                 // matches FontTools handling @ https://github.com/fonttools/fonttools/blob/3b9a73ff8379ab49d3ce35aaaaf04b3a7d9d1655/Lib/fontTools/pens/pointPen.py#L321-L323
                 // FontTools has an else case to support UFO glif's choice to not include 'move' for closed paths that does not apply here.
-                if contour.num_points() > 1 && contour.last() == contour.first() {
+                if contour.num_points() > 1 && contour.last().eq(contour.first()) {
                     contour.remove_last();
                 }
             }
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -1175,6 +1175,26 @@ mod tests {
         simple_glyphs_from_kurbo(&[path1, path2]).unwrap();
     }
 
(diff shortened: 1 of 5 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -105,6 +109,283 @@ impl OtPoint for (i16, i16) {
...
+    }
+
+    fn distance(&self, other: &Self) -> f64 {
+        self.point.distance(other.point)
+    }
+}
+
+impl From<ContourPoint> for CurvePoint {
+    fn from(pt: ContourPoint) -> Self {
+        let (x, y) = pt.point.get();
```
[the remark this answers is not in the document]
rsheeter: Create PR is crashing, suggested diff shown in https://github.com/googlefonts/fontations/compare/filter-implied-on-curves...otround

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
-    fn get(self) -> (i16, i16);
+/// Indices of on-curve points that are at equal distance between two off-curve points
+fn impliable_oncurve_points(points: &[ContourPoint], tolerance: f64) -> HashSet<usize> {
+    let mut result = HashSet::new();
+    for (i, p1) in points.iter().enumerate() {
+        if !p1.on_curve {
+            continue;
+        }
+        let prv = if i > 0 { i - 1 } else { points.len() - 1 };
+        let nxt = if i < points.len() - 1 { i + 1 } else { 0 };
```
rsheeter, line 243: I suggest you implement a wrapping read function and use it here, e.g. prev = wrapping_read(i-1), etc

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
+                            None
+                        } else {
+                            Some(CurvePoint::from(pt))
+                        }
+                    })
+                    .collect(),
+            ))
+        }
+        contours
+    }
```
rsheeter, line 232: I find myself wondering if we can implement what the comment says more directly. Rather than allocating hash sets do a forward walk and filter if all points at index are implicit.

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
+/// Iterator that iterates over a vector of iterators simultaneously
+struct MultiZip<I: Iterator>(Vec<I>);
+
+impl<I: Iterator> Iterator for MultiZip<I> {
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.iter_mut().map(Iterator::next).collect()
     }
 }
```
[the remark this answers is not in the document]
anthrotype: yeah, didn't know where exactly to put it so.. any suggestions?

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
+    /// Add a line segment to all contours
+    fn line_to(&mut self, pts: Vec<kurbo::Point>) {
+        assert_eq!(pts.len(), self.len());
+        for (i, pt) in pts.into_iter().enumerate() {
+            self.0[i].push(ContourPoint::on_curve(pt));
+        }
+    }
+
+    /// Add a quadratic curve segment to all contours
+    fn quad_to(&mut self, p0s: Vec<kurbo::Point>, p1s: Vec<kurbo::Point>) {
```
dfrg, line 162: Can we take `&[(kurbo::Point, kurbo::Point)]` here instead? It would avoid the need for the zip and the length check.

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
+
+        // elements is never empty (if it were, MultiZip would have stopped), hence the unwrap
+        let first_el = elements.first().unwrap();
+        match first_el {
+            kurbo::PathEl::MoveTo(_) => {
+                // we have a new contour, flush the current one
+                if let Some(prev) = current.take() {
+                    contours.push(prev);
+                }
+                let mut pts = Vec::with_capacity(num_glyphs);
```
dfrg, line 305: If the `new`, `line_to` and `quad_to` methods are changed to take slices, these allocations can be reused. I don’t want to stumble into premature optimization but we did have a recent patch to address malloc in a profile and these per point allocations are likely to be high frequency.

write-fonts/src/tables/glyf.rs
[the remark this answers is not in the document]
dfrg: I think having a generic `util` module at the crate root is reasonable.

write-fonts/src/tables/glyf.rs
rsheeter: #377 shows an exploratory implementation

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
-/// A helper trait for converting other point types to open-type compatible reprs
-pub trait OtPoint {
-    fn get(self) -> (i16, i16);
+/// Indices of on-curve points that are at equal distance between two off-curve points
+fn impliable_oncurve_points(points: &[ContourPoint], tolerance: f64) -> HashSet<usize> {
+    let mut result = HashSet::new();
+    for (i, p1) in points.iter().enumerate() {
+        if !p1.on_curve {
+            continue;
+        }
```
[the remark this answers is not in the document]
rsheeter: #377 covers this

write-fonts/src/tables/glyf.rs
rsheeter: #377 covers this

write-fonts/src/tables/glyf.rs
```diff
@@ -123,15 +387,15 @@ impl Contour {
     }
 
     /// Add a line segment
-    pub fn line_to(&mut self, pt: impl OtPoint) {
-        let (x, y) = pt.get();
+    pub fn line_to(&mut self, pt: impl OtRound<(i16, i16)>) {
```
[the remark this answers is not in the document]
rsheeter: > Contour::{line_to,quad_to} are no longer used by internal code

If fontmake-rs compiles with those methods gone I'd vote delete. No code is the most maintainable code :)

> continue to work like it did before when one passes in an already tuple of (i16, 16)

I didn't implement OtRound on i16 at all so no. My test compile didn't seem to hit that case. We certainly implement it as a nop if we need to.

write-fonts/src/tables/glyf.rs
```diff
@@ -939,6 +1170,469 @@ mod tests {
...
+        let mut path2 = BezPath::new();
+        path2.move_to((3., 3.));
+        path2.quad_to((4., 4.), (5., 5.));
+        path2.line_to((3., 3.));
+        path2.close_path();
+
+        simple_glyphs_from_kurbo(&[path1, path2]).unwrap();
+    }
+
+    fn make_interpolatable_paths(
```
rsheeter, line 1213: this is long enough I can't tell what it does - what will the resulting paths look like - at a glance, perhaps a brief comment is warranted?

write-fonts/src/tables/glyf.rs
```diff
@@ -939,6 +1170,469 @@ mod tests {
         assert_eq!(points[3].y, -20);
     }
 
+    #[test]
+    #[should_panic(expected = "UnequalNumberOfElements")]
```
rsheeter, line 1174: asserting the expected result might be nicer than unwrap to panic?

write-fonts/src/tables/glyf.rs
```diff
@@ -86,29 +88,291 @@ pub enum BadKurbo {
...
+
+/// A helper struct for building interpolatable contours
+///
+/// Holds a vec of contour, one contour per glyph.
+#[derive(Clone, Debug, PartialEq)]
+struct InterpolatableContourBuilder(Vec<Vec<ContourPoint>>);
+
+impl InterpolatableContourBuilder {
+    /// Create new set of interpolatable contours beginning at the provided points
+    fn new(move_pts: Vec<kurbo::Point>) -> Self {
```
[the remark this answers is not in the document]
anthrotype: well, it's doing into_iter internally so I believe it is consuming the vecs. Note this whole struct is private, it doesn't need to be nice. These points are not usually already in a continuous slice, the caller has to make a vector anyway since the points belong to distinct BezPaths. Maybe I'm missing something else..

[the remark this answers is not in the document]
anthrotype: the "reused" part is what I don't quite understand at the moment (by what, where?) but I will try to do what you say and hopefully rustc will guide me through it ;)

write-fonts/src/tables/glyf.rs
dfrg: No worries! You captured the bulk of it. I’ll do a PR against this branch demonstrating what I mean about reusing the vecs.

dfrg: Hopefully #378 adds some clarity!

write-fonts/src/tables/glyf.rs
```diff
@@ -939,6 +1170,469 @@ mod tests {
...
+                CurvePoint::on_curve(4, 4),
+                CurvePoint::off_curve(5, 5),
+                // CurvePoint::on_curve(6, 6),  // implied
+                CurvePoint::off_curve(7, 7),
+            ]],
+        );
+    }
+
+    #[test]
+    fn simple_glyphs_from_kurbo_2_quads_1_line_implictly_closed() {
```
rsheeter, line 1351: general nit: I didn't always find it obvious what each test was testing from the name. This was true across many of the tests. For example, here eliminate_implied_on_curve_between_quads might better describe it?

write-fonts/src/tables/glyf.rs
```diff
@@ -939,6 +1170,469 @@ mod tests {
...
+                CurvePoint::on_curve(4, 4),
+                CurvePoint::on_curve(5, 5),
+                CurvePoint::on_curve(3, 3),
+            ]],
+        );
+    }
+
+    #[test]
+    fn simple_glyphs_from_kurbo_4_lines_closed_duplicate_last_pt() {
+        for implicit_closing_line in &[true, false] {
```
rsheeter, line 1540: Optional, could we use one of the libraries you were exploring for parameterized tests here?

rsheeter (approved):
Couple of comments, broadly lgtm

dfrg (approved):
LGTM as well

write-fonts/src/tables/glyf.rs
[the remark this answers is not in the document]
anthrotype: I'll remove the unused (resolving this)

write-fonts/src/tables/glyf.rs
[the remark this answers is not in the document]
anthrotype: ok I'll move it to a new write_fonts::util. I made them pub so we could also use it in fontmake-rs (e.g. when iterating over multiple BezPaths for cu2quing them interpolatably)

write-fonts/src/tables/glyf.rs
anthrotype: I added a docstring in https://github.com/googlefonts/fontations/pull/368/commits/57d9d719e57de249504931dad8db44394ea78a3c

write-fonts/src/tables/glyf.rs
anthrotype: done https://github.com/googlefonts/fontations/pull/368/commits/7d0f30686ef31edb069760b1c8d0b0fe14c12731

write-fonts/src/tables/glyf.rs
anthrotype: yeah, but maybe in a different PR

Issue #370, opened 2023-04-26
ReverseContourPen: duplicate LineTo after MoveTo when contour ends w/ CurveTo


anthrotype:
Take this simple closed contour that starts and ends with a CurveTo:

```rust
let contour = BezPath::from_vec(vec![
    PathEl::MoveTo((0.0, 0.0).into()),
    PathEl::CurveTo(
        (1.0, 1.0).into(),
        (2.0, 2.0).into(),
        (3.0, 3.0).into(),
    ),
    PathEl::CurveTo(
        (4.0, 4.0).into(),
        (5.0, 5.0).into(),
        (0.0, 0.0).into(),
    ),
    PathEl::ClosePath,
]);
let mut bez_pen = BezPathPen::new();
let mut rev_pen = ReverseContourPen::new(&mut bez_pen);
write_to_pen(&contour, &mut rev_pen);
rev_pen
    .flush()
    .unwrap();
let reversed = bez_pen.into_inner();
println!("{:#?}", reversed);
```

After reversing with ReverseContourPen, a duplicate LineTo segment appears after the MoveTo which was not there to begin with.

```diff
BezPath(
    [
        MoveTo(
            (0.0, 0.0),
        ),
-        LineTo(
-           (0.0, 0.0),
-       ),
        CurveTo(
            (5.0, 5.0),
            (4.0, 4.0),
            (3.0, 3.0),
        ),
        CurveTo(
            (2.0, 2.0),
            (1.0, 1.0),
            (0.0, 0.0),
        ),
    ],
)
```

This will produce an extra point if we compile that BezPath to a glyf SimpleGlyph with fontc. We need to find where that exactly gets added and get rid of it.

rsheeter:
IIRC this was meant to help retain interpolation compatibility when reversing.

anthrotype:
no no, b/c fontmake does not add them and contours are still interpolation compatible, so we do have some mismatch somewhere

anthrotype:
we don't want the duplicate points to be added, they only appear when ReverseContourPen gets used, so we should fix ReverseContourPen to match the python equivalent (which has no qualms about interpolation compatibility)

anthrotype:
[...]

PR #371, opened 2023-04-26
add (failing) test reversing contour starting & ending w/ CurveTo

anthrotype:
meant to reproduce #370 

Currently this fails because ReverseContourPen seems to add an extra LineTo duplicating the MoveTo...

### DIFF
--- a/write-fonts/src/pens.rs
+++ b/write-fonts/src/pens.rs
@@ -362,9 +362,9 @@ impl Pen for RecordingPen {
 #[cfg(test)]
 mod tests {
     use font_types::{Pen, PenCommand};
-    use kurbo::Affine;
+    use kurbo::{Affine, BezPath, PathEl};
 
-    use super::{BezPathPen, RecordingPen, ReverseContourPen, TransformPen};
+    use super::{write_to_pen, BezPathPen, RecordingPen, ReverseContourPen, TransformPen};
 
     fn draw_open_test_shape(pen: &mut impl Pen) {
         pen.move_to(10.0, 10.0);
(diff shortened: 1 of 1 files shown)


anthrotype:
I implemented @behdad's idea of starting from the last point of the contour so we retain any implied closing line in the original contour. If the original contour has an explicit closing line then the reversed contour is identical to the one that FontTools' ReverseContourPen would produce.

With this (plus https://github.com/googlefonts/fontc/pull/340) I finally got Oswald's glyf table to be 100% identical to the one fontmake produces! 🎉

write-fonts/src/pens.rs
```diff
@@ -221,16 +230,14 @@ fn flush_subpath<T: Pen>(commands: &[PenCommand], pen: &mut T) -> Result<(), Con
                 x: end_x,
                 y: end_y,
             },
-            // Close is a line from (end_x,end_y)=>(start_x, start_y) so reversed it's a line to end x/y
-            PenCommand::Close => PenCommand::LineTo { x: end_x, y: end_y },
+            PenCommand::Close => {
+                panic!("Subpath should have 0 or 1 close, and it should already have been removed")
```
rsheeter, line 234: I would prefer this be an error, akin to subpath does not start with moveto above

write-fonts/src/pens.rs
```diff
@@ -169,31 +169,40 @@ pub struct ReverseContourPen<'a, T: Pen> {
...
-    let (end_x, end_y) = commands
-        .last()
-        .unwrap()
-        .end_point()
-        .unwrap_or((start_x, start_y));
+    // When reversed, the move is to the end point of the last segment,
+    // i.e. the last command in an open contour, or the penultimate command
+    // in a closed contour.
+    let is_closed = *commands.last().unwrap() == PenCommand::Close;
+    let last_segment_idx = if is_closed {
```
rsheeter, line 192: I think it would be worth a comment to note that -2 must be a valid index because we know that first is move and last is close so there must be at least move/close. Since you are headed on vacation I will add the comment and merge.

write-fonts/src/pens.rs
rsheeter: Since you are headed on vacation I will adjust then merge.

write-fonts/src/pens.rs
```diff
@@ -548,4 +556,260 @@ mod tests {
         assert_eq!(Some(Rect::new(50.0, 50.0, 350.0, 300.0)), pen.bounds());
         assert!(pen.bounds().unwrap().area() > bez.bounding_box().area());
     }
+
+    // Direct port of fonttools' reverseContourPen_test.py::test_reverse_pen, adapted to rust,
+    // excluding test cases that don't apply because we don't implement outputImpliedClosingLine=False.
+    // https://github.com/fonttools/fonttools/blob/85c80be/Tests/pens/reverseContourPen_test.py#L6-L467
+    #[rstest]
+    #[case::closed_last_line_not_on_move(
```
rsheeter, line 564: Given our offset use I think we should add a test with only a move and a test with only move/close. Will add prior to merging.

write-fonts/src/pens.rs
rsheeter: Done

PR #372, opened 2023-04-28
Update var store delta computation

dfrg:
Changed to use 64-bit precision when computing scaled deltas.

Also changed `ItemVariationStore::compute_delta()` to return `i32` rather than `Fixed`. The usage of the fixed type was based on an assumption that no longer holds with the variation deltas stored in the `COLR` table.

The changes required a few tests to be updated to match a small difference in rounding behavior that we inherited from updating some code to match FreeType 2.13.


cmyr (approved):
makes sense that we shake out some issues through use.

PR #374, opened 2023-04-28
Fix miscalculation of segment map size in avar

cmyr:
This is a curious case; avar is unique in that it has a hand written impl of the `VarSize` trait, and that impl was not accounting for the length of the count field itself when determining the size. This meant that the size could be zero, which would prevent us from advancing while iterating, leading to a hang.

This works fine when there was only a single segment map, since we do not truncate the data when reading the type; but it would cause us to read incorrect values when tables contained multiple axes.

This fixes the issue, and adds a test case with multiple axes.

The big question I have, though, is: why haven't we hit this before? Or maybe we have hit this before, and I haven't noticed?

cc @dfrg, want to double check that this makes sense?

fixes #367

### DIFF
--- a/read-fonts/src/tables/avar.rs
+++ b/read-fonts/src/tables/avar.rs
@@ -35,7 +35,10 @@ impl<'a> VarSize for SegmentMaps<'a> {
     type Size = u16;
 
     fn read_len_at(data: FontData, pos: usize) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


dfrg (approved):
This was definitely an oversight on my part and certainly demonstrates the value of codegen when these bugs pop up in hand written code. The fix looks solid, thanks!

rsheeter:
This also perhaps supports hooking up fuzzing.

Issue #375, opened 2023-04-28
Testing support for COLR outlines


dfrg:
Testing is currently non-existent for color outlines. The plan is to build something similar to [extract_glyphs.py](https://github.com/googlefonts/fontations/blob/c84689133f0565d50e02a90e8824450b8aab2a9a/font-test-data/test_data/extract_glyphs.py) and use FreeType to generate serialized paint graphs that can be tested against our code.

We'll likely want to start with the test font available at https://github.com/googlefonts/color-fonts/blob/main/fonts/test_glyphs-glyf_colr_1_variable.ttf. The [build script](https://github.com/googlefonts/color-fonts/blob/main/config/test_glyphs-glyf_colr_1.py) for that font has useful info.

PR #376, opened 2023-04-28
[read-fonts] Add useful helpers for reading COLR glyphs

dfrg:
This adds an intermediate "resolved" state for the paint graph where all values are promoted to `Fixed`, variation deltas are applied and transforms are collapsed into a single variant for each type (transform, translate, scale, rotate, skew).

The result is roughly equivalent to FreeType's [FT_COLR_Paint](https://freetype.org/freetype2/docs/reference/ft2-layer_management.html#ft_colr_paint) type. While the structure is slightly different, but the value of numeric fields _should_ be identical.

Also adds helper methods to the `Colr` table to retrieve base glyphs, layers, clip boxes, etc.

Lots of code, but the bulk of it is just mechanical type conversions.

There are no tests here yet. Opened #375 to track.

### DIFF
--- a/read-fonts/src/tables/colr.rs
+++ b/read-fonts/src/tables/colr.rs
@@ -109,7 +109,7 @@ pub struct ColrInstance<'a> {
 }
 
 impl<'a> ColrInstance<'a> {
-    /// Creates a new resolver for the given `COLR` table and normalized variation
+    /// Creates a new instance for the given `COLR` table and normalized variation
     /// coordinates.
     pub fn new(colr: Colr<'a>, coords: &'a [F2Dot14]) -> Self {
         let index_map = colr.var_index_map().and_then(|res| res.ok());
@@ -207,6 +207,10 @@ pub struct ResolvedColorStop {
 }
 
 /// Collection of [`ColorStop`] or [`VarColorStop`].
+// Note: only one of these fields is used at any given time, but this structure
+// was chosen over the obvious enum approach for simplicity in generating a
+// single concrete type for the `impl Iterator` return type of the `resolve`
+// method.
 pub struct ColorStops<'a> {
     stops: &'a [ColorStop],
     var_stops: &'a [VarColorStop],
--- a/read-fonts/src/tables/colr.rs
+++ b/read-fonts/src/tables/colr.rs
@@ -1,5 +1,696 @@
 //! The [COLR](https://docs.microsoft.com/en-us/typography/opentype/spec/colr) table
 
-use super::variations::{DeltaSetIndexMap, ItemVariationStore};
+use super::variations::{DeltaSetIndex, DeltaSetIndexMap, ItemVariationStore};
+use types::BoundingBox;
+
+use std::ops::Deref;
 
 include!("../../generated/generated_colr.rs");
+
+/// Identifier used for representing a paint on the recursion blacklist.
+pub type PaintId = usize;
+
+impl<'a> Colr<'a> {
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/colr.rs
```diff
@@ -1,5 +1,696 @@
...
+/// variation deltas.
+#[derive(Clone, Debug)]
+pub struct ResolvedColorStop {
+    pub offset: Fixed,
+    pub palette_index: u16,
+    pub alpha: Fixed,
+}
+
+/// Collection of [`ColorStop`] or [`VarColorStop`].
+pub struct ColorStops<'a> {
```
anthrotype, line 210: if it _either_ contains one _or_ the other, never both, should this be a enum with two variants?

read-fonts/src/tables/colr.rs
```diff
@@ -1,5 +1,696 @@
...
+                    center_y: gradient.center_y().apply_delta(deltas[1]),
+                    start_angle: gradient.start_angle().apply_delta(deltas[2]),
+                    end_angle: gradient.end_angle().apply_delta(deltas[3]),
+                    color_stops: color_line.into(),
+                    extend,
+                }
+            }
+            Self::Glyph(glyph) => ResolvedPaint::Glyph {
+                glyph_id: glyph.glyph_id(),
+                paint: glyph.paint()?,
```
anthrotype, line 462: is it intentional that this does not resolve all sub-paints recursively? probably so, but I figured I'd ask.
Is it even possible for a Rust struct (e.g. ResolvedPaint) to have a field that is of the same type as itself?

read-fonts/src/tables/colr.rs
anthrotype: no problems, I just asked out of curiousity (and based on my understanding that color stops can only ever be defined as either all variable or none at all). Just do whatever feels more convenient here

read-fonts/src/tables/colr.rs
dfrg: Yes, this is intentional. Fully resolving the graph up front would require boxing the recursive type.

The relationship between `Paint` and `ResolvedPaint` is similar to `FT_OpaquePaint` and `FT_COLR_Paint`.

read-fonts/src/tables/colr.rs
```diff
@@ -1,5 +1,696 @@
...
+    index_map: Option<DeltaSetIndexMap<'a>>,
+    var_store: Option<ItemVariationStore<'a>>,
+    coords: &'a [F2Dot14],
+}
+
+impl<'a> ColrInstance<'a> {
+    /// Creates a new resolver for the given `COLR` table and normalized variation
+    /// coordinates.
+    pub fn new(colr: Colr<'a>, coords: &'a [F2Dot14]) -> Self {
+        let index_map = colr.var_index_map().and_then(|res| res.ok());
```
cmyr, line 115: just trivia but you can also write this as `colr.var_index_map().transpose().ok().flatten()`.

cmyr (approved):
I trust that the actual choices in the API are directly motivated by your needs as a consumer, so I won't second guess you there (and nothing really jumped out in any case!) and I don't really have any nits either, so lgtm!

read-fonts/src/tables/colr.rs
dfrg: I _think_ I had the transpose version in there first and clippy yelled at me about it 🤷‍♂️

PR #377, opened 2023-05-04
Filter implied points directly

rsheeter:
Exporing one of my own suggestions on #368


anthrotype (approved):
I like this, much better thanks! LGTM (need clippy fixes)

anthrotype:
I'm going to merge so I can apply the rest of my changes to my branch and avoid conflicts

PR #378, opened 2023-05-04
Reuse temporary allocations for implied on curve filtering

dfrg:
Allocates temporary  `pts` and `quad_pts` buffers in `simple_glyphs_from_kurbo` that are reused for each path element.

Also changes the `first` and `last` contour point methods in `InterpolatableContourBuilder` to return iterators which can be compared with the `Iterator::eq` method (rather than allocating vecs for comparison).


anthrotype (approved):
Ah yes, I see what you meant now! LGTM

PR #384, opened 2023-05-05
Add log support to table packing

cmyr:
Logging will be important for things like table splitting and promotion, so this makes the log crate a non-dev dependency in `write-fonts`.


cmyr:
nm, I'll bundle this with another commit in #385

PR #385, opened 2023-05-05
Refactor topo sort method

cmyr:
Splits this in two, and adds a separate 'has_overflows' method that is slightly faster.

This includes the commit from #384, makes sense to just PR these together.

### DIFF
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["text-processing", "parsing", "graphics"]
 [dependencies]
 font-types = { version = "0.1.5", path = "../font-types" }
 read-fonts = { version = "0.2.0", path = "../read-fonts" }
+log = "0.4"
 bitflags = "1.3"
 kurbo = "0.9.4"
 
@@ -21,5 +22,4 @@ diff = "0.1.12"
 ansi_term = "0.12.1"
 font-test-data = { path = "../font-test-data" }
 read-fonts = { version = "0.2.0", path = "../read-fonts", features = [ "codegen_test"] }
-log = "0.4"
 env_logger = "0.10.0"
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -203,24 +203,67 @@ impl Graph {
         }
     }
 
+    /// Attempt to pack the graph.
+    ///
+    /// This involves finding an order for objects such that all offsets are
+    /// resolveable.
+    ///
+    /// In the simple case, this just means finding a topological ordering.
+    /// In exceptional cases, however, this may require us to significantly
+    /// modify the graph.
+    ///
+    /// Our implementation is closely modeled on the implementation in the
+    /// HarfBuzz repacker; see the [repacker docs] for further detail.
+    ///
     /// returns `true` if a solution is found, `false` otherwise
-    pub(crate) fn topological_sort(&mut self) -> bool {
-        log::info!("attempting to pack {} objects", self.objects.len());
+    ///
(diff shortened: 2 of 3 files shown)


write-fonts/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["text-processing", "parsing", "graphics"]
 [dependencies]
 font-types = { version = "0.1.5", path = "../font-types" }
 read-fonts = { version = "0.2.0", path = "../read-fonts" }
+log = "0.4"
```
rsheeter, line 16: I wonder, do we actually unconditionally want log? Suppose we had a subsetter for use in pdf processing, should log be required?

write-fonts/src/graph.rs
```diff
@@ -203,17 +203,62 @@ impl Graph {
...
+    /// This involves finding an order for objects such that all offsets are
+    /// resolveable.
+    ///
+    /// In the simple case, this just means finding a topological ordering.
+    /// In exceptional cases, however, this may require us to significantly
+    /// modify the graph.
+    ///
     /// returns `true` if a solution is found, `false` otherwise
-    pub(crate) fn topological_sort(&mut self) -> bool {
+    pub(crate) fn pack_objects(&mut self) -> bool {
```
rsheeter, line 216: this should reference prior and relevant art. I believe in this case that means harfbuzz and in particular Garret's writing on the topic?

write-fonts/src/graph.rs
```diff
@@ -203,17 +203,62 @@ impl Graph {
...
+        self.sort_shortest_distance();
+
+        let n_overflows = self.find_overflows().len();
+        if n_overflows > 0 {
+            log::info!("packing failed with {n_overflows} overflows");
+        }
+        n_overflows == 0
+    }
+
+    /// Initial sorting operation. Attempt kahn, falling back to shortest distance.
```
rsheeter, line 231: Kahn. Also link to https://github.com/harfbuzz/harfbuzz/blob/main/docs/repacker.md#topological-sorting-algorithms? And perhaps the HarfBuzz implementation of the equivalent?

write-fonts/src/write.rs
```diff
@@ -43,7 +43,7 @@ pub fn dump_table<T: FontWrite + Validate>(table: &T) -> Result<Vec<u8>, Validat
     let mut writer = TableWriter::default();
     table.write_into(&mut writer);
     let mut graph = writer.finish();
-    if !graph.topological_sort() {
+    if !graph.pack_objects() {
         //TODO: implement extension promotion and table splitting
         panic!("could not find a graph packing that handles all offsets, aborting");
```
rsheeter, line 48: Still perplexed at the panic over Result

write-fonts/src/graph.rs
```diff
@@ -203,17 +203,62 @@ impl Graph {
...
+                if link.len.max_value() < rel_off {
+                    return true;
+                }
+            }
         }
-        self.find_overflows().is_empty()
+        false
     }
 
     fn find_overflows(&self) -> Vec<(ObjectId, ObjectId)> {
```
rsheeter, line 264: Are the tests in this file original or ported from HB? If they are original, was it not possible to port a test suite from HB? If they are ported, please add refs to the corresponding original test to each one.

rsheeter (approved):
LGTM provided the refs to prior art are improved (or will be in a followon)

write-fonts/Cargo.toml
cmyr: This will also come up in #387, where I add a dependency to write graphviz files. I could put both of these behind a 'diagnostics' feature, if desired?

Issue #192, opened 2022-12-21
Implement a good integer set ala (`hb_set`)


cmyr:
This wants to be an issue, presumably one that says "make me a memory safe hb_set"?

_Originally posted by @rsheeter in https://github.com/googlefonts/fontations/pull/188#discussion_r1053834809_

PR #386, opened 2023-05-05
More closely follow hb-repacker when isolating/assigning spaces

cmyr:
My previous implementation was a looser interpretation of hb, but for the promotion/splitting work I want to have as much of the general code match HB has possible, so I can know bugs are in the implementation, not in the design.

based on #385, which should go in first.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -54,6 +54,7 @@ impl Space {
     /// The first space used for assignment to specific subgraphs.
     const INIT: Space = Space(2);
 
+    #[cfg(test)]
     const fn is_custom(self) -> bool {
         self.0 >= Space::INIT.0
     }
@@ -223,7 +224,7 @@ impl Graph {
             return true;
         }
 
-        self.assign_32bit_spaces();
+        self.assign_spaces_hb();
         self.sort_shortest_distance();
 
         let n_overflows = self.find_overflows().len();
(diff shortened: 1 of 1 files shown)


write-fonts/src/graph.rs
```diff
@@ -408,157 +409,208 @@ impl Graph {
...
 
-        // assign all nodes reachable from 16/24 bit edges to space 0.
-        self.assign_space_0();
+        log::debug!("found {} space roots to isolate", roots.len());
 
-        while !roots.is_empty() {
-            let root = *roots.iter().next().unwrap();
-            self.isolate_and_assign_space(root);
-            roots.remove(&root);
+        // we want to *invert* the visited set, but we don't have a fancy hb_set_t
```
rsheeter, line 430: should we file an issue to write an hb_set_t? - I believe some of hb's structures are very carefully tuned to work well for text.

write-fonts/src/graph.rs
```diff
@@ -408,157 +409,208 @@ impl Graph {
...
+        let _all_ids = self.order.iter().copied().collect::<HashSet<_>>();
```
rsheeter, line 431: why did you need to bind _all_ids? It looks like you only use it the only time?

rsheeter (approved):
I won't claim full understanding of the algorithms - if you want that Garret needs to review - but the Rust code seems reasonable enough.

LGTM with the same request as the prior CL: lets be very dilligent about linking to the prior art on a fine grained basis. https://github.com/googlefonts/fontmake-rs/blob/main/fontir/src/variations.rs shows my attempt to do this for some of the py:rust conversion.

write-fonts/src/graph.rs
cmyr: We've got one (#192)

PR #387, opened 2023-05-05
Add graphviz support for debugging

cmyr:
This adds the ability to write out a graphviz file of the object graph, so we can visualize what is happening when packing fails (or when we're curious).

To make the graph more useful, I've also added a method to the `FontWrite` trait that lets tables include a name; this way we can display the name in the graph.

### DIFF
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -9,6 +9,7 @@ readme = "README.md"
 categories = ["text-processing", "parsing", "graphics"]
 
 [features]
+default = ["dot2"]
 
 [dependencies]
 font-types = { version = "0.1.5", path = "../font-types" }
@@ -16,6 +17,7 @@ read-fonts = { version = "0.2.0", path = "../read-fonts" }
 log = "0.4"
 bitflags = "1.3"
 kurbo = "0.9.4"
+dot2 = { version = "1.0", optional = true }
 
 [dev-dependencies]
 diff = "0.1.12"
--- a/write-fonts/src/graph/graphviz.rs
+++ b/write-fonts/src/graph/graphviz.rs
@@ -1,0 +1,133 @@
+//! Support for generating graphviz files from our object graph
+
+use super::{Graph, ObjectId, OffsetLen, Space};
+
+pub struct GraphVizGraph<'a> {
+    graph: &'a Graph,
+    edges: Vec<GraphVizEdge>,
+}
+
+impl<'a> GraphVizGraph<'a> {
+    pub(crate) fn from_graph(graph: &'a Graph) -> Self {
+        let mut edges = Vec::new();
+        for (parent_id, table) in &graph.objects {
+            let parent = &graph.nodes[parent_id];
+            for link in &table.offsets {
+                let child = &graph.nodes[&link.object];
+                let len = child.position - parent.position;
+                edges.push(GraphVizEdge {
+                    source: *parent_id,
+                    target: link.object,
+                    len,
+                    type_: link.len,
+                });
+            }
+        }
+        GraphVizGraph { graph, edges }
+    }
+
+    /// Write out this graph as a graphviz file to the provided path.
+    ///
+    /// Overwrites any existing file at this location.
+    pub fn write_to_file(&self, path: impl AsRef<std::path::Path>) -> std::io::Result<()> {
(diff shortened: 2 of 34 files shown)


write-fonts/Cargo.toml
```diff
@@ -16,6 +16,7 @@ read-fonts = { version = "0.2.0", path = "../read-fonts" }
 log = "0.4"
 bitflags = "1.3"
 kurbo = "0.9.4"
+dot2 = { version = "1.0" }
```
rsheeter, line 19: I immediately wonder if this should be an opt-in feature ?

write-fonts/src/graph/graphviz.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    fn node_id(&'a self, n: &Self::Node) -> dot2::Result<dot2::Id<'a>> {
+        dot2::Id::new(format!("N{}", n.0))
+    }
+
+    fn node_label<'b>(&'b self, n: &Self::Node) -> dot2::Result<dot2::label::Text<'b>> {
+        let obj = &self.graph.objects[n];
+        let node = &self.graph.nodes[n];
+
+        let name = if obj.name == "mock data" {
```
rsheeter, line 105: why is it necessary to special case this?

write-fonts/src/write.rs
```diff
@@ -44,6 +47,18 @@ pub fn dump_table<T: FontWrite + Validate>(table: &T) -> Result<Vec<u8>, Validat
...
+        let time = std::time::SystemTime::now()
+            .duration_since(std::time::SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        let path = format!("{}_failure_{time}.dot", table.name());
+        if !std::path::Path::new(&path).exists() {
+            graph
+                .write_graph_viz(path)
+                .expect("failed to save graphviz");
+        }
```
rsheeter, line 61: I think we should fix the fixme, having this unconditional happen, or even be possible to happen, feels off.

write-fonts/src/write.rs
```diff
@@ -178,12 +195,28 @@ impl Default for TableWriter {
 }
 
 /// The encoded data for a given table, along with info on included offsets
-#[derive(Debug, Default, Clone, Hash, PartialEq, Eq)]
+#[derive(Debug, Default, Clone)] // DO NOT DERIVE MORE TRAITS! we want to ignore name field
```
rsheeter, line 198: At a glance I would think if you let the name be included the practical outcome is that you can't reuse the bytes of one table as a valid table of some other type? Is that something we actually do? Are there other implications?

write-fonts/src/write.rs
rsheeter: Also I dislike that it blindly tries to write; what if I specifically want to do an entirely in-memory compile on a server? - I don't want it spewing graphviz everywhere.

write-fonts/src/write.rs
```diff
@@ -44,6 +47,18 @@ pub fn dump_table<T: FontWrite + Validate>(table: &T) -> Result<Vec<u8>, Validat
...
+            .unwrap()
+            .as_secs();
+        let path = format!("{}_failure_{time}.dot", table.name());
+        if !std::path::Path::new(&path).exists() {
+            graph
+                .write_graph_viz(path)
+                .expect("failed to save graphviz");
+        }
         //TODO: implement extension promotion and table splitting
         panic!("could not find a graph packing that handles all offsets, aborting");
```
rsheeter, line 63: panic in a method with a Result?

write-fonts/src/write.rs
cmyr: Correct, we will currently reuse a table as long as the bytes are identical. I do not have any stats on how often this happens in practice, but it is part of the design.

write-fonts/src/write.rs
cmyr: resolved in #389

write-fonts/src/graph/graphviz.rs
cmyr: This came as a result of debugging some test failures; in tests all objects are named "mock data", but it's more useful if we can see the object ids.

cmyr:
Okay this is updated to address the concerns made here, I think it should be gtg @rsheeter

PR #388, opened 2023-05-05
Implement subgraph splitting for overflow resolution

cmyr:
A port of the harfbuzz implementation; when overflows occur in subgraphs with multiple 32-bit roots (e.g. extension subtables sharing coverage tables or similar) we will iteratively select some subset of those roots and move them to another subgraph, duplicating and dependencies.

~based on #386, which should go in first.~

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -57,7 +57,6 @@ impl Space {
     /// The first space used for assignment to specific subgraphs.
     const INIT: Space = Space(2);
 
-    #[cfg(test)]
     const fn is_custom(self) -> bool {
         self.0 >= Space::INIT.0
     }
@@ -90,6 +89,7 @@ pub struct Graph {
     distance_invalid: bool,
     positions_invalid: bool,
     next_space: Space,
+    num_roots_per_space: HashMap<Space, usize>,
 }
 
 #[derive(Debug)]
@@ -117,6 +117,14 @@ struct Distance {
 //TODO: remove me? maybe? not really used right now...
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 struct Priority(u8);
+
+/// A record of an overflowing offset
+pub(crate) struct Overflow {
+    parent: ObjectId,
+    child: ObjectId,
+    distance: u32,
+    offset_type: OffsetLen,
+}
 
 impl Priority {
     const ZERO: Priority = Priority(0);
@@ -206,6 +214,7 @@ impl Graph {
             distance_invalid: true,
             positions_invalid: true,
             next_space: Space::INIT,
+            num_roots_per_space: Default::default(),
         }
     }
 
@@ -232,11 +241,32 @@ impl Graph {
         self.assign_spaces_hb();
         self.sort_shortest_distance();
 
-        let n_overflows = self.find_overflows().len();
-        if n_overflows > 0 {
-            log::info!("packing failed with {n_overflows} overflows");
+        if !self.has_overflows() {
+            return true;
         }
-        n_overflows == 0
+
+        // now isolate spaces in a loop, until there are no more left:
+        let overflows = loop {
+            let overflows = self.find_overflows();
+            if overflows.is_empty() {
+                // we're done
+                return true;
+            }
+            log::info!(
+                "failed with {} overflows, current size {}",
+                overflows.len(),
+                self.debug_len()
+            );
+            if !self.try_splitting_subgraphs(&overflows) {
+                log::info!("finished isolating all subgraphs without solution");
+                break overflows;
+            }
+            self.sort_shortest_distance();
+        };
+
+        assert!(!overflows.is_empty());
+        self.debug_overflows(&overflows);
+        false
     }
 
     /// Initial sorting operation. Attempt Kahn, falling back to shortest distance.
@@ -272,7 +302,7 @@ impl Graph {
         false
     }
 
-    pub(crate) fn find_overflows(&self) -> Vec<(ObjectId, ObjectId)> {
+    pub(crate) fn find_overflows(&self) -> Vec<Overflow> {
         let mut result = Vec::new();
         for (parent_id, data) in &self.objects {
             let parent = &self.nodes[parent_id];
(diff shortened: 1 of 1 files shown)


write-fonts/src/graph.rs
```diff
@@ -232,11 +241,33 @@ impl Graph {
...
+            if !self.try_splitting_subgraphs(&overflows) {
+                log::info!("finished isolating all subgraphs without solution");
+                break overflows;
+            }
+            self.sort_shortest_distance();
+        };
+
+        assert!(!overflows.is_empty());
+        self.debug_overflows(&overflows);
+        //FIXME: return some kind of error here
```
rsheeter, line 269: I suggest we might err on the side of issues over comments for this sort of thing, with possible exception for cases where we're going to fix it in O(days)

write-fonts/src/graph.rs
```diff
@@ -281,13 +312,47 @@ impl Graph {
...
+                        offset_type: link.len,
+                    });
                 }
             }
         }
         result
     }
 
+    fn debug_overflows(&self, overflows: &[Overflow]) {
+        let (parents, children): (HashSet<_>, HashSet<_>) =
```
rsheeter, line 328: if debug is not enabled early exit

rsheeter: also should be behind log feature per other conversation? - np if that's out of scope here though.

write-fonts/src/graph.rs
```diff
@@ -615,6 +681,63 @@ impl Graph {
...
+    ///
+    /// [_try_isolating_subgraphs]: https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-repacker.hh#L182
+    fn try_splitting_subgraphs(&mut self, overflows: &[Overflow]) -> bool {
+        let mut to_isolate = HashMap::new();
+        for overflow in overflows {
+            let child_space = self.nodes[&overflow.child].space;
+            // we only isolate subgraphs in wide-space
+            if !child_space.is_custom() || self.num_roots_per_space[&child_space] <= 1 {
+                continue;
+            }
```
rsheeter, line 699: Optional, looks like this would fit very naturally as `.iter().filter()`?

write-fonts/src/graph.rs
```diff
@@ -615,6 +681,63 @@ impl Graph {
...
+            debug_assert_eq!(self.nodes[&root].space, child_space);
+            to_isolate
+                .entry(child_space)
+                .or_insert_with(HashSet::new)
+                .insert(root);
+        }
+
+        if to_isolate.is_empty() {
+            return false;
+        }
```
rsheeter, line 710: At a glance it looks like you could  write this as to_isolate = overflows.iter.filter.map.collect?

write-fonts/src/graph.rs
```diff
@@ -900,6 +1025,66 @@ mod tests {
...
+        //      |                              |
+        //      2          (Lookup)            2
+        //     / \                            / \
+        //  ╔═3   4═╗   (ext subtables)    ╔═3   4═╗
+        //  ║       ║                      ║       ║   (long offsets)
+        //  5─┐   ┌─6    (subtables)       5       6
+        //  │ └─8─┘ │                     / \     / \
+        //  │       │    (cov tables)    7'  8'  7   8
+        //  └───7───┘
+        //
```
rsheeter, line 1046: Diagram appreciated, without it the test would be inscrutable

rsheeter (approved):

write-fonts/src/graph.rs
cmyr: Oops this comment is out of date; we check the return and convert to an error in the `dump_table` fn.

write-fonts/src/graph.rs
```diff
@@ -232,11 +241,33 @@ impl Graph {
...
-        let n_overflows = self.find_overflows().len();
-        if n_overflows > 0 {
-            log::info!("packing failed with {n_overflows} overflows");
+        if !self.has_overflows() {
+            return true;
         }
-        n_overflows == 0
+
+        // now isolate spaces in a loop, until there are no more left:
+        let overflows = loop {
```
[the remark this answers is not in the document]
cmyr: The problem is that we aren't always going to continue until there are no overflows left; sometimes overflow is unavoidable.

The control flow here is:
- check if there are overflows
- if there are none, return
- if there are some, try to make progress on resolving them
- if we don't make progress, break the loop and print diagnostics

This should always terminate, since we exit the loop if either there are no overflows, *or* if there are overflows but there are no splittable subgraphs remaining.

write-fonts/src/graph.rs
cmyr: yea we don't have a log feature yet, will revisit that when I make that change (opened  #397 to track this)

write-fonts/src/graph.rs
cmyr: Agreed, I found it impossible to come up with good test cases without drawing them out first.

dfrg:
QQ: `ObjectId`s are allocated with a global `usize` counter. Should we change this to an explicit `u64` (or allocate ids locally in the graph) to avoid potential overflows if someone does decide to run this on a 32-bit system?

cmyr:
> QQ: `ObjectId`s are allocated with a global `usize` counter. Should we change this to an explicit `u64` (or allocate ids locally in the graph) to avoid potential overflows if someone does decide to run this on a 32-bit system?

Hmm... I would expect the process to only be running for a specific compilation, which even in the case of a very large multi-access font it's hard to imagine there being 4.2e9 unique objects? And in that case I would definitely expect us to be on a 64-bit system. That said, it wouldn't *hurt*...

write-fonts/src/graph.rs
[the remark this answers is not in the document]
rsheeter: The comment above reads as if we are going to continue until there are none so that might need a fix then. IMHO while primary-exit-condition is still nicer, even if there is another path toward failure.

What about using `Result<(), SomeErrType>` so you can use normal control flow, e.g. `try_splitting_subgraphs(...)?`?

write-fonts/src/graph.rs
rsheeter: It looks like the early exit for log disabled - avoiding the iterations and zips to produce logs we never write - was not done?

PR #389, opened 2023-05-08
Return an error if packing fails

cmyr:
Currently not very useful, but we will expose the graphviz support through this error type.


rsheeter (approved):
LGTM. This is breaking for write-fonts right? Should we bump minor here or just remember to do it later?

cmyr:
Good note, I think we do it later since this batch of work my motivate additional changes.

PR #391, opened 2023-05-09
TableWriter and serialization tweaks

cmyr:
This is a pair of commits split out from the work on promotion and table splitting.

The first of these is a few small tweaks to the `TableWriter` type, renaming `write` to `write_bytes` and adding a `write` method for writing a scalar, which we will use when we do the manual table writing, as well as adding a convenience method for generating a graph from a root table.

The second of these commits moves the logic for actually writing out the final buffer into a `serialize` method on the `Graph` type, instead of being a free function in the `write` module. This was mainly motivated by making it easier during testing for us to dump out the bytes for a given graph, so that we can reparse them and verify things are as we expect.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -1,5 +1,7 @@
 //! A graph for resolving table offsets
 
+use font_types::Uint24;
+
 use super::write::TableData;
 use std::{
     collections::{BinaryHeap, HashMap, HashSet, VecDeque},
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -82,9 +84,9 @@ impl ObjectStore {
 
 #[derive(Debug)]
 pub struct Graph {
```
rsheeter, line 86: I think perhaps this one deserves a comment, even if it's just a "go read here to understand me"? - Graph of objects, nodes, and order with some invalid bits is extremely general and has non-obvious interpretation

write-fonts/src/graph.rs
```diff
@@ -209,6 +211,64 @@ impl Graph {
         }
     }
 
+    /// Write out the serialized graph.
+    ///
+    /// This is not public API, and you are responsible for ensuring that
+    /// the graph is sorted before calling.
```
rsheeter, line 217: how do I ensure it's sorted? what would happen if I didn't?

rsheeter: Not crazy about all the panic pathways, perhaps serialize needs a Result? - write panicking would be Very Bad in some proposed use cases and while I appreciate this _shouldn't_ happen ... we're making it more possible it can.

rsheeter (approved):
LGTM, with some concern about panics. But perhaps write vs panic is a larger conversation?

write-fonts/src/graph.rs
cmyr: You don't call this method; it's called by something like `dump_table`, which is responsible for creating the graph, trying to pack it, and then serializing if packing is successful.

There isn't actually anything in here that can panic unless the API is misused, and then it should panic every time it gets called.

PR #394, opened 2023-05-09
[write-fonts] Error tweaks

cmyr:
Some fixups motivated by trying to actually use these errors in fea-rs.

- Reexport errors at top-level
- Don't impl Debug for Graph, since it ends up being unintelligable

JMM™️

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -86,7 +86,7 @@ impl ObjectStore {
 ///
 /// This type is used during compilation, to determine the final write order
 /// for the various subtables.
-#[derive(Debug)]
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -80,7 +80,6 @@ impl ObjectStore {
     }
 }
 
-#[derive(Debug)]
 pub struct Graph {
```
rsheeter, line 83: I suggest a comment wrt why we don't derive Debug, otherwise it may be added back

rsheeter (approved):
LGTM but has conflicts so did not merge

PR #396, opened 2023-05-09
Add 'serde' feature to read/write-fonts

cmyr:
This is intended to be used to serialize test cases in write-fonts; it is also added in read-fonts since write-fonts reuses flags and enums declared in read-fonts.

One funny thing: (I think) this hit a serde bug around using non-literals to provide default values for const generics, which we do in the OffsetMarker types in write-fonts; I've opened an issue upstream at https://github.com/serde-rs/serde/issues/2449

### DIFF
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -11,6 +11,7 @@ categories = ["text-processing", "parsing", "graphics"]
 [features]
 default = ["dot2"]
 read = []
+serde = ["dep:serde", "font-types/serde", "read-fonts/serde"]
 
 [dependencies]
 font-types = { version = "0.4.0", path = "../font-types" }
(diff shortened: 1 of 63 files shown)


write-fonts/Cargo.toml
```diff
@@ -10,6 +10,7 @@ categories = ["text-processing", "parsing", "graphics"]
 
 [features]
 default = ["dot2"]
```
rsheeter, line 12: I don't think we want dot2 by default do we? [I know it's not changed here]

rsheeter (approved):

write-fonts/Cargo.toml
cmyr: not in the long-term, but for now it's nice to not need to pass an extra flag at the command line if I want to write out the graph in a test or similar 🤷

cmyr:
I'm holding out on this because I don't want to merge until I have a clear use. My *intended* use for this is to create binary dumps of real-world tables that require special handling (extension promotion or table splitting) and then use those in integration tests, but I'm leaving this as a draft until I actually get around to building that infrastructure.

PR #399, opened 2023-05-10
Add iterators for cmap

### LINKED ISSUE #390: Get a list of supported codepoints
Both `skrifa` and `read-fonts` offer to map a codepoint to a glyph for me, but I just want a list of what codepoints are supported by a font. Ideally, I'd just love a hashmap or similar mapping codepoints to glyph IDs.

Currently I'm having to do this, which does not feel ergonomic:

```rust
pub fn codepoints(&self) -> HashSet<u16> {
        let cmap = self.fontref().cmap().expect("Couldn't parse cmap table");
        let mut points = HashSet::new();
        let records = cmap.encoding_records();
        for record in records {
            if let Ok(subtable) = record.subtable(cmap.offset_data()) {
                match subtable {
                    CmapSubtable::Format4(c) => {
                        for (start, end) in c.start_code().iter().zip(c.end_code().iter()) {
                            for point in start.get()..end.get() {
                                points.insert(point);
                            }
                        }
                    }
                    CmapSubtable::Format12(_) => todo!(),
                    _ => {}
                }
            }
        }
        points
    }
```

dfrg:
Adds iterators to retrieve mapping data for cmap formats 4, 12 and 14 in read-fonts.

Also exposes this functionality in skrifa's `Charmap`.

Closes #390

### DIFF
--- a/read-fonts/src/tables/cmap.rs
+++ b/read-fonts/src/tables/cmap.rs
@@ -2,7 +2,7 @@
 
 include!("../../generated/generated_cmap.rs");
 
-/// Result of the mapping a codepoint with a variation selector.
+/// Result of mapping a codepoint with a variation selector.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum MapVariant {
     /// The variation selector should be ignored and the default mapping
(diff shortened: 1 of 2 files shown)


rsheeter (approved):
Not enormously enthused about exposing the iter types but I also don't see a convenient way around it given that the obvious path leads directly to try to return multiple types as `impl Iterator` from the charmap method.

PR #400, opened 2023-05-11
Implement GPOS/GSUB extension promotion

### LINKED ISSUE #303: thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()

cmyr:
The implementation here does not involve `write` types; instead we manually modify the graph in place. The implementation closely follows HarfBuzz.

This adds a `TableType` enum so that we can retain some type information after initial compilation; this somewhat overlaps with the `FontWrite::name` method added in #387, and maybe these could be combined?

This should get additional tests, and I'm working on porting some of the tests from the hb-repacker test suite.

closes #303.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -1,5 +1,6 @@
 //! A graph for resolving table offsets
 
+use crate::{table_type::TableType, tables::layout::LookupType};
 use font_types::Uint24;
 
 use super::write::TableData;
@@ -286,6 +287,10 @@ impl Graph {
         out
     }
 
+    fn can_potentially_promote_subtables(&self) -> bool {
+        [TableType::GSUB, TableType::GPOS].contains(&self.objects[&self.root].type_)
+    }
+
     /// Attempt to pack the graph.
     ///
     /// This involves finding an order for objects such that all offsets are
@@ -306,6 +311,11 @@ impl Graph {
             return true;
         }
 
+        if self.can_potentially_promote_subtables() {
+            self.try_promoting_subtables();
+        }
+
+        log::info!("assigning spaces");
         self.assign_spaces_hb();
         self.sort_shortest_distance();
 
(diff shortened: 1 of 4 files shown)


font-codegen/src/record.rs
```diff
@@ -217,6 +217,14 @@ pub(crate) fn generate_compile_impl(
         }
     };
 
+    let maybe_table_type = attrs.tag.is_some().then(|| {
+        quote! {
+            fn type_(&self) -> TableType {
```
dfrg, line 222: Just a nit, but the `type_` symbol is pretty unfortunate considering how pervasive it is in the code. Maybe just call it `table_type`?

write-fonts/src/graph.rs
```diff
@@ -809,6 +820,210 @@ impl Graph {
...
+
+        impl LookupSize {
+            // I could impl Ord but then I need to impl PartialEq and it ends
+            // up being way more code
+            fn sort_key(&self) -> impl Ord {
+                let bytes_per_subtable = self.subtable_count as f64 / self.subgraph_size as f64;
+                // f64 isn't ord, so we turn it into an integer,
+                // then reverse, because we want bigger things first
+                std::cmp::Reverse((bytes_per_subtable * 1e9) as u64)
+            }
```
dfrg, line 934: Seems like HB falls back to comparing lookup list order when the bytes per subtable sizes are equal. I'm not sure how common this is and it doesn't affect correctness, but may be something to keep in mind if we're directly comparing output and find discrepancies.

write-fonts/src/graph.rs
```diff
@@ -809,6 +820,210 @@ impl Graph {
...
+    }
+
+    fn find_subgraph_size(
+        &self,
+        queue: &mut VecDeque<ObjectId>,
+        visited: &mut HashSet<ObjectId>,
+    ) -> usize {
+        let mut size = 0;
+        visited.clear();
+        while !queue.is_empty() {
```
dfrg, line 1013: +1 to using iteration over recursion here. We should do more of this.

dfrg (approved):
LGTM with a few comments.

I don't fully grok the metric for determining when to promote, but it matches HB and the graph and extension promotion code look solid.

write-fonts/src/graph.rs
cmyr: Good point. cc @garretrieger, is there any particular rationale for explicitly using lookup index as a tie-break? I mean here:

https://github.com/harfbuzz/harfbuzz/blob/4056315c868693c655373856a83603fdc1587aab/src/hb-repacker.hh#L56-L69

garretrieger: I don't think this isn't particularly important from a repacking perspective, but using lookup idx to break ties should guarantee a stable sort order. ie. you're not at the mercy of the chosen sorting algorithm to order tie breaks for you.

PR #401, opened 2023-05-11
Add table splitting for PairPos format1

cmyr:
Built on top of #400, this adds the ability to split overflowing PairPos format 1 lookups.

I wanted to implement splitting of at least one lookup subtable on top of the promotion work in order to be confident that the approach I'd chosen for promotion would be suitable for splitting as well. Looks like it is?

### DIFF
--- a/write-fonts/src/graph/splitting.rs
+++ b/write-fonts/src/graph/splitting.rs
@@ -1,4 +1,4 @@
-//! splitting layout subtables
+//! splitting layout (GPOS) subtables
 
 use std::collections::{HashMap, HashSet};
 
@@ -17,14 +17,17 @@ pub(super) fn split_pair_pos(graph: &mut Graph, lookup: ObjectId) {
     split_subtables(graph, lookup, split_pair_pos_subtable)
 }
 
-/// A common impl handling updating the lookuup with the new subtables
+/// A common impl handling updating the lookup with the new subtables
 fn split_subtables(
     graph: &mut Graph,
     lookup: ObjectId,
     split_fn: fn(&mut Graph, ObjectId) -> Option<Vec<ObjectId>>,
 ) {
     let data = graph.objects.remove(&lookup).unwrap();
-    debug_assert!(data.reparse::<rgpos::PositionLookup>().is_ok());
+    debug_assert!(
+        data.reparse::<rgpos::PositionLookup>().is_ok(),
+        "table splitting is only relevant for GPOS?"
+    );
     log::debug!("trying to split subtables in {lookup:?}, '{}'", data.type_);
 
     let mut new_subtables = HashMap::new();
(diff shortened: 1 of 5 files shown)


write-fonts/src/graph/splitting.rs
```diff
@@ -0,0 +1,402 @@
...
+}
+
+/// A common impl handling updating the lookuup with the new subtables
+fn split_subtables(
+    graph: &mut Graph,
+    lookup: ObjectId,
+    split_fn: fn(&mut Graph, ObjectId) -> Option<Vec<ObjectId>>,
+) {
+    let data = graph.objects.remove(&lookup).unwrap();
+    debug_assert!(data.reparse::<rgpos::PositionLookup>().is_ok());
```
dfrg, line 28: This function is called `split_subtables` and doesn’t seem to actually depend on it being a GPOS subtable. This feels like a bit of a footgun if/when we extend to GSUB.

write-fonts/src/graph/splitting.rs
```diff
@@ -0,0 +1,402 @@
...
+fn split_pair_pos_subtable(graph: &mut Graph, lookup: ObjectId) -> Option<Vec<ObjectId>> {
+    let data = &graph.objects[&lookup];
+    let format: u16 = data.read_at(0).unwrap();
+    match format {
+        1 => split_pair_pos_format_1(graph, lookup),
+        2 => {
+            log::info!("table splitting not yet implemented for PairPos format 2");
+            None
+        }
+        _ => unreachable!("we only write valid formats"),
```
dfrg, line 78: Are we certain that this won’t be encountered when subsetting an arbitrary font? We don’t want to panic in that case. 

Maybe we can filter out subtables with invalid formats and emit a warning.

write-fonts/src/graph/splitting.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    if split_points.is_empty() {
+        return None;
+    }
+    split_points.push(data.offsets.len() - 1);
+    // okay, now we have a list of split points.
+
+    let mut new_subtables = Vec::new();
+    let mut prev_idx = 0;
+    for idx in &split_points {
```
dfrg, line 135: Small style nit: just consume the vec here? I don’t think it’s used after this.

dfrg (approved):
Seems a bit tricky in places so I’ve traced through the logic a few times and this looks solid to me.

A few comments inline but nothing major.

write-fonts/src/graph/splitting.rs
cmyr: My understanding is that table splitting is only relevant for GPOS, where individual subtables can exceed 64k bytes. in GSUB the main problem is with having so many subtables that you get offset overflows, and that's handled by extension promotion.

That said I agree I should make that more explicit :)

PR #406, opened 2023-05-15
Make write-fonts Error Send

rsheeter:
fea-rs and fontmake-rs both expect this.

While we're at it, rig to release write-fonts.

Will need to update https://github.com/cmyr/fea-rs/pull/135 once published.


rsheeter:
Merging per IM discussion.

PR #407, opened 2023-05-16
Make it possible to write sparse tuple stores

rsheeter:
IIUC we currently require that all deltas are present which seems like it woulud make sparse points for iup difficult


anthrotype (approved):
LGTM

anthrotype (approved):
LGTM

PR #415, opened 2023-05-18
Use control box for glyf

rsheeter:
See https://github.com/googlefonts/fontmake-rs/issues/285#issuecomment-1552385007


dfrg (approved):
LGTM. It might also be reasonable to add a `control_box()` method to `Shape` in kurbo.

anthrotype (approved):

rsheeter:
I wasn't sure if wanting control box was a common need, would be happy to have it ofc.

PR #419, opened 2023-05-22
Use u64 to store ObjectId

### LINKED ISSUE #398: Use a u64 for ObjectId
Because maybe once day this does get used in a long-running process or something?

              > QQ: `ObjectId`s are allocated with a global `usize` counter. Should we change this to an explicit `u64` (or allocate ids locally in the graph) to avoid potential overflows if someone does decide to run this on a 32-bit system?

Hmm... I would expect the process to only be running for a specific compilation, which even in the case of a very large multi-access font it's hard to imagine there being 4.2e9 unique objects? And in that case I would definitely expect us to be on a 64-bit system. That said, it wouldn't *hurt*...

_Originally posted by @cmyr in https://github.com/googlefonts/fontations/issues/388#issuecomment-1542515221_

cmyr:
- fixes #398

(JMM)

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -5,16 +5,17 @@ use font_types::Uint24;
 use super::write::TableData;
 use std::{
     collections::{BinaryHeap, HashMap, HashSet, VecDeque},
-    sync::atomic::AtomicUsize,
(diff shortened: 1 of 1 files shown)


dfrg (approved):
Looks good, thanks!

PR #421, opened 2023-05-22
Forbid unsafe in font-types

rsheeter:
JMM


cmyr (approved):
good idea to make this policy.

PR #427, opened 2023-05-23
Cleanup cmap test assertions

cmyr:
I'm sorry Rod, but this patch is going to make you mad.

Noticed this when reviewing old PRs; I thought it shouldn't be necessary to make everything a vec, since there should be PartialEq impls for comparing BigEndian types to their native values, and that should mean you can then compare slices directly, since that's provided by std.

And this works! With one small, infuriating caveat:

*We can impl PartialEq<T> for BigEndian<T>, but not the other way around*.

What this means is that you can write, `be_val == ne_val` but not `ne_val == be_val`. This is related to the orphan rules (https://doc.rust-lang.org/error_codes/E0210.html) and the tl;dr there is that we can't impl a foreign trait (PartialEq) on a foreign type (T). So we are stuck with this weird asymetry.

That said, I thought it would be useful to PR this patch so that other folks are aware of this fact, and can avoid confusion in the future.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -131,7 +131,7 @@ impl Cmap {
 
 #[cfg(test)]
 mod tests {
-    use font_types::{BigEndian, GlyphId, Scalar};
+    use font_types::GlyphId;
     use read::{
         tables::cmap::{Cmap, CmapSubtable, PlatformId},
         FontData, FontRead,
@@ -141,10 +141,6 @@ mod tests {
         dump_table,
         tables::cmap::{self as write, UNICODE_BMP_ENCODING, WINDOWS_BMP_ENCODING},
     };
-
-    fn to_vec<T: Scalar>(bees: &[BigEndian<T>]) -> Vec<T> {
-        bees.iter().map(|be| be.get()).collect()
-    }
 
     fn assert_generates_simple_cmap(mappings: Vec<(char, GlyphId)>) {
         let cmap = write::Cmap::from_mappings(mappings);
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/cmap.rs
```diff
@@ -181,20 +177,11 @@ mod tests {
...
-                to_vec(cmap4.end_code())
-            );
+            assert_eq!(cmap4.start_code(), &[10u16, 30u16, 153u16, 0xffffu16]);
+            assert_eq!(cmap4.end_code(), &[20u16, 90u16, 480u16, 0xffffu16]);
             // The example starts at gid 1, we're starting at 0
-            assert_eq!(vec![-10i16, -19i16, -81i16, 1i16], to_vec(cmap4.id_delta()));
-            assert_eq!(
-                vec![0u16, 0u16, 0u16, 0u16],
-                to_vec(cmap4.id_range_offsets())
-            );
```
rsheeter, line 197: I'm not sure I follow why you need to "fix" this, wouldn't the test be completely fine as it was before still?

cmyr: The old version is functionally equivalent, just unnecessarily verbose.

rsheeter (approved):
LGTM

write-fonts/src/tables/cmap.rs
cmyr: yes, that's the annoyance. If you prefer having that ordering over having less code then by all means you should keep it as-is and we can close this 🤷

rsheeter: My personal inclination would be to merge partialeq but not update the test. It's also fine to just update the test :)

cmyr: What do you mean 'merge PartialEq'?

rsheeter: Basically revert write-fonts/src/tables/cmap.rs. I think I misread this PR as _adding_ the partialeq as opposed to just a comment.

The order of asserts is not so important so I'm also entirely fine if you just merge as-is

PR #428, opened 2023-05-23
Replace FontWrite 'name' method with 'table_type' method

cmyr:
*this was originally part of #400, but I'm splitting it out here*

So as part of the table splitting and promotion work I needed some way to track the original source type of compiled bytes in order to possibly modify them afterwards. This ended up overlapping significantly with the `name` method on `FontWrite`, and I think it makes sense to combine them; in the general case tables can store their name directly in the result as a `&'static str` (like in the name method) and in the special cases we can manually provide a reasonable name.

So this does that; it replaces the `name` method with a `table_type` method, and then adds the basics of this new `TableType` enum; this will be used to handle promotion in #400.

### DIFF
--- a/font-codegen/src/record.rs
+++ b/font-codegen/src/record.rs
@@ -217,6 +217,14 @@ pub(crate) fn generate_compile_impl(
         }
     };
 
+    let maybe_table_type = attrs.tag.is_some().then(|| {
+        quote! {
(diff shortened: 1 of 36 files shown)


dfrg (approved):
Very nice to see the name and type properties unified and thanks for limiting the use of `type_` to a single field. LGTM!

PR #431, opened 2023-05-24
Use #[compile] to set avar version

### LINKED ISSUE #373: Use #[compile] to set the table version in avar
So I'm way late on this, but the pattern we use elsewhere is to explicitly compute the version, using the `#[compile]` annotation. As written, it would be easy to accidentally write out an `avar` table with the version (0, 0), which is invalid.

_Originally posted by @cmyr in https://github.com/googlefonts/fontations/pull/304#discussion_r1180530097_

cmyr:
closes #373.

This was a minor oversight when this table was first added; making this change means we don't need to manually set the version for this table, since it is always 1.0.

Unfortunately this is a breaking change, since it removes the 'version' field from the public API.

### DIFF
--- a/resources/codegen_inputs/avar.rs
+++ b/resources/codegen_inputs/avar.rs
@@ -5,6 +5,7 @@
 table Avar {
     /// Major version number of the axis variations table — set to 1.
     /// Minor version number of the axis variations table — set to 0.
+    #[compile(MajorMinor::VERSION_1_0)]
     version: MajorMinor,
     /// Permanently reserved; set to zero.
     #[skip_getter]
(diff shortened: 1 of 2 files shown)


rsheeter:
If this breaks fontmake-rs please release with a minor version bump and update there too; it's a nasty surprise when you want to use a git dep on fontations for something only to find out something unrelated broke your compile.

rsheeter (approved):
lgtm

cmyr:
agree, that is annoying. It would be nice if we could come up with some process such that we don't need to cut a new release on every single breaking commit, though I'm not sure what that looks like?

PR #438, opened 2023-05-26
Add types to represent typed arrays of offsets

cmyr:
This was a long-standing papercut: in codegen we generate typed getters for offsets, but for arrays of offsets this was limited to generating an iterator.

This was a slightly annoying API, since an iterator cannot be retained, and does not provide random access.

This patch addresses this by adding a pair of concrete types that can wrap arrays of offsets and nullable offsets, and can dynamically resolve those offsets as they are accessed.

I've had this branch written for months and sitting around, so we might as well get it in now.

This is a breaking change.

### DIFF
--- a/read-fonts/src/offset_array.rs
+++ b/read-fonts/src/offset_array.rs
@@ -1,0 +1,148 @@
+//! Arrays of offsets with dynamic resolution
+//!
+//! This module provides a number of types that wrap arrays of offsets, dynamically
+//! resolving individual offsets as they are accessed.
+
+use crate::offset::ResolveNullableOffset;
+use font_types::{BigEndian, Nullable, Offset16, Scalar};
+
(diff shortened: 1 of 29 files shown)


read-fonts/src/offset_array.rs
```diff
@@ -0,0 +1,136 @@
...
+//!
+//! This module provides a number of types that wrap arrays of offsets, dynamically
+//! resolving individual offsets as they are accessed.
+
+use crate::offset::ResolveNullableOffset;
+use font_types::{BigEndian, Nullable, Offset16, Scalar};
+
+use crate::{FontData, FontReadWithArgs, Offset, ReadArgs, ReadError, ResolveOffset};
+
+/// An array of offsets that can be resolved on access.
```
rsheeter, line 11: I suggest explaining here what this types job is, why we can't just give you the slice of offsets directly. To my understanding the answer is we could, but then the caller would have to pass in data and read args to resolve items which would be ugly.

Issue #439, opened 2023-05-29
[read-fonts] Overflow in `Transform::compute_flags`


behdad:
This line can (and does for me) overflow:
```
        } else if (self.xx - self.yy) != F2Dot14::ZERO { 
```

Just check for inequality instead?

behdad:
Fixing that I can build `NotoSans` and `NotoSans-Arabic` from designspace.

PR #442, opened 2023-05-30
[read-fonts/glyf] Avoid overflow in Transform::compute_flags()

behdad:
Fixes https://github.com/googlefonts/fontations/issues/439


rsheeter (approved):
LGTM!

PR #443, opened 2023-05-30
[skrifa] add variation API

dfrg:
Adds `AxisCollection` and `NamedInstanceCollection` types (along with their associated elements) as a high level API for accessing variation metadata.

Also adds an owned version of the `LocationRef` type called `Location` for storing computed normalized coordinates.

Both `AxisCollection` and `NamedInstance` have a `location` method converting variation settings into normalized coordinates.

In addition, there's a new `SmallArray` utility type which is really just a limited version of the type available in the `smallvec` crate. This is used as the backing store for `Location` and for temporary storage for filtering variation settings.


skrifa/src/variation.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+}
+
+/// Collection of axes in a variable font.
+///
+/// See the [`Axis`] type for more detail.
+#[derive(Clone)]
+pub struct AxisCollection<'a> {
+    fvar: Option<Fvar<'a>>,
+    avar: Option<Avar<'a>>,
```
drott, line 83: Do we need that / does avar need to be public in `AxisCollection` as client API?

skrifa/src/util.rs
```diff
@@ -0,0 +1,151 @@
+//! Internal helpers.
```
drott, line 1: I think this is a useful addition. I'd suggest to rename this to smallarray.rs to avoid generic names such a `util` or `helper` and land this in a separate CL.

skrifa/src/util.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use core::fmt;
+
+/// SmallVec like implementation but for fixed sized arrays whose
+/// size is only known at runtime.
+///
+/// Used to avoid allocations when possible for small or temporary
+/// arrays.
+///
+/// Consider taking on a smallvec dependency in the future.
```
drott, line 11: FWIW, just checked at least it's not approved and included as a third_party dependency in Chromium yet. Would be easier if it was. https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/?q=third_party%2Frust%2F%20

skrifa/src/variation.rs
```diff
@@ -0,0 +1,539 @@
...
+    ///
+    /// ```rust
+    /// # use skrifa::prelude::*;
+    /// # fn wrapper(font: &FontRef) {
+    /// let axes = font.axes();
+    /// let mut location = vec![NormalizedCoord::default(); axes.len()];
+    /// axes.location_to_slice(&[("wght", 250.0), ("wdth", 75.0)], &mut location);
+    /// # }
+    /// ```
+    pub fn location_to_slice<I>(&self, settings: I, location: &mut [NormalizedCoord])
```
drott, line 184: Not sure about Rust conventions, mainly curious: Are there idiomatic function definition styles related to this that inform about a success outcome instead of panicking? Can the reference parameter / output sliced be somehow combined with a `Result<>` type? - but I don't have a strong opinion on that vs. panicking yet.

skrifa/src/variation.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// let axes = font.axes();
+    /// let filtered: Vec<_> = axes
+    ///     .filter(&[("wght", 400.0), ("opsz", 100.0), ("wght", 1200.0)])
+    ///     .collect();
+    /// // The first "wght" and "opsz" settings are dropped and the final
+    /// // "wght" axis is clamped to the maximum value of 900.
+    /// assert_eq!(&filtered, &[("wght", 900.0).into()]);
+    /// # }
+    /// ```
+    pub fn filter<I>(&self, settings: I) -> impl Iterator<Item = VariationSetting> + Clone
```
drott, line 243: This and the previous function are both highly useful for the Skia context. Thanks for adding these.

drott (commented):
Big thanks for adding this so quickly, some comment below.

skrifa/src/variation.rs
dfrg: I’m not sure I understand the question. It’s an implementation detail here for computing the final location in normalized space. My intention is that users shouldn’t have to deal with axis remapping explicitly at this level.

skrifa/src/util.rs
dfrg: What about `util::smallarray`? It’s really common in rust to shove these small internal implementation details into a generic module like this to avoid polluting the crate root. For me personally, it makes browsing the code easier because I can distinguish between “real” code and ancillary types at a glance.

Happy to move this to a separate PR.

skrifa/src/util.rs
```diff
@@ -0,0 +1,151 @@
...
+#[derive(Clone)]
```
cmyr, line 12: Might also be worth deriving `Default` and `PartialEq`/`Eq`?

skrifa/src/util.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl<T, const N: usize> fmt::Debug for SmallArray<T, N>
+where
+    T: Copy + fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_list().entries(self.as_slice().iter()).finish()
+    }
+}
```
cmyr, line 67: The only other traits I'd consider implementing here are `AsRef`/`AsMut`, and maybe `Deref`/`DerefMut` also for completeness we could consider `Borrow`/`BorrowMut`, although those are not used that often and we can probably wait until we actually need them.


Also: If we implement `Deref` we might not need or own iter impls, if we can just inherit the ones on the slice type?

cmyr (approved):
Some nits and notes but overall this looks great!

skrifa/src/util.rs
dfrg: Good call on the comparison traits. Default is effectively meaningless however because it will always construct a zero length array and, unlike SmallVec, the length can’t be changed after construction.

skrifa/src/util.rs
dfrg: Since this is an internal type, basically a stopgap until we can take on smallvec, I didn’t want to flesh it out with more code than is necessary to implement our required behavior. Less code = fewer bugs?

skrifa/src/variation.rs
```diff
@@ -0,0 +1,539 @@
...
+
+    /// Returns the maximum value of the axis.
+    pub fn max_value(&self) -> f32 {
+        self.record.max_value().to_f64() as _
+    }
+
+    /// Returns a normalized coordinate for the given user coordinate. The value will be
+    /// clamped to the range specified by the minimum and maximum values.
+    ///
+    /// This does not apply any axis variation remapping.
```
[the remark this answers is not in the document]
dfrg: Thanks for keeping me in check on these :)

skrifa/src/util.rs
cmyr: If it's internal, can more things be `pub(crate)` instead of `pub`? I would prefer to minimize the public API as much as possible to avoid situations where we unintentionally make breaking changes.

cmyr:
I'm sorry my suggestion broke rustfmt 😢

skrifa/src/variation.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// # use skrifa::prelude::*;
+    /// # fn wrapper(font: &FontRef) {
+    /// let location = font.axes().location(&[("wght", 250.0), ("wdth", 75.0)]);
+    /// # }
+    /// ```
+    pub fn location<I>(&self, settings: I) -> Location
+    where
+        I: IntoIterator,
+        I::IntoIter: 'a + Clone,
+        I::Item: Into<VariationSetting>,
```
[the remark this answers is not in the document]
dfrg: Yes, this is the same signature as the one in `ScalerBuilder` (which I'll probably just remove now that we have this method). I'm holding firm on this style because I think there is real value in being able to express variations in a low friction way. 

If you look at the tests (and doc comments) in this PR, you'll see the settings specified as something like
```rust
[("wght", 100.0), ("wdth", 75.0)]
```
 which I believe is simple, clear, and easy to understand for anyone familiar with the CSS `font-variation-settings` property. 

The alternative is
```rust
[VariationSetting::new(Tag::new(b"wght"), 100.0), VariationSetting::new(Tag::new(b"wdth"), 75.0)]
```
which is incredibly noisy and difficult to read. It is still useful to have the `VariationSetting` type though both as a place to attach documentation, and as a target for `Into`.

I did however notice that the `I::IntoIter: 'a + Clone` bound is unnecessary here so I'll remove that and it should make the signature more approachable. At least, I think `I: IntoIterator, I::Item: Into<T>` is a common enough pattern in Rust that it should be clear.

skrifa/src/util.rs
dfrg: Moved to `small_array` mod and merged in #445

skrifa/src/util.rs
dfrg: Done in #445! I also went ahead and added `Deref`/`DerefMut` based on your suggestion.

skrifa/src/variation.rs
drott: So, `AxisCollection` is public, but the fields are not? Is that right?

skrifa/src/variation.rs
drott: Thanks for the explanation, sgtm to leave as is.

drott (approved):
LGTM, thanks for splitting out `SmallArray`.

skrifa/src/variation.rs
dfrg: Correct! In Rust, all fields are private unless otherwise specified.

skrifa/src/variation.rs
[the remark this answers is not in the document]
cmyr: yea the other alternative that jumps out at me is ignoring the iterator bit and just expecting it to be a slice? but just thinking out loud, if this works then great. :)

skrifa/src/variation.rs
rsheeter: I think skrifa should try very very hard to not panic

rsheeter: (and therefore so should read-fonts and font-types)

PR #444, opened 2023-05-31
[write-fonts] Fix writing of PackedPointNumbers

behdad:
Fixes https://github.com/googlefonts/fontc/issues/313

### DIFF
--- a/write-fonts/src/tables/variations.rs
+++ b/write-fonts/src/tables/variations.rs
@@ -231,7 +231,7 @@ impl FontWrite for PackedPointNumbers {
         // compute the actual count:
         match self.as_slice().len() {
             len @ 0..=127 => (len as u8).write_into(writer),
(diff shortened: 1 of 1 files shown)


rsheeter:
LGTM but can we add a test?

rsheeter (approved):

behdad:
> LGTM but can we add a test?

That might be above my paygrade currently :D. I'll take a look.

behdad:
> > LGTM but can we add a test?
> 
> That might be above my paygrade currently :D. I'll take a look.

Done.

behdad:
The read-fonts counterpart to this fix is here btw:
https://github.com/googlefonts/fontations/blob/e6f31f06d7c40cde5e9059aa3dd7769875adb41e/read-fonts/src/tables/variations.rs#L253

PR #445, opened 2023-05-31
[skrifa] add SmallArray type

dfrg:
Pulls out the `SmallArray` type from https://github.com/googlefonts/fontations/pull/443 with suggested improvements applied.

If this looks good, JMM!


rsheeter (approved):
LGTM but ... are we sure we can't use smallvec?

PR #449, opened 2023-06-01
CFF: add parsing for INDEX data

dfrg:
Adds support for reading data from CFF INDEX arrays.

### DIFF
--- a/font-types/src/uint24.rs
+++ b/font-types/src/uint24.rs
@@ -26,6 +26,11 @@ impl Uint24 {
         }
     }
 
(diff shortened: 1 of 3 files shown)


cmyr (approved):
This all looks reasonable. One note: I think in codegen you can actually indicate that a field is `FontData` instead of `[u8]` which might be a slight ergonomics improvement?

dfrg:
TIL! Thanks, I’ll keep that in mind as I move forward and swap it out in cases where it’s helpful.

PR #457, opened 2023-06-01
Add lifetime to T for generated ArrayOfOffsets

dfrg:
We currently generate return types like:
```rust
ArrayOfNullableOffsets<'a, ItemVariationData, Offset32>
```
where `ItemVariationData`  may need an explicit `'a` lifetime to be useful. This changes codegen to add those lifetimes when the target is not generic. @cmyr  please double check me on this.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -816,12 +816,15 @@ impl Field {
                 quote!(ArrayOfOffsets)
             };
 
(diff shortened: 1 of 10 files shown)


cmyr (approved):
Oops yes this makes sense, I think I skipped this for the arrays portion because it didn't seem to be used. 😅

PR #458, opened 2023-06-02
[read-fonts] add infra for CFF2 blending

dfrg:
Adds a new `BlendState` type that handles computation of variation region scalars to support the DICT/charstring blend operator.


font-test-data/src/lib.rs
```diff
@@ -59,3 +59,26 @@ pub mod post {
         0x4, 0x68, 0x6f, 0x6c, 0x61, // 4, h o l a
     ];
 }
+
+pub mod cff2 {
+    /// CFF2 example font
```
rsheeter, line 64: it's not a *font*, but a cff2 table right?

read-fonts/src/tables/postscript.rs
```diff
@@ -14,7 +16,9 @@ pub enum Error {
     /// The `off_size` field in an INDEX contained an invalid value.
     InvalidIndexOffsetSize(u8),
     /// An INDEX contained a zero offset.
-    ZeroOffset,
+    ZeroOffsetInIndex,
+    /// Variation store index referenced an invalid variation region.
+    InvalidVsIndex(u16),
     /// Underlying parsing error.
```
rsheeter, line 22: The comments on the variants here seem a bit redundant with the names?

read-fonts/src/tables/postscript.rs
```diff
@@ -14,7 +16,9 @@ pub enum Error {
     /// The `off_size` field in an INDEX contained an invalid value.
     InvalidIndexOffsetSize(u8),
     /// An INDEX contained a zero offset.
-    ZeroOffset,
+    ZeroOffsetInIndex,
+    /// Variation store index referenced an invalid variation region.
+    InvalidVsIndex(u16),
```
rsheeter, line 21: Suggest InvalidVariationStoreIndex in full; Vs is a bit obtuse.

read-fonts/src/tables/postscript/blend.rs
```diff
@@ -0,0 +1,166 @@
+//! Support for the dictionary and charstring blend operator.
+
+use font_types::{BigEndian, F2Dot14, Fixed};
+
+use super::Error;
+use crate::tables::variations::{ItemVariationData, ItemVariationStore, VariationRegionList};
+
+/// The maximum number of region scalars that we precompute.
+const MAX_PRECOMPUTED_SCALARS: usize = 16;
```
rsheeter, line 9: Why 16?

read-fonts/src/tables/postscript/blend.rs
```diff
@@ -0,0 +1,166 @@
...
+/// precomputed scalars are stored internally and the overflow is computed
+/// as needed.
+///
+/// The `MAX_PRECOMPUTED_SCALARS` constant determines the size of the
+/// internal buffer (currently 16).
+pub struct BlendState<'a> {
+    store: ItemVariationStore<'a>,
+    regions: VariationRegionList<'a>,
+    coords: &'a [F2Dot14],
+    vs_index: u16,
```
rsheeter, line 25: store_index might make the relationship I think I'm seeing clearer?

read-fonts/src/tables/postscript/blend.rs
```diff
@@ -0,0 +1,166 @@
...
+///
+/// To avoid allocation, scalars are computed on demand but this can be
+/// prohibitively expensive in charstrings where blends are applied to
+/// large numbers of elements. To amortize the cost, a fixed number of
+/// precomputed scalars are stored internally and the overflow is computed
+/// as needed.
+///
+/// The `MAX_PRECOMPUTED_SCALARS` constant determines the size of the
+/// internal buffer (currently 16).
+pub struct BlendState<'a> {
```
rsheeter, line 21: Link to both spec and any implementation(s) we referenced?

read-fonts/src/tables/postscript/blend.rs
```diff
@@ -0,0 +1,166 @@
...
+    store: ItemVariationStore<'a>,
+    regions: VariationRegionList<'a>,
```
rsheeter, line 23: Why is it necessary to store regions? - access should be fast and you only use it to call self.regions.variation_regions.

Perhaps we should not store it (preferred) or store the final result of self.regions.variation_regions?

read-fonts/src/tables/postscript/blend.rs
```diff
@@ -0,0 +1,166 @@
...
+    coords: &'a [F2Dot14],
+    vs_index: u16,
+    needs_update: bool,
```
rsheeter, line 26: Not a huge fan of this. Needs a comment to explain why it's a good idea.

I don't know the usage pattern but I'd prefer we avoided it unless that is somehow infeasible. Could we:

1. Have two types - or an enum with two variants - one for the unupdated/stale state and one for the updated state. Provide methods to move between them.
   * Makes sense if it's desirable to change vs_index repeatedly and only call update once. If you more or less call update_if_needed every change seems silly.
1. Have one type that is always updated
   * Makes sense if the usage pattern is to essentially always call update_if_needed when you set a vs index

rsheeter (commented):
Couple of nits plus I'd really like to see needs_update go away, or to understand better why it can't. At a glance it feels like a code smell.

font-test-data/src/lib.rs
dfrg: Yes, it's a table. The spec referred to it as a font so I mindlessly copied it. Fixed.

read-fonts/src/tables/postscript.rs
dfrg: Done

read-fonts/src/tables/postscript/blend.rs
dfrg: I tried to keep the names matching the spec, but store_index works for me. Changed.

read-fonts/src/tables/postscript/blend.rs
dfrg: Added a link to CFF2 blending section in the spec. This isn't based on any other implementations.

read-fonts/src/tables/postscript/blend.rs
dfrg: Changed to read regions from IVS when needed.

read-fonts/src/tables/postscript/blend.rs
dfrg: Fair enough. I changed this to compute scalars on construction and update them directly when the store index changes.

rsheeter (approved):
LGTM provided comments are addressed

read-fonts/src/tables/postscript/blend.rs
dfrg: Completely made up number chosen to balance size with trying to capture as many precomputed regions as possible. I'll measure when I have a larger set of CFF2 fonts to test against and adjust accordingly.

rsheeter: ^ would make a lovely comment :)

dfrg: will do :)

PR #459, opened 2023-06-02
Don't panic!

### LINKED ISSUE #456: Don't panic in skrifa
And transitively in read-fonts and font-types.

We will host it in things that may handle panic by dying and we don't want that.

Context: https://github.com/googlefonts/fontations/pull/443#discussion_r1211342698

dfrg:
Fixes #456

### DIFF
--- a/skrifa/src/variation.rs
+++ b/skrifa/src/variation.rs
@@ -165,11 +165,9 @@ impl<'a> AxisCollection<'a> {
     ///     before normalization.
     /// * If more than one setting for an axis is provided, the last one is
     ///     used.
-    /// * Omitted settings are set to 0.0, representing the default position
+    /// * If no setting for an axis is provided, the associated coordinate is
+    ///     set to the normalized value 0.0, representing the default position
     ///     in variation space.
-    ///
-    /// If `location.len() < self.len()`, the omitted coordinates will assume
-    /// their default positions.
     ///
     /// # Examples
     ///
@@ -333,9 +331,6 @@ impl<'a> NamedInstance<'a> {
 
     /// Computes a location in normalized variation space for this instance and
     /// stores the result in the given slice.
-    ///
-    /// If `location.len() < self.user_coords().count()`, the omitted
-    /// coordinates will assume their default positions.
     ///
     /// # Examples
     ///
--- a/skrifa/src/variation.rs
+++ b/skrifa/src/variation.rs
@@ -168,9 +168,8 @@ impl<'a> AxisCollection<'a> {
     /// * Omitted settings are set to 0.0, representing the default position
     ///     in variation space.
     ///
-    /// # Panics
-    ///
-    /// Panics if `location.len() < self.len()`.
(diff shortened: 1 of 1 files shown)


skrifa/src/variation.rs
```diff
@@ -168,9 +168,8 @@ impl<'a> AxisCollection<'a> {
     /// * Omitted settings are set to 0.0, representing the default position
     ///     in variation space.
     ///
-    /// # Panics
-    ///
-    /// Panics if `location.len() < self.len()`.
+    /// If `location.len() < self.len()`, the omitted coordinates will assume
+    /// their default positions.
```
rsheeter, line 172: Isn't this comment duplicative of the "Omitted settings are set to ..." part?

skrifa/src/variation.rs
```diff
@@ -168,9 +168,8 @@ impl<'a> AxisCollection<'a> {
     /// * Omitted settings are set to 0.0, representing the default position
     ///     in variation space.
```
rsheeter, line 169: I suggest you might explicitly specify set to the _normalized_ value 0.0, which is equivalent to the user value at the default in fvar; if you start from user coordinates and read to here I think it would be easy to miss that you said they'd be normalized. Also precious few people grok font coordinates and as much as possible we'd like this all to be somewhat noob accessible.

skrifa/src/variation.rs
rsheeter: Hm. Also it's wrong because location.len could be >= self.len while still not specifying positions on all axes so ... just delete and rely on the Ommitted bit?

skrifa/src/variation.rs
rsheeter: Also, optional, perhaps change "Omitted settings" to "If no setting for an axis is provided" to match teh language used for the prior bullet?

skrifa/src/variation.rs
```diff
@@ -333,9 +334,8 @@ impl<'a> NamedInstance<'a> {
     /// Computes a location in normalized variation space for this instance and
     /// stores the result in the given slice.
     ///
-    /// # Panics
-    ///
-    /// Panics if `location.len() < self.user_coords().count()`.
+    /// If `location.len() < self.user_coords().count()`, the omitted
+    /// coordinates will assume their default positions.
```
rsheeter, line 338: I like the If not setting is provided... language over the len comparison for reasons noted in the comment above.

rsheeter (approved):
LGTM w/nits

skrifa/src/variation.rs
dfrg: Thanks, I like this working better. Changed to:
```
If no setting for an axis is provided, the associated coordinate is
set to the normalized value 0.0, representing the default position
in variation space.
```

skrifa/src/variation.rs
dfrg: Just deleted this as per suggestion above. In this case, all settings should be provided because they come from the named instance.

PR #461, opened 2023-06-02
[read-fonts] add CFF operand stack

dfrg:
Adds a new `Stack` type that implements the operand stack for storing and processing intermediate values while parsing CFF DICTs and charstrings.

### DIFF
--- a/read-fonts/src/tables/postscript/stack.rs
+++ b/read-fonts/src/tables/postscript/stack.rs
@@ -84,12 +84,11 @@ impl Stack {
         self.value_is_fixed[..self.top].reverse();
     }
 
-    pub fn push_i32(&mut self, value: i32) -> Result<(), Error> {
-        self.push(value, false)
-    }
-
-    pub fn push_fixed(&mut self, value: Fixed) -> Result<(), Error> {
-        self.push(value.to_bits(), true)
+    pub fn push(&mut self, number: impl Into<Number>) -> Result<(), Error> {
+        match number.into() {
+            Number::I32(value) => self.push_impl(value, false),
+            Number::Fixed(value) => self.push_impl(value.to_bits(), true),
+        }
     }
 
     /// Returns the 32-bit integer at the given index on the stack.
@@ -160,6 +159,17 @@ impl Stack {
                     Fixed::from_i32(*value)
                 }
             })
+    }
+
+    /// Returns an iterator yielding all elements on the stack as number
+    /// values.
+    ///
+    /// This is useful for capturing the current state of the stack.
+    pub fn number_values(&self) -> impl Iterator<Item = Number> + '_ {
+        self.values[..self.top]
+            .iter()
+            .zip(&self.value_is_fixed)
+            .map(|(value, is_real)| Number::from_stack(*value, *is_real))
     }
 
     /// Apply a prefix sum to decode delta-encoded numbers.
@@ -260,7 +270,7 @@ impl Stack {
         Ok(())
     }
 
-    fn push(&mut self, value: i32, is_fixed: bool) -> Result<(), Error> {
+    fn push_impl(&mut self, value: i32, is_fixed: bool) -> Result<(), Error> {
         if self.top == MAX_STACK {
             return Err(Error::StackOverflow);
         }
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/postscript/stack.rs
```diff
@@ -0,0 +1,381 @@
...
+    pub fn get_fixed(&self, index: usize) -> Result<Fixed, Error> {
+        let value = *self
+            .values
+            .get(index)
+            .ok_or(Error::InvalidStackAccess(index))?;
+        Ok(if self.value_is_fixed[index] {
+            Fixed::from_bits(value)
+        } else {
+            Fixed::from_i32(value)
+        })
```
cmyr, line 126: this is funny, what's with the asymmetry?

read-fonts/src/tables/postscript/stack.rs
```diff
@@ -0,0 +1,381 @@
...
+        self.value_is_fixed[..self.top].reverse();
+    }
+
+    pub fn push_i32(&mut self, value: i32) -> Result<(), Error> {
+        self.push(value, false)
+    }
+
+    pub fn push_fixed(&mut self, value: Fixed) -> Result<(), Error> {
+        self.push(value.to_bits(), true)
+    }
```
cmyr, line 93: One other API possibility here is to use an enum in the API, but not in the internal storage.

This would look like,

```rust
struct Operand {
    I32(i32),
    Fixed(Fixed),
}

// in our impl

pub fn push(&mut self, value: impl Into<Operand>) -> Result<(), Error> {
    match value.into() {
        Operand::I32(val) => self.push(val, false),
        Operand::Fixed(val) => self.push(val.to_bits(), false),
    }
}
```

again, just spitballing.

cmyr (approved):
Looks good! a few ignorable comments and one QQ inline.

read-fonts/src/tables/postscript/stack.rs
```diff
@@ -0,0 +1,381 @@
...
+/// is also known (based on the operator) when reading from the stack, so the
+/// conversion is performed on demand at read time.
+///
+/// Storing the entries as an enum would require 8 bytes each and since these
+/// objects are created on the _stack_, we reduce the required size by storing
+/// the entries in parallel arrays holding the raw 32-bit value and a flag that
+/// tracks which values are fixed point.
+pub struct Stack {
+    values: [i32; MAX_STACK],
+    value_is_fixed: [bool; MAX_STACK],
```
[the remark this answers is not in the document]
dfrg: I had considered packing the flags into bits but figured the extra complexity wasn’t worth it. Your comment on that one rings true to me :)

I did not think of `repr(packed)` but that effectively gives us the same memory requirements but with potentially worse performance.

read-fonts/src/tables/postscript/stack.rs
dfrg: If you mean the fact that pop_int requires an integer but pop_fixed does a conversion, the only answer I can give is “that’s what FreeType does.” In some ways, it makes sense— when we’re asking for an integer from the stack, it’s almost always used as an offset or index in which case we generally want the full 32 bits.

read-fonts/src/tables/postscript/stack.rs
dfrg: Good call! I actually have a `Number` enum in code that hasn’t been pushed yet that looks just like this. I’ll add it to this PR and use your unified push method.

read-fonts/src/tables/postscript/stack.rs
```diff
@@ -0,0 +1,381 @@
...
+
+use super::{BlendState, Error};
+
+/// Maximum size of the operand stack.
+///
+/// "Operators in Top DICT, Font DICTs, Private DICTs and CharStrings may be
+/// preceded by up to a maximum of 513 operands."
+///
+/// <https://learn.microsoft.com/en-us/typography/opentype/spec/cff2#table-9-top-dict-operator-entries>
+const MAX_STACK: usize = 513;
```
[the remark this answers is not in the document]
dfrg: Your explanation makes more sense than anything I can come up with :)

PR #462, opened 2023-06-02
[font-types] Rework raw traits

cmyr:
As part of the safety work, I wanted to simplify our font-types traits where possible.

In particular, this removes the `ReadScalar` trait, which was intended to provide a unified way to read both `Scalar` types as well as their `BigEndian` wrappers; but the former case is not common, and the extra trait was confusing, so I have separated this functionality.

This means that we will need separate methods for when we want to actually read a BigEndian value, but this doesn't seem too bad.

In addition this adds a new trait to more narrowly encode the condition that the Raw associated type on the Scalar trait should only ever be a fixed-size byte array.

### DIFF
--- a/font-types/src/raw.rs
+++ b/font-types/src/raw.rs
@@ -6,14 +6,23 @@
 ///
 /// You do not need to implement this trait directly; it is an implemention
 /// detail of the [`BigEndian`] wrapper.
-pub trait Scalar {
+pub trait Scalar: Sized {
     /// The raw byte representation of this type.
-    type Raw: Copy + AsRef<[u8]>;
+    type Raw: sealed::BeByteArray;
 
     /// Create an instance of this type from raw big-endian bytes
     fn from_raw(raw: Self::Raw) -> Self;
+
     /// Encode this type as raw big-endian bytes
     fn to_raw(self) -> Self::Raw;
+
+    /// Attempt to read a scalar from a slice.
+    ///
+    /// This will always succeed if `slice.len() == Self::RAW_BYTE_LEN`, and will
+    /// always return `None` otherwise.
+    fn read(slice: &[u8]) -> Option<Self> {
+        sealed::BeByteArray::from_slice(slice).map(Self::from_raw)
+    }
 }
 
 /// A trait for types that have a known, constant size.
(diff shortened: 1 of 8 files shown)


font-types/src/raw.rs
```diff
@@ -57,7 +71,15 @@ impl<T: Scalar> BigEndian<T> {
     pub fn new(raw: T::Raw) -> BigEndian<T> {
         BigEndian(raw)
     }
-    /// Read a copy of this type from raw bytes.
+
+    /// Attempt to construct a new raw value from this slice.
+    ///
+    /// This will always fail if the slice is the correct length.
```
rsheeter, line 77: nit: always is not necessary.

Also ... shouldn't it say is *not* the correct length? E.g. Fails if the slice is not the correct length ?

font-types/src/raw.rs
```diff
@@ -32,18 +41,23 @@ pub trait FixedSize: Sized {
...
+/// we hide this trait; it isn't part of the public API, and this clarifies
+/// the guarantee that it is only implemented for [u8; N]
+mod sealed {
+    /// A trait representing any fixed-size big-endian byte array.
     ///
-    /// This should use all the provided bytes; bounds checking is performed upstream.
-    fn read(bytes: &[u8]) -> Option<Self>;
+    /// This is only used in `Scalar`, as a way of expressing the condition that the
+    /// `Raw` type is always a fixed-size byte array.
+    pub trait BeByteArray: Copy + AsRef<[u8]> {
```
rsheeter, line 51: pub on something that isn't visible is weird particularly if you are newer to Rust. Would pub(crate) be ok?

rsheeter (approved):

font-types/src/raw.rs
cmyr: This is a Funny Rust Thing. `BeByteArray` needs to be `pub`, because it appears in another public type (the `Scalar` trait). However, we don't want anyone else to ever be able to implement this trait, because that would compromise our safety guarantees.

By placing the public type in the private module, it is still technically public, but can never be named, and thus cannot be implemented.

I was looking for a source for this pattern and I can't find exactly what I was looking for; the closest I can find is the [sealed traits](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) pattern, which is similar but includes an extra trait.

font-types/src/raw.rs
cmyr: oh yes, good catch.

font-types/src/raw.rs
rsheeter: > By placing the public type in the private module, it is still technically public, but can never be named, and thus cannot be implemented.

This is very clever but perhaps not very obvious. Can you put a comment on it? - could have the exact words you used in your comment :)

PR #463, opened 2023-06-02
Tighten up safety guarantees

### LINKED ISSUE #228: Audit unsafe in read-fonts
* Enumerate them all and get reviews

### LINKED ISSUE #434: read_ref_at and read_array should be marked `unsafe` or guarantee soundness
This function comments explains why the functions are not memory safe to call for any input.

```rs
    //NOTE: this is definitely unsound, since FixedSized isn't private,
    // and we don't enforce all soundness requirements: for instance, you could
    // use this to create an enum with an invalid discriminant, which is UB.
    //
    // In practice I believe my *current* use is correct, as it is all in auto-generated
    // code, and I know the invariants, but this should be revisited.
    pub fn read_ref_at<T: FixedSize>(&self, offset: usize) -> Result<&'a T, ReadError> {
        assert_ne!(std::mem::size_of::<T>(), 0);
        assert_eq!(std::mem::align_of::<T>(), 1);
        self.bytes
            .get(offset..offset + T::RAW_BYTE_LEN)

        unsafe { Ok(self.read_ref_unchecked(offset)) }
    }
```
```rs
    //NOTE: unsound, see the note on read_ref_at
    pub fn read_array<T: FixedSize>(&self, range: Range<usize>) -> Result<&'a [T], ReadError> {
        assert_ne!(std::mem::size_of::<T>(), 0);
        assert_eq!(std::mem::align_of::<T>(), 1);
        let bytes = self
            .bytes
            .get(range.clone())
        if bytes.len() % std::mem::size_of::<T>() != 0 {
            return Err(ReadError::InvalidArrayLen);
        };
        unsafe { Ok(self.read_array_unchecked(range)) }
    }
```

### LINKED ISSUE #435: unsafe blocks should come with a SAFETY comment
It is best practice for:

- unsafe functions to describe their preconditions in a `# Safety` section
- unsafe blocks to describe how they meet the preconditions in a guaranteed way in a `// SAFETY:` comment. this also implies that the unsafe block is encapsulated or controlled enough that it can guarantee safety.

See also: https://chromium.googlesource.com/chromium/src/+/master/docs/security/rule-of-2.md#unsafe-code-in-safe-languages

### LINKED ISSUE #436: Soundness of read_ref_unchecked
[`read_ref_unchecked`](https://github.com/googlefonts/fontations/blob/ea58ee9f6262f85ce84ca4fad2289f2614066b97/read-fonts/src/font_data.rs#L128C15-L131) will create a reference into a memory buffer, changing the representation from &[u8] to &T.

The number of bytes used for this &[u8] is T::RAW_BYTE_LEN which is [defined to be](https://github.com/googlefonts/fontations/blob/ea58ee9f6262f85ce84ca4fad2289f2614066b97/font-types/src/raw.rs#L21C9-L32) the number of bytes required to represent the type in a font file, which may be smaller than the size of the native type when the native type has tail padding (for alignment purposes).

There are 2 possible problematic outcomes:

## Overlapping refs

Two consecutive T elements in the buffer may overlap in memory. Let's say T has `size_of` equal to 2, but T::RAW_BYTE_LEN is 1.

```
| T1 | T2 | T3 | ... |
```

- `read_ref_unchecked(0)` will return `&T` that points to `| T1 | T2 |`.
- `read_ref_unchecked(1)` will return `&T` that points to `| T2 | T3 |`.

If we were to `*read_ref_unchecked(0) = another_t`, we'd expect `size_of<T>` bytes to be memcpy'd which would clobber `T2`.

This is happily mitigated by the fact that there *is no mutable reference version* of this function. I think we should document why there is no mutable reference version somewhere in the code (rustdocs) so that one does not appear in the future.

## Invalid refs

The last T element in the buffer actually goes off the end of the buffer. Again, let's say T has `size_of` equal to 2, but T::RAW_BYTE_LEN is 1.

```
| T1 | T2 | ... | T9 |
```

A call to `read_ref_unchecked(8)` will return `&T` that points to `| T9 | invalid memory |`. If that `&T` is moved (which it can't be, see [Overlapping refs](#overlapping-refs)) or copied, we'd see two bytes of memory copied, which would involve a read out of bounds, which is a memory safety bug.

I _think_ this could be mitigated by requiring T to not be `Copy`, or by requiring `size_of<T>() - T::RAW_BYTE_LEN` bytes of padding to exist at the back of the buffer?

cmyr:
This is based on top of #462, but can be reviewed separately.

This addresses the remaining concerns around our use of unsafe.

- It introduces a new trait, `JustBytes`, which is our equivalent of `FromBytes` in zerocopy. This trait is marked unsafe, and is intended to only be implemented during codegen, as well as for `BigEndian<T>` and `u8`.
- It adds `# Safety` and `// SAFETY` documentation blocks and comments where required.

In practice, the most important part of this PR is that we no longer use the general-purpose `FixedSize` trait as the trait bound on methods that involve reinterpreting bytes; this fixes the unsoundness where someone could call one of those methods with a type that implemented that trait, but violated the safety invariants.

- closes #228 
- closes #435 
- references #434,  #436

### DIFF
--- a/font-codegen/src/record.rs
+++ b/font-codegen/src/record.rs
@@ -40,6 +40,10 @@ pub(crate) fn generate(item: &Record) -> syn::Result<TokenStream> {
             impl FixedSize for #name {
                 const RAW_BYTE_LEN: usize = #( #inner_types::RAW_BYTE_LEN )+*;
             }
+
+            unsafe impl JustBytes for #name {
+                fn this_trait_should_only_be_implemented_in_generated_code() {}
+            }
         }
     });
     let maybe_impl_read_with_args = (has_read_args).then(|| generate_read_with_args(item));
--- a/font-codegen/src/record.rs
+++ b/font-codegen/src/record.rs
@@ -41,7 +41,8 @@ pub(crate) fn generate(item: &Record) -> syn::Result<TokenStream> {
                 const RAW_BYTE_LEN: usize = #( #inner_types::RAW_BYTE_LEN )+*;
             }
 
-            unsafe impl JustBytes for #name {
+            /// SAFETY: see the [`FromBytes`] trait documentation.
+            unsafe impl FromBytes for #name {
                 fn this_trait_should_only_be_implemented_in_generated_code() {}
             }
         }
--- a/read-fonts/src/read.rs
+++ b/read-fonts/src/read.rs
@@ -1,6 +1,6 @@
 //! Traits for interpreting font data
 
-use types::{FixedSize, Scalar, Tag};
+use types::{BigEndian, FixedSize, Scalar, Tag};
 
 use crate::font_data::FontData;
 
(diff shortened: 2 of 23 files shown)


font-codegen/src/record.rs
```diff
@@ -40,6 +40,11 @@ pub(crate) fn generate(item: &Record) -> syn::Result<TokenStream> {
             impl FixedSize for #name {
                 const RAW_BYTE_LEN: usize = #( #inner_types::RAW_BYTE_LEN )+*;
             }
+
+            /// SAFETY: I am repr(c) + repr(packed), and all of my fields impl `JustBytes`.
```
rsheeter, line 44: Not all readers will be Rusty enough to understand the full implications of this so despite being a nuisance I suggest spelling it out in a little more detail what all this means. For example, I would not have known repr(packed) => alignment 1 => can be converted from bytes at any offset without looking up the documentation.

rsheeter: Also iiuc the codegen is what is actually ensuring the things asserted here. The trait is merely a marker we claim we will only put on things that are so structured. To me that's a bit different than what the comment says.

font-codegen/src/record.rs
```diff
@@ -40,6 +40,11 @@ pub(crate) fn generate(item: &Record) -> syn::Result<TokenStream> {
...
+            unsafe impl JustBytes for #name {
```
rsheeter, line 45: Optional, how do you feel about FromBytes or some other name that more clearly indicates it's safe to convert this from raw bytes?

rsheeter (approved):
Nice, ty. Few questions inline but overall LGTM.

read-fonts/src/read.rs
```diff
@@ -94,6 +94,44 @@ pub trait VarSize {
...
+/// A marker trait for types that can read from a big-endian buffer without copying.
+///
+/// WARNING: Do not implement this trait manually. Implementations are created
+/// where appropriate during code generation, and there should be no conditions
+/// under which this trait could be implemented, but cannot be implemented by
+/// codegen.
+///
+/// # Safety
+///
+/// If a type `T` implements `JustBytes` than unsafe code may assume that it is
```
dfrg, line 106:
```suggestion
/// If a type `T` implements `JustBytes` then unsafe code may assume that it is
```

font-codegen/src/record.rs
cmyr: This comment is really just about satisfying #435; I think a better solution is to centralize all the docs around how this works in the trait, and then I'll just link to the trait from the generated code.

PR #465, opened 2023-06-02
Add another test of Name sort order

cmyr:
Something funny is up in fontc, where it looks like NameRecords are being sorted incorrectly in some circumstances? See https://github.com/googlefonts/fontc/pull/321.

This test is probably redundant, but it should more clearly assert that the order of records in a vec does not influence the order they end up in in the name table.


dfrg (approved):
LGTM

PR #467, opened 2023-06-04
[read-fonts] CFF SID and standard strings

dfrg:
Adds the `SID` operand type, named `StringId` here along with the ability to resolve it to either an entry in the standard string set or an index into the string INDEX in the CFF table (confusing, eh?).

Also adds a `Latin1String` type because these things technically use that encoding even if they're almost always ASCII.

### DIFF
--- a/read-fonts/src/tables/postscript/string.rs
+++ b/read-fonts/src/tables/postscript/string.rs
@@ -1,0 +1,504 @@
+//! PostScript string identifiers.
+
+/// PostScript string identifier (SID).
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+pub struct StringId(u16);
+
+impl StringId {
+    pub fn new(raw: u16) -> Self {
+        Self(raw)
+    }
+
+    /// Resolves the identifier as a standard string.
+    ///
+    /// If the identifer represents a standard string, returns `Ok(string)`,
+    /// otherwise returns `Err(index)` with the index that should be used to
+    /// retrieve the string from the CFF string INDEX.
+    ///
+    /// The standard string set is available in the section
+    /// "Appendix A - Standard Strings" at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf>.
+    pub fn standard_string(self) -> Result<Latin1String<'static>, usize> {
+        let ix = self.0 as usize;
+        if let Some(string) = STANDARD_STRINGS.get(ix) {
+            // The standard strings are all ASCII so it's safe to intepret them
+            // as Latin-1. This is verified in a unit test.
+            Ok(Latin1String::new(string.as_bytes()))
+        } else {
+            Err(ix - STANDARD_STRINGS.len())
+        }
+    }
+}
+
+impl From<i32> for StringId {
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/postscript/string.rs
```diff
@@ -0,0 +1,504 @@
...
+
+    /// Resolves the identifier as a standard string.
+    ///
+    /// If the identifer represents a standard string, returns `Ok(string)`,
+    /// otherwise returns `Err(index)` with the index that should be used to
+    /// retrieve the string from the CFF string INDEX.
+    ///
+    /// The standard string set is available in the section
+    /// "Appendix A - Standard Strings" at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf>.
+    pub fn standard_string(self) -> Result<Latin1String<'static>, usize> {
```
cmyr, line 20: Curious about how this API ends up being used, it feels like it's mixing concerns a little bit? I would be tempted to split this up into separate methods, like

```rust
fn is_standard_string(&self) -> bool;
fn standard_string(&self) -> Option<Latin1String<'static>>;
fn index(&self) -> Option<usize>;
```

or something along these lines? And depending on what the types look like you could maybe move the whole "standard or index" functionality into a method on whatever type encapsulates the string storage?

read-fonts/src/tables/postscript/string.rs
```diff
@@ -0,0 +1,504 @@
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        for ch in self.chars() {
+            write!(f, "{}", ch)?;
+        }
+        Ok(())
+    }
+}
+
+/// See "Appendix A - Standard Strings" in <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf>
+const STANDARD_STRINGS: &[&str] = &[
```
cmyr, line 80: does this want to be `static` instead of `const`? Does using `const` possible duplicate the storage of the strings?


Also sort of an aside but it would be nice to have this declared in one place in the ecosystem, since we also need it during compilation.

read-fonts/src/tables/postscript/string.rs
dfrg: This was just for economy of expression. If you have one of these things, you either want the standard string or the index to look up the custom one.  In this case, `Result` seemed like a nice option to represent “either” similar in style to the `binary_search` methods 🤷🏻‍♂️

In most cases, the user will just pass this id to some `string()` method on the CFF table which will do all this internally.

read-fonts/src/tables/postscript/string.rs
dfrg: The str elements are really pointers to separate storage so LLVM will dedup them independently regardless of how this is declared.

Good call on making this available to write-fonts. I’ll make it pub and re-export it.

read-fonts/src/tables/postscript/string.rs
dfrg: Going to go ahead and merge this as-is because it is a good match for the current use cases. We can revisit if the API proves cumbersome for future uses.

PR #468, opened 2023-06-04
[read-fonts] CFF DICT parsing/decoding

dfrg:
This adds supporting types and two new iterator functions for parsing DICTs:

* `dict::raw_entries()` yields numeric operands and operators in the order in which they appear in the data stream (operands first). In the spirit of read-fonts, this is the low level parsing routine.

* `dict::entries()` (built on raw_entries) uses an operand stack and blend state to associate operands with their respective operators and apply any necessary processing such as type conversion, delta decoding and blending.

Depends on #467

### DIFF
--- a/read-fonts/src/tables/postscript/dict.rs
+++ b/read-fonts/src/tables/postscript/dict.rs
@@ -1,0 +1,724 @@
+//! Parsing for PostScript DICTs.
+
+use std::ops::Range;
+
+use super::{BlendState, Error, Number, Stack, StringId};
+use crate::{
+    tables::variations::ItemVariationStore,
+    types::{F2Dot14, Fixed},
+    Cursor,
+};
+
+/// PostScript DICT operator.
+///
+/// See "Table 9 Top DICT Operator Entries" and "Table 23 Private DICT
+/// Operators" at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf>
+#[derive(Copy, Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Operator {
+    Version,
+    Notice,
+    FullName,
+    FamilyName,
+    Weight,
+    FontBbox,
+    CharstringsOffset,
+    PrivateDictRange,
+    VariationStoreOffset,
+    Copyright,
+    IsFixedPitch,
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/postscript/dict.rs
```diff
@@ -0,0 +1,718 @@
...
+        Self::Operand(value.into())
+    }
+}
+
+/// Given a byte slice containing DICT data, returns an iterator yielding
+/// raw operands and operators.
+///
+/// This does not perform any additional processing such as type conversion,
+/// delta decoding or blending.
+pub fn tokens(dict_data: &[u8]) -> impl Iterator<Item = Result<Token, Error>> + '_ + Clone {
```
cmyr, line 178: do we expect this to be actually used elsewhere?

read-fonts/src/tables/postscript/dict.rs
```diff
@@ -0,0 +1,718 @@
...
+        }
+        snaps
+    }
+
+    pub fn values(&self) -> &[Fixed] {
+        &self.values[..self.len as usize]
+    }
+}
+
+fn parse_int(cursor: &mut Cursor, b0: i32) -> Result<i32, Error> {
```
cmyr, line 498: why do we pass in `b0` as an i32?

cmyr (approved):
Very cool to see this all coming together!

read-fonts/src/tables/postscript/dict.rs
dfrg: Good question. I exposed this to keep with the theme that read-fonts tries to provide low level access to encoded data.

The `entries` function is a sort of high level interface in that it consumes some information (specifically the blend vectors).

I was thinking we’d want a mid level version of this in write-fonts (where we can allocate) that can capture a more complete version of the data that can be round-tripped. This is a useful primitive for building that.

read-fonts/src/tables/postscript/dict.rs
dfrg: It’s used as i32 in some of the match arms. This is probably an artifact of my extracting the code into a function. I’ll switch to u8 and do the cast internally.

PR #472, opened 2023-06-06
check that p0,p1,p2 are both collinear AND equidistant before making p1 implied

anthrotype:
Before we were simply testing that the euclidean distance between p1 and p0 was the same as the distance between p2 and p1, without actually enforcing that the three points were on the same line.
Here we check that the two vectors p1p0 and p2p1's x/y components are almost the same.
FontTools also does the same, see:
https://github.com/fonttools/fonttools/blob/66ce7c0e0c0ffdd11b2b5011323305f166ee3b85/Lib/fontTools/ttLib/tables/_g_l_y_f.py#L1599-L1601

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -118,8 +118,14 @@ impl ContourPoint {
         Self::new(point, false)
     }
 
-    fn distance(&self, other: &Self) -> f64 {
-        self.point.distance(other.point)
+    /// Return True if p0, p1 (self) and p2 are collinear and p1 is equidistant from p0 and p2
+    fn is_midpoint_between(&self, p0: &Self, p2: &Self) -> bool {
+        let p1 = self;
+        let p1p0 = p1.point - p0.point;
+        let p2p1 = p2.point - p1.point;
+        // should tolerance be a parameter?
+        let eps = f32::EPSILON as f64;
+        (p1p0.x - p2p1.x).abs() < eps && (p1p0.y - p2p1.y).abs() < eps
     }
 }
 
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -118,8 +118,14 @@ impl ContourPoint {
...
-    fn distance(&self, other: &Self) -> f64 {
-        self.point.distance(other.point)
+    /// Return True if p0, p1 (self) and p2 are collinear and p1 is equidistant from p0 and p2
+    fn is_midpoint_between(&self, p0: &Self, p2: &Self) -> bool {
+        let p1 = self;
+        let p1p0 = p1.point - p0.point;
+        let p2p1 = p2.point - p1.point;
+        // should tolerance be a parameter?
+        let eps = f32::EPSILON as f64;
+        (p1p0.x - p2p1.x).abs() < eps && (p1p0.y - p2p1.y).abs() < eps
```
rsheeter, line 128: Can this be written more tersely? - the correct mid is (p0+p2) / 2 and we want to know if p1 is within epsilon of it. If we multiply p1 instead of dividing isn't that:

```rust
let delta_mid = p0 + p2 - 2 * p1; // how far from the midpoint we are
mid.x.abs() < f32::EPSILON && mid.y.abs() < f32::EPSILON
```

anthrotype: But p0, p1 and P2 are points not vectors, you can't sum them up or multiply by a scalar I think..

rsheeter: In looking closer kurbo implements midpoint, subtraction, and multiply by affine so it should have all the parts. Let me try real quick.

rsheeter: https://github.com/googlefonts/fontations/pull/474 exhibits; maybe the affine and the tolerance might want to be constants if we keep it. Feel free to merge or ignore.

anthrotype:
I pushed a failing test case and will follow up with the fix using your suggested more terse apprach

PR #474, opened 2023-06-06
Playing with midpoint

rsheeter:
Feel free to merge or ignore

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -5,7 +5,7 @@ use crate::{
     util::WrappingGet,
     OtRound,
 };
-use kurbo::{BezPath, Rect};
+use kurbo::{Affine, BezPath, Rect};
 
 use read_fonts::{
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/glyf.rs
```diff
@@ -120,12 +120,9 @@ impl ContourPoint {
 
     /// Return True if p0, p1 (self) and p2 are collinear and p1 is equidistant from p0 and p2
     fn is_midpoint_between(&self, p0: &Self, p2: &Self) -> bool {
-        let p1 = self;
-        let p1p0 = p1.point - p0.point;
-        let p2p1 = p2.point - p1.point;
-        // should tolerance be a parameter?
+        let delta_mid = p0.point.midpoint(p2.point) - Affine::scale(2.0) * self.point;
```
anthrotype, line 123: the scale is unnecessary (in fact the tests are failing). But I'll use your suggestion thanks

rsheeter: I had in mind that if p1 is supposed to roughly equal (p0+p2)/2 then 2*p1 should be roughly equal to (p0+p2). Admittedly I didn't actually test it :)

PR #476, opened 2023-06-07
[font-types] Add some pen adapters

dfrg:
Adds new `SvgPen` and `BufferPen` types that impl `Pen` and support output to SVG path data and buffers (containing `PenCommand`s), respectively.

These are implemented in terms of core traits so don't require std.

### DIFF
--- a/font-types/src/pen.rs
+++ b/font-types/src/pen.rs
@@ -1,11 +1,17 @@
+//! Types for working with sequences of path commands.
+
+use core::{
+    fmt,
+    iter::{self, Extend},
+};
+
 /// Interface for accepting a sequence of path commands.
 ///
 /// This is a general abstraction to unify ouput for processes that decode and/or
 /// transform outlines.
 ///
-/// /// AbstractPen in Python terms.
-/// <https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/basePen.py#L54>.
-/// Implementations
+/// Roughly equivalent to [AbstractPen](https://github.com/fonttools/fonttools/blob/78e10d8b42095b709cd4125e592d914d3ed1558e/Lib/fontTools/pens/basePen.py#L54)
+/// in FontTools.
 pub trait Pen {
     /// Emit a command to begin a new subpath at (x, y).
     fn move_to(&mut self, x: f32, y: f32);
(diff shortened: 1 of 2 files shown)


font-types/src/pen.rs
```diff
@@ -85,3 +91,146 @@ impl PenCommand {
...
+        let _ = write!(self.target, "{}Z", self.space);
+        self.space = " ";
+    }
+}
+
+/// Pen adapter that collects commands into a target buffer.
+///
+/// The target may be any type that implements `Extend<PenCommand>` such as
+/// `Vec<PenCommand>`.
+pub struct BufferPen<T> {
```
anthrotype, line 149: how about calling this RecordingPen like FontTools does?
Would be also nice if this had a "replay" method that takes another pen as argument and re-plays the recording onto that.

anthrotype (approved):

font-types/src/pen.rs
dfrg: That’s a better name, ty. Will change that and add a consuming replay method that is available when `T: IntoIterator<Item=PenCommand>`.

font-types/src/pen.rs
```diff
@@ -85,3 +91,146 @@ impl PenCommand {
...
+    pub fn new(target: T) -> Self {
+        Self { target }
+    }
+
+    pub fn into_inner(self) -> T {
+        self.target
+    }
+}
+
+impl<T: Extend<PenCommand>> Pen for BufferPen<T> {
```
cmyr, line 163: I think this type bound makes it possible to create a BufferPen that has unexpected behaviour?

for instance I'm not sure if `PenCommand` implements `Eq` and `Ord`, but if so, if you used a `BTreeSet` as the internal storage you could end up storing commands in a different order than they were provided, or you could end up removing duplicate commands that were meaningful.

cmyr (commented):
Looks good conceptually; my one note is that I wonder if the use of generics is maybe solving a problem we don't have?

Concretely, I'd be curious if there is a specific use case for either of these wherein the source is something other than `String` or `Vec`. If not, I feel like we would end up with a better API by just fixing the internal storage.

I think this is more of a concern for the `BufferPen`, since `Extend` is implemented by types with quite different behaviour. It makes more sense for me with `SvgPen`, where the pattern is familiar from other types that have a `W: Write` param.

font-types/src/pen.rs
dfrg: Fair enough! I don’t have any concrete use cases at the moment beyond providing some actual useful impls of Pen that are available out of the box. The skrifa scaler docs would benefit from not having to define an impl inline to demonstrate usage but I don’t think that is a strong enough case to override your concerns.

The other motivation was trying to expose some of the pen types that are hidden away in write-fonts, but pursuing iterator based path transforms is probably a more interesting route in the long term.

I’ll leave this open for a bit and let it simmer but based on your comments, I’m leaning towards closing and revisiting in the future.

font-types/src/pen.rs
rsheeter: I suggest also explicitly linking to recording pen and noting they do the same thing here and anywhere else we made a pen that matches, or is similar to, one fonttools ships.

font-types/src/pen.rs
rsheeter: > I wonder if the use of generics is maybe solving a problem we don't have?

+1, this looks overengineered at a glance.

rsheeter: > this seems to already exist in write-fonts (backed by a Vec)

We should delete them?

rsheeter:
I don't find it entirely obvious these things belong in font-types at all? SvgPen in particular doesn't feel like a fundamental type for interaction with fonts. I wonder if they might fit better in either skrifa or a new crate...?

dfrg:
Based on all the feedback, it seems these things don't adequately address any real need, so I'll close this and approach it differently if/when required.

Issue #478, opened 2023-06-07
Add extracted otf glyphs to test data


dfrg:
Make sure `extract_glyphs.py` exports the correct data for CFF fonts, add the constants to font-test-data and use them to actually test charstring evaluation.

dfrg:
Also add unit tests for each charstring operator!

dfrg:
Extracted glyphs added in #519. Testing is done in #520. 

A test for all operators was added in #479.

PR #479, opened 2023-06-07
[read-fonts] CFF charstring evaluation

dfrg:
Adds support for charstring evaluation. Results are passed to an impl of the `CommandSink` trait that provides hooks for processing both path commands and hinting data.

This doesn't do scaling or hinting, but both can be built on top of the sink trait.

Removes the existing "low level" parsing code for now because that did not properly handle the hint/counter masks args in the data stream that depend on tracking the stem hint count (and thus, also recursing into subroutines which can apply stem hints). I'll think about how to properly address this.

Testing is very light. We need a bit more infra for building a full PostScript "font" object from the CFF/CFF2 tables to support more sophisticated evaluation. Opened #478 to track testing.


read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
+                        while i < count {
+                            let (x1, x2, x3, y1, y2, y3);
+                            if alternate {
+                                x1 = x + self.stack.get_fixed(i)?;
+                                y1 = y;
+                                x2 = x1 + self.stack.get_fixed(i + 1)?;
+                                y2 = y1 + self.stack.get_fixed(i + 2)?;
+                                y3 = y2 + self.stack.get_fixed(i + 3)?;
+                                x3 = if count - i == 5 {
+                                    x2 + self.stack.get_fixed(i + 4)?
```
dfrg, line 476: Oops, this is missing an `i += 1`. Same for the similar branch below. I have a patch ready for this.

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
-            _ => Token::Operator(
-                Operator::from_opcode(b0).ok_or(Error::InvalidCharstringOperator(b0))?,
-            ),
-        }
-    })
+/// Charstring operators.
+/// See <https://learn.microsoft.com/en-us/typography/opentype/spec/cff2charstr#appendix-a-cff2-charstring-command-codes>
+// TODO: This is currently missing legacy math and logical operators.
+// fonttools doesn't even implement these: <https://github.com/fonttools/fonttools/blob/65598197c8afd415781f6667a7fb647c2c987fff/Lib/fontTools/misc/psCharStrings.py#L409>
+mod ops {
```
cmyr, line 529: I would probably replace this with an enum that could represent both the one-byte and two-byte operators, and then have a little convenience method that parses that out, but this works too.

cmyr (approved):
Did a read through and nothing really jumped out at me; I did not do a close reading of the CFF spec for each operation, so going to trust that any typos get shaken out when we get testing in place.

read-fonts/src/tables/postscript/charstring.rs
rsheeter: Plausible to have tests that catch?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
-    HFlex,
-    Flex,
-    HFlex1,
-    Flex1,
+/// See "Appendix B Type 2 Charstring Implementation Limits" in
+/// <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf>
+pub const NESTING_DEPTH_LIMIT: u32 = 10;
+
+/// Trait for processing commands resulting from charstring evaluation.
+#[allow(unused_variables)]
```
rsheeter, line 13: can we link to some sort of list of these commands and their definitions? See page 42 of some adobe pdf?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
-/// Either a PostScript charstring operator or a (numeric) operand.
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-pub enum Token {
-    Operator(Operator),
-    Operand(Number),
+/// Evaluates the given charstring and emits the resulting commands to the
+/// specified sink.
+///
+/// If evaluating a CFF2 charstring and the top-level table contains an
+/// item variation store, then `blend_state` must be provided.
```
rsheeter, line 66: Is there any reasonable default blend state? Comment should advise what happens if I don't provide a blend state for CFF2?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
+pub fn evaluate(
+    charstring_data: &[u8],
+    global_subrs: Index,
+    subrs: Option<Index>,
+    blend_state: Option<BlendState>,
+    sink: &mut impl CommandSink,
+) -> Result<(), Error> {
+    let mut evaluator = Evaluator::new(global_subrs, subrs, blend_state);
+    evaluator.evaluate(charstring_data, Fixed::ZERO, Fixed::ZERO, sink, 0)?;
```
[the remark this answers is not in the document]
rsheeter: +1, this sticks out ot me too. A comment on evaluator might help.

read-fonts/src/tables/postscript/charstring.rs
rsheeter: +1

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
+            let op = cursor.read::<u8>()?;
+            if op == ESCAPE {
+                let two_byte_op = cursor.read::<u8>()?;
+                // The following "flex" operators are intended to emit
+                // either two curves or a straight line depending on
+                // a "flex depth" parameter and the distance from the
+                // joining point to the chord connecting the two
+                // end points. In practice, we just emit the two curves,
+                // following FreeType:
+                // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/psaux/psintrp.c#L335>
```
rsheeter, line 129: Optional, wdyt about adding something like

```rust
    pub fn get_multiple_fixed<const COUNT: usize>(&self, first_index: usize) -> Result<[Fixed; COUNT], Error> {
        let mut result = [Fixed::from_i32(0); COUNT];
        for i in 0..COUNT {
            result[i] = self.get_fixed(first_index + i)?;
        }
        Ok(result)
    }
```

So you can write things like

```rust
                        let stack: [Fixed; 6] = self.stack.get_multiple_fixed(0)?;
                        let dx1 = x + stack[0];
                        let dy1 = y;
                        let dx2 = dx1 + stack[1];
```

Compiler will check you don't walk off the end, little bit less noise in the endless streams of dxn = something + stack.

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1,203 +1,644 @@
...
+                match two_byte_op {
+                    HFLEX => {
```
rsheeter, line 131: Request links to the specific page of documentation, e.g. https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf#page=19, at the top of each of these. 

It's a hard to get from the docs to our stream of something = something + something. For example, hflex says `dx1 dx2 dy2 dx3 dx4 dx5 dx6 `. OK, so we start at (x,y) and then the next point is defined only by dx1 so we don't update y, etc. If our sink was friendly to points we could encode this much more directly. For example, HFLEX would become something like:

```rust
let rdr = PointReader::new(&self.stack, x, y, 0);
sink.curve_to(rdr.dx(), rd.dxdy(), rdr.dx());

// PointReader apis:
// dx(): read a point that has only a dx from stack, added to curr x/y
// dxdy(): has x and y from stack
// etc
```

read-fonts/src/tables/postscript/charstring.rs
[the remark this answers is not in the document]
dfrg: This is transient state. Moved most of these parameters to fields, but left `charstring_data` and `nesting_depth` as those are directly dependent on the recursive calls. The cursor doesn't fit as a field because it's only valid for one stack frame.

Split out operator processing to a separate method from general flow and number parsing.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Done

read-fonts/src/tables/postscript/charstring.rs
dfrg: Added comments for both optional parameters (blend_state and subrs) noting the errors they'll return and why they would fail.

There's no reasonable default for blend state. It carries the item variation store from the CFF2 table and the operands for the blend operator can't even be parsed without knowing the region counts.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Yep, ty! I like this and it's very similar to what I did for handling variation deltas in the COLR table. Done.

read-fonts/src/tables/postscript/charstring.rs
dfrg: (links to specific pages added)

I want to like this one too, but I'm not sure this can be captured with the generality you're assuming. There are a lot of little exceptions to the "delta streaming" model across the operators. In your example, hflex requires resetting y to the initial value at dy5 which would require special processing.

I don't mind playing with something like this, but I'd like to not block this PR on it.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Done

read-fonts/src/tables/postscript/charstring.rs
dfrg: Yes, I can add units tests for each path construction operator. Updated #478 to capture this.

dfrg: Added a test with a charstring that contains all of the path construction operators. We currently match FreeType. This will hopefully ensure we continue to do so.

read-fonts/src/tables/postscript/charstring.rs
rsheeter: Per f2f I still think we're obscuring the logic with details and should at least try doing it in a clearer way. Fine to be in a followon. Tests can ensure we match FT.

rsheeter (approved):
LGTM subject to updating to less noisy point reading, either here or in a followon

read-fonts/src/tables/postscript/charstring.rs
dfrg: Will explore this in another PR.

PR #480, opened 2023-06-08
test if oncurves can be implied both before and after rounding

anthrotype:
and also, the same algorithm as python's math.isclose to compare floats when dropping implied oncurves

with this we now match fonttools 100% and produce identical glyf tables (except for a couple of known issues which we'll fix separately).


cmyr (approved):
Cool, I appreciate the thoroughness of this solution.

PR #482, opened 2023-06-08
Add SAFETY comment and remove an unsafe

danakj:
This adds a SAFETY comment on one use of unsafe, and in the other case a comment could not be written to explain how it is guaranteed to be safe, so remove the use of the unsafe method and use the safe from_utf8() instead.

Some followup work to address https://github.com/googlefonts/fontations/issues/435

### DIFF
--- a/otexplorer/src/print.rs
+++ b/otexplorer/src/print.rs
@@ -27,7 +27,7 @@ impl<'a> std::io::Write for PrettyPrinter<'a> {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         let len = self.writer.write(buf)?;
         let wrote_buf = &buf[..len];
-        let wrote_str = std::String::from_utf8_lossy(wrote_buf);
+        let wrote_str = String::from_utf8_lossy(wrote_buf);
         self.line_pos = match wrote_buf.iter().rev().position(|b| *b == b'\n') {
             Some(pos) => {
                 assert_eq!(pos, 0);
--- a/otexplorer/src/print.rs
+++ b/otexplorer/src/print.rs
@@ -27,7 +27,7 @@ impl<'a> std::io::Write for PrettyPrinter<'a> {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         let len = self.writer.write(buf)?;
         let wrote_buf = &buf[..len];
-        let wrote_str = unsafe { std::str::from_utf8_unchecked(wrote_buf) };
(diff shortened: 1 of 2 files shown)


otexplorer/src/print.rs
```diff
@@ -27,7 +27,7 @@ impl<'a> std::io::Write for PrettyPrinter<'a> {
     fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
         let len = self.writer.write(buf)?;
         let wrote_buf = &buf[..len];
-        let wrote_str = unsafe { std::str::from_utf8_unchecked(wrote_buf) };
```
cmyr, line 30: this won't compile since the type has changed from `&str` to `Result<&str, _>`. I agree this should be changed, but let's just do `String::from_utf8_lossy`?

read-fonts/src/tables/post.rs
```diff
@@ -73,6 +73,8 @@ impl<'a> FontRead<'a> for PString<'a> {
             .get(1..len as usize + 1)
             .ok_or(ReadError::OutOfBounds)?;
         if pstring.is_ascii() {
+            // SAFETY: `pstring` must be valid UTF-8, which is known to be the
+            // case since ASCII is a subset of UTF-8 and `is_ascii()` is true.
```
cmyr, line 77: Thanks!

cmyr (commented):
Thanks for this, one little note inline.

otexplorer/src/print.rs
danakj: Oh my bad, thanks.

danakj (commented):
Use from_unsafe_lossy to avoid Result type.

cmyr:
yea I find the web editor is more pain than pleasure, much of the time 😒

cmyr (approved):
Thanks!

otexplorer/src/print.rs
```diff
@@ -27,7 +27,7 @@ impl<'a> std::io::Write for PrettyPrinter<'a> {
...
+        let wrote_str = std::String::from_utf8_lossy(wrote_buf);
```
cmyr, line 30: oops, one more time?

```suggestion
        let wrote_str = String::from_utf8_lossy(wrote_buf);
```

danakj: Ah yes.. I knew it looked wrong.

danakj:
Maybe we can squash all of this when merging :sweat_smile:

PR #485, opened 2023-06-09
Expose outline format in TableProvider

### LINKED ISSUE #477: [font-types] Typo in `CFF_SFTN_VERSION`
I assume this was meant to be `CFF_SFNT_VERSION` instead.

https://github.com/googlefonts/fontations/blob/1d8564345b4499127c46991bd588bbd214749bc1/font-types/src/lib.rs#LL54C11-L54C27

dfrg:
Adds a new `OutlineFormat` enum and `outline_format` method to `TableProvider`. Impl in `FontRef` bases this on `sfntVersion`.

While this was generally accessible from `FontRef`, the skrifa scaler only accesses fonts through `TableProvider` and this info will be necessary for passing some of the Unicode text rendering tests (those that include both `glyf` and `CFF` tables and expect the choice to be made based on `sfntVersion`) in the future.

Also fixes #477

### DIFF
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -242,7 +242,7 @@ impl<'a> FontRef<'a> {
         data: FontData<'a>,
         table_directory: TableDirectory<'a>,
     ) -> Result<Self, ReadError> {
-        if [TT_SFNT_VERSION, CFF_SFTN_VERSION].contains(&table_directory.sfnt_version()) {
+        if [TT_SFNT_VERSION, CFF_SFNT_VERSION].contains(&table_directory.sfnt_version()) {
             Ok(FontRef {
                 data,
                 table_directory,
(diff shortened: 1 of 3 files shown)


read-fonts/src/lib.rs
```diff
@@ -257,4 +257,19 @@ impl<'a> TableProvider<'a> for FontRef<'a> {
     fn data_for_tag(&self, tag: Tag) -> Option<FontData<'a>> {
         self.table_data(tag)
     }
+
+    fn outline_format(&self) -> Option<OutlineFormat> {
+        Some(match self.table_directory.sfnt_version() {
+            TT_SFNT_VERSION => OutlineFormat::TrueType,
+            CFF_SFNT_VERSION => OutlineFormat::PostScript,
+            _ => return None,
```
cmyr, line 265: hmmm.

For FontRef, at least, if the table directory has an unknown sfnt version we return an error.

Are there other `TableProviders` for which this isn't true?

It might make sense to be maximally defensive, but I think it's probably reasonable to say we don't handle unknown `sfnt` values, since it's really not clear what we should do in that case?

dfrg: Good point. I could imagine a `TableProvider` impl that doesn't provide any outlines at all I suppose. The `table_directory` field on `FontRef` is also public so it's conceivable to construct one with an sfnt version that doesn't match what we expect.

One possible solution would be returning `PostScript` for `CFF_SFNT_VERSION` here and just defaulting to `TrueType` for any other values.

PR #488, opened 2023-06-09
Remove panicking paths from Tag

cmyr:
Copy-pasting the first commit message:

his was ill-considered, and is not a good match for the ways this type
is used in practice.

More broadly, this change represents a loosening of constraints on the
contents of a Tag, to better match what we are actually doing anyway.
Concretely, this means that we are taking the spec as a *suggestion*,
but fully support invalid tags.

This also adds a `validate` method to Tag, which can be used to check
the validity of a tag that was created through one of the unchecked
paths.

The second commit changes our printing behaviour, so that if a tag is non-ascii we will print the numerical value of the non-ascii bytes.

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -194,14 +194,14 @@ impl AsRef<[u8]> for Tag {
 
 impl Display for Tag {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
-        // a dumb no-std way of ensuring this string is valid utf-8
-        let mut bytes = [b'-'; 4];
-        for (i, b) in self.0.iter().enumerate() {
-            if b.is_ascii() {
-                bytes[i] = *b;
+        for byte in self.0 {
+            if (0x20..=0x7E).contains(&byte) {
+                write!(f, "{}", byte as char)?;
+            } else {
+                write!(f, "{{0x{:02X}}}", byte)?;
             }
         }
-        Display::fmt(&std::str::from_utf8(&bytes).unwrap(), f)
+        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


font-types/src/tag.rs
```diff
@@ -6,31 +6,35 @@ use std::{
 
 /// An OpenType tag.
 ///
-/// A tag is a 4-byte array where each byte is in the printable ascii range
-/// (0x20..=0x7E).
+/// [Per the spec][spec], A tag is a 4-byte array where each byte is in the
```
dfrg, line 9:
```suggestion
/// [Per the spec][spec], a tag is a 4-byte array where each byte is in the
```

font-types/src/tag.rs
```diff
@@ -6,31 +6,35 @@ use std::{
...
     ///
-    /// If the input has fewer than four bytes, it will be padded with spaces
-    /// (`0x20`).
-    ///
-    /// # Panics
-    ///
-    /// This method panics if the tag is not valid per the requirements above.
+    /// If the slice is longer than four bytes, it will be truncated. If it is
+    /// shorter, the missing bytes will be replaced with the space character (0x20).
     pub const fn new(src: &[u8]) -> Tag {
```
dfrg, line 29: Can we just change this to take `&[u8; 4]`? That works nicely with typical usage where you pass something like `b"wght"` when you want a const tag. I'm not sure if this breaks any actual uses.

dfrg (approved):
Looks good with a comment update and a question.

font-types/src/tag.rs
dfrg: Thinking a bit more, there are some places where the short form is used. `CFF` and `cvt` come to mind. I did define the CFF tag as "CFF " when I added it to codegen out of habit. I think skrifa uses `b"cvt"` but that could be changed.

cmyr: honestly I didn't think the `b"xx"` syntax created fixed-size arrays, I thought it always created a static slice? I'm happy to make this change, since the silent truncation behaviour was one of the things I liked the least.

dfrg: Sounds good to me!

PR #492, opened 2023-06-13
[read-fonts] CFF charstring less noisy point reading

dfrg:
Try to do fancy things when emitting curves


rsheeter (approved):
Much more readable imho, ty.

dfrg:
Thanks for the quick reviews!

Late commit is just a small simplification to `emit_curves` that I spotted while doing a final pass.

PR #495, opened 2023-06-13
[read-fonts] Better CFF/CFF2 tables

dfrg:
This changes codegen inputs to generate fake "headers" for the CFF/CFF2 tables and then hand defines new top level tables for each that parse all data that doesn't require DICT evaluation.

This makes these tables follow a similar style to the rest of read-fonts.


cmyr (approved):
This looks nice and principled.

Not sure if it's helpful but a reminder that you can actually use the `FontData` identifier as a field type in codegen, iirc.

PR #497, opened 2023-06-14
[skrifa] use gvar for metric deltas if hvar is missing

dfrg:
@drott found that skrifa is computing incorrect advance widths for some variable fonts. Turns out that we didn't handle the case where the `HVAR` table is missing. This updates `GlyphMetrics` to fall back to `gvar` in that case.


cmyr (approved):
Looks good!

PR #501, opened 2023-06-16
Tweak docs for Skia build

rsheeter:
Based on a test run through. Note that the `bazelisk run` on the viewer fails but I think probably we should fix that.

```shell
$ bazelisk run --sandbox_base=/dev/shm --with_fontations //tools/viewer -- --slide GM_typeface_fontations_roboto
...
Resource "resources/images/gamut.png" not found.
Resource "resources/fonts/Roboto-Regular.ttf" not found.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Releasing involves the following steps:
    ./resources/scripts/release.sh read-fonts write-fonts
    ```
 
-## skia and chromium builds
+## Skia and Chromium builds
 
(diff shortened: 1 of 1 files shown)


drott (approved):
The Chromium instructions are currently incomplete. The CL landed, but that default disabled the functionality, and I am in the process of landing a Chromium chrome://flags switch to activate it. That's intended to bring the behavior closer to what it will be like if we start default building the Fontations backend  in Canary. I'll update those to show how to enable the flag.

PR #503, opened 2023-06-21
Update build instructions for Chromium

drott:


yisibl:
Can this only be turned on in a custom build of Chromium? When will it be available in Chrome canary?

drott:
> Can this only be turned on in a custom build of Chromium? When will it be available in Chrome canary?

Yes and we're working on that.

PR #510, opened 2023-06-23
Add DeviceOrVariationIndex enum

cmyr:
In the first pass of this stuff I basically ignored the `VariationIndex` table, and just used the `Device` table everywhere. This patch replaces this with a 'format table', `DeviceOrVariationIndex`, that is an enum with `Device` and `VariationIndex` members. We parse these as appropriate based on the `DeltaFormat`.

I have mixed feelings! On the plus side, this is consistent with the rest of the codebase and makes good use of codegen etc. On the downside, this will be more annoying to use without much clear benefit, since you'll need to use a `match` statement or similar in order to access the actual inner table that you want; this is especially annoying since the `Device` table seems to be more or less ignored now.

There are some alternatives that might make this less of a headache:

- I could keep this code, but add some convenience methods to the enum for accessing the inner values, without unwrapping
- I could remove this, just keeping the device table, and leaving it up to the user to interpret the fields correctly as needed.

There are probably other reasonable options, but those two feel like a good place to start the conversation.

### DIFF
--- a/docs/codegen-tour.md
+++ b/docs/codegen-tour.md
@@ -439,11 +439,16 @@ You then also declare that you want to create an enum, providing an explicit
 format, and listing which tables should be included:
 
 ```rust
-format u16 MyTable {
+format u16[@N] MyTable {
     Format1(MyTableFormat1),
     Format2(MyTableFormat2),
 }
 ```
+
+the 'format' keyword is followed by the type that represents the format, and
+optionally a position  at which to read it (indicated by the '@' token, followed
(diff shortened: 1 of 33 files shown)


dfrg:
This honestly seems like a good approach to me. The design is a bit unfortunate but the fault for that lies with the spec and this seems to accurately capture the concept with nice codegen. 

All the abstraction around `ValueRecord` _might_ be a performance problem for shaping but we’ll profile when we get to that and add an escape hatch if necessary.

PR #511, opened 2023-06-23
[read-fonts] TrueType scaler that works from a memory block

dfrg:
This adds support for scaling TrueType outlines to read-fonts. The scaler can compute the full memory requirements up front and then operate within an arbitrary block of bytes of the required size.

The primary motivation is supporting embedded use cases where dynamic heap allocation is not available. This code is general enough, however, to replace the implementation in skrifa and has the benefit of enabling use of custom host allocators (in Skia, for example).

This contains new unsafe code (`alloc_slice` in mem.rs). I've added checks, constraints and comments that hopefully address the safety concerns. This needs thorough review.

The diff is somewhat large, but much of the code can be ignored. The `deltas.rs` module is copied verbatim from skrifa. The `scale.rs` module is also copied, but with two modifications: 1) required memory is sliced from the provided buffers rather than using `Vec`s and 2) hinting is exposed by calling a provided closure rather than an internal stub. There are no real functional changes and this code produces the same results as skrifa (and FreeType) for the Noto + Google Fonts collections.

Thus, focus for review should be on both `mem.rs` which contains the unsafe code for buffer allocation and `scale.rs::Scaler::outline_info_rec` where the memory requirements are actually computed.

~~If we decide to move forward with this, I'll move the "extracted glyph" testing code from skrifa to read-fonts and adapt the current outline tests before merging. This is trivial but I didn't want to increase the diff size even further.~~ Moved in #519. Will add tests later.

API usage looks like this:
```rust
let scaler = Scaler::new(&font);
// Returns a ScalerGlyph type that contains the raw glyph data as 
// well as metadata for allocating the necessary buffers.
// The second parameter specifies whether we intend to hint the glyph.
let glyph = scaler.glyph(glyph_id, false)?;
let buf_size = glyph.required_buffer_size();
// Nice, potentially reusable, generic buffer!
let mut buf = vec![0u8, buf_size];
// Allocate typed slices from our buffer.
let memory = glyph.memory_from_buffer(&mut buf)?;
// Load the actual outline.
let outline = scaler.outline(memory, &glyph, font_size, coords)?;
outline.to_path(&mut pen)?;
```


read-fonts/src/tables/truetype.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Errors that may occur when scaling glyphs.
+#[derive(Clone, Debug)]
+pub enum Error {
+    /// The requested glyph was not present in the font.
+    GlyphNotFound(GlyphId),
+    /// Exceeded a recursion limit when loading a glyph.
+    RecursionLimitExceeded(GlyphId),
+    /// Exceeded memory limits when loading a glyph.
+    InsufficientMemory,
```
rsheeter, line 34: Shouldn't we know what glyph, what limit, and perhaps how much we needed?

read-fonts/src/tables/truetype/deltas.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub points: &'a [Point<i32>],
+    pub flags: &'a mut [PointFlags],
+    pub contours: &'a [u16],
+}
+
+/// Compute a set of deltas for the points in a simple glyph.
+///
+/// This function will use interpolation to infer missing deltas for tuples
+/// that contain sparse sets. The `working_points` buffer is temporary storage
+/// used for this and the length must be >= glyph.points.len().
```
rsheeter, line 44: shouldn't we have a guard clause to check this and error if it's not true?

read-fonts/src/tables/truetype.rs
dfrg: Unfortunately, the full context isn’t known at all places where this error is generated so we can’t provide useful information in the general case. Presumably, the user has computed the required size up front and provided a buffer >= that size (or returned their own error if unable to do so). If the buffer size does meet our requirement and we’re still failing then it’s a logic bug in our code and the only reasonable thing to do is file an issue.

read-fonts/src/tables/truetype/deltas.rs
dfrg: Yes, good catch! I’ll remove the potential panics.

dfrg: Fixed

read-fonts/src/tables/truetype/deltas.rs
```diff
@@ -0,0 +1,350 @@
...
+    let SimpleGlyph {
+        points,
+        flags,
+        contours,
+    } = glyph;
+    // Include the first phantom point if the font is missing variable metrics
+    // for left side bearings. The adjustment made here may affect the final
+    // shift of the outline.
+    let actual_len = if has_var_lsb {
+        points.len() - 4
```
cmyr, line 73: is it possible for `points` to be shorter than expected?

read-fonts/src/tables/truetype/mem.rs
```diff
@@ -0,0 +1,188 @@
...
+/// This must only be implemented for types that contain no internal padding
+/// and for which all bit patterns are valid.
+unsafe trait TransmuteElement: Sized + Copy {}
+
+unsafe impl TransmuteElement for u16 {}
+unsafe impl TransmuteElement for i32 {}
+unsafe impl TransmuteElement for Fixed {}
+unsafe impl TransmuteElement for F26Dot6 {}
+unsafe impl TransmuteElement for PointFlags {}
+unsafe impl<T> TransmuteElement for Point<T> where T: TransmuteElement {}
```
cmyr, line 73: I don't think this is safe? `Point` does not have a `repr`, and could contain padding.

read-fonts/src/tables/truetype/mem.rs
```diff
@@ -0,0 +1,188 @@
...
+/// # Safety
+/// This must only be implemented for types that contain no internal padding
+/// and for which all bit patterns are valid.
+unsafe trait TransmuteElement: Sized + Copy {}
+
+unsafe impl TransmuteElement for u16 {}
+unsafe impl TransmuteElement for i32 {}
+unsafe impl TransmuteElement for Fixed {}
+unsafe impl TransmuteElement for F26Dot6 {}
+unsafe impl TransmuteElement for PointFlags {}
```
cmyr, line 72: I'd need a second opinion on this, but I'm not sure if `PointFlags` is safe. I would mark it `#[repr(transparent)]` to guarantee that it has the same layout as a `u8` (although obviously it should??)

read-fonts/src/tables/truetype/scale.rs
```diff
@@ -0,0 +1,718 @@
...
+        gvar::Gvar,
+        hmtx::Hmtx,
+        hvar::Hvar,
+        loca::Loca,
+    },
+    types::{F26Dot6, F2Dot14, Fixed, GlyphId, Point},
+    TableProvider,
+};
+
+/// Scaler state for TrueType outlines.
```
cmyr, line 21: Assuming this is all actually public API I think it would benefit from a bit more documentation around how exactly the consumer should use it.

cmyr (commented):
Impressive work chad! I have a few maybe-pedantic notes on the unsafe, but overall this is a great milestone :)

read-fonts/src/tables/truetype/deltas.rs
dfrg: We always push phantom points before calling into this so it should be at least 4, but yes, if we make this public, we’ll need an extra check. I’ll add one to be safe.

read-fonts/src/tables/truetype/mem.rs
dfrg: Good catch. Point always contains two values of the same type so cannot contain padding unless the coordinate type does. I think making this `repr(C)` should guarantee that?

read-fonts/src/tables/truetype/mem.rs
dfrg: Another good catch. Will add `repr(transparent)`

read-fonts/src/tables/truetype/mem.rs
dfrg: By “cannot” I mean “should not in any sane compiler” :)

read-fonts/src/tables/truetype/deltas.rs
dfrg: Done

read-fonts/src/tables/truetype/mem.rs
dfrg: Done. Also added this to the fixed point types. I thought we had already done it for some reason.

read-fonts/src/tables/truetype/mem.rs
dfrg: Fixed

dfrg:
@danakj would you mind taking a glance at the new unsafe code in this PR? It's localized to the `alloc_slice` function in `mem.rs` and is all private to that module.

For context, we've had an internal request to support TrueType fonts on embedded microcontrollers where heap allocation is not available and this is designed to support that.

read-fonts/src/tables/truetype/mem.rs
```diff
@@ -0,0 +1,192 @@
...
+    let len_in_bytes = len * size_of::<T>();
+    if len_in_bytes > aligned_buf.len() {
+        return None;
+    }
+    let (slice_buf, rest) = aligned_buf.split_at_mut(len_in_bytes);
+    // SAFETY: Alignment and size requirements have been checked in 1) and
+    // 2) above, respectively.
+    // Element types are limited by the `TransmuteElement` trait which
+    // defines requirements for transmutation and is private to this module.
+    let slice = unsafe { std::slice::from_raw_parts_mut(slice_buf.as_mut_ptr() as *mut T, len) };
```
danakj, line 99: I believe this is fine on the basis that:
- this function takes a borrow on buf
- thus the returned slice borrow keeps the original borrow from the caller
- the returned borrows do not overlap, as provided by split_at_mut() and that this slice's length in bytes was used to do the splitting
- getting the borrow as a pointer then making a new borrow is fine, the original one is invalidated but was thrown away, slice_buf is not usable thereafter.

danakj (approved):
Sorry for taking a bit, I lost this window underneath others after starting to look.

read-fonts/src/tables/truetype/mem.rs
dfrg: Thanks for the additional analysis. This all looks correct to me as well.

dfrg:
> Sorry for taking a bit, I lost this window underneath others after starting to look.

No worries! The review is much appreciated.

dfrg:
This was delayed pending discussion on whether the functionality should be public. We resolved to encapsulate this code as a private module in skrifa and rethink the higher level API there to accommodate clients with differing requirements.

Thanks all for the feedback!

PR #515, opened 2023-06-28
[font-types] Serde Tag as str when target is text

cmyr:
That is, if the target format is human-readable we will serialize Tag as a string, and otherwise we will serialize it as raw bytes.

Does this make sense? mmaaaaybe? It's definitely nice to serialize as a string. Whether we even need to think about supporting non-utf8 tags is unclear to me; they may exist in fonts in the wild, but they shouldn't be used in new fonts.

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -18,7 +18,6 @@ use std::{
 ///
 /// [spec]: https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types
 #[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Tag([u8; 4]);
 
 impl Tag {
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
Very helpful to fontc, the tag bytes are quite unreadable, ty

font-types/src/tag.rs
```diff
@@ -233,6 +232,42 @@ impl Default for Tag {
...
+#[cfg(feature = "serde")]
+impl serde::Serialize for Tag {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        if serializer.is_human_readable() {
+            std::str::from_utf8(&self.0)
+                .map_err(serde::ser::Error::custom)?
+                .serialize(serializer)
```
rsheeter, line 249: Optional and IMHO can wait until we have a real example: you could fallback to the non-string path if string fails. Deserialize would similarly attempt string, presumably failing if the initial serialize failed, and fallback to parsing `[u8; 4]`.

I suppose you could stuff it into an enum to serialize/deserialize to avoid deserialize having to attempt multiple formats.

cmyr: I looked into that and while possible it's kind of annoying, and so I ended up reaching the same conclusion you have here, e.g. let's wait and see if this is ever actually an issue.

Issue #490, opened 2023-06-09
Consistently use #[inline] in font-types


cmyr:
This attribute is poorly documented and inconsistently used, but potentially important.

For some more information, see https://github.com/rust-lang/hashbrown/pull/119#issuecomment-537539046.

The TL;DR is that `#[inline]` has two tasks: *within* a crate, it functions as a hint. Between crate boundaries, though, it is required in order for inlining to be possible at all.

This last point is particularly important in the `font-types` crate; it contains main small methods, which are really only called from external crates. In many cases, the code produced by inlining these functions is not going to be much larger than the code required to setup and execute a function call.

Based on this, I suspect that the majority of the small functions in font-types should be marked `#[inline]`; and this should definitely be true of functions like the `get` methods on `BigEndian`.

The downside of this is that each crate that uses font-types will end up with a separate copy of these functions. I believe that these should ultimately be deduplicated, but it will minorly impact compile times.

PR #516, opened 2023-06-28
[font-types] inline more things

dfrg:
Based on #490, add more inline attributes to functions in `font-types`. This was measurably affecting performance in the TrueType scaler.

Also adds lossy f32 conversions to the fixed point types and updates `glyf::to_path` to use them.

JMM

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -135,6 +135,8 @@ macro_rules! fixed_mul_div {
     ($ty:ty) => {
         impl $ty {
             /// Multiplies `self` by `a` and divides the product by `b`.
+            // This one is specifically not always inlined due to size and
+            // frequency of use. We leave it to compiler discretion.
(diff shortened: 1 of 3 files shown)


font-types/src/fixed.rs
```diff
@@ -122,6 +135,7 @@ macro_rules! fixed_mul_div {
     ($ty:ty) => {
         impl $ty {
             /// Multiplies `self` by `a` and divides the product by `b`.
+            #[inline]
```
rsheeter, line 138: not always?

rsheeter: if intentional please comment why

dfrg: This one is large and not used that often, so I thought it best to leave it to compiler discretion

rsheeter (approved):
lgtm

PR #518, opened 2023-06-29
Perf improvements for the new scaler

dfrg:
This is based on #511 and is a separate PR just to ease review.

Changes the API so that we only do a load from the `glyf` table once for the top-level glyph. Also contains various other optimizations for traversing composite glyphs.

The performance improvements from this are significant and with these changes, the "new scaler" is now (sometimes substantially) faster than the current scaler.

Just as an example, API use looks something like this:

```rust
let scaler = Scaler::new(&font);
// Returns a ScalerGlyph type that contains the actual glyph data as 
// well as metadata for allocating the necessary buffers.
// The second parameter specifies whether we intend to hint the glyph.
let glyph = scaler.glyph(glyph_id, false)?;
let buf_size = glyph.required_buffer_size();
// Nice, potentially reusable, generic buffer!
let mut buf = vec![0u8, buf_size];
// Allocate typed slices from our buffer.
let memory = glyph.memory_from_buffer(&mut buf)?;
// Load the actual outline.
let outline = scaler.outline(memory, &glyph, font_size, coords)?;
outline.to_path(&mut pen)?;
```

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -582,7 +582,7 @@ impl Iterator for ComponentIter<'_> {
 }
 
 /// Iterator that only returns glyph identifiers for each component.
-/// 
+///
 /// Significantly faster in cases where we're just processing the glyph
 /// tree, counting components or accessing instructions.
 #[derive(Clone)]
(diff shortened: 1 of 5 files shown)


dfrg:
The diff here is noisier than I expected due to some renaming and not particularly useful in isolation. Closing this and pushing the changes to the main PR instead.

PR #519, opened 2023-06-29
Move outline testing infra from skrifa to read-fonts

dfrg:
This simply moves the "extracted glyph" parsing code from skrifa to read-fonts (behind the scaler_test feature) and adds some new test data to prep for further CFF work.


cmyr (approved):
~yolo~ LGTM

PR #520, opened 2023-06-29
[read-fonts] Add scaler for CFF/CFF2

dfrg:
This is a unified scaler for charstrings in CFF/CFF2 tables. The API attempts to be similar in style to the new TrueType scaler. Infrastructure for hinting is here, but completing it will require a few more PRs. #519 added extracted glyph outlines for static and variable fonts. Those are tested here.

After this lands  ~~hooking it up to skrifa will be trivial and~~ we'll magically support .otf fonts :)

edit: diff for skrifa integration was small so I went ahead and pushed it here


read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -55,21 +55,21 @@ where
...
-            cx1.to_f64() as f32,
-            cy1.to_f64() as f32,
-            x.to_f64() as f32,
-            y.to_f64() as f32,
+            cx0.to_f32(),
+            cy0.to_f32(),
+            cx1.to_f32(),
+            cy1.to_f32(),
+            x.to_f32(),
+            y.to_f32(),
```
rsheeter, line 72: Nit: to_f32 would have been a nice standalone JMM PR. I wouldn't bother to hoist it out now tho.

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -78,6 +78,179 @@ where
...
+
+impl<'a, S> ScalingSink26Dot6<'a, S> {
+    pub fn new(sink: &'a mut S, scale: Fixed) -> Self {
+        Self { scale, inner: sink }
+    }
+
+    fn scale(&self, coord: Fixed) -> Fixed {
+        if self.scale != Fixed::ONE {
+            // The following dance is necessary to exactly match FreeType's
+            // application of scaling factors:
```
rsheeter, line 99: if it's possible to tell, perhaps note _why_ we think FT does it the way it does it. Or if not perhaps note that :)

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -78,6 +78,179 @@ where
...
+    }
+
+    fn close(&mut self) {
+        self.inner.close();
+    }
+}
+
+/// Command sink adapter that simplifies path operations.
+///
+/// This currently removes degenerate moves and lines.
```
rsheeter, line 161: Comment would be more helpful if it offered:

* Why? 
* Do we think it might change in future? If so ... why?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -78,6 +78,179 @@ where
...
+            self.scale(y),
+        );
+    }
+
+    fn close(&mut self) {
+        self.inner.close();
+    }
+}
+
+/// Command sink adapter that simplifies path operations.
```
rsheeter, line 159: The opening line basically restates what the type says which is not hugely helpful. Further, "simplify" is a loaded term given things like simplify(path). I suggest an alternate name.

NopFilteringSink ... something that directly conveys what it does.

read-fonts/src/tables/postscript/charstring.rs
rsheeter: Why only moves and lines, why don't we filter all commands whose points are coincident? That pattern could also inform a more informative name?

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+    version: Version<'a>,
+    top_dict: ScalerTopDict<'a>,
+    units_per_em: u16,
+}
+
+impl<'a> Scaler<'a> {
+    /// Creates a new scaler for the given font.
+    ///
+    /// This will choose an underyling CFF2 or CFF table from the font, in that
+    /// order.
```
rsheeter, line 59: Globally or can we pull some glyphs from cff, some from cff2? ...is that even a legal setup...?

The example suggests the user must be very aware of the fonts internals. What should a noob do, can we have a Just Works scaler that doesn't expose so many of the internals? 

Think I "just" want to get a damn outline and I don't much care what outline table it came from.

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    /// Returns the subfont (or Font DICT) index for the given glyph
+    /// identifier.
+    pub fn subfont_index(&self, glyph_id: GlyphId) -> u32 {
+        self.top_dict
+            .fd_select
+            .as_ref()
+            .and_then(|select| select.font_index(glyph_id))
+            // Missing FDSelect assumes either a single Font DICT at index 0,
+            // or a Private DICT entry in the Top DICT.
```
rsheeter, line 120: feels like there should be citation?

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+/// # fn example(font: FontRef, coords: &[F2Dot14], pen: &mut impl Pen) -> Result<(), Error> {
+/// let scaler = Scaler::new(&font)?;
+/// let glyph_id = GlyphId::new(24);
+/// // Retrieve the subfont index for the requested glyph.
+/// let subfont_index = scaler.subfont_index(glyph_id);
+/// // Construct a subfont with the given configuration.
+/// let size = 16.0;
+/// let coords = &[];
+/// let with_hinting = false;
+/// let subfont = scaler.subfont(subfont_index, size, coords, with_hinting)?;
```
rsheeter, line 42: When would the user use a subfont index other than the one computed for the glyph id? If the answer is never, why are we inflicting all this tiresome detail on them?

rsheeter: If we really must inflict the details can we offer an interface that doesn't as well?

For context, I want people to be able to write code like https://github.com/rsheeter/read-fonts-wasm/blob/main/src/lib.rs and have it Just Work for truetype, cff, or cff2. Ideally I also want to interact with COLR similarly: if it says I need an outline for a glyph I can ask for it without having to care about the details of which outline table I'm using. At most I want to tell you something like what I consider the outline table priority to be, e.g. the order of tags.

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+
+    /// Loads and scales an outline for the given subfont instance, glyph
+    /// identifier and normalized variation coordinates.
+    ///
+    /// Before calling this method, use [`subfont_index`](Self::subfont_index)
+    /// to retrieve the subfont index for the desired glyph and then
+    /// [`subfont`](Self::subfont) to create an instance of the subfont for a
+    /// particular size and location in variation space.
+    /// Creating subfont instances is not free, so this process is exposed in
+    /// discrete steps to allow for caching.
```
rsheeter, line 184: This seems like it's exposing exactly the sort of detail Context claims to abstract away from me? - not necessarily a problem but perhaps we should explain why.

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+            let mut range = None;
+            for entry in dict::entries(font_dict_data, None) {
+                if let dict::Entry::PrivateDictRange(r) = entry? {
+                    range = Some(r);
+                    break;
+                }
+            }
+            range
+        } else {
+            // Otherwise, assume the top dict provided a private dict range
```
rsheeter, line 254: "assume" threw me here; it seems to not quite be true, we error if not ok below. I think perhaps it's more like `// Last chance, a privatedict range from the top dict` than "assume"?

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+            self.top_dict.private_dict_range.clone()
+        }
+        .ok_or(Error::MissingPrivateDict)
+    }
+}
+
+enum Version<'a> {
```
rsheeter, line 261: ref https://learn.microsoft.com/en-us/typography/opentype/spec/cff and https://learn.microsoft.com/en-us/typography/opentype/spec/cff2 respectively?

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+    Version1(Cff<'a>),
+    Version2(Cff2<'a>),
+}
+
+/// Specifies local subroutines and hinting parameters for some subset of
+/// glyphs in a CFF or CFF2 table.
+///
+/// This type is designed to be cacheable to avoid re-evaluating the private
+/// dict every time a charstring is processed.
```
rsheeter, line 270: again we seem to be presuming our user will cache. Where would one find instructions on implementing a correct, high performance, implementation and ... shouldn't we just give them one?

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+    fn check_blues(blues: &Blues, expected_values: &[(f64, f64)]) {
+        for (i, blue) in blues.values().iter().enumerate() {
+            let expected = expected_values[i];
+            assert_eq!(blue.0, Fixed::from_f64(expected.0));
+            assert_eq!(blue.1, Fixed::from_f64(expected.1));
+        }
+    }
+
+    #[test]
+    fn read_noto_serif_display() {
```
rsheeter, line 414: Suggest naming based on what the test is examining, for example `read_cff_static` would tell me more than this name and a pattern would emerge as you go on to `read_cff2_static`, etc.

read-fonts/src/tables/postscript/scale.rs
```diff
@@ -0,0 +1,541 @@
...
+        );
+    }
+
+    #[test]
+    fn noto_serif_display_outlines() {
+        compare_glyphs(
+            font_test_data::NOTO_SERIF_DISPLAY_TRIMMED,
+            font_test_data::NOTO_SERIF_DISPLAY_TRIMMED_GLYPHS,
+        );
+    }
```
rsheeter, line 498: I cannot readily tell where the expected values come from when I read the unit test. I assume the answer is freetype, perhaps through naming or comment we could clarify. For example, I can follow this somewhat more easily:

```rust
    #[test]
    fn cff_static_outline_matches_freetype() {
        compare_glyphs(
            font_test_data::NOTO_SERIF_DISPLAY_TRIMMED,
            font_test_data::NOTO_SERIF_DISPLAY_TRIMMED_GLYPHS,
        );
    }

    #[test]
    fn cff2_static_outline_matches_freetype() { ... }
   
   // etc
```

...that's making me realize, it looks like we test nothing more than some outline for each outline format. Shouldn't we build up test assets that use all the drawing commands? E.g. test that a cff with an hstem matches the freetype outline with an hstem, perhaps making a test font with many test glyphs such that each glyph uses "basic" (line, curve, etc) commands plus one unusual (hstem, vstem, etc) command . We could them compare those glyphs for (cff, cff2) x (static, variable).

rsheeter (approved):
Code lgtm but I did have a fair number of questions and I'm skeptical the unit tests are adequate (details in comment). I realize the "real" test is a large corpus of glyphs but I think we can get more mileage out of unit testing by constructing a targeted corpus. 

Some of the questions might imply additional work that doesn't make sense to do in this PR (assuming it makes sense at all); feel free to file issues for followon PRs if you prefer not to address here. I'm fond of this approach when the core of the thing seems good and I need to make followon changes, (e.g. https://github.com/googlefonts/fontc/issues/362.

skrifa/src/scale/scaler.rs
```diff
@@ -198,32 +208,48 @@ impl<'a> Scaler<'a> {
...
+            Err(Error::NoSources)
+        }
     }
 }
 
-/// Outline glyph scalers.
-struct Outlines<'a> {
-    glyf: Option<(glyf::Scaler<'a>, &'a mut glyf::Outline)>,
+// Clippy doesn't like the size discrepancy between the two variants. Ignore
+// for now: we'll replace this with a real cache.
```
rsheeter, line 226: suggest filing an issue for this

skrifa/src/scale/scaler.rs
```diff
@@ -122,22 +123,30 @@ impl<'a> ScalerBuilder<'a> {
     pub fn build(mut self, font: &impl TableProvider<'a>) -> Scaler<'a> {
```
rsheeter, line 123: I wonder if the scaler builder should let me set what outline tables I prefer in prioritized order, probably defaulting to [glyf, cff2, cff]?

skrifa/src/scale/scaler.rs
```diff
@@ -122,22 +123,30 @@ impl<'a> ScalerBuilder<'a> {
...
         self.resolve_variations(font);
         let coords = &self.context.coords[..];
-        let glyf = if let Ok(glyf) = glyf::Scaler::new(
+        let size = self.size.ppem().unwrap_or_default();
+        let outlines = if let Ok(glyf) = glyf::Scaler::new(
```
rsheeter, line 127: I think this means that if I have cff *and* glyf with most of the glyphs in cff my font will perform exceedingly poorly. I'll grant you this is a weird font, but I think it should work. Maybe scaler should have a vector of Outlines in prioritized order; most of the time you would have only one but a particularly exciting font could have 0 or > 1.

rsheeter: Oh hey or a COLR font that uses outlines from *both*! That's definitely a good idea right?

read-fonts/src/tables/postscript/scale.rs
rsheeter: Aha! That's at the very bottom. I think perhaps we could say here that we anticipate most users will use this via skrifa which will provide the caching?

read-fonts/src/tables/postscript/charstring.rs
dfrg: My best guess is that this is a combination of FT being a TrueType first implementation and the fact that the Adobe code was added later… so it was all plugged in to fit the existing model. I’ll note that this was likely done for historical and compatibility reasons 🤷🏻‍♂️

read-fonts/src/tables/postscript/charstring.rs
dfrg: Like so many other things, it’s just what FT does so we follow.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Will rename this.

read-fonts/src/tables/postscript/scale.rs
dfrg: My assumption is that anyone using read-fonts directly will definitely need to have a solid understanding of font internals. To my mind, that’s the whole point of the library. Skrifa is supposed to be the “easy mode” API.

read-fonts/src/tables/postscript/scale.rs
dfrg: Once this lands, the code you linked will just work with CFF/CFF2 fonts :)

read-fonts/src/tables/postscript/scale.rs
dfrg: I’ll add a note to the Scaler docs

read-fonts/src/tables/postscript/scale.rs
dfrg: We actually test a number of outlines at multiple sizes and (for VFs) locations in variation space. That was the whole point of adding the extract glyphs script. 

The charstring module has a handcrafted test that contains all operators except hinting. We can’t test those until hinting support lands.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Done

read-fonts/src/tables/postscript/scale.rs
dfrg: Per IM, mix and match isn't plausible due to the dense nature of the charstring index (and loca table).

read-fonts/src/tables/postscript/scale.rs
dfrg: Added links to the spec for each variant

read-fonts/src/tables/postscript/scale.rs
dfrg: Done

read-fonts/src/tables/postscript/scale.rs
dfrg: Renamed the tests as suggested

read-fonts/src/tables/postscript/scale.rs
dfrg: Renamed these as suggested. We can increase coverage by adding more glyphs or test fonts to font-test-data. Or additionally, but adjusting the outline instances generated by the extract glyphs script.

skrifa/src/scale/scaler.rs
dfrg: The spec doesn't support this. Both `loca` and the charstrings index are defined to have `maxp.numGlyphs` entries and empty entries are valid (think any space glyph), so there is no way to specify sparse glyph sets.

skrifa/src/scale/scaler.rs
dfrg: I'm going to agree with earlier Rod in saying that skrifa users probably shouldn't care about this level of detail. We should also likely be using `sfntVersion` to determine priority instead (that is, prefer CFF2->CFF->glyf when the version is OTTO and glyf->CFF2->CFF otherwise). We don't yet have enough information to do that here. See #485

skrifa/src/scale/scaler.rs
dfrg: #535

dfrg:
Per other discussions the scaler code has been moved from read-fonts to a private module in skrifa. We can likely move some of this code back to read-fonts to accommodate usage as requested in #541 but this PR has already been delayed long enough and I wanted to keep churn low.

PR #521, opened 2023-07-02
[scaler-bench] Do some work in the callbacks

behdad:
Such that the calculations for points are not all optimized away.


dfrg (approved):
Thanks for taking a look. This LGTM!

dfrg:
Looks like most of the clippy/fmt errors are mine. I’ll fix them up in a new commit.

PR #525, opened 2023-07-04
[codegen] Generate From impls for format tables

cmyr:
That is, if a table `T` has multiple formats, for each format `S` we generate `From<S>` for `T`.

This is a minor quality-of-life improvement that should let me offer a better API in some parts of write-fonts (it's motivated by the case of Device/VariationIndex tables, but I've been meaning to do this for a while)


dfrg (approved):
LGTM. I was surprised to see this only being done for write-fonts but then I couldn’t think of any actual use case for read.

PR #527, opened 2023-07-04
[font-types] Ensure Tag impls DeserializeOwned

cmyr:
I ran into a bit of a surprise when trying to update fontc to the our latest release: it was unable to deserialize our Tag type, complaining that,

> Error("axes[0].tag: invalid type: string \"wght\", expected a borrowed string")

This caused some headscratching, since I couldn't reproduce it in tests. It turned out to be a funny set of things:

- My implementation involved temporarily deserializing to a &str
- The implementation of Deserialize for &str only works when the input can be borrowed
- The deserialization code in fontc uses yaml::from_reader
- from_reader involves reading from a byte stream, which is transient
- ergo you cannot deserialize &str from a reader

The solution is to do a slightly more principled Deserialize implementation, where I explicitly implement the Visitor trait.

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -259,9 +259,22 @@ impl<'de> serde::Deserialize<'de> for Tag {
     where
         D: serde::Deserializer<'de>,
     {
+        struct TagStrVisitor;
+        impl<'de> serde::de::Visitor<'de> for TagStrVisitor {
+            type Value = Tag;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+                write!(formatter, "a four-byte ascii string")
+            }
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

font-types/src/tag.rs
```diff
@@ -259,9 +259,22 @@ impl<'de> serde::Deserialize<'de> for Tag {
     where
         D: serde::Deserializer<'de>,
     {
+        struct TagStrVisitor;
```
anthrotype, line 262: noob question: any particular reason why this struct def/impl section is inlined in the `deserialized` function body and not at the top level as usual? Is it because it's only used in here? I presume there aren't any downsides to this (struct is defined only once, not each time one calls `deserialize`).

cmyr: correct, this is just a pattern you see sometimes, where small helper functions or types are just defined in the function body. It makes it clear that they are only referenced here.

cmyr: (and also correct, this compiles exactly the same as if it were declared at the top level; the only practical difference is that this type is not visible outside of this function)

PR #528, opened 2023-07-04
[font-types] Make serde feature explicitly require std

cmyr:
Previously it would have been possible to disable default features and also enable the serde feature, which would have caused compilation to fail. This now makes explicit that the serde feature requires std.

JMM

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -311,6 +311,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "std")]
     fn display() {
         let bad_tag = Tag::new(&[0x19, b'z', b'@', 0x7F]);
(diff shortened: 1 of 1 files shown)


dfrg:
Do we actually need this? Do any of the types in this crate actually require std for serde?

I’m also curious about having std as a default feature. There doesn’t appear to be any way to turn this off from read-fonts.

cmyr:
Good Q actually? The thing that caused this to jump out at me was the custom deserialize for `Tag`, where one of the methods requires formatting with `std::fmt` types, but maybe those are also available in core?

dfrg:
I think `core::fmt` covers the same functionality as std but I’m not sure how that interacts with serde and custom (de)serialization.

cmyr:
okay you're totally right. I've changed this patch to now just fix one broken test case, and removed the serde->std requirement.

PR #530, opened 2023-07-05
Implement FromObjRef, FromTableRef for CompositeGlyph

rsheeter:
Enabled fontc to dump_table, write to a file, and recover the write-fonts type by reading the read-fonts type then from_table_ref'ing it. SimpleGlyph already knows this trick.


dfrg (approved):
LGTM

PR #550, opened 2023-08-01
[write-fonts] Improve builder API

cmyr:
This makes a number of tweaks to the FontBuilder API, intended to improve ergonomics:

- A 'copy_missing_tables' method was added to simplify the common step of bulk copying existing tables from a FontRef
- The existing 'add_table' method was renamed 'add_raw',
- and a new 'add_table' method was added that removes the need for the caller to manually compile the table and provide the appropriate tag to the builder.

### DIFF
--- a/write-fonts/src/font_builder.rs
+++ b/write-fonts/src/font_builder.rs
@@ -1,8 +1,9 @@
 //!  A builder for top-level font objects
 
-use std::borrow::Cow;
 use std::collections::BTreeMap;
+use std::{borrow::Cow, fmt::Display};
 
+use read::{FontRef, TableProvider};
 use types::{Tag, TT_SFNT_VERSION};
 
 include!("../generated/generated_font.rs");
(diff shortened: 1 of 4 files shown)


write-fonts/src/font_builder.rs
```diff
@@ -37,16 +48,57 @@ impl TableDirectory {
...
+        for record in font.table_directory.table_records() {
+            let tag = record.tag();
+            if !self.tables.contains_key(&tag) {
+                let data = match font.data_for_tag(tag) {
+                    Some(data) => data,
+                    None => {
+                        log::warn!("data for '{tag}' is malformed");
+                        Default::default()
+                    }
+                };
```
dfrg, line 87: My preference would be to skip the table here (or return an error) rather than adding an empty one.

dfrg (approved):
This looks great and the rmw example is pretty slick now, thanks! One nit inline but feel free to ignore.

PR #552, opened 2023-08-02
Fixup breakage between two outstanding PRs

cmyr:
This has now happened twice, so I'm inclined to turn on the setting that requires branches to be up to date in order to be mergeable.


cmyr:
I don't have permissions to manage this setting, but it's under "branches", where you set protection rules: "require branches to be up to date before merging".

dfrg (approved):

PR #555, opened 2023-08-02
[write-fonts] Impl FontRead for glyf types

cmyr:
In codegen we impl this trait for any type that impls FromTableRef, but this was missing for these types (which don't use codegen)

Noticed this was missing when I saw that fontc was jumping through hoops to load glyfs.

JMM


dfrg (approved):
Must have missed these when I added the original impls. Seems like a nice QOL improvement!

PR #557, opened 2023-08-02
[write-fonts] Split tables::glyf into two modules

cmyr:
I want to add the glyf/loca compilation logic to write-fonts and this
file was already pushing 2k lines, so I think it's worth breaking up.

----
based on #555


dfrg (approved):
Divide and conquer!

PR #559, opened 2023-08-03
Add GlyfLocaBuilder

cmyr:
This attempts to be a nice ergonomic way to build the glyf & loca tables.

Based on #558.

I have a few questions about this, and the fontc impl (maybe @anthrotype can answer these?)

- I've added special logic here so that if a glyph contains no outline/contour, we write no data for that glyph, but still add a (zero-length) entry to the loca table. I was looking at fontc's current impl, and it doesn't seem to do this? I assume it is an oversight there, but I'd love someone to double-check my work.
- Additionally the spec says that the first glyph is always the placeholder/notdef glyph. Should we be enforcing this somehow, or is it up to the caller? It feels like it has to be up to the caller, since we don't really have the relevant information...

### DIFF
--- a/write-fonts/src/error.rs
+++ b/write-fonts/src/error.rs
@@ -56,6 +56,12 @@ impl std::fmt::Debug for PackingError {
 impl std::error::Error for PackingError {}
 impl std::error::Error for Error {}
 
+impl From<ValidationReport> for Error {
+    fn from(value: ValidationReport) -> Self {
+        Error::ValidationFailed(value)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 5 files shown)


anthrotype:
> if a glyph contains no outline/contour, we write no data for that glyph, but still add a (zero-length) entry to the loca table

yeah empty glyphs contain no glyf data, so their location is the same as the preceding glyph

> I was looking at fontc's current impl, and it doesn't seem to do this?

well, what does it do differently?

> enforcing .notdef... up to the caller?

yeah, sounds good to me

anthrotype (approved):

cmyr:
In fontc, currently, if a glyph contains no outlines we still write the header.

PR #560, opened 2023-08-03
[write-fonts] SimpleGlyph construction tweaks

cmyr:
- avoid using 'kurbo' in method names: this feels likely to confuse people who are unfamiliar with the kurbo crate
- rename 'BadKurbo' error to 'MalformedPath' (ditto)
- wrap simple_glyphs_from_kurbo with a method on Glyph: this saves us having to rexport another top-level item
- add docs to this method (I didn't initially understand the invariants)

-----
I think this is the last bit of breakage I have planned, will make sure all of this works with fontc and then do a release.

### DIFF
--- a/write-fonts/src/tables/glyf/simple.rs
+++ b/write-fonts/src/tables/glyf/simple.rs
@@ -69,11 +69,12 @@ impl SimpleGlyph {
     /// The paths are expected to be preprocessed, and interpolation compatible
     /// (i.e. they should have the same number and type of points, in the same
     /// order.) They should contain only line and quadratic segments; the caller
-    /// is responsible for converting cubics to quadratics  as needed.
+    /// is responsible for converting cubics to quadratics as needed.
     ///
     /// This method is provided for use when compiling variable fonts.
     /// The inputs are expected to be different instances of the same named
-    /// glyph, each corresponding to a different location in the designspace.
+    /// glyph, each corresponding to a different location in the variation
+    /// space.
     pub fn interpolatable_glyphs_from_bezpaths(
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/glyf/simple.rs
```diff
@@ -58,10 +58,26 @@ impl SimpleGlyph {
...
+            .map(|mut x| x.pop().unwrap())
+    }
+
+    /// Attempt to create a set of interpolation-compatible glyphs from a set
+    /// of paths.
+    ///
+    /// The paths are expected to be preprocessed, and interpolation compatible
+    /// (i.e. they should have the same number and type of points, in the same
+    /// order.) They should contain only line and quadratic segments; the caller
+    /// is responsible for converting cubics to quadratics  as needed.
```
dfrg, line 72:
```suggestion
    /// is responsible for converting cubics to quadratics as needed.
```

write-fonts/src/tables/glyf/simple.rs
```diff
@@ -58,10 +58,26 @@ impl SimpleGlyph {
...
+    ///
+    /// This method is provided for use when compiling variable fonts.
+    /// The inputs are expected to be different instances of the same named
+    /// glyph, each corresponding to a different location in the designspace.
```
dfrg, line 76:
```suggestion
    /// glyph, each corresponding to a different location in the variation
    /// space.
```
This reads a bit better to me as I associate "designspace" with the file format.

dfrg (approved):
Couple comment suggestions. Otherwise looks good!

PR #562, opened 2023-08-03
[write-fonts] Migrate LocaFormat tests from fontc

cmyr:
If I'm moving this type into write-fonts I might as well grab the tests at the same time...


dfrg (approved):
🚢

PR #565, opened 2023-08-09
iup_delta_optimize: fix interpretation of contour_ends

### LINKED ISSUE #564: IUP optimization bug with Oswald's glyph "2"
Compare the three fonts built from the same Oswald.glyphs source:

https://github.com/googlefonts/fontations/assets/6939968/da6e42a3-dd63-40c6-a0be-bb96b0e1a350

The first one was built with fontmake 3.7.1, the second one with fontc (main as of https://github.com/googlefonts/fontc/commit/287e6bc6c9f39229ba12f5d48ad62e740700dbab), and the third one with the IUP optimization forcibly disabled like so:

```diff
diff --git a/fontbe/src/glyphs.rs b/fontbe/src/glyphs.rs
index c3d4570..7fa21b2 100644
--- a/fontbe/src/glyphs.rs
+++ b/fontbe/src/glyphs.rs
@@ -598,7 +598,7 @@ impl CheckedGlyph {
     fn should_iup(&self) -> bool {
         match self {
             CheckedGlyph::Composite { .. } => false,
-            CheckedGlyph::Contour { .. } => true,
+            CheckedGlyph::Contour { .. } => false,
         }
     }
 }
```

The glyph "2" looks odd when IUP is enabled, but looks ok when it is disabled.

anthrotype:
Fixes #564 

the input and expected result is taken from what fontmake/fonttools spit out for Oswald's glyph "two"

EDIT: Commit [36a8929](https://github.com/googlefonts/fontations/pull/565/commits/36a8929947c1147cb5f56c2d6117b1bb9b28ee3b) also fixes https://github.com/googlefonts/fontations/issues/567

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,5 +10,9 @@ members = [
     "skrifa",
 ]
 
+[workspace.dependencies]
+# dev dependencies
+pretty_assertions = "1.3.0"
+
 [workspace.metadata.release]
 allow-branch = ["main"]
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -26,3 +26,4 @@ font-test-data = { path = "../font-test-data" }
 read-fonts = { version = "0.10.0", path = "../read-fonts", features = [ "codegen_test"] }
 env_logger = "0.10.0"
 rstest = "0.18.0"
+pretty_assertions.workspace = true
--- a/write-fonts/src/tables/variations/iup.rs
+++ b/write-fonts/src/tables/variations/iup.rs
@@ -47,7 +47,7 @@ pub fn iup_delta_optimize(
     let expected_num_coords = contour_ends
         .last()
         .copied()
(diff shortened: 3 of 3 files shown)


anthrotype:
this will a breaking change for fontc, should I also bump write-fonts minor version within this same PR, or you prefer a follow-up?

dfrg (approved):
Not sure if this is ready but the fix lgtm.

dfrg:
> this will a breaking change for fontc, should I also bump write-fonts minor version within this same PR, or you prefer a follow-up?

Probably a follow-up but current policy is merge without approval for version bumps.

anthrotype:
thanks for the review, yes it's ready but we need to update write-fonts first before we can integrate this

anthrotype:
> we need to update write-fonts first before we can integrate this

sorry, I meant before we can integrate https://github.com/googlefonts/fontc/pull/390

PR #566, opened 2023-08-09
[skrifa] CFF hinting part 1: scaling the blues

dfrg:
This code implements the first stage of CFF hinting. It uses hinting parameters captured from the Private DICT (blue values, fuzz, shift, etc) and combines them with a desired scale to generate a set of "blue zones" that will (in future PRs) be used along with per-glyph stem hints to generate the final hint map for grid fitting.

This is based on `cf2_blues_init` at https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/psaux/psblues.c#L66

### DIFF
--- a/read-fonts/src/tables/postscript/dict.rs
+++ b/read-fonts/src/tables/postscript/dict.rs
@@ -430,7 +430,7 @@ pub struct Blues {
 }
 
 impl Blues {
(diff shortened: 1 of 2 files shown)


cmyr (approved):
Looks good, appreciate the inline comments & sourcing. :)

PR #572, opened 2023-08-11
iup_contour_optimize: ensure encoded indices match rotated list of deltas

anthrotype:
the expected correct result comes from fontTools `iup_delta_optimize` which we want to match.
I am still trying to figure out the root cause of the divergence.

in ttx form, this is the diff

```diff
--- GoogleSans-VF-fontmake.ttf
+++ GoogleSans-VF-fontc.ttf
       <tuple>
         <coord axis="GRAD" value="1.0"/>
+        <delta pt="0" x="2" y="0"/>
         <delta pt="1" x="4" y="0"/>
-        <delta pt="2" x="8" y="-1"/>
         <delta pt="3" x="10" y="-1"/>
-        <delta pt="4" x="10" y="0"/>
-        <delta pt="5" x="-14" y="25"/>
-        <delta pt="6" x="-8" y="34"/>
-        <delta pt="7" x="-3" y="38"/>
-        <delta pt="8" x="-5" y="35"/>
         <delta pt="9" x="-7" y="35"/>
-        <delta pt="10" x="6" y="35"/>
-        <delta pt="11" x="22" y="27"/>
-        <delta pt="12" x="29" y="11"/>
         <delta pt="14" x="29" y="-13"/>
         <delta pt="15" x="22" y="-29"/>
         <delta pt="16" x="8" y="-37"/>
@@ -121,25 +112,31 @@
         <delta pt="21" x="-19" y="-28"/>
         <delta pt="22" x="8" y="0"/>
         <delta pt="23" x="8" y="0"/>
+        <delta pt="24" x="6" y="0"/>
         <delta pt="25" x="4" y="0"/>
+        <delta pt="27" x="0" y="0"/>
+        <delta pt="28" x="-5" y="0"/>
+        <delta pt="29" x="-8" y="0"/>
+        <delta pt="30" x="-10" y="0"/>
         <delta pt="31" x="-10" y="0"/>
+        <delta pt="32" x="-8" y="0"/>
+        <delta pt="33" x="-5" y="0"/>
+        <delta pt="34" x="-1" y="0"/>
       </tuple>
```


anthrotype:
Found it! In `iup_contour_optimize`, there's an `if` branch where the deltas/coords are rotated rightwards so that the last point is encoded; then we get the correct resulting set of encoded indices from `iup_contour_optimize_dp`, but... the fonttools original version applies the solution (i.e. replace unencoded points with None) to the deltas and _then_ rotates them back to the left, whereas in our rust code we rotate the deltas back _before_ we apply the solution itself at the end.. So the indices did not match!
I fixed it by rotating the set of encoded indices as well, so they match the rotated list of deltas, see https://github.com/googlefonts/fontations/pull/572/commits/68cbc79f080f44b955797f00cb74fa674d495b56

behdad:
LGTM.

PR #576, opened 2023-08-15
[write-fonts] Post builder fixups

### LINKED ISSUE #548: write-fonts: 'post' table's version should be computed
Just noticed that this is missing.

cmyr:
was looking for a little task while waiting for something else to finish and saw #548.

This closes #548, although not exactly as intended.

This is also a breaking change, so I'd prefer to hold it back until we have some more breaking stuff and we can merge it all together.

-----

This makes it so that we compute the num_glyphs value, and provide defaults for the legacy 'min_mem_XXX' fields.

It also replaces the `new_v2` method with methods for adding string data to an existing table. I couldn't decide on what the actual API should be here, but I didn't like that `new_v2` added defaults for the standard fields.

### DIFF
--- a/resources/codegen_inputs/post.rs
+++ b/resources/codegen_inputs/post.rs
@@ -7,8 +7,6 @@ table Post {
     /// 0x00025000 for version 2.5 (deprecated) 0x00030000 for version
     /// 3.0
     #[version]
-    // NOTE: we will need some sort of builder to compile post tables, and that
-    // builder should set the version. This attribute is a placeholder.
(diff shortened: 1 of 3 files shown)


cmyr:
ah okay good catch!

cmyr:
That's actually an interesting complication for codegen, this is the first time where we will have a lower version of a table (2.0) that has more fields than a higher version (3.0). We make a bunch of assumptions about fields only being added with versions, so I'll have to think about this some.

PR #578, opened 2023-08-16
[write-fonts] Add more tests of ValueRecord compilation

cmyr:
Sanity checking that these tables are writing correctly in the simple case.

----

this is part of my looking into https://github.com/googlefonts/fontc/issues/402; in inspecting the generated font with otexplorer it reported that two of the VariationIndex tables were missing, e.g. the offsets were out of bounds for the data in the table. These tests pass, but we might as well keep them.


dfrg (approved):
More tests are always better!

Issue #581, opened 2023-08-21
[Skrifa] Provide access to `glyf` overlap flags in scaler


drott:
In order to mimic what [Skia does for FreeType](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/ports/SkFontHost_FreeType_common.cpp;l=2065), we need to be able to simplify overlapped paths based on the presence of the overlap flag(s) in `glyf`. 

This is used for stroking a glyph outline - which looks not as intended when path overlap. 

It'd be useful if the scaler of skrifa or a part of read_fonts could expose a flag that's similar to the [`FT_OUTLINE_OVERLAP`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/freetype/src/include/freetype/ftimage.h;l=409;drc=6d6ff81c27d10c12e3308b6bd488b042cf8967ac) flag in behaviour. (For this one I am not sure if it is only set based on OVERLAP_SIMPLE or also on OVERLAP_COMPOUND speaking in OpenType terms).

References:
* https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/ports/SkFontHost_FreeType_common.cpp;l=2065
* https://learn.microsoft.com/en-us/typography/opentype/spec/glyf OVERLAP_SIMPLE, OVERLAP_COMPOUND

PR #582, opened 2023-08-23
Capture overlap bits from TrueType glyphs

dfrg:
First step in addressing #581. Captures simple and composite overlap bits and plumbs them through internally.

Still needs to be exposed in skrifa public API. This will be done in a follow up.


read-fonts/src/tables/glyf.rs
```diff
@@ -945,6 +958,52 @@ mod tests {
...
+            let glyph = match loca.get_glyf(GlyphId::new(gid), &glyf) {
+                Ok(Some(Glyph::Simple(glyph))) => glyph,
+                _ => continue,
+            };
+            if gid == 3 {
+                // Only GID 3 has the overlap bit set
+                assert!(glyph.has_overlapping_contours())
+            } else {
+                assert!(!glyph.has_overlapping_contours())
+            }
```
rsheeter, line 977: This seems like the verbose version of  `assert_eq!(gid == 3, glyph.has_overlapping_contours())` ?

rsheeter: Hm. Actually even then it's a bit more vulnerable to broken test data than I like; I propose we directly assert that only gid 3 has the bit:

`assert_eq!(vec![3], (0..glyph_count).filter(...retain only if has overlaps).collect())`

read-fonts/src/tables/glyf.rs
```diff
@@ -945,6 +958,52 @@ mod tests {
...
+        }
+    }
+
+    #[test]
+    fn composite_overlapping_contour_flag() {
```
rsheeter, line 982: The loading of test data, num glyphs, etc is repetitive noise; perhaps a test helper that lets you easily iterate over (gid, &Glyph) would help tidy?

read-fonts/src/tables/glyf.rs
```diff
@@ -945,6 +958,52 @@ mod tests {
...
+        let font = FontRef::new(font_test_data::VAZIRMATN_VAR).unwrap();
+        let loca = font.loca(None).unwrap();
+        let glyf = font.glyf().unwrap();
+        let glyph_count = font.maxp().unwrap().num_glyphs();
+        for gid in 0..glyph_count {
+            let glyph = match loca.get_glyf(GlyphId::new(gid), &glyf) {
+                Ok(Some(Glyph::Composite(glyph))) => glyph,
+                _ => continue,
+            };
+            // Only GID 2, component 1 has the overlap bit set
```
rsheeter, line 992: As above I might prefer it as something like `assert_eq!(vec![(2, 1)], ...collect (gid, component index) of things with overlap compound)`

skrifa/src/scale/glyf/scaler.rs
```diff
@@ -733,3 +735,22 @@ impl<'a, H> ScalerInstance<'a, H> {
...
+    fn overlap_flags() {
+        let font = FontRef::new(font_test_data::VAZIRMATN_VAR).unwrap();
+        let scaler = Scaler::new(&font).unwrap();
+        let glyph_count = font.maxp().unwrap().num_glyphs();
+        for gid in 0..glyph_count {
+            let glyph = scaler.glyph(GlyphId::new(gid), false).unwrap();
+            // GID 2 is a composite glyph with the overlap bit on a component
+            // GID 3 is a simple glyph with the overlap bit on the first flag
+            assert_eq!(glyph.has_overlaps, gid == 2 || gid == 3);
+        }
```
rsheeter, line 754: As above, I'd prefer `assert_eq!(vec[2, 3], ...collect gids with overlap...)`

rsheeter (approved):
LGTM

read-fonts/src/tables/glyf.rs
dfrg: I like this style as well. Fixed!

read-fonts/src/tables/glyf.rs
dfrg: The refactoring part of my brain seems to turn off when writing tests. Good call, ty!

dfrg:
Oops, I pushed all the fixes requested here to the branch for #583. Since both PRs are approved and related, I'll consider these "fixed in a followon" :)

PR #583, opened 2023-08-23
[skrifa] return overlap status from scaler 

dfrg:
Changes `Scaler::outline` to return a new `ScalerMetrics` struct that contains the overlap flag. This also has optional fields to represent metrics that may be adjusted by hinting.

Based on https://github.com/googlefonts/fontations/pull/582 which must be merged first.
Fixes https://github.com/googlefonts/fontations/issues/581


skrifa/src/scale/mod.rs
```diff
@@ -246,6 +246,24 @@ mod tests {
...
+            is_cff: false,
+        };
+        let mut scaler = cx.new_scaler().build(&font);
+        let glyph_count = font.maxp().unwrap().num_glyphs();
+        for gid in 0..glyph_count {
+            let metrics = scaler.outline(GlyphId::new(gid), &mut path).unwrap();
+            // GID 2 is a composite glyph with the overlap bit on a component
+            // GID 3 is a simple glyph with the overlap bit on the first flag
+            assert_eq!(metrics.has_overlaps, gid == 2 || gid == 3);
+        }
```
rsheeter, line 264: Same preference for how to assert this as #582

rsheeter (approved):

skrifa/src/scale/mod.rs
dfrg: Aye

PR #589, opened 2023-08-31
Improve PairPos validation

cmyr:
In write-fonts we have a 'validation' pass before compilation, which is an opportunity to go and inspect the tables we're writing and make sure that the data is sane. A bunch of these checks are generated automatically, but a lot of the validation logic we would want is case-specific; in this case we have a mechanism for naming additional methods to be called in particular places during validation.

We haven't made a ton of use of this custom validation logic, but we probably should, since it helps us be much more confident in the data that we're writing.

This adds some custom validation logic to the two PairPos subtable formats; there was something fishy happening here that was messing up my table packing.

Another thing I'm reminding of while doing this work is how tricky it is to construct these tables correctly without the assistance of a builder; there's a bunch of builders in `fea-rs`, and they should probably eventually migrate to `write-fonts`.

### DIFF
--- a/write-fonts/src/tables/gpos.rs
+++ b/write-fonts/src/tables/gpos.rs
@@ -107,6 +107,27 @@ impl PairPosFormat1 {
             .map(|rec| rec.value_record2.format())
             .unwrap_or(ValueFormat::empty())
     }
+
+    fn check_format_consistency(&self, ctx: &mut ValidationCtx) {
+        let vf1 = self.compute_value_format1();
+        let vf2 = self.compute_value_format2();
+        ctx.in_array(|ctx| {
+            for pairset in &self.pair_sets {
+                ctx.array_item(|ctx| {
+                    ctx.in_field("pair_value_records", |ctx| {
+                        ctx.in_array(|ctx| {
+                            if pairset.pair_value_records.iter().any(|pairset| {
+                                pairset.value_record1.format() != vf1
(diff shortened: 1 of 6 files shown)


rsheeter:
I suggest filing issues for additional logic as we identify specific cases

+1 to enhance write-fonts w/builders

write-fonts/src/tables/gpos.rs
```diff
@@ -107,6 +107,27 @@ impl PairPosFormat1 {
...
+                                pairset.value_record1.format() != vf1
+                                    || pairset.value_record2.format() != vf2
+                            }) {
+                                ctx.report("all ValueRecords must have same format")
+                            }
+                        })
+                    })
+                })
+            }
+        })
```
rsheeter, line 129: I don't love the nesting, how might we structure this more like an iter chain where it tends to extend down more than right?

rsheeter (approved):

write-fonts/src/tables/gpos.rs
cmyr: Agreed. currently the vast majority of validation code is auto-generated, which makes this less painful. I think that improving the API is out of scope for this PR, but I'll open an issue to look into improving this somewhat. I'm not sure that an iter-like API is possible since it doesn't really support scoping, but we might be able to at least reduce the amount of nesting required.

PR #591, opened 2023-09-06
Graph isolation fix

cmyr:
This is a roll up of two commits:

- the first is a set of changes to ensure that graph packing is deterministic; in practice this means replacing `HashSet` and `HashMap` with `BTreeSet` and `BTreeMap` in places where we later iterate over the collection.
- the second fixes a subtle bug in our logic when we are attempting to isolate some subgraph: we were not tracking whether we had visited a node previously when counting edges, which meant that in certain circumstances, when a node pointed to the same child multiple times, we would count subsequent links in that subgraph multiple times, which could cause us to fail to correctly duplicate nodes when needed. This bug was a large part of what I was banging my head against last week.

cc @garretrieger; the relevant stuff here is in 1bc5900.


rsheeter (approved):

cmyr:
updated this to copy the harfbuzz impl, which reuses the subgraph map to check for 'visited' state, saving us from needing a second data structure.

PR #595, opened 2023-09-07
[write-fonts] Include children in pairpos f1 size

cmyr:
This fixes a bug where we were not including the size of subtables when deciding on split points for pairpos format 1 tables.

This also removes the 'can_potentially_promote_subtables' check; if we reach this point we have already overflowed, which means we are probably in GPOS and GSUB, and this method complicated testing, since splitting & expansion wouldn't be triggered on a lookup subtable if it was not an ancestor of GPOS/GSUB.

### DIFF
--- a/write-fonts/src/graph/splitting.rs
+++ b/write-fonts/src/graph/splitting.rs
@@ -102,9 +102,22 @@ fn split_pair_pos_format_1(graph: &mut Graph, subtable: ObjectId) -> Option<Vec<
     let mut accumulated = BASE_SIZE;
     let mut split_points = Vec::new();
 
-    for (i, pair_set) in data.offsets.iter().skip(1).enumerate() {
-        let table_size = if visited.insert(pair_set.object) {
-            graph.objects[&pair_set.object].bytes.len()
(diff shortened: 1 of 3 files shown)


rsheeter:
> means we are probably in GPOS and GSUB

The "probably" causes me a degree of anxiety here, what if we were wrong about this?

write-fonts/src/graph/splitting.rs
```diff
@@ -402,4 +415,83 @@ mod tests {
         // ensure that the PairSet tables at the split boundaries are as expected
         assert_eq!(sub1.pair_set_count() + sub2.pair_set_count(), N_GLYPHS);
     }
+
+    fn make_pp1_with_device_tables(g1_count: u16, g2_count: u16) -> PairPos {
```
rsheeter, line 419: I suggest not shortening pairpos1; not everyone will know what a pp1 is

cmyr:
> > means we are probably in GPOS and GSUB
> 
> The "probably" causes me a degree of anxiety here, what if we were wrong about this?

it's a strong probably, but just acknowledging that one can construct pathological examples of other tables that would fail packing.

rsheeter (approved):

PR #597, opened 2023-09-07
[write-fonts] Split PairPos format 2 subtables

cmyr:
Based on top of #595.

This implements subtable splitting for PairPos format 2 (class-based) subtables. It is based on the implementation in hb-repacker.

This is one of the more complicated tables to split, and writing this patch uncovered a number of other bugs in the existing implementation, which have been fixed separately.

With this patch, we are once again compiling our target fonts.

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -124,3 +124,40 @@ impl From<DeltaFormat> for i64 {
         value as u16 as _
     }
 }
+
+impl ClassDefFormat1<'_> {
+    /// Get the class for this glyph id
+    pub fn get(&self, gid: GlyphId) -> u16 {
+        if gid < self.start_glyph_id() {
+            return 0;
+        }
+        let idx = gid.to_u16() - self.start_glyph_id().to_u16();
+        self.class_value_array()
+            .get(idx as usize)
+            .map(|x| x.get())
+            .unwrap_or(0)
+    }
+}
+
+impl ClassDefFormat2<'_> {
+    /// Get the class for this glyph id
+    pub fn get(&self, gid: GlyphId) -> u16 {
+        self.class_range_records()
+            .iter()
+            .find_map(|record| {
+                (record.start_glyph_id() >= gid && record.end_glyph_id() <= gid)
+                    .then_some(record.class())
+            })
+            .unwrap_or(0)
+    }
+}
+
+impl ClassDef<'_> {
+    /// Get the class for this glyph id
+    pub fn get(&self, gid: GlyphId) -> u16 {
+        match self {
+            ClassDef::Format1(table) => table.get(gid),
+            ClassDef::Format2(table) => table.get(gid),
+        }
+    }
+}
--- a/write-fonts/src/graph/splitting.rs
+++ b/write-fonts/src/graph/splitting.rs
@@ -1,15 +1,14 @@
 //! splitting layout (GPOS) subtables
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
-use font_types::{GlyphId, Offset16};
+use font_types::{FixedSize, GlyphId, Offset16};
+use read_fonts::tables::{gpos as rgpos, layout as rlayout};
 
 use super::{Graph, ObjectId};
-
-use crate::types::FixedSize;
-use crate::{tables::layout as wlayout, write::TableData};
-
-use read_fonts::tables::{gpos as rgpos, layout as rlayout};
+use crate::{
+    tables::layout as wlayout, write::OffsetRecord, write::TableData, FontWrite, TableWriter,
+};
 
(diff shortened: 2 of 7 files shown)


write-fonts/src/graph/splitting.rs
```diff
@@ -188,6 +186,247 @@ fn split_off_ppf1(graph: &mut Graph, subtable: ObjectId, start: u16, end: u16) -
...
+    let class_def_1 = class_def_1.reparse::<rlayout::ClassDef>().unwrap();
+
+    let n_records = (end - start) + 1;
+    log::trace!("splitting off {n_records} class1records ({start}..={end})");
+
+    let class_map = coverage
+        .iter()
+        .filter_map(|gid| {
+            let glyph_class = class_def_1.get(gid);
+            (start..=end)
```
dfrg, line 335: I'm used to seeing a range with an exclusive end in Rust unless otherwise specified. I think this (specifically `(end - start) + 1` and `(start..=end)`) might be a bit more natural if we don't subtract one in the calling code?

write-fonts/src/graph/splitting.rs
```diff
@@ -188,6 +186,247 @@ fn split_off_ppf1(graph: &mut Graph, subtable: ObjectId, start: u16, end: u16) -
...
+        prev_subtable_start = subtable_start;
+        next_device_offset += offsets_used;
+        new_subtables.push(graph.add_object(new_subtable));
+    }
+
+    Some(new_subtables)
+}
+
+// returns the new table, + the number of non-null device offsets encountered.
+fn split_off_ppf2(
```
dfrg, line 313: This function seems to be a combination of some `PairPosFormat2::split_context_t` methods and `actuate_subtable_split`. I like the simplifications you've managed here but future readers might appreciate a couple comments with links to relevant portions of the HB code.

dfrg (approved):
Really nice work! This looks like a faithful implementation of the HarfBuzz logic (minus the shrinking?). The usual nits, comments and questions inline :)

write-fonts/src/graph/splitting.rs
cmyr: really good callout chad! There was 100% an off-by-one here, and in the process of investigating that I found two other straight-up mistakes (missing or incorrect fields) in the generated subtables.

write-fonts/src/graph/splitting.rs
```diff
@@ -188,6 +186,247 @@ fn split_off_ppf1(graph: &mut Graph, subtable: ObjectId, start: u16, end: u16) -
...
+            }
+        }
+
+        accumulated += accumulated_delta;
+        let largest_obj = coverage_size.max(class_def_1_size).max(class_def2_size);
+        let total = accumulated + coverage_size + class_def_1_size + class_def2_size
+            // largest obj packs last and can overflow
+            - largest_obj;
+
+        if total > MAX_TABLE_SIZE {
```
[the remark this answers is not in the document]
cmyr: in hb it's `>= 1 << 16`, which is equivalent to `> u16::MAX`.

read-fonts/src/tables/layout.rs
```diff
@@ -124,3 +124,40 @@ impl From<DeltaFormat> for i64 {
...
+            .unwrap_or(0)
+    }
+}
+
+impl ClassDefFormat2<'_> {
+    /// Get the class for this glyph id
+    pub fn get(&self, gid: GlyphId) -> u16 {
+        self.class_range_records()
+            .iter()
+            .find_map(|record| {
```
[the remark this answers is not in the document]
cmyr: good note, will open an issue to optimize this.

write-fonts/src/graph/splitting.rs
cmyr: yea there are some specific details of the harfbuzz implementation I haven't followed here; in particular harfbuzz shrinks the original table and retains it, whereas I just generate a full set of new tables. There are a few comments scattered about but I can try to rationalize that.

write-fonts/src/graph/splitting.rs
```diff
@@ -188,6 +186,247 @@ fn split_off_ppf1(graph: &mut Graph, subtable: ObjectId, start: u16, end: u16) -
...
+    for (idx, class1rec) in pp2.class1_records().iter().enumerate() {
+        let mut accumulated_delta = class1_record_size;
+        coverage_size += estimator.increment_coverage_size(idx as _);
+        class_def_1_size += estimator.increment_class_def_size(idx as _);
+        max_coverage_size = max_coverage_size.max(coverage_size);
+        max_class_def_1_size = max_class_def_1_size.max(class_def_1_size);
+
+        // NOTE:
+        //I'm finding that we generate slightly more splits than I would expect,
+        //and I want to look into that, but i also want to get this merged.
```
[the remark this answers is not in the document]
cmyr: The latter, although I haven't compared directly with harfbuzz.

Issue #593, opened 2023-09-07
[skrifa] CBDT/CBLC and SBIX bitmap color font access interfaces


drott:
For bitmap color font support in the Fontations `SkTypeface` backend, interfaces in Skrifa or read-fonts are needed to access the bitmap strikes of CBDT/CBLC and SBIX fonts. (In principle EBDT/EBLC might be needed but in practice I don't think we encounter such fonts so much, except for some Windows Chinese system fonts which are not high priority). 

I have not analyzed FreeType's support or interfaces or what `SkTypeface` does with them yet, but I suspect we might want some control over retrieving what strike sizes are available and selecting them. And/or and automatic mode might be useful where the next higher strike size or the highest available strike size is selected according to configured scaled font size.

.2 cents considerations regarding priority, I wonder if this make sense to do even before hinting and autohinting support. It depends on how much work this is. If it's a smaller amount of work, this might unblock me from starting the work on the Skia side so that work could start in parallel while hinting and autohinting is being worked on.

drott:
As a basis for discussion and invitation for feedback, here's my current thinking on how to hook up bitmap glyphs.

## Alternative 1: New type of `ColorGlyphCollection`

### Option a)

Instantiate SBIX for example, add format for SBIX to `ColorGlyphFormat`, then support calling `paint()` on it. Problem: `paint()` operates in font units, so it's ambiguous which strike size should be chosen. This could be solved by adding 
a getter to `ColorPainter`, for example `bitmap_output_ppem_size()` or `strike_size()` or similar. This would make the `ColorPainter` aware of what output strike size is expected. It could then find the nearest bigger or the max strike size bitmap, send a `push_transform()` with a scale to bridge between font unit space and 1:1 bitmap output space, and combine that neutralizing transform with a small scale transform if needed to go from the strike size found in the font to the requested output strike size. 

Add methods to ColorPainter for `paint_png...` and `paint_jpg...` or `paint_a8_bitmap_buf...` or some such. 

Other advantages: 
* Brings us closer to something like a `PaintBitmap` operation for COLRv2. 
* Allows us to address SBIX glyph placement logic and generally handle more of the glyph placement and scaling inside the library
Disadvantage: 
* Somewhat odd to ask the `ColorPainter` for the final size. 

### Option b)

[...]

PR #609, opened 2023-09-12
[read-fonts] initial codegen for parsing sbix table

dfrg:
Starting work on bitmap support and figured I'd tackle the low hanging fruit first. This includes basic codegen for parsing sbix along with a tiny helper method to load `GlyphData` for a given glyph id.

Working towards #593


resources/codegen_inputs/sbix.rs
```diff
@@ -0,0 +1,45 @@
...
+#[read_args(num_glyphs: u16)]
+#[tag = "sbix"]
+table Sbix {
+    /// Table version number — set to 1.
+    #[compile(1)]
+    version: u16,
+    /// Bit 0: Set to 1.
+    /// Bit 1: Draw outlines.
+    /// Bits 2 to 15: reserved (set to 0).
+    flags: u16,
```
cmyr, line 13: we might want to define a flags type for this?

cmyr (approved):
looks good, I think matching the glyf/loca API is the right idea (not that there is another clear alternative, so 🤷)

resources/codegen_inputs/sbix.rs
dfrg: Good call. The unnamed bit 0 that must always be set to 1 seems problematic for the codegen flags type. Wdyt about wrapping in a record instead and implementing default and appropriate methods for the “render outlines bit” instead?

cmyr: I think an easier option would be to use flags, but use the `compile_with` annotation to provide a fn that ensures that bit one is always set when writing.

dfrg: Ah, will do!

dfrg: Since I'm adding `compile_with`, went ahead and extended this for write support. Should be good to go.

PR #610, opened 2023-09-12
[write-fonts] Fix miscalculation of ppf2 subgraph size

cmyr:
I had run into a probley friday afternoon that I didn't have time to dig into, and I didn't want to block on getting fontc working again. This fixes that.

It turns out were incrementing the current index only when the subtable had not been seen before, which meant we would never advance the index and would not visit all subtables.

### DIFF
--- a/write-fonts/src/graph/splitting.rs
+++ b/write-fonts/src/graph/splitting.rs
@@ -204,6 +204,7 @@ fn split_off_ppf1(graph: &mut Graph, subtable: ObjectId, start: u16, end: u16) -
 // based off of
 // <https://github.com/harfbuzz/harfbuzz/blob/f380a32825a1b2c51bbe21dc7acb9b3cc0921f69/src/graph/pairpos-graph.hh#L207>
 fn split_pair_pos_format_2(graph: &mut Graph, subtable: ObjectId) -> Option<Vec<ObjectId>> {
+    // the minimum size of a format 2 subtable
     const BASE_SIZE: usize = 8 * u16::RAW_BYTE_LEN;
     let data = &graph.objects[&subtable];
 
@@ -279,12 +280,10 @@ fn split_pair_pos_format_2(graph: &mut Graph, subtable: ObjectId) -> Option<Vec<
         }
 
         accumulated += accumulated_delta;
-        let _largest_obj = coverage_size.max(class_def_1_size).max(class_def2_size);
-        let total = accumulated + coverage_size + class_def_1_size + class_def2_size;
-        //NOTE: harfbuzz has this optimization but I was seeing an overflow with
-        //the last object in some cases, so currently disabled
-        // largest obj packs last and can overflow
-        //- _largest_obj;
+        let largest_obj = coverage_size.max(class_def_1_size).max(class_def2_size);
+        let total = accumulated + coverage_size + class_def_1_size + class_def2_size
+        // largest obj packs last and can overflow (we only point to the start)
+        - largest_obj;
 
         if total > MAX_TABLE_SIZE {
             split_points.push(idx);
@@ -630,12 +629,10 @@ fn size_of_value_record_children(
         .filter_map(|offset| (!offset.is_null()).then_some(*offset.offset()))
         .map(|_| {
             let obj = offsets[*next_offset_idx].object;
-            if seen.insert(obj) {
-                *next_offset_idx += 1;
-                graph.objects[&obj].bytes.len()
-            } else {
-                0
-            }
+            *next_offset_idx += 1;
+            seen.insert(obj)
+                .then(|| graph.objects[&obj].bytes.len())
+                .unwrap_or(0)
         })
         .sum()
 }
@@ -668,7 +665,7 @@ mod tests {
                 Class1Record, Class2Record, PairPos, PairSet, PairValueRecord, PositionLookup,
                 ValueRecord,
             },
-            layout::{CoverageTableBuilder, DeviceOrVariationIndex, VariationIndex},
+            layout::{CoverageTableBuilder, Device, DeviceOrVariationIndex, VariationIndex},
         },
         FontWrite, TableWriter,
     };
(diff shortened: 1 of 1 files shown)


write-fonts/src/graph/splitting.rs
```diff
@@ -279,12 +280,10 @@ fn split_pair_pos_format_2(graph: &mut Graph, subtable: ObjectId) -> Option<Vec<
         }
 
         accumulated += accumulated_delta;
-        let _largest_obj = coverage_size.max(class_def_1_size).max(class_def2_size);
-        let total = accumulated + coverage_size + class_def_1_size + class_def2_size;
-        //NOTE: harfbuzz has this optimization but I was seeing an overflow with
-        //the last object in some cases, so currently disabled
+        let largest_obj = coverage_size.max(class_def_1_size).max(class_def2_size);
+        let total = accumulated + coverage_size + class_def_1_size + class_def2_size
         // largest obj packs last and can overflow
```
rsheeter, line 285: I think explaining why, such as because we only need to be able to point to the start and never the end, as well

write-fonts/src/graph/splitting.rs
```diff
@@ -630,12 +629,9 @@ fn size_of_value_record_children(
...
             let obj = offsets[*next_offset_idx].object;
-            if seen.insert(obj) {
-                *next_offset_idx += 1;
-                graph.objects[&obj].bytes.len()
-            } else {
-                0
-            }
+            let seen = seen.insert(obj);
+            *next_offset_idx += 1;
+            seen.then(|| graph.objects[&obj].bytes.len()).unwrap_or(0)
```
rsheeter, line 634: I dont' immediately see why you need to bind seen, wouldn't this be fine as:

```rust
            let obj = offsets[*next_offset_idx].object;
            *next_offset_idx += 1;
            seen.insert(obj)
                    .then(|| graph.objects[&obj].bytes.len())
                    .unwrap_or(0)
```

Hm. Actually your `map` always takes an offset and increments next which sounds a lot like iteration so just zip?

```rust
    subtables
        .iter()
        .filter_map(|offset| (!offset.is_null()).then_some(*offset.offset()))
       .zip(offsets.iter()).
        .map(|(offset, _)| {
            seen.insert(offset.object)
                    .then(|| graph.objects[&offset.object].bytes.len()).unwrap_or(0)
        })
        .sum()
```

write-fonts/src/graph/splitting.rs
```diff
@@ -1066,4 +1063,81 @@ mod tests {
             .sum();
         assert_eq!(total_c2recs, expected_n_c2_recs);
     }
+
+    #[test]
+    fn size_of_value_record_children_sanity() {
+        // let's have  single class1class, and three class2 classes
+        // we want a duplicate varidx, a null varidx, and a device table?
+
+        fn make_val_record(
```
rsheeter, line 1072: Optional, this would be easier to read hoisted out like make_class_def. 

It's also hard to tell what the args are in the code. What if it took only the advance so this:

```rust
                make_val_record(5, None),
                make_val_record(6, DeviceOrVariationIndex::variation_index(4, 20)),
```

read more like this:

```rust
                val_with_xadvance(5),
                val_with_xadvance(6).with_x_advance_device(DeviceOrVariationIndex::variation_index(4, 20)),
```

make_class_def is similar, what's a class def 3,3,1?!

write-fonts/src/graph/splitting.rs
```diff
@@ -1066,4 +1063,81 @@ mod tests {
...
+        let ppf2_data = &graph.objects[&root_id];
+        let ppf2 = ppf2_data.reparse::<rgpos::PairPosFormat2>().unwrap();
+        assert_eq!(ppf2.class1_records().len(), 1);
+        let c1rec = ppf2.class1_records().get(0).unwrap();
+        let mut visited = HashSet::new();
+        let mut next_device_offset = 3;
+        assert_eq!(c1rec.class2_records.len(), 3);
+
+        // a little helper so we don't have to have this huge fn call in each assert
+        let mut test_closure = |record: &rgpos::ValueRecord| -> usize {
```
rsheeter, line 1122: "test_closure" is perhaps nto the most descriptive name, size_of_children would make the tests read much more clearly

write-fonts/src/graph/splitting.rs
```diff
@@ -1066,4 +1063,81 @@ mod tests {
...
+        assert_eq!(test_closure(c2rec1.value_record1()), 0, "no subtables");
+        assert_eq!(test_closure(c2rec1.value_record2()), 6, "one new varidx");
+        let c2rec2 = c1rec.class2_records().get(1).unwrap();
+        assert_eq!(test_closure(c2rec2.value_record1()), 10, "a device table");
+        assert_eq!(test_closure(c2rec2.value_record2()), 0, "duplicate table");
+        let c2rec3 = c1rec.class2_records().get(2).unwrap();
+        assert_eq!(test_closure(c2rec3.value_record1(),), 6, "new varidx table");
+        assert_eq!(test_closure(c2rec3.value_record2(),), 0, "a null offset");
+        assert_eq!(next_device_offset, 7, "we visited all offsets");
+    }
```
rsheeter, line 1142: IMHO this would be nicer as one assert that there are the values. Less noisy bindings, error-prone get(index) calls and no need to check we visited everything.

```rust
assert_eq!(vec![
  (0, 6),  // no subtables, one new var idx
  (10, 0),// a device table with a duplicate table
  (6, 0),  // a new varidx table with a null offset
],
c1rec.class2_records().iter().map(|r| (r.value_record1(), r.value_record2()).collect::<Vec<_>>()
);
```

rsheeter (approved):
LGTM subject to comments being addressed

write-fonts/src/graph/splitting.rs
cmyr: This is defined inline since it's only used in this one test, whereas make_class_def is shared. :)

write-fonts/src/graph/splitting.rs
cmyr: huh interesting, I find personally find that slightly less clear? I prefer fine-grained asserts because it makes explicit exactly what has failed.

The check for the value of `next_device_offset` is important, because this we want to ensure that this value is incremented correctly, e.g. it is incremented for each non-null offset in each value record. we visit.

rsheeter: The single assert tells you when lots of values were wrong, when you have the wrong number of values, and makes it much harder to `get(the wrong value)`. I would argue it's equivalent in terms of telling you what failed while giving greater context (were ALL the values borked? just one?) but I'll grant you it's stylistic so you could view this one as optional.

PR #612, opened 2023-09-13
[write-fonts] Graphviz improvements

cmyr:
This adds the ability to set an env var ('FONTC_PRUNE_GRAPH') that tells write-fonts to prune a generated dotviz file to only include spaces that contain an overflow. This was very helpful for debugging our PairPos format 2 packing failures.

It also cleanups up our graphviz code to not use subgraphs, which weren't adding much if anything.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -1169,7 +1169,10 @@ impl Graph {
 
     #[cfg(feature = "dot2")]
     pub(crate) fn write_graph_viz(&self, path: impl AsRef<std::path::Path>) -> std::io::Result<()> {
-        graphviz::GraphVizGraph::from_graph(self).write_to_file(path)
+        // if this is set then we prune the generated graph
+        const PRUNE_GRAPH_ENV_VAR: &str = "FONTC_PRUNE_GRAPH";
+        let try_trim_graph = std::env::var_os(PRUNE_GRAPH_ENV_VAR).is_some();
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -1169,7 +1169,10 @@ impl Graph {
 
     #[cfg(feature = "dot2")]
     pub(crate) fn write_graph_viz(&self, path: impl AsRef<std::path::Path>) -> std::io::Result<()> {
-        graphviz::GraphVizGraph::from_graph(self).write_to_file(path)
+        // if this is set then we prune the generated graph
+        const PRUNE_GRAPH_ENV_VAR: &str = "FONTC_PRUNE_GRAPH";
```
anthrotype, line 1173: (bikeshedding) I feel the var name ties too much write-fonts with fontc.. conceivable another tool might find it useful.. maybe WRITE_FONTS_PRUNE_GRAPH

cmyr: I was thinking that `WRITE_` was confusing as a prefix, as it's a verb. Given that we own both projects, and that this is undocumented/private API, I thiiiink I prefer this version? open to more ideas though..

anthrotype: `FONTATIONS_PRUNE_GRAPH`

anthrotype: but it's fine like that, since it's undocumented/private api anyway

Issue #614, opened 2023-09-14
gvar: use shared point numbers


cmyr:
Another thing that may be causing our gvar table to not match fonttools: we are not currently never opting to write the 'shared point numbers' that may optionally follow the GlyphVariationData header, and which can be shared between multiple tuple variations.

Issue #392, opened 2023-05-09
Add (optional) serde support to write-fonts


cmyr:
I would like to be able to write tests for write-fonts that involve attempting to serialize specific real-world failures, and the best way I can think of doing this is to add optional serde support; this way I can save some specific table to disk and then load and compile it in tests.

This will require also adding (optional) serde support to font-types; see #311.

PR #616, opened 2023-09-19
Only use IUP representation when more compact

### LINKED ISSUE #568: should prefer non-IUP-optimized deltas when more compact
as I noted in https://github.com/googlefonts/fontations/issues/564#issuecomment-1671634698

> fontTools.varLib doesn't always write out the gvar tuples as they come out of iup_delta_optimize, but only when doing so produces fewer bytes (not just fewer explicit deltas):
https://github.com/fonttools/fonttools/blob/57fb47de3a41c724c3bf1051d00715e08d86141e/Lib/fontTools/varLib/__init__.py#L370-L382

fontc (or write-fonts?) should try to do something like that, but am not sure how best to do this

cmyr:
This fixes #568; when presented with deltas that include interpolatable values, we will generate both representations and then choose the more compact.

This is mostly straightforward, but there is some API to bikeshed, particularly around the `iup_delta_optimize` API.

Previously, this returned the deltas as `Vec<Option<Vec2>>`, with `None` values representing deltas that could be interpolated. This meant that we did not have access to these deltas, which are necessary in order for us to generate the non-IUP representation.

We have some options here. The simplest one would be to return the `GlyphDelta` type that we use when building gvar; the only concern here is that this type represents a point as`(i16, i16)`, where we are currently using `Vec2`. I personally think this is fine, since as far as I can tell we do no subsequent processing between when we check IUP and when we compile the final binary, but I want to minimize changes for the time being.

Another option would be to add another new type that is basically `(Vec2, bool)`; a third option is to just return `Vec<bool>` with an entry for each input item, indicating if it can be interpolated.

I chose this last option, but after trying to integrate into `fontc` it feels bad; fontc passes ownership of the deltas to `iup_delta_optimize`, which means we need an additional clone of all the deltas in `fontc` in order to convert from our bools back to some useful representation.

Finally there's an additional headache; if we define a type in `write-fonts` we can't easily store that type in `fontc` contexts because `write-fonts` types do not implement the serde traits; this has come up before and so maybe is another argument for merging #392.

tl;dr: life is easiest of `iup_delta_optimize` returns `GlyphDelta` directly, and if this type implements `Serialize` and `Deserialize`?

On a happy note, if I use this patch in fontc than the gvar diff for our fonts drops dramatically, and it looks like the only remaining difference would be fixed by #614.

progress on https://github.com/googlefonts/fontc/issues/422

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -26,7 +26,24 @@ pub struct GlyphDeltas {
     // start and end tuples of optional intermediate region
     intermediate_region: Option<(Tuple, Tuple)>,
     // (x, y) deltas or None for do not encode. One entry per point in the glyph.
-    deltas: Vec<Option<(i16, i16)>>,
+    deltas: Vec<GlyphDelta>,
+}
+
+/// A delta for a single value in a glyph.
+///
+/// This includes a flag indicating whether or not this delta is required (i.e
+/// it cannot be interpolated from neighbouring deltas and coordinates).
+/// This is only relevant for simple glyphs; interpolatable points may be omitted
+/// in the final binary when doing so saves space.
+/// See <https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#inferred-deltas-for-un-referenced-point-numbers>
+/// for more information.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct GlyphDelta {
+    pub x: i16,
+    pub y: i16,
+    /// This delta must be included, i.e. cannot be interpolated
+    pub required: bool,
 }
 
 /// An error representing invalid input when building a gvar table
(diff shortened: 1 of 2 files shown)


anthrotype:
I also think it would be fine if iup_delta_optimize API returns GlyphDeltas (which carry along the required flag), it's fine if they get rounded to integer as the caller doesn't need to process them further. And I don't like too much that we need to clone the deltas before calling iup_delta_optimize. Could we pass a reference or slice instead?

If we decide to return rounded GlyphDeltas, maybe the rounding method itself could be a parameter. Right now I think we don't match fonttools, which instead of the usual OtRound in the case of rounding of deltas it uses the built-in Python round() function (which uses [the Banker's method](https://docs.rs/rust_decimal/latest/rust_decimal/enum.RoundingStrategy.html#variant.MidpointNearestEven) which is more statistically robust), but I am digressing and this belong to a separate issue..

cmyr:
I've added serde and updated this on top of it to return `GlyphDelta`s, and I think it's much cleaner overall.

PR #624, opened 2023-09-25
Also build format 12 subtables in Cmap::from_mappings if any char > BMP

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/451

This is intended to match fontmake/fonttools output.
If the input mappings contain any characters beyond the basic multilingual plane (> 0xFFFF) we also build two cmap12 subtables (one for Unicode platform, another identical one for WIndows platform, both pointing to the same format 12 subtable) which will be a superset of the BMP-only cmap4 subtables.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -27,6 +27,7 @@ fn size_of_cmap12(num_groups: u32) -> u32 {
 
 impl CmapSubtable {
     /// Create a new format 4 `CmapSubtable` from a list of `(char, GlyphId)` pairs.
+    ///
     /// The pairs are expected to be already sorted by codepoint.
     /// Characters beyond the BMP are ignored. If all characters are beyond the BMP
     /// then `None` is returned.
@@ -116,6 +117,7 @@ impl CmapSubtable {
     }
 
     /// Create a new format 12 `CmapSubtable` from a list of `(char, GlyphId)` pairs.
+    ///
     /// The pairs are expected to be already sorted by chars.
     /// In case of duplicate chars, the last one wins.
     fn create_format_12(mappings: &[(char, GlyphId)]) -> Self {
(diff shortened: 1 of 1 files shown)


anthrotype:
> intended to match fontmake/fonttools output

in reality this PR differs slightly in that it will _not_ produce empty cmap format 4 subtables in the unlikely scenario the input mappings _only_ contain characters beyond the BMP, which is what fontmake would do in this case, but an empty cmap subtable should be as good as an altogether absent one...

write-fonts/src/tables/cmap.rs
```diff
@@ -4,78 +4,45 @@
...
-            entry_selector,
-            range_shift,
-            end_code,
-            start_code,
-            id_deltas,
-            id_range_offsets,
-            vec![], // becauseour idRangeOffset's are 0 glyphIdArray is unused
-        ))
-    }
+fn size_of_cmap12(num_groups: u32) -> u32 {
```
rsheeter, line 23: Optional, in re-reading I tend to think both size_of fns should have a link to the cmap spec

anthrotype: Done

write-fonts/src/tables/cmap.rs
```diff
@@ -105,27 +74,148 @@ impl Cmap {
...
+                bmp_subtable.clone(),
+            ));
+            win_records.push(EncodingRecord::new(
+                PlatformId::Windows,
+                WINDOWS_BMP_ENCODING,
+                bmp_subtable,
+            ));
+        }
+
+        if mappings.iter().any(|(cp, _)| *cp > '\u{FFFF}') {
```
rsheeter, line 198: `> 0xFFFF` is meaningless unless you know unicode; I suggest a function is-bmp, perhaps as a trait similar to ot_round so this becomes something like `.any(|(cp, _)| !cp.is_basic_multilingual_plan())`

rsheeter (approved):

write-fonts/src/tables/cmap.rs
anthrotype: I'll add a comment. I don't think that trait would be used much except here for cmap

PR #626, opened 2023-09-26
[skrifa] update metrics scaling to match FreeType

### LINKED ISSUE #590: [skrifa] Small advance differences in Noto Sans between FT and Fontations
I haven't reduced this to FT calls vs. Fontations calls yet, but in my comparison of Skia backends I get:

```
fontations: 
gid: 36 advance: 15.33600044250488281250 gid: 68 advance: 13.46399974822998046875 gid: 47 advance: 12.57600021362304687500 gid: 79 advance: 6.19199991226196289062 
ft: 
gid: 36 advance: 15.33595275878906250000 gid: 68 advance: 13.46395874023437500000 gid: 47 advance: 12.57595825195312500000 gid: 79 advance: 6.19198608398437500000 
```

For Noto Sans which triggers in the tests that I am developing for path and metrics equality.

Both `SkFont` objects are configured in Skia as follows:

```
                font.setSize(24);
                font.setEdging(SkFont::Edging::kSubpixelAntiAlias);
                font.setSubpixel(true);
                font.setHinting(SkFontHinting::kNone);
```

dfrg:
Updates the `Metrics` and `GlyphMetrics` types to use an intermediate fixed point calculation for scaling that matches FreeType.

Fixes #590

### DIFF
--- a/skrifa/src/instance.rs
+++ b/skrifa/src/instance.rs
@@ -1,5 +1,7 @@
 //! Helpers for selecting a font size and location in variation space.
 
+use read_fonts::types::Fixed;
+
 use crate::small_array::SmallArray;
 
 /// Type for a normalized variation coordinate.
@@ -48,6 +50,15 @@ impl Size {
             self.0 / units_per_em as f32
         } else {
             1.0
+        }
+    }
+
+    /// Computes a fixed point linear scale factor that matches FreeType.
+    pub(crate) fn fixed_linear_scale(self, units_per_em: u16) -> Fixed {
+        if self.0 > 0.0 && units_per_em > 0 {
+            Fixed::from_bits((self.0 * 64.) as i32) / Fixed::from_bits(units_per_em as i32)
+        } else {
+            Fixed::from_bits(0x10000 * 64)
         }
     }
 }
--- a/skrifa/src/instance.rs
+++ b/skrifa/src/instance.rs
@@ -55,9 +55,18 @@ impl Size {
 
     /// Computes a fixed point linear scale factor that matches FreeType.
     pub(crate) fn fixed_linear_scale(self, units_per_em: u16) -> Fixed {
+        // FreeType computes a 16.16 scale factor that converts to 26.6.
+        // This is done in two steps, assuming use of FT_Set_Pixel_Size:
+        // 1) height is multiplied by 64:
+        //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3596>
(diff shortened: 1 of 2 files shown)


dfrg:
To avoid adding clutter, this doesn't address the issue that the `Size` type can't represent a ppem of 0. Punted to #627

skrifa/src/instance.rs
```diff
@@ -50,6 +52,15 @@ impl Size {
             1.0
         }
     }
+
+    /// Computes a fixed point linear scale factor that matches FreeType.
+    pub(crate) fn fixed_linear_scale(self, units_per_em: u16) -> Fixed {
```
rsheeter, line 57: Would be nice to link to the corresponding FT code and explain the rationale if it's not obvious from the corresponding FT code. Currently it's a wee bit magical.

skrifa/src/metrics.rs
```diff
@@ -545,6 +565,29 @@ mod tests {
         assert_eq!(expected, &result[..]);
     }
```
rsheeter, line 567: We changed all that scaling and the only test we need is that Size::unscaled() and Size::new(upem) are the same?

Shouldn't there be a unit test that we fail without your modifications to prove the scaling works?

rsheeter: At the very least a test of some scale factor applications?

rsheeter (approved):
LGTM subject to tests to prove it works. Naively I'd hve expected at least a unit test with some same scales that are confirmed to match FreeType where previously they did not.

skrifa/src/instance.rs
dfrg: Wee bit is probably an understatement :) Will do.

skrifa/src/metrics.rs
dfrg: Oops, I forgot to note that I’m going to change the glyph extraction script to grab advances too so that the tests are data driven. I’ll do a follow up for that. 

I’ll add an additional unit test here that specifically tests the scaling code with the values reported in the issue.

skrifa/src/metrics.rs
```diff
@@ -116,29 +116,29 @@ impl Metrics {
             ..Default::default()
         };
         let coords = location.into().coords();
-        let scale = size.linear_scale(metrics.units_per_em);
+        let scale = ScaleFactor(size.fixed_linear_scale(metrics.units_per_em));
```
drott, line 119: Nit: Could the name `scale` and `ScaleFactor` include a hint that we're talking about `Fixed` type computations?

dfrg: Indeed. I’ll update the names

PR #628, opened 2023-09-26
Use 'shared point numbers' in gvar

### LINKED ISSUE #614: gvar: use shared point numbers

cmyr:
in gvar you can optionally store on set of point numbers that are referenced by multiple regions for the same glyph; we were not previously using this very efficiently.

- based off of #616
- fixes #614

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -27,7 +27,7 @@ pub struct GlyphDeltas {
     intermediate_region: Option<(Tuple, Tuple)>,
     // (x, y) deltas or None for do not encode. One entry per point in the glyph.
     deltas: Vec<GlyphDelta>,
-    point_numbers: PackedPointNumbers,
+    best_point_packing: PackedPointNumbers,
 }
 
 /// A delta for a single value in a glyph.
@@ -225,20 +225,19 @@ impl GlyphVariations {
         for deltas in &self.variations {
             // for each set points, get compiled size + number of occurances
             let (_, count) = point_number_counts
-                .entry(&deltas.point_numbers)
+                .entry(&deltas.best_point_packing)
                 .or_insert_with(|| {
-                    let size = deltas.point_numbers.compute_size();
+                    let size = deltas.best_point_packing.compute_size();
                     (size as usize, 0usize)
                 });
             *count += 1;
         }
 
         // find the one that saves the most bytes
-        let (pts, (size, count)) = point_number_counts
+        let (pts, (_, count)) = point_number_counts
             .into_iter()
             .filter(|(_, (_, count))| *count > 1)
             .max_by_key(|(_, (size, count))| *count * *size)?;
-        log::trace!("max shared pts ({size}B, {count}n), {pts:?}");
 
         // no use sharing points if they only occur once
         (count > 1).then(|| pts.to_owned())
@@ -276,6 +275,10 @@ impl GlyphDelta {
     pub fn optional(x: i16, y: i16) -> Self {
         Self::new(x, y, false)
     }
+
+    fn to_tuple(self) -> (i16, i16) {
+        (self.x, self.y)
+    }
 }
 
 impl GlyphDeltas {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/gvar.rs
```diff
@@ -197,19 +198,63 @@ impl GlyphVariations {
...
+    ///
+    /// If multiple tuple variations for a given glyph use the same point numbers,
+    /// it is possible to store this in the glyph table, avoiding duplicating
+    /// data.
+    ///
+    /// This implementation is currently based on the one in fonttools, where it
+    /// is part of the compileTupleVariationStore method:
+    /// <https://github.com/fonttools/fonttools/blob/0a3360e52727cdefce2e9b28286b074faf99033c/Lib/fontTools/ttLib/tables/TupleVariation.py#L641>
+    ///
+    /// # Note
```
rsheeter, line 211: Suggest moving the Note to an issue; buried in comments it's basically lost

write-fonts/src/tables/gvar.rs
```diff
@@ -197,19 +198,63 @@ impl GlyphVariations {
...
+                    (size as usize, 0usize)
+                });
+            *count += 1;
+        }
+
+        // find the one that saves the most bytes
+        let (pts, (_, count)) = point_number_counts
+            .into_iter()
+            .filter(|(_, (_, count))| *count > 1)
+            .max_by_key(|(_, (size, count))| *count * *size)?;
```
behdad, line 240: Multiply by `(*count - 1)` here?

behdad:
LGTM except for the one comment.

write-fonts/src/tables/gvar.rs
```diff
@@ -197,19 +198,63 @@ impl GlyphVariations {
...
+
+        // no use sharing points if they only occur once
+        (count > 1).then(|| pts.to_owned())
```
rsheeter, line 243: Maybe I'm misreading the sea of brackets and underscroes but it seems like you filter for count > 1 and return None otherwise so this has to be true, the comment should move to the filter and this should just say pts.to_owned()?

write-fonts/src/tables/gvar.rs
```diff
@@ -688,4 +760,158 @@ mod tests {
...
+        assert_eq!(
+            built.per_tuple_data[0].private_point_numbers,
+            Some(PackedPointNumbers::All)
+        );
+        assert_eq!(
+            built.per_tuple_data[1].private_point_numbers,
+            Some(PackedPointNumbers::Some(vec![1, 3]))
+        );
+    }
 }
```
rsheeter, line 917: compute_shared_points explicitly avoids sharing points that only occur once, perhaps there should be a direct test of this?

write-fonts/src/tables/gvar.rs
```diff
@@ -248,80 +297,102 @@ impl GlyphDeltas {
...
-            GlyphTupleVariationData {
-                private_point_numbers: None,
-                x_deltas: PackedDeltas::new(x_deltas),
-                y_deltas: PackedDeltas::new(y_deltas),
-            }
+    // this is a type method just to expose it for testing, we call it before
+    // we finish instantiating self.
+    //
+    // NOTE: we do a lot of duplicate work here with creating & throwing away
+    // buffers, and that can be improved at the cost of a bit more complexity
```
rsheeter, line 317: Issue please, comments get lost.

write-fonts/src/tables/gvar.rs
```diff
@@ -688,4 +760,158 @@ mod tests {
...
+        let built = variations.build(&tups);
+        assert_eq!(
+            built.per_tuple_data[0].private_point_numbers,
+            Some(PackedPointNumbers::All)
+        );
+        assert_eq!(
+            built.per_tuple_data[1].private_point_numbers,
+            Some(PackedPointNumbers::Some(vec![1, 3]))
+        );
+    }
```
rsheeter, line 916: This looks suspiciously like many tests shoved into one because they share test data. Make a helper to provide the data and split up the tests?

rsheeter (approved):
LGTM subject to comments being addressed.

write-fonts/src/tables/gvar.rs
cmyr: opened as #634

write-fonts/src/tables/gvar.rs
cmyr: you're right, good catch

write-fonts/src/tables/gvar.rs
cmyr: Opened as #635

write-fonts/src/tables/gvar.rs
cmyr: yes fair. this test is noisy because I was using it to debug something. The only thing we really care about here is the very last assertions, that we use all the points for the first tuple but sparse points for the second; all the intermediate stuff is just leftovers from my debugging, can clean that up.

PR #630, opened 2023-09-26
[read-fonts] Fix possible crash in PackedPointNumbersIter

cmyr:
This would only be encountered if the user manually tried to iterate packed points with a single 0x0 byte (signifying that there are points for all deltas) and did not also know how many points there were; but there's no reason not to be more careful here and at least stop iterating after u16::MAX items.

JMM

### DIFF
--- a/read-fonts/src/tables/variations.rs
+++ b/read-fonts/src/tables/variations.rs
@@ -362,7 +362,7 @@ impl Iterator for PackedPointNumbersIter<'_> {
         // if our count is zero, we keep incrementing forever
         if self.count == 0 {
             let result = self.last_val;
-            self.last_val += 1;
+            self.last_val = self.last_val.checked_add(1)?;
             return Some(result);
         }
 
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/variations.rs
```diff
@@ -362,7 +362,7 @@ impl Iterator for PackedPointNumbersIter<'_> {
         // if our count is zero, we keep incrementing forever
         if self.count == 0 {
             let result = self.last_val;
-            self.last_val += 1;
+            self.last_val = self.last_val.checked_add(1)?;
```
rsheeter, line 365: Why no test?

rsheeter (approved):

read-fonts/src/tables/variations.rs
cmyr: To take the question at face value: I didn't add a test here because I don't see it adding value that exceeds the costs; it would be hard to regress this, and in the case where you did you'd still never find this crash during normal use. (The condition where a crash is possible (exhausting the iterator) can only be encountered when misusing this API, because you need to know how many points exist in the glyph in order to make sense of the gvar data.)

To take the question as an indirect request for a test case: sure, I'll happily add a test case :)

PR #631, opened 2023-09-26
[write-fonts] iup: fixup vector distance calculation

cmyr:
We were checking the axes independently, instead of checking the overall length of the vector.


rsheeter (approved):

behdad:
Nit: If you use hypot2 or equivalent you can save a sqrt. LGTM otherwise.

behdad:
> Nit: If you use hypot2 or equivalent you can save a sqrt. LGTM otherwise.

Should it also do the `powi(2)`  once outside the loop?

Issue #633, opened 2023-09-27
Move VariationStoreBuilder class from fea-rs to write-fonts


anthrotype:
Colin made this for building variable GDEF/GPOS:

https://github.com/cmyr/fea-rs/blob/main/fea-rs/src/compile/tables/var_store.rs

but since it's going to be useful for other non-FEA realted tables like HVAR or COLR I think we would like to move to write-fonts so fontc can use it directly

PR #636, opened 2023-09-27
Copy fea-rs VariationStoreBuilder to write-fonts

anthrotype:
and make it `pub` so that fontc can use it to build stuff like HVAR or COLR

Fixes https://github.com/googlefonts/fontations/issues/633

fea-rs var_store.rs module was copied from this commit:
https://github.com/cmyr/fea-rs/blob/497f51c66530f0b685646fae1db192502cdab1fe/fea-rs/src/compile/tables/var_store.rs

### DIFF
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -20,6 +20,7 @@ log = "0.4"
 kurbo = "0.9.4"
 dot2 = { version = "1.0", optional = true }
 serde = { version = "1.0", features = ["derive"], optional = true }
+indexmap = "2.0"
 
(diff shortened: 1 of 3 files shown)


rsheeter (approved):
LGTM based on the assumption you moved it w/o significant changes

anthrotype:
> based on the assumption you moved it w/o significant changes

first commit literally copied from https://github.com/cmyr/fea-rs/blob/497f51c66530f0b685646fae1db192502cdab1fe/fea-rs/src/compile/tables/var_store.rs

following commits simply adjust the import paths and make it pub

PR #637, opened 2023-09-27
[read-fonts] parse embedded bitmap tables

dfrg:
This can now read Noto Color Emoji bitmaps with code like the following:
```rust
    let data = std::fs::read("path/to/NotoColorEmoji.ttf").unwrap();
    let font = FontRef::from_index(&data, 0).unwrap();
    let ch = '🫶';
    let ch = '✨';
    let ch = '\u{2764}';
    let cmap = font.cmap().unwrap();
    let cblc = font.cblc().unwrap();
    let cbdt = font.cbdt().unwrap();
    let gid = cmap.map_codepoint(ch).unwrap();
    let size = &cblc.bitmap_sizes()[0];
    let location = size.location(cblc.offset_data(), gid).unwrap();
    let bitmap = cbdt.data(&location).unwrap();
    std::fs::write("emoji_image.png", bitmap.data).unwrap();
```


font-codegen/src/fields.rs
```diff
@@ -402,7 +402,7 @@ fn traversal_arm_for_field(
             quote!(Field::new(#name_str, traversal::FieldType::var_array(#typ_str, self.#name()#maybe_unwrap, #data)))
         }
         // HACK: who wouldn't want to hard-code ValueRecord handling
```
rsheeter, line 404: I know it's not new but lets change // HACK to // issue link to fix the value record hacks

font-codegen/src/fields.rs
```diff
@@ -475,10 +475,15 @@ impl Field {
     }
 
     pub(crate) fn is_zerocopy_compatible(&self) -> bool {
-        matches!(
-            self.typ,
-            FieldType::Scalar { .. } | FieldType::Offset { .. }
-        )
+        // hack: we want to add `FieldType::Struct` here but don't want to
+        // catch `ValueRecord` so use this attribute to ignore it.
+        // Fields that require args for reading can't be read "zerocopy"
```
rsheeter, line 480: I'm not quite sure what meaning to take from this. Do you mean anything with a variable number of something has to be copied? - I didn't think that was true.

font-codegen/src/record.rs
```diff
@@ -52,10 +52,15 @@ pub(crate) fn generate(item: &Record, all_items: &Items) -> syn::Result<TokenStr
         }
     });
     let maybe_impl_read_with_args = (has_read_args).then(|| generate_read_with_args(item));
-    let maybe_extra_traits = item
+    let mut maybe_extra_traits = item
         .gets_extra_traits(all_items)
         .then(|| quote!(PartialEq, Eq, PartialOrd, Ord, Hash));
-
+    if item.name == "SbitLineMetrics" {
```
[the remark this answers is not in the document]
rsheeter: Fix as in remove hardcoded typespecific behavior

font-test-data/test_data/ttx/embedded_bitmaps.ttx
```diff
@@ -0,0 +1,313 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rsheeter, line 1: I feel like some commentary on how this was created and what's in it might be helpful? - at a glance it's not super clear what's here or what the significance of each glyph is.

Glyphnames that were explanatory would go a long way, for example "noto_emoji_style_cbdt" would be more informative than "numbersign"

font-test-data/test_data/ttx/embedded_bitmaps.ttx
```diff
@@ -0,0 +1,313 @@
...
+    <strikedata index="0">
+      <cbdt_bitmap_format_17 name="numbersign">
+        <SmallGlyphMetrics>
+          <height value="128"/>
+          <width value="136"/>
+          <BearingX value="0"/>
+          <BearingY value="101"/>
+          <Advance value="136"/>
+        </SmallGlyphMetrics>
+        <rawimagedata>
```
rsheeter, line 203: is this a real image? Does it need to be so long or could it be a much shorter fragment of example data?

font-codegen/src/fields.rs
dfrg: It means if we enable this for ValueRecord it breaks everything. I’ll add an issue for addressing codegen hacks around embedded records but I’ll have to defer to @cmyr about what the best approach might be.

dfrg: It means if we enable this for ValueRecord it breaks everything. I’ll add an issue for addressing codegen hacks around embedded records but I’ll have to defer to @cmyr about what the best approach might be.

read-fonts/src/tables/cbdt.rs
```diff
@@ -0,0 +1,124 @@
...
+                height: 128,
+                width: 136,
+                bearing_x: 0.into(),
+                bearing_y: 101.into(),
+                advance: 136,
+            },
+        )];
+        for (gid, data, metrics) in expected {
+            let location = size.location(cblc.offset_data(), *gid).unwrap();
+            // all glyphs have data format == 17
```
rsheeter, line 106: this says exactly what the line below says, an explanation of the significance might be more helpful?

font-test-data/test_data/ttx/embedded_bitmaps.ttx
dfrg: Good point. I’ll make them after the formats they cover.

read-fonts/src/tables/cbdt.rs
```diff
@@ -0,0 +1,124 @@
...
+        let cblc = font.cblc().unwrap();
+        let cbdt = font.cbdt().unwrap();
+        let size = &cblc.bitmap_sizes()[0];
+        // Metrics for size at index 0
+        assert_eq!(size.hori.ascender(), 101);
+        assert_eq!(size.hori.descender(), -27);
+        assert_eq!(size.hori.width_max(), 136);
+        assert_eq!(size.vert.ascender(), 101);
+        assert_eq!(size.vert.descender(), -27);
+        assert_eq!(size.vert.width_max(), 136);
```
rsheeter, line 30: Optional, I prefer the half dozen or so asserts here collapsed to one

font-test-data/test_data/ttx/embedded_bitmaps.ttx
dfrg: It is real. I thought having a real PNG would make sense but now realize that it doesn’t matter. I can shorten this.

rsheeter (approved):

font-codegen/src/fields.rs
dfrg: #659 for tracking all codegen hacks

font-test-data/test_data/ttx/embedded_bitmaps.ttx
dfrg: Done!

font-test-data/test_data/ttx/embedded_bitmaps.ttx
dfrg: I was worried that ttx might validate the PNG but it appears not. Truncated to the first 8 bytes.

read-fonts/src/tables/cbdt.rs
dfrg: There's no real significance, so I just removed the redundant comment.

read-fonts/src/tables/cbdt.rs
dfrg: Done

dfrg:
#660 for missing format tests

font-codegen/src/fields.rs
cmyr: To clarify there are more issues than just `ValueRecord`; the issue is *any* record that is not fixed size, of which there are about ten (including  things like `PairValueRecord` `Tuple`, `VariationRegion`, `Mark2Record`, etc).

I also think that this change is incorrect, since a struct can only be included as a field of another record *if* it is copy, which this does not guarantee.

PR #639, opened 2023-09-28
[write-fonts] Use short offsets in gvar where appropriate

### LINKED ISSUE #615: gvar: use short offsets when appropriate
In gvar, the offsets to the variation data for each glyph can be either 16- or 32-bit. Currently we just always use the long format since it got us compiling *something*, but we should be using short offsets where appropriate.

cmyr:
This was something I skipped in my initial implementation, but should save us a few kb in the common case.

closes #615

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -262,15 +262,15 @@ impl GlyphVariations {
             .map(|tup| tup.build(shared_tuple_map, shared_points.as_ref()))
             .unzip();
 
-        let mut this = GlyphVariationData {
+        let mut temp = GlyphVariationData {
             tuple_variation_headers: tuple_headers,
             shared_point_numbers: shared_points,
             per_tuple_data: tuple_data,
             length: 0,
         };
 
-        this.length = this.compute_size();
-        this
+        temp.length = temp.compute_size();
+        temp
     }
 }
 
@@ -432,7 +432,7 @@ pub struct GlyphVariationData {
     /// calculated length required to store this data
     ///
     /// we compute this once up front because we need to know it in a bunch
-    /// of different places
+    /// of different places (u32 because offsets are max u32)
     length: u32,
 }
 
@@ -502,7 +502,7 @@ impl FontWrite for GlyphDataWriter<'_> {
         for glyph in self.data {
             if !glyph.is_empty() {
                 glyph.write_into(writer);
-                if !self.long_offsets && glyph.length % 2 != 0 {
+                if !self.long_offsets {
                     writer.pad_to_2byte_aligned();
                 }
             }
@@ -970,7 +970,9 @@ mod tests {
         );
     }
 
-    fn make_39_bytes_of_variation_data() -> Vec<GlyphDeltas> {
+    // when using short offsets we store (real offset / 2), so all offsets must
+    // be even, which means when we have an odd number of bytes we have to pad.
+    fn make_31_bytes_of_variation_data() -> Vec<GlyphDeltas> {
         vec![
             GlyphDeltas::new(
                 Tuple::new(vec![F2Dot14::from_f32(-1.0), F2Dot14::from_f32(-1.0)]),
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/gvar.rs
```diff
@@ -123,8 +123,16 @@ impl Gvar {
...
 
     fn compute_flags(&self) -> GvarFlags {
-        //TODO: use short offsets sometimes
-        GvarFlags::LONG_OFFSETS
+        let max_offset = self
+            .glyph_variation_data_offsets
+            .iter()
+            .fold(0, |acc, val| acc + val.length);
+
+        if max_offset / 2 < (u16::MAX as u32) {
```
behdad, line 131: This doesn't consider the padding, does it?

behdad: Also, `<=`?

behdad:
Thanks. Logic LGTM.

write-fonts/src/tables/gvar.rs
```diff
@@ -934,4 +969,119 @@ mod tests {
...
+        let gvar = Gvar::new(a_small_number_of_variations).unwrap();
+        assert_eq!(gvar.compute_flags(), GvarFlags::empty());
+
+        let writer = gvar.compile_variation_data();
+        let mut sink = TableWriter::default();
+        writer.write_into(&mut sink);
+
+        let bytes = sink.into_data().bytes;
+        let expected_len = 6 * u16::RAW_BYTE_LEN // offsets: n_glyphs + 1
+                             + 32 * 5; // rounded size of each glyph, * 5 typhs
```
anthrotype, line 1033: is "typhs" a typo?

cmyr: um yes, that is supposed to say 'glyphs', i imagine..

write-fonts/src/tables/gvar.rs
```diff
@@ -934,4 +969,120 @@ mod tests {
             Some(PackedPointNumbers::Some(vec![1, 3]))
         );
     }
+
+    fn make_39_bytes_of_variation_data() -> Vec<GlyphDeltas> {
```
rsheeter, line 973: Why does it matter that it's 39 bytes?

write-fonts/src/tables/gvar.rs
```diff
@@ -934,4 +969,120 @@ mod tests {
...
+    #[test]
+    fn who_tests_the_testers() {
+        let variations = GlyphVariations::new(GlyphId::NOTDEF, make_39_bytes_of_variation_data());
+        // to get our real size we need to mock up the shared tuples:
+        let mut tupl_map = HashMap::new();
+        tupl_map.insert(&variations.variations[0].peak_tuple, 1);
+        tupl_map.insert(&variations.variations[1].peak_tuple, 2);
+
+        let built = variations.clone().build(&tupl_map);
+        // we need an odd number for this to make sense
```
rsheeter, line 1012: explain why

write-fonts/src/tables/gvar.rs
anthrotype: guess because it's % 2 != 0 so must be 2-byte aligned

write-fonts/src/tables/gvar.rs
```diff
@@ -934,4 +969,120 @@ mod tests {
...
+                .iter()
+                .map(|off| off.unwrap().get())
+                .collect::<Vec<_>>(),
+            vec![0u32, 32, 64, 96, 128, 160]
+        );
+    }
+
+    #[test]
+    fn use_long_offsets_when_necessary() {
+        // 2**16 * 2 / (31 + 1 padding) (bytes per tuple) = 4096 so this just overflows
```
rsheeter, line 1054: I suggest you test the most possible short offsets as well, for example loop over N_GLYPHS 4099, 4100 and confirm that their respective flags are SHORT_OFFSETS, LONG_OFFSETS with the expected sizes

write-fonts/src/tables/gvar.rs
```diff
@@ -934,4 +969,120 @@ mod tests {
...
+        assert_eq!(loaded.flags(), GvarFlags::LONG_OFFSETS);
+        assert_eq!(
+            loaded
+                .glyph_variation_data_offsets()
+                .iter()
+                .take(6)
+                .map(|off| off.unwrap().get())
+                .collect::<Vec<_>>(),
+            vec![0u32, 31, 62, 93, 124, 155]
+        )
```
rsheeter, line 1086: The majority of prefer_short_offsets and use_long_offsets_when_necessary is duplicative noise; if you diff them they differ only in a few key values. Make a helper, e.g. `assert_gvar(num glyphs, expected flags, expected offset type, expected first values)`

That will make testing the maximum number of short offsets as it's own test trivial too.

write-fonts/src/tables/gvar.rs
```diff
@@ -254,11 +262,15 @@ impl GlyphVariations {
             .map(|tup| tup.build(shared_tuple_map, shared_points.as_ref()))
             .unzip();
 
-        GlyphVariationData {
+        let mut this = GlyphVariationData {
```
rsheeter, line 265: Nit: the name "this" is perhaps suboptimal to readers with a C++ or Java background.

write-fonts/src/tables/gvar.rs
```diff
@@ -417,6 +429,11 @@ pub struct GlyphVariationData {
     // optional; present if multiple variations have the same point numbers
     shared_point_numbers: Option<PackedPointNumbers>,
     per_tuple_data: Vec<GlyphTupleVariationData>,
+    /// calculated length required to store this data
+    ///
+    /// we compute this once up front because we need to know it in a bunch
+    /// of different places
+    length: u32,
```
rsheeter, line 436: usize seems the more natural type, I suppose you eventually want to assign it somewhere u32. Perhaps note why it's u32 here?

write-fonts/src/tables/gvar.rs
```diff
@@ -457,19 +474,37 @@ struct GlyphDataWriter<'a> {
...
+                last.write_into(writer);
+            }
         }
         // then write the actual data
         for glyph in self.data {
             if !glyph.is_empty() {
                 glyph.write_into(writer);
+                if !self.long_offsets && glyph.length % 2 != 0 {
+                    writer.pad_to_2byte_aligned();
+                }
```
rsheeter, line 507: You dont' need && glyph.length % 2 here as pad_to_2byte_aligned will happily nop if htat's true?

rsheeter (approved):
LGTM once comments addressed

PR #641, opened 2023-09-29
[codegen] Add missing debug impls to generic groups

### LINKED ISSUE #605: we aren't generating debug impls for fancy layout enums
This is an oversight. In particular I mean `[Position|Substitution]Lookup` and `[Position|Substitution]Extension`.

cmyr:
Aka the lookup and extension enums in gpos/gsub.

fixes #605

This also renames an internal method to match the name used elsewhere in codegen.

### DIFF
--- a/font-codegen/src/table.rs
+++ b/font-codegen/src/table.rs
@@ -190,7 +190,7 @@ pub(crate) fn generate_group(item: &GenericGroup) -> syn::Result<TokenStream> {
 
     let mut variant_decls = Vec::new();
     let mut read_match_arms = Vec::new();
-    let mut as_some_table_arms = Vec::new();
+    let mut dyn_inner_arms = Vec::new();
     for var in &item.variants {
         let var_name = &var.name;
         let type_id = &var.type_id;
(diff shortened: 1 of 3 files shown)


font-codegen/src/table.rs
```diff
@@ -190,15 +190,15 @@ pub(crate) fn generate_group(item: &GenericGroup) -> syn::Result<TokenStream> {
 
     let mut variant_decls = Vec::new();
     let mut read_match_arms = Vec::new();
-    let mut as_some_table_arms = Vec::new();
+    let mut dyn_inner_arms = Vec::new();
```
rsheeter, line 193: Nit: Feels like Hungarian notation which isn't really necessary in Rust. Would anything be particularly horrible about dropping dyn_ from various names? - we know it's dyn, it's in the signature.

rsheeter (approved):

font-codegen/src/table.rs
cmyr: nothing terrible, but the change here is just making this small portion of the code match the naming convention used everywhere else. This method is a private implementation detail related to the traversal code, and I don't have particularly strong feelings either way (the other name I had used here was 'as_some_table')

in any case if we want to change this it would be a significantly bigger diff and should be a separate PR.

PR #647, opened 2023-10-05
[read-fonts] Add 'get' method to coverage tables

cmyr:
I wanted this for testing the table splitting in Mark2Base, and it's something we will want anyway.

This also fixes a little mixup in the signature of CoverageTable::iter; the elided lifetime here meant the compiler tied the iterator to the lifetime of the table object, whereas we really want to tie it to the lifetime of the underlying data.

JMM

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -3,11 +3,13 @@
 #[path = "./lookupflag.rs"]
 mod lookupflag;
 
+use core::cmp::Ordering;
+
 pub use lookupflag::LookupFlag;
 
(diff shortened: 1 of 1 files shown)


cmyr:
ooh, looks like we have a new rustc/clippy.

Also I'm wondering if this shouldn't return `usize`, since we expect coverage indices to be used mostly to index into collections?

dfrg (approved):
Happy to see these and thanks for doing the binary search!

I’m ambivalent on the return type. As `u16` matches the spec, this seems good to me as is.

anthrotype (approved):

PR #649, opened 2023-10-05
Table splitting for MarkBase

cmyr:
This is an initial port of the table splitting impl for this subtable.

Unlike with PairPos, we don't have any real-world failure cases to test this on yet, but I think it's ready as a checkpoint.

This is based on #647, which should go in first. We also need to resolve clippy & rustdoc complains (#648)

### DIFF
--- a/write-fonts/src/graph/splitting/mark2base.rs
+++ b/write-fonts/src/graph/splitting/mark2base.rs
@@ -98,6 +98,8 @@ fn split_off_mark_pos(
     let base_coverage_id = data.offsets[1].object;
     let mark_array_id = data.offsets[2].object;
     let base_array_id = data.offsets[3].object;
+    let old_mark_array = &graph.objects[&mark_array_id];
+    let old_base_array = &graph.objects[&base_array_id];
     let mark_class_count = (end - start) as u16;
     let mut new_subtable = TableData::new(data.type_);
 
@@ -111,10 +113,10 @@ fn split_off_mark_pos(
         .filter_map(|(i, gid)| mark_glyphs_by_cov_id.contains(&i).then_some(gid))
         .collect();
     let new_mark_coverage = super::make_table_data(&new_mark_coverage);
+    let new_mark_array = split_off_mark_array(old_mark_array, start as u16, &mark_glyphs_by_cov_id);
+    let new_base_array = split_off_base_array(old_base_array, start, end, class_info.len());
     let new_mark_coverage_id = graph.add_object(new_mark_coverage);
-    let new_mark_array = split_off_mark_array(graph, mark_array_id, start, &mark_glyphs_by_cov_id);
     let new_mark_array_id = graph.add_object(new_mark_array);
-    let new_base_array = split_off_base_array(graph, base_array_id, start, end, class_info.len());
     let new_base_array_id = graph.add_object(new_base_array);
 
     new_subtable.write(1u16); // format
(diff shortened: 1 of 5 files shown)


anthrotype:
> we don't have any real-world failure cases to test this on yet

in the equivalent fonttools PR https://github.com/fonttools/fonttools/pull/1297 from a few years back, some Noto fonts are mentioned. In any case, fontc probably will have to emit some MarkToBase lookups before we can test this with real fonts

write-fonts/src/graph/splitting/mark2base.rs
```diff
@@ -0,0 +1,680 @@
...
+        .reparse_with_args(&(old_mark_class_count as u16))
+        .unwrap();
+
+    let mut next_offset_idx = 0;
+    // because offsets may be null, and there is no pattern, we visit each one
+    for base_record in base_array.base_records().iter() {
+        let base_record = base_record.unwrap();
+        for (mark_class, offset) in base_record.base_anchor_offsets().iter().enumerate() {
+            let has_offset = !offset.get().is_null();
+            let in_range = (start..end).contains(&mark_class);
```
dfrg, line 182: I think this comment should be on the inner for loop. 

I'm assuming this means that we iterate all and check the range rather than slicing `base_anchor_offsets()` on `start..end` because we need to track `next_offset_idx`? Could we break out of the loop when `mark_class >= end`? (not suggesting we do so, just checking my understanding of what's going on)

write-fonts/src/graph/splitting/mark2base.rs
```diff
@@ -0,0 +1,680 @@
...
+                    debug_assert!(graph.objects[&id.object].offsets.is_empty());
+                    visited
+                        .insert(id.object)
+                        .then(|| graph.objects[&id.object].bytes.len())
+                        .unwrap_or(0)
+                })
+                .sum::<usize>();
+            base_size + children_size
+        })
+        .sum()
```
dfrg, line 240: No issues.. just wanted to say that I <3 this code.

dfrg (approved):
Apologies for letting this one sit so long. There's some realllly annoying inherent complexity here so I both compared to HB and traced the logic according to the spec and... this seems solid, nice work! A couple comments/questions inline but I didn't see anything blocking.

write-fonts/src/graph/splitting/mark2base.rs
```diff
@@ -0,0 +1,680 @@
...
+            accumulated = min_subtable_size + accumulated_delta;
+            partial_coverage_size = 4 + u16::RAW_BYTE_LEN * info.marks.len();
+            visited.clear();
+        }
+    }
+
+    log::debug!(
+        "nothing to split, size '{}'",
+        accumulated + partial_coverage_size
+    );
```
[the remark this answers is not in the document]
cmyr: it's supposed to be a companion to the `trace!` above, which prints each split; this prints the size of the final subtable (it's sort of a fenceposts-vs-fences thing)

write-fonts/src/graph/splitting/mark2base.rs
cmyr: noted, reading this back I agree that there could be more explanation here. Basically: what is tricky here is that the anchor tables may or may not have subtables, and there is no pattern around how subtables are distributed across the set of anchor tables, so we need to visit each offset and see if it's null or not to figure out whether we are copying over a subtable.

The reason we can't break early is because even if we're out of range, if we saw an offset we need to increment `next_offset_idx` so that once we're back in range we are copying the correct subtable. 🤕

dfrg: Ohh, yep, I see it now. The comment seems like it's in the proper place then... I was just missing that the offset list is associated with the base array and not the base record (obvious in hindsigh). Thanks!

PR #652, opened 2023-10-06
[VariationStoreBuilder] Fix bug computing VarData's region_indexes

anthrotype:
I found a bug in the VariationStoreBuilder in the way it's computing the ItemVariationData.region_indexes (array of indices into the VariationRegionList).

Looks like we are writing these indices always in increasing order, whereas they should match the order of the VarData column that each region is mapped to.

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -877,4 +877,47 @@ mod tests {
         builder.add_deltas(vec![(r1.clone(), 0), (r2.clone(), 0)]);
         let _ = builder.build();
     }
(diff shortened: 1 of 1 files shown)


anthrotype:
as expected the test fails: https://github.com/googlefonts/fontations/actions/runs/6434109825/job/17472635920?pr=652

```
---- tables::variations::ivs_builder::tests::vardata_region_indices_order stdout ----
thread 'tables::variations::ivs_builder::tests::vardata_region_indices_order' panicked at write-fonts/src/tables/variations/ivs_builder.rs:906:9:
  left: [0, 1]
 right: [1, 0]
```

PR #653, opened 2023-10-06
Fix Incorrect region indexes in ItemVariationStore

cmyr:
I ended up very confused in the course of investigating #652, which led me to try and improve a bunch of names and docs in the ItemVariationStore builder code. This includes the test from #652 as well as a fix that is slightly more efficient than the one there, although that solution was also correct :)

cc @anthrotype

- closes #652


anthrotype (approved):
LGTM!

PR #663, opened 2023-10-17
[iup] Fix mismatch in behaviour with fonttools

### LINKED ISSUE #662: IUP failing on certain inputs
It looks like we're failing on certain glyphs in googlesans; I see this reported for at least 'm.sc' and 'servicemark', and am investigating further.

_Originally posted by @madig in https://github.com/googlefonts/googlesans/issues/586#issuecomment-1755594306_

cmyr:
This patch fixes an issue with iup not resolving, although I do not fully understand the problem.

Basically: in comparing our implementation with the one in fonttools, there is clear divergence in some of the numerical code, where we are using Option<usize> in places where python is just using an integer. More importantly, we are using None where Python is using -1, but our comparison code is never checking if a value == None, whereas python *does* check if a value == -1, which means that python is setting the solution in cases where we are not.

This can be fixed by using `checked_sub` instead of `saturating_sub`, at least for this input.

It feels like this is probably fine, since the structure of this loop is such that the only negative value that is possible is -1 (at the start) but it feels a little loose.

- fixes #662 
- cc https://github.com/googlefonts/googlesans/issues/586

### DIFF
--- a/write-fonts/src/tables/gvar/iup.rs
+++ b/write-fonts/src/tables/gvar/iup.rs
@@ -527,7 +527,8 @@ fn iup_contour_optimize(
                 solution.insert(idx % n);
                 i = dp_result.chain[idx];
             }
-            if i == Some(start.saturating_sub(n)) {
+            // this should only be None for the first loop, when start is n - 1
+            if i == start.checked_sub(n) {
                 // Python reads [-1] to get 0, usize doesn't like that
                 let cost = dp_result.costs[start]
                     - if n < start {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/gvar/iup.rs
```diff
@@ -1279,4 +1279,65 @@ mod tests {
...
+            (0.0, 0.0),
+            (0.0, 0.0),
+        ]
+        .into_iter()
+        .map(|(x, y)| Point::new(x, y))
+        .collect::<Vec<_>>();
+
+        let contour_ends = vec![17];
+
+        iup_delta_optimize(deltas, coords, 0.5, &contour_ends).unwrap();
```
rsheeter, line 1341: can the test be further simplified? Do we need the entire outline to reproduce the problem?

(my intuition is it should simplify)

rsheeter (approved):

PR #664, opened 2023-10-18
[write-fonts] Tweak VariationStoreBuilder for HVAR case

cmyr:
This change lets us build the 'unoptimized' version of this table which can be a more efficient choice in HVAR, in some cases.

cc @anthrotype, is this fit for your needs?

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -21,13 +21,16 @@ type TemporaryDeltaSetId = u32;
 pub struct VariationStoreBuilder {
     // region -> index map
     all_regions: HashMap<VariationRegion, usize>,
-    // we use an index map so that we have a deterministic ordering,
-    // which lets us write better tests
-    //NOTE: we could store IndexMap<RowShape, IndexMap<DeltaSet, TemporaryDeltaSetId>> here,
-    //if desired, but that would require us to know the total number of regions
-    //when we construct the builder.
-    delta_sets: IndexMap<DeltaSet, TemporaryDeltaSetId>,
-    next_id: TemporaryDeltaSetId,
+    delta_sets: DeltaSetStorage,
+}
+
+/// A collection of delta sets.
+#[derive(Clone, Debug)]
+enum DeltaSetStorage {
+    // only for hvar: we do not deduplicate deltas, and store one per glyph id
+    Direct(Vec<DeltaSet>),
+    // the general case, where each delta gets a unique id
+    Deduplicated(IndexMap<DeltaSet, TemporaryDeltaSetId>),
 }
 
 /// A map from the temporary delta set identifiers to the final values.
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -46,21 +49,34 @@ pub struct VariationIndexRemapping {
 struct DeltaSet(Vec<(u16, i32)>);
 
 impl VariationStoreBuilder {
-    pub fn add_deltas<T: Into<i32>>(&mut self, deltas: Vec<(VariationRegion, T)>) -> u32 {
+    /// Create a builder that will optimize delta storage
+    pub fn new() -> Self {
```
rsheeter, line 53: Nit: I might prefer optimized for symmetry with unoptimized.

rsheeter (approved):

anthrotype (approved):
Much better, thank you!

I just tested building a direct HVAR for NotoSansCherokee and we match the same output as fontmake.py

write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -942,4 +1037,40 @@ mod tests {
...
+
+        let mut builder = VariationStoreBuilder::unoptimized();
+        builder.add_deltas(vec![(r0.clone(), 1), (r1.clone(), 2)]);
+        // 256 won't fit in a u8 thus we expect the deltas for the column corresponding
+        // to r1 will be packed as u16
+        builder.add_deltas(vec![(r0.clone(), 1), (r1.clone(), 2)]);
+        builder.add_deltas(vec![(r0.clone(), 3), (r1.clone(), 256)]);
+        builder.add_deltas(vec![(r0.clone(), 3), (r1.clone(), 256)]);
+
+        let (ivs, _) = builder.build();
```
anthrotype, line 1054: maybe the test could check that the returned varidx_map (currently unused `_`) contains an identity mapping from temp ids to final varindices?

write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -83,28 +99,64 @@ impl VariationStoreBuilder {
...
+        // then encode everything with that encoding.
+        let encoding = Encoding {
+            shape,
+            deltas: self.delta_sets.iter().collect(),
+        };
+        debug_assert!(
+            encoding.deltas.len() <= u16::MAX as usize,
+            "unmapped variation store supports at most u16::MAX items"
+        );
+        encoding.encode(key_map, 0)
```
anthrotype, line 145: should we assert as sanity-check that `key_map` only contains an identity mapping so that gid:varidx relationship holds for the unoptimized builder?

write-fonts/src/tables/variations/ivs_builder.rs
anthrotype: maybe add both methods and have default new constructor call optimized

anthrotype: then VariationStoreBuilder would have three constructors that do exactly the same thing, syntactic sugar addicted

cmyr: yea I went back and forth on this, e.g. considered removing `default` and only having explicit constructors, then didn't want to break existing code, then wanted names that made it clear that one of the constructors really *is* preferred in the general case, then realized this was also a good argument for default... 🥴 

I also don't like `optimized` since the tense is wrong. I hate all options equally?

write-fonts/src/tables/variations/ivs_builder.rs
cmyr: my expectation is that the key map isn't used in this case, but can certainly add this to the test, we do set the values.

write-fonts/src/tables/variations/ivs_builder.rs
cmyr: okay I've gone with `new` and `new_with_implicit_indices`.

PR #665, opened 2023-10-20
[write-fonts] Allow to construct DeltaSetIndexMap from VariationIndex iterator

anthrotype:
fontc needs to build a DeltaSetIndexMap for the HVAR table and would like to not bother with encoding details like entry format or whether to build Format0 or Format1 depending on the number of mappings, etc.

The VariationStoreBuilder returns, alongside the ItemVariationStore, a VariationIndexRemapping (which maps from TemporaryDeltaSetId (u32) to VariationIndex).
After this PR, fontc can simply collect all these VariationIndex values into a new DeltaSetIndexMap.

### DIFF
--- a/write-fonts/src/tables/variations.rs
+++ b/write-fonts/src/tables/variations.rs
@@ -370,9 +370,82 @@ impl Tuple {
     }
 }
 
+impl DeltaSetIndexMap {
+    /// Return the most compact entry format that can represent this mapping.
+    ///
+    /// EntryFormat is a packed u8 field that describes the compressed representation
+    /// of delta-set indices. For more info, see:
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#associating-target-items-to-variation-data>
+    // This is a direct port from fonttools' DeltaSetMap.getEntryFormat:
+    // https://github.com/fonttools/fonttools/blob/6d531f/Lib/fontTools/ttLib/tables/otTables.py#L644-L666
+    fn get_entry_format(mapping: &[u32]) -> EntryFormat {
+        let ored = mapping.iter().fold(0, |acc, idx| acc | *idx);
+
+        let inner = (ored & 0xFFFF) as u16;
+        let inner_bits = (16 - inner.leading_zeros() as u8).max(1);
+        assert!(inner_bits <= 16);
+
+        let ored = (ored >> (16 - inner_bits)) | (ored & ((1 << inner_bits) - 1));
+        let entry_size = if ored <= 0x000000FF {
+            1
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/variations.rs
```diff
@@ -370,9 +370,82 @@ impl Tuple {
...
+    /// EntryFormat is a packed u8 field that describes the compressed representation
+    /// of delta-set indices. For more info, see:
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#associating-target-items-to-variation-data>
+    // This is a direct port from fonttools' DeltaSetMap.getEntryFormat:
+    // https://github.com/fonttools/fonttools/blob/6d531f/Lib/fontTools/ttLib/tables/otTables.py#L644-L666
+    fn get_entry_format(mapping: &[u32]) -> EntryFormat {
+        let ored = mapping.iter().fold(0, |acc, idx| acc | *idx);
+
+        let inner = (ored & 0xFFFF) as u16;
+        let inner_bits = (16 - inner.leading_zeros() as u8).max(1);
```
cmyr, line 385: could we use [`u16::count_ones`](http://doc.rust-lang.org/1.73.0/std/primitive.u16.html#method.count_ones)?

cmyr (approved):
Looks good! A couple comments inline but nothing blocking :)

write-fonts/src/tables/variations.rs
anthrotype: no we can't use count_ones here, that one simply returns the number of ones in the binary representation of self, whereas here we want to count the maximum number of bits that are sufficient to represent this u16.

```rust
>> let inner: u16 = 0b00000000_00010010;
>> inner.count_ones()
2
>> (16 - inner.leading_zeros() as u8).max(1)
5

File: write-fonts/src/tables/variations.rs [hunk 2]
```diff
@@ -370,9 +370,82 @@ impl Tuple {
...
+        let ored = (ored >> (16 - inner_bits)) | (ored & ((1 << inner_bits) - 1));
+        let entry_size = if ored <= 0x000000FF {
+            1
+        } else if ored <= 0x0000FFFF {
+            2
+        } else if ored <= 0x00FFFFFF {
+            3
+        } else {
+            4
+        };
```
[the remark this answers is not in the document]
anthrotype: thanks, I'll do that. I think it has a typo though:

```diff
 let entry_size = match ored {
     0..=0xFF => 1,
     0x100..=0xFFFF => 2,
-     0x1000..=0xFFFFFF => 3,
+     0x10000..=0xFFFFFF => 3,
     _ => 4,
 };
```

PR #670, opened 2023-10-23
Reproduce OutOfBounds error when hvar index map has fewer entries than glyphs

anthrotype:
https://github.com/googlefonts/fontations/pull/668#issuecomment-1775536793

/cc @drott


anthrotype:
fails as expected:

https://github.com/googlefonts/fontations/actions/runs/6617440064/job/17973745815?pr=670#step:5:127 

```
thread 'tables::hvar::tests::advance_deltas_from_hvar_with_truncated_adv_index_map' panicked at read-fonts/src/tables/hvar.rs:108:22:
called `Result::unwrap()` on an `Err` value: OutOfBounds
```

drott (approved):

PR #671, opened 2023-10-23
[write-fonts] Match fonttools sorting in var store

cmyr:
With this patch we will use a consistent ordering to sort both encodings (which correspond to outer indices, in a VariationIndex table) and individual delta sets (which correspond to inner indices.)

This PR is more complicated than expected because we had to jump through some hoops in order to precisely match the sorting behaviour in fonttools.

With this, https://github.com/googlefonts/fontc/pull/490 now produces identical output for HVAR.

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -335,9 +335,8 @@ impl<'a> Encoder<'a> {
             .for_each(|enc| enc.deltas.sort_unstable());
         // and then sort the encodings themselves; order doesn't matter,
         // but we want to match fonttools output when comparing ttx
-        self.encodings.sort_unstable_by(|a, b| {
-            (a.shape.row_cost(), &a.shape).cmp(&(b.shape.row_cost(), &b.shape))
-        });
+        self.encodings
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -329,6 +329,15 @@ impl<'a> Encoder<'a> {
...
         }
         self.encodings = to_process.into_iter().flatten().collect();
+        // now sort the items in each individual encoding
+        self.encodings
+            .iter_mut()
+            .for_each(|enc| enc.deltas.sort_unstable());
+        // and then sort the encodings themselves; order doesn't matter,
+        // but we want to match fonttools output when comparing ttx
+        self.encodings.sort_unstable_by(|a, b| {
+            (a.shape.row_cost(), &a.shape).cmp(&(b.shape.row_cost(), &b.shape))
```
behdad, line 339: I think this line won't match fonttools. As we discussed, comparing `shape` here compares lower indices first. Whereas in fonttools the `bits` encoding compares higher indices first, because it's compared as a big integer. No?

behdad:
LGTM. Thanks!

PR #672, opened 2023-10-24
[ivs_builder] Fix all-zero delta sets not being de-duplicated

anthrotype:
this test currently fails because the VarStore ends up with two identical [0] delta-sets instead of one. The builder is treating glyphs without any variations (add_deltas([]) and glyphs with explicitly defined 0 deltas (add_deltas([0]) as different, whereas it should treat them the same and have all point to the same no-op delta-set row.

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -1339,20 +1339,24 @@ mod tests {
 
     #[test]
     fn no_duplicate_zero_delta_sets() {
-        let r0 = VariationRegion::new(vec![reg_coords(0.0, 1.0, 1.0)]);
+        let r0 = VariationRegion::new(vec![reg_coords(0.0, 5.0, 1.0)]);
+        let r1 = VariationRegion::new(vec![reg_coords(0.5, 1.0, 1.0)]);
         let mut builder = VariationStoreBuilder::new();
         let varidxes = vec![
             // first glyph has no variations (e.g. .notdef only defined at default location)
             // but we still need to add it to the variation store to reserve an index so
             // we add an empty delta set
             builder.add_deltas::<i32>(Vec::new()),
-            builder.add_deltas(vec![(r0.clone(), 100)]),
-            // this glyph has an explicit master that is the same as the default (delta is 0);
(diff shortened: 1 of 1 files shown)


anthrotype:
```diff
--- fontmake
+++ fontc

   <VarData index="0">
-    <!-- ItemCount=2 -->
+    <!-- ItemCount=3 -->
     <NumShorts value="0"/>
     <!-- VarRegionCount=1 -->
     <VarRegionIndex index="0" value="0"/>
     <Item index="0" value="[0]"/>
-    <Item index="1" value="[100]"/>
+    <Item index="1" value="[0]"/>
+    <Item index="2" value="[100]"/>
   </VarData>
```

cmyr:
hmm yes I knew that it was possible for us to end up with duplicate data as a result of the merge operation, and wasn't sure if it would be an issue in practice.

One funny thing here is that I imagine I *do* want to preserve the duplicates in the case where we are using the implicit ordering?

anthrotype:
Correct. But this specific test is for the optimized builder

cmyr:
Okay I've added a commit here that should solve this issue.

write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -78,8 +78,15 @@ impl VariationStoreBuilder {
     ) -> TemporaryDeltaSetId {
         let mut delta_set = Vec::with_capacity(deltas.len());
         for (region, delta) in deltas {
+            let delta = delta.into();
+            // a delta of zero is the same as not including this region;
```
anthrotype, line 82: this is incorrect. A delta of zero is _not_ the same as not including a region. Imagine a glyph defines three masters (or regions): ({wght:400}, 100) is the default, then there are ({wght:500}, 100) and ({wght:700}, 200). The deltas for these two are respectively [0, 100], but the 0 there cannot be simply skipped as if it was not there. It has the effect of anchoring that location (or restoring it back) to the default value. If it was not there, there would be a linear interpolation between ({wght:400}, 100) and ({wght:700}, 200).
Only when _all_ deltas in a delta set are 0 they can be treated the same as them being empty.

anthrotype: I'll modify the test to show what I mean

cmyr (approved):
okay I see what you mean, this looks better :)

PR #674, opened 2023-10-25
Add COLR test fonts

drott:
Use COLRv0/v1 test fonts from [1] to aid in the work of developing a COLRv1/COLRv0 painter callback interface, as outlined in [2].

[1] https://github.com/googlefonts/color-fonts
[2] https://github.com/googlefonts/oxidize/pull/48#issuecomment-1755759028


dfrg (approved):
Glad to see these arrive and looking forward to color support!

One nit: our policy is to check in the associated ttx files as well in the ttx directory. Once that’s done, this lgtm!

PR #675, opened 2023-10-25
Add derived Debug implementation for ResolvedPaint

drott:
Enable debug printing of ResolvedPaint by adding a derived implementattion for this Enum as well as for ColorStops.


drott:
Looks like this needs a decision and work on what to do with `Paint` types - that didn't seem to be a problem locally when I used it. Will look into it.

drott:
Solved this differently locally.a

PR #678, opened 2023-10-27
[skrifa] cff: correct "scaling" for unscaled outlines 

dfrg:
Make us match `FT_LOAD_NO_SCALE` for CFF outlines.

The original code returned the exact unscaled values from charstring evaluation but FT actually performs a fixed point multiply by 1/64 followed by truncation. We now do the same.


cmyr (approved):
It doesn't look *un*reasonable

PR #682, opened 2023-10-31
[read-fonts] Add convenience methods to AnchorTable

cmyr:
This should save the client from needing to interact with the enum in the general case.

Way back when I was first doing codegen I thought it might be nice to autogenerate getters for shared fields, but I never got around to it. Maybe it's worth thinking about again?

This was motivated by https://github.com/googlefonts/fontc/pull/492, which was adding a helper method to do the same thing.


anthrotype:
> might be nice to autogenerate getters for shared fields

+1

anthrotype (approved):

rsheeter:
+1 to doing this automatically, otherwise users will almost assuredly write such things. Shall we file an issue if we don't have one already?

cmyr:
opened #687 to track the general case.

PR #684, opened 2023-10-31
[codegen] Improve generated docs for typed offset getters

### LINKED ISSUE #681: It's not obvious which FontData to give
Suppose you want to loop over your base anchors to see their coordinates. It's a bit more indirect than you'd hoped but np. And then you hit this:

```rust
read_fonts::tables::gpos::BaseRecord
pub fn base_anchors(&self, data: FontData<'a>) -> ArrayOfNullableOffsets<'a, AnchorTable<'a>, Offset16>
```

Great, which FontData does it want? Pity it's not apparent from type but that's OK, the comment will help right?

```rust
    /// A dynamically resolving wrapper for [`base_anchor_offsets`][Self::base_anchor_offsets].
    pub fn base_anchors(
        &self,
        data: FontData<'a>,
    ) -> ArrayOfNullableOffsets<'a, AnchorTable<'a>, Offset16> {
        let offsets = self.base_anchor_offsets();
        ArrayOfNullableOffsets::new(offsets, data, ())
    }
```

To be fair you can tell the answer from the spec but it's a bit tiresome. It would be nice to make it easier. At the very least add comments to these methods that say where to get the correct FontData. Bonus points for having types so you can't pass the wrong one without making an effort.

For context see https://github.com/googlefonts/fontc/pull/492 GPOS tests such as compile_basic_gpos_mark_base.

cmyr:
motivated by #681

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -779,6 +779,29 @@ impl Field {
         quote!( self.shape.#shape_range_fn_name() #try_op )
     }
 
+    fn typed_offset_getter_docs(&self, has_data_arg: bool) -> TokenStream {
+        let raw_name = &self.name;
+        // If there's no arguments than we just link to the raw offset method
+        if !has_data_arg {
+            let docs = if self.is_array() {
+                format!(" A dynamically resolving wrapper for [`{raw_name}`][Self::{raw_name}].")
+            } else {
+                format!(" Attempt to resolve [`{raw_name}`][Self::{raw_name}].")
+            };
+            return quote!(#[doc = #docs]);
+        }
+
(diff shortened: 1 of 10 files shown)


rsheeter:
I think this is an improvement but I'm not sold it closes the issue. I would ideally like at least one of the following to close:

* Say which specific thing to get the data from
* Resolve the issue of being trivially able to pass the wrong data
   * Could we use the type system here?
   * If, for example, `mark_anchor(data)` took a data typed MarkArrayData it would be trivial to find where to get one of those.

rsheeter (approved):
LGTM but I suggest not closing the issue

PR #686, opened 2023-11-01
[skrifa] cff hintmap

dfrg:
Adds the HintMap type (CF2_HintMap in FreeType).

This is basically the core of the CFF hinting engine. It generates a list of sorted edges from the "blues" alignment zones and per-glyph stem hints which can be used to transform from character space (font units) to device space (grid fitted pixels).

I'd expect one or two more PRs to hook it all up, but this should be the worst of it. Almost there!


cmyr (approved):
I am not going to pretend to have much understanding of what's going on here, but nothing jumps out at me as being obviously bad. 👼

dfrg:
When it comes to hinting, I'll take "not obviously bad" as high praise :)

Issue #232, opened 2023-02-09
Reconsider F26Dot6 ops


dfrg:
Based on the comment [here](https://github.com/googlefonts/fontations/pull/231#discussion_r1101719975) in #231, audit use of mul/div operators for `F26Dot6` and either remove them (and use conversions to `Fixed` where necessary) or add correct implementations.

PR #688, opened 2023-11-03
[skrifa] cff hinting sink

dfrg:
This adds a new `HintingSink` type that pulls all the pieces together and feeds the hinting engine from charstring evaluation.

### DIFF
--- a/skrifa/src/scale/cff/hint.rs
+++ b/skrifa/src/scale/cff/hint.rs
@@ -945,11 +945,10 @@ impl<'a, S: CommandSink> HintingSink<'a, S> {
     fn maybe_close_subpath(&mut self) {
         // This requires some explanation. The hint mask can be modified
         // during charstring evaluation which changes the set of hints that
-        // are applied. FreeType ensures that the closing line (this logic
-        // doesn't apply to cubics *shrug*) for any subpath is transformed
-        // with the same hint map as the starting point for the subpath.
-        // This is done by stashing a copy of the hint map that is active
-        // when a new subpath is started. Unlike FreeType, we make use
+        // are applied. FreeType ensures that the closing line for any subpath
+        // is transformed with the same hint map as the starting point for the
+        // subpath. This is done by stashing a copy of the hint map that is
+        // active when a new subpath is started. Unlike FreeType, we make use
         // of close elements, so we can cheat a bit here and avoid the
         // extra hintmap. If we're closing an open subpath and have a pending
         // line and the line is not equal to the start point in character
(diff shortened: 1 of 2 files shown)


skrifa/src/scale/cff/hint.rs
```diff
@@ -904,6 +907,182 @@ fn msb_mask(bit: usize) -> u8 {
...
+    pub fn finish(&mut self) {
+        self.maybe_close_subpath();
+    }
+
+    fn maybe_close_subpath(&mut self) {
+        // This requires some explanation. The hint mask can be modified
+        // during charstring evaluation which changes the set of hints that
+        // are applied. FreeType ensures that the closing line (this logic
+        // doesn't apply to cubics *shrug*) for any subpath is transformed
+        // with the same hint map as the starting point for the subpath.
```
cmyr, line 950: "for any subpath is transformed" is missing some word somewhere

skrifa/src/scale/cff/hint.rs
```diff
@@ -904,6 +907,182 @@ fn msb_mask(bit: usize) -> u8 {
...
+        // We emit close commands based on the sequence of moves.
+        // See `maybe_close_subpath`
+    }
+}
+
+/// FreeType returns values in 26.6 but we keep them in 16.6. This
+/// simply zeroes out the low 10 bits so that we match.
+fn trunc(value: Fixed) -> Fixed {
+    Fixed::from_bits(value.to_bits() & !0x3FF)
+}
```
cmyr, line 1084: So to clarify, we're using `Fixed` to represent `26.6`? I vaguely recall this, but has there been any thought towards a `F26Dot6` type?

cmyr (approved):
Very nice to see this coming together!

skrifa/src/scale/cff/hint.rs
dfrg: The parenthetical seems to completely destroy the flow of this sentence. I’ll clean it up.

skrifa/src/scale/cff/hint.rs
dfrg: The actual value here is 16.16 but we zero the low 10 bits so that conversion to float has the same precision as if it was 26.6. 

We do have `F26Dot6` now but we can’t actually implement it correctly for our purposes (see #232). The unfortunate reality is that FT works (sometimes) in 26.6 but uses 16.16 mul/div functions on carefully crafted values to get the expected results. If we try to do the principled thing, we’ll break compatibility. In that case, I’d rather just move to floating point (someday!)

rsheeter: ^ seems worth capturing in comments. Also we should file a move to float issue if we haven't already.

dfrg: There’s already a comment on this function that was cut off in the review. I’ll add the note about conversion to float to clarify the purpose. 

+1 on the issue. Will do.

PR #692, opened 2023-11-07
[skrifa] glyph metrics fixes

dfrg:
A few changes to fix correctness and/or match FreeType:
1. Make unscaled case pass-through. Avoids overflow on large source values.
2. Cast advance width to `u16` before scaling to match FT
3. Don't reject empty `glyf` glyphs when reading metric deltas from `gvar`

### DIFF
--- a/skrifa/src/instance.rs
+++ b/skrifa/src/instance.rs
@@ -54,20 +54,20 @@ impl Size {
     }
 
     /// Computes a fixed point linear scale factor that matches FreeType.
-    pub(crate) fn fixed_linear_scale(self, units_per_em: u16) -> Fixed {
-        // FreeType computes a 16.16 scale factor that converts to 26.6.
-        // This is done in two steps, assuming use of FT_Set_Pixel_Size:
-        // 1) height is multiplied by 64:
-        //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3596>
-        // 2) this value is divided by UPEM:
-        //    (here, scaled_h=height and h=upem)
-        //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3312>
-        if self.0 > 0.0 && units_per_em > 0 {
-            Fixed::from_bits((self.0 * 64.) as i32) / Fixed::from_bits(units_per_em as i32)
+    pub(crate) fn fixed_linear_scale(self, units_per_em: u16) -> Option<Fixed> {
+        if self.0 > 0.0 && units_per_em > 0 && self.0 != units_per_em as f32 {
+            // FreeType computes a 16.16 scale factor that converts to 26.6.
+            // This is done in two steps, assuming use of FT_Set_Pixel_Size:
+            // 1) height is multiplied by 64:
+            //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3596>
+            // 2) this value is divided by UPEM:
+            //    (here, scaled_h=height and h=upem)
+            //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3312>
+            Some(Fixed::from_bits((self.0 * 64.) as i32) / Fixed::from_bits(units_per_em as i32))
         } else {
-            // This is an identity scale for the pattern
-            // `mul_div(value, scale, 64)`
-            Fixed::from_bits(0x10000 * 64)
+            // Just pass through in unscaled case. Avoids potential overflow
+            // for large values.
+            None
         }
     }
 }
--- a/skrifa/src/metrics.rs
+++ b/skrifa/src/metrics.rs
@@ -308,7 +308,10 @@ impl<'a> GlyphMetrics<'a> {
         } else if self.gvar.is_some() {
             advance += self.metric_deltas_from_gvar(glyph_id)[1];
         }
-        Some(self.fixed_scale.apply(advance))
+        // FT casts advance to FT_UShort after applying variation delta and
+        // before applying scale.
+        // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/sfnt/ttmtx.c#L331>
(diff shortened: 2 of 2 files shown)


skrifa/src/instance.rs
```diff
@@ -54,20 +54,20 @@ impl Size {
...
+            //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3596>
+            // 2) this value is divided by UPEM:
+            //    (here, scaled_h=height and h=upem)
+            //    <https://gitlab.freedesktop.org/freetype/freetype/-/blob/49781ab72b2dfd0f78172023921d08d08f323ade/src/base/ftobjs.c#L3312>
+            Some(Fixed::from_bits((self.0 * 64.) as i32) / Fixed::from_bits(units_per_em as i32))
         } else {
-            // This is an identity scale for the pattern
-            // `mul_div(value, scale, 64)`
-            Fixed::from_bits(0x10000 * 64)
+            // Just pass through in unscaled case. Avoids potential overflow
```
drott, line 68: Not quite clear from the comment how returning None here is a pass through? Is that handled in the caller?

skrifa/src/metrics.rs
```diff
@@ -308,7 +308,10 @@ impl<'a> GlyphMetrics<'a> {
         } else if self.gvar.is_some() {
             advance += self.metric_deltas_from_gvar(glyph_id)[1];
         }
-        Some(self.fixed_scale.apply(advance))
+        // FT casts advance to FT_UShort after applying variation delta and
```
drott, line 311: For which cases is this FT emulation needed, and do we still provide correct behavior? Does it lead to an intentional truncation? 

Background: We are just discussing a new COLRv1 lack of precision issues due to FT's limited internal representation around variation deltas:
https://gitlab.freedesktop.org/freetype/freetype/-/issues/1257
See also
https://gitlab.freedesktop.org/freetype/freetype/-/commit/1178227b39a4eeecbefad226d4a753b345a16eb1

skrifa/src/instance.rs
```diff
@@ -54,20 +54,20 @@ impl Size {
     }
```
drott, line 54: Which tests are covering these behavior changes? Fauntlet runs?

skrifa/src/metrics.rs
dfrg: This was actually triggered by a case where FT wraps a negative value. This one is difficult to accommodate in testing because the scale is applied after the fact. 

And yes, I’ve been tracking that issue and discussed it with Rod yesterday. We agreed that COLR seems like a good place to experiment with breaking compatibility for more accurate results so we’d be fine with making those changes here.

skrifa/src/instance.rs
dfrg: I’ll update the comment to reflect this.

skrifa/src/metrics.rs
drott: The issue I was quoting from FT originates from COLRv1 but applies to all variation values (MVAR, HVAR, etc., [listed in the issue](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1257#note_2114118)) that use `tt_var_get_item_delta` - is that where the truncations happens? 

What kind of value was that that was wrapped? This really sounds like a bug in FT - I lean towards not emulating that.

drott: Further thinking: If that issue in FT was fixed (for example running FT with @tursunova's patch (which still needs a bit of work, but should cover most cases)) - would then the fauntlet runs and FT match with regards to this line?

dfrg: That patch doesn’t seem to touch the unsigned short cast so we’d still have a diff. This case might be a buggy font because it generates a negative advance width. I think those are generally unexpected in text layout. 

On the broader question, we do FT style truncation for deltas at the call site rather than in the IVS code so we can adjust as needed. We do also have the rounding though, which I ported from your 64 bit accumulation patch.

drott: > which I ported from your 64 bit accumulation patch.

Which one do you mean, or which point in the code in FT is that?

> On the broader question, we do FT style truncation for deltas at the call site rather than in the IVS code so we can adjust as needed. 

I have trouble understanding that sentence. What do you mean by "adjust as needed" in this context?

dfrg: That cast occurs in `tt_face_get_metrics` which returns the advance through a pointer to `FT_UShort`.

dfrg: I meant the rounding in FT_MulAddFix here: https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/src/base/ftcalc.c?ref_type=heads#L1083

wrt call site vs IVS, we don’t have any generic “item delta” function that does truncation. Wherever we are applying a delta we truncate explicitly, usually with a comment that this is done to match FT. For example: https://github.com/googlefonts/fontations/blob/d4651c5e31e81dbddac66163de66ab7fdf84488c/skrifa/src/metrics.rs#L304

drott:
With this applied, when comparing advances of FT for the COLRv1 test font, in the `colrv1.cpp` gm test in Skia, at font size 120, I get

**Fontations:**
```
Glyph advance for codepoint 984320 0xf0500: 119.999893
[...]
Glyph advance for codepoint 984328 0xf0508: 119.999893
```

**FreeType:**
```
Glyph advance for codepoint 984320 0xf0500: 120.000000
[...]
Glyph advance for codepoint 984328 0xf0508: 120.000000
```

In `hmtx` these glyphs have width 1000, units per em is 1000.

drott:
> Hrm, yes, this patch matches linearHoriAdvance. I’m guessing Skia is pulling from the glyph slot advance field which is calculated differently and rounded somewhere along the way.

Yes, looks like it
```
mx.advance.fX =  SkFDot6ToFloat(fFace->glyph->advance.x);
```
in [`SkFontHost_FreeType.cpp:315`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/src/ports/SkFontHost_FreeType.cpp;l=1315)

drott:
> By “haven’t hooked it up” I mean that field is an option and is not populated yet :(

Not to rush, but any updates on that? I was feeling like I needed a break from bitmap glyphs for a day and came back to this.

dfrg:
Yes, this is computed now for TrueType outlines. It will still be `None` for CFF because that scaler doesn’t modify the advance width.

drott:
> Yes, this is computed now for TrueType outlines.

Thanks for the update. Do you mean in shipping/merged code? Is this in `GlyphMetrics::advance_width` or in `AdjustedMetrics`?

dfrg:
It’s in `AdjustedMetrics` and published to crates.io.

PR #697, opened 2023-11-09
[skrifa] floating point COLR

dfrg:
Internal interface for color outline loading in full floating point.

This copies `ColrInstance` & co from read-fonts to skrifa and updates it to process all values/deltas in floating point. Additionally, adds methods to `ItemVariationStore` to compute scalars/deltas in floating point. It's f32 all the way down.

This doesn't remove the code in read-fonts and also contains a module for printing the paint graph so it's not ready to merge. 

Printing the graph for a glyph:
```rust
use skrifa::prelude::*;
use skrifa::raw::TableProvider;

let font_data = std::fs::read("/path/to/colrv1font.ttf").unwrap();
let font = FontRef::new(&font_data).unwrap();
let colr = font.colr().unwrap();
let coords = [NormalizedCoord::from_f32(-0.6021); 64];
skrifa::scale::print_color_glyph(&colr, GlyphId::new(94), LocationRef::new(&coords));
```

For the COLRv1 test font, glyph 94 (radial_horizontal_gradient_extend_mode_pad) with all coords set to -0.6021 loads as follows with skrifa maintaining fractional values from deltas:

freetype:
```
Glyph 2
    RadialGradient p0: (-202.000000,-102.000000) r0: -502.000000 p1: (98.000000,-102.000000) r1: -402.000000 extend: Pad stops: (-1.204224 3 1.000000), (-0.704224 9 1.000000), (-0.204224 0 1.000000)
```

skrifa:
```
Glyph GID_2
    RadialGradient p0: (-202.11182,-102.11182) r0: -502.11182 p1: (97.88818,-102.11182) r1: -402.11182 extend: Pad stops: (-1.2042236 3 1), (-0.70422363 9 1), (-0.20422363 0 1)
```

### DIFF
--- a/read-fonts/src/tables/variations.rs
+++ b/read-fonts/src/tables/variations.rs
@@ -577,6 +577,66 @@ impl<'a> ItemVariationStore<'a> {
         }
         Ok(((accum + 0x8000) >> 16) as i32)
     }
+
+    /// Computes the delta value in floating point for the specified index and set
+    /// of normalized variation coordinates.
+    ///
+    /// This requires a type parameter specifying the type of target value for the
+    /// delta.
+    pub fn compute_delta_f32<T: ItemDeltaTarget>(
+        &self,
+        index: DeltaSetIndex,
+        coords: &[F2Dot14],
+    ) -> Result<f32, ReadError> {
+        let data = match self.item_variation_data().get(index.outer as usize) {
+            Some(data) => data?,
+            None => return Ok(0.0),
+        };
+        let regions = self.variation_region_list()?.variation_regions();
+        let region_indices = data.region_indexes();
+        // Compute deltas in 32-bit floating point.
+        let mut accum = 0f32;
+        for (i, region_delta) in data.delta_set(index.inner).enumerate() {
+            let region_index = region_indices
+                .get(i)
+                .ok_or(ReadError::MalformedData(
+                    "invalid delta sets in ItemVariationStore",
+                ))?
+                .get() as usize;
+            let region = regions.get(region_index)?;
+            let scalar = region.compute_scalar_f32(coords);
+            accum += T::delta_to_f32(region_delta) * scalar;
+        }
+        Ok(accum)
+    }
+}
+
+pub trait ItemDeltaTarget {
+    fn delta_to_f32(delta: i32) -> f32;
+}
+
+impl ItemDeltaTarget for Fixed {
+    fn delta_to_f32(delta: i32) -> f32 {
+        Fixed::from_bits(delta).to_f32()
+    }
+}
+
+impl ItemDeltaTarget for FWord {
+    fn delta_to_f32(delta: i32) -> f32 {
+        delta as f32
+    }
+}
+
+impl ItemDeltaTarget for UfWord {
+    fn delta_to_f32(delta: i32) -> f32 {
+        delta as f32
+    }
+}
+
+impl ItemDeltaTarget for F2Dot14 {
+    fn delta_to_f32(delta: i32) -> f32 {
+        F2Dot14::from_bits(delta as i16).to_f32()
+    }
 }
 
 impl<'a> VariationRegion<'a> {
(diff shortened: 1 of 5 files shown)


drott:
Thanks for this - the fact that FreeType returns rounded values is the bug in FreeType at the moment that [@tursunova is in the process of fixing](https://gitlab.freedesktop.org/freetype/freetype/-/merge_requests/309)  - so it's a bit hard to tell how war the fixed computation would be off.  I am a bit concerned about the code duplication - if it's COLR related, could we perhaps shift read_fonts to do this in floats right away? I don't think we have a huge legacy there from FreeType (rather, we have bugs that we need to fix, like https://bugs.chromium.org/p/chromium/issues/detail?id=1473020

dfrg:
I am also curious to see how the issue will be resolved in FreeType and how big the diffs might be.

re duplication: if we go this route, my intent was to just remove this functionality from read-fonts. I think your new `ColorPainter` code does a better job of providing a nice semantic model of a paint graph. For people wanting low level access, that would still be available from our codegen'd types in read-fonts.

skrifa/src/scale/colr/instance.rs
```diff
@@ -0,0 +1,653 @@
+//! COLR table instance.
```
drott, line 1: What about putting this in `skrifa/src/colr/instance.rs`? Or `skrifa/src/color/instance.rs` - My thinking is: We're not really scaling here, mostly applying variations. And then `ColorPainter` can go in the same directory when it is ready.

skrifa/src/scale/colr/instance.rs
```diff
@@ -0,0 +1,653 @@
...
+                };
+                *delta = compute_delta(ty, delta_ix);
+            }
+        }
+        deltas
+    }
+}
+
+#[derive(Copy, Clone)]
+enum DeltaTy {
```
drott, line 92: Does `Ty` mean Type? Is that a convention we're using. Can we spell it out? `DeltaType`?

skrifa/src/scale/colr/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod instance;
```
drott, line 1: In line with the comment above, move this to `src/colr/mod.rs`?

read-fonts/src/tables/variations.rs
```diff
@@ -577,6 +577,66 @@ impl<'a> ItemVariationStore<'a> {
         }
         Ok(((accum + 0x8000) >> 16) as i32)
     }
+
+    /// Computes the delta value in floating point for the specified index and set
+    /// of normalized variation coordinates.
+    ///
+    /// This requires a type parameter specifying the type of target value for the
+    /// delta.
+    pub fn compute_delta_f32<T: ItemDeltaTarget>(
```
drott, line 586: Is it possible to add some standalone file level tests for this?

skrifa/src/scale/colr/instance.rs
dfrg: This seems reasonable but the details are a bit tricky. I’m thinking forward to adding bounding box computation which will require the scaling code. As of now, the “metadata level” types don’t have any dependencies on the scale module and I find that to be a clean separation. I suppose we can maintain that by just taking a closure for computing per glyph bounding boxes. Let me put together a quick proposal for this API.

skrifa/src/scale/colr/instance.rs
dfrg: Oops, yes. Since `type` is a reserved word in rust, this is fairly idiomatic and I follow it by habit but I’m happy to spell it out.

read-fonts/src/tables/variations.rs
dfrg: Will do!

skrifa/src/scale/colr/instance.rs
dfrg: Done

read-fonts/src/tables/variations.rs
dfrg: We don't have a reference for correctness here so I added a sanity test to ensure that the floating point results are within a reasonable margin to the delta produced by the FT compatible computation. Specifically, it ensures that either the rounded or truncated floating point delta matches the FT compatible delta. We have to account for both because fractional accumulation can push us to either side of the 0.5 boundary in some cases.

skrifa/src/scale/colr/instance.rs
dfrg: #702 is a proposal for this that tries to fit it into the current design for metadata access. I see your point about the matrix library, but it seems like `ColorPainter` needs the struct regardless as a parameter for transform callbacks. I believe your code already has the constructors for translate/rotate/skew, so we'd really only need to add `mat * mat` and `mat * point` operators to make this work.

drott:
Can we close this, is this superceded by #725?

PR #698, opened 2023-11-13
Packing fixups

cmyr:
Two little changes motivated by investigating https://github.com/googlefonts/fontc/issues/559.

- the first tweaks how we assign the 'table type' for extension subtables, for a better debugging experience,
- and the second modifies some assertions in our graph packing code

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -862,7 +862,7 @@ impl Graph {
     fn actually_promote_subtables(&mut self, to_promote: &[ObjectId]) {
         fn make_extension(type_: LookupType, subtable_id: ObjectId) -> TableData {
             const EXT_FORMAT: u16 = 1;
-            let mut data = TableData::new(type_.promote().into());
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -862,7 +864,7 @@ impl Graph {
     fn actually_promote_subtables(&mut self, to_promote: &[ObjectId]) {
         fn make_extension(type_: LookupType, subtable_id: ObjectId) -> TableData {
             const EXT_FORMAT: u16 = 1;
-            let mut data = TableData::new(type_.promote().into());
+            let mut data = TableData::new(TableType::Named("ExtensionPosFormat1"));
```
rsheeter, line 867: The named by string scares me, we can't use some sort of type here?

rsheeter (approved):

write-fonts/src/graph.rs
cmyr: This is just used for diagnostics, and in every other case is set during codegen.

PR #699, opened 2023-11-14
[codegen] derive Clone for format tables

dfrg:
This derives `Clone` for format switching tables matching what we do for other records and tables.

Been meaning to fix this for a while and I've been coding around it, but I now foresee a future issue with variable color stops where this will be a real pain.

If it looks good, JMM.


cmyr (approved):
Totally, looks like it's just an oversight.

PR #700, opened 2023-11-14
Re-enable clippy and fix lints

### LINKED ISSUE #651: turn clippy back on in CI once upstream bug gets fixed
https://github.com/googlefonts/fontations/pull/650

> There's a new clippy with new lints, but when we fix these lints we're crashing the compiler, so let's just turn it off until this gets resolved.
> I've opened https://github.com/rust-lang/rust-clippy/issues/11619 upstream, and we can turn clippy back on when that gets resolved.

dfrg:
Found a workaround for the clippy crash so this re-enables it in CI and fixes issues that snuck in while it was disabled.

Fixes #651

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,20 +30,19 @@ jobs:
       - name: check no println! or eprintln! statements
         run: resources/scripts/check_no_println.sh
 
-  # disabled temporarily pending fixing a crash in clippy
-  #clippy-lint:
-    #name: Clippy lints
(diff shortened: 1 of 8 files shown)


cmyr (approved):
looks good, maybe open an issue to remove these `allow()` attributes next week?

dfrg:
> looks good, maybe open an issue to remove these `allow()` attributes next week?

clippy wants me to use `println`/`eprintln` :(

PR #701, opened 2023-11-14
[read-fonts] Decoding Device table values

cmyr:
This is not something we really expect to see in the wild very much, but it feels like something we will want for completeness in any case, so here we are.

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -236,6 +236,53 @@ impl ClassDef<'_> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
I didn't check the arithmatic in great detail but looks reasonable at a glance. Checking against how font tools understands your values to inform the tests might be helpful?

cmyr:
The values in the tests come from the spec, so I'm reasonably confident that this is working as intended. I looked at fonttools but there doesn't seem to be a trivial way to generate a device table directly from binary without having compiled a full font.

PR #702, opened 2023-11-15
[skrifa] public color API

dfrg:
https://github.com/googlefonts/fontations/pull/697#discussion_r1394224889 requests moving the color instance code from the scale module to the top level "metadata section" of the crate. This PR contains a proposal for such an API that meshes well with the current design of `MetadataProvider`.

This depends on #697 but is pulled out here for separate discussion.

One key point is the method for bounding box computation. It'd be nice to avoid a hard dependency on anything in the `scale` module so the function here takes a closure for that purpose.

@drott thoughts? There's some overlap with your `ColorPainter` work but I think this provides a nice entry point for that.

### DIFF
--- a/skrifa/src/color/mod.rs
+++ b/skrifa/src/color/mod.rs
@@ -20,26 +20,29 @@ pub trait ColorPainter {}
 /// Affine transformation matrix.
 #[derive(Copy, Clone, Debug)]
 pub struct Transform {
-    xx: f32,
-    yx: f32,
-    xy: f32,
-    yy: f32,
-    dx: f32,
-    dy: f32,
+    pub xx: f32,
+    pub yx: f32,
+    pub xy: f32,
+    pub yy: f32,
+    pub dx: f32,
+    pub dy: f32,
 }
 
 /// Reference to a paint graph that represents a color glyph outline.
 #[derive(Clone)]
 pub struct ColorOutline<'a> {
     colr: colr::Colr<'a>,
-    glyph_id: GlyphId,
     kind: ColorOutlineKind<'a>,
 }
 
 impl<'a> ColorOutline<'a> {
-    /// Returns the glyph identifier that was used to retrieve this outline.
-    pub fn glyph_id(&self) -> GlyphId {
-        self.glyph_id
+    /// Returns the version of the color table from which this outline was
+    /// selected.
+    pub fn version(&self) -> u32 {
+        match &self.kind {
+            ColorOutlineKind::V0(..) => 0,
+            ColorOutlineKind::V1(..) => 1,
+        }
     }
 
     /// Evaluates the paint graph at the specified location in variation space
@@ -87,19 +90,14 @@ enum ColorOutlineKind<'a> {
 /// Collection of color outlines.
 #[derive(Clone)]
 pub struct ColorOutlineCollection<'a> {
-    glyph_count: u16,
     colr: Option<colr::Colr<'a>>,
 }
 
 impl<'a> ColorOutlineCollection<'a> {
     /// Creates a new color outline collection for the given font.
     pub fn new(font: &impl TableProvider<'a>) -> Self {
-        let glyph_count = font
-            .maxp()
-            .map(|maxp| maxp.num_glyphs())
-            .unwrap_or_default();
         let colr = font.colr().ok();
-        Self { glyph_count, colr }
+        Self { colr }
     }
 
     /// Returns the color outline for the given glyph identifier.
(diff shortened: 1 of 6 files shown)


skrifa/src/color/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Evaluates the paint graph at the specified location in variation space
+    /// and returns the bounding box for the full scene.
+    ///
+    /// The `glyph_bounds` closure will be invoked for each clip node in the
+    /// graph and should return the bounding box for the given glyph, location
+    /// and transform.
+    pub fn bounding_box(
+        &self,
+        location: impl Into<LocationRef<'a>>,
+        glyph_bounds: impl FnMut(GlyphId, &LocationRef, Transform) -> Option<BoundingBox>,
```
drott, line 54: In trying to understand your intention, first some questions: So the difference from a "degenerate" `ColorPainter` implementation on the client side (to build a bounding box there) would be that the transform matrix is accumulated internally and we would call back less frequently, not for each node in the graph? Since there is no mention in the comment about what to do with compositions where the path through the graph forks, do you mean this to track the accumulated transform internally, and call back for each clip node with a fully spelled-out transform? 

A previously configured, client-side transform (on the viewport for example) would be taken into account by the client multiplying that in when it receives the callback? 

So the client would perform own path retrieval, combine its own possible viewport transform with the accumulated transform from the callback, and compute the bounding box of the path, with transform applied, and return that? 

I do appreciate that this would disconnect the color module from the scale module - which I think make sense, as we discussed in the review of 

This works as an interface, and we can shift slightly more code to the Rust side. I am trying to weigh whether this somewhat unusual API pattern is worth defining in addition to the to the paint traversal call, with which the client can compute a bounding box as well: it receives more callback and the matrix tracking would need to happen on the client side, but since we have that on the 2D graphics library side - that's doable there just as well.

skrifa/src/color/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+        let glyph_count = font
+            .maxp()
+            .map(|maxp| maxp.num_glyphs())
+            .unwrap_or_default();
+        let colr = font.colr().ok();
+        Self { glyph_count, colr }
+    }
+
+    /// Returns the color outline for the given glyph identifier.
+    pub fn get(&self, glyph_id: GlyphId) -> Option<ColorOutline<'a>> {
```
drott, line 106: I think the direct access is more important than the iterator - so perhaps we can do without the `maxp()` access? 

As discussed, in https://github.com/googlefonts/oxidize/pull/48 - we need the (cheaply acquirable) information in Skia whether a glyph is COLRv0 and COLRv1. I think that's doable well with the type-based distinction of `ColorOutlineKind`.

skrifa/src/color/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Evaluates the paint graph at the specified location in variation space
+    /// and emits the results to the given painter.
+    pub fn paint(
+        &self,
+        location: impl Into<LocationRef<'a>>,
+        painter: &mut impl ColorPainter,
+    ) -> Result<(), ReadError> {
+        let instance = instance::ColrInstance::new(self.colr.clone(), location.into().coords());
+        match &self.kind {
```
drott, line 67: I am aware this is a draft - but how where does `self.kind` get initialized here - that would be a `ColorOutlineKind` retrieved for `self.glyph_id`, right?

skrifa/src/provider.rs
```diff
@@ -47,6 +48,11 @@ pub trait MetadataProvider<'a>: raw::TableProvider<'a> + Sized {
     fn charmap(&self) -> Charmap<'a> {
         Charmap::new(self)
     }
+
+    /// Returns the collection of color outlines.
+    fn color_outlines(&self) -> ColorOutlineCollection<'a> {
```
drott, line 53: Yes, I think this works as a good entry point. And is perhaps a bit more idiomatic than just a "do_traverse_paint" function that takes a glyph id somewhere. It also answers the "what glyph format is available" question.

drott (commented):
Thanks for this proposal, some thoughts below.

skrifa/src/color/mod.rs
dfrg: The answer to all of your questions is “yes” with the exception that this wouldn’t do any caching of paths internally although the client could certainly do so if they’d like to trade space for time. Seems to be a point in favor of this approach?

I hadn’t considered a client provided root transform. We can accept that as a parameter here and fold it in if you think that’d be more useful. Otherwise, yes, a client could apply it in the callback.

I have two general concerns:

1. Handling edge cases (composition nodes, detecting unbounded subgraphs) is not exactly trivial so it would be nice to offer a correct implementation of this instead of pushing that burden to all potential clients. 

2. While not a top priority atm, I’m trying to keep the embedded case in mind wrt heap allocations. This interface would allow the caller to decide how to handle memory requirements for loading glyphs. Aside: that does leave the recursion blacklist as a potential problem— Rod and I discussed adding a cfg feature that uses a max recursion depth instead but that’s not something we need to address now.

skrifa/src/color/mod.rs
dfrg: I’d like to keep the iterator but I can implement it without maxp. Good catch.

dfrg: Re detection: I can add a `version()` method to `ColorOutline` that returns 0 or 1

skrifa/src/color/mod.rs
dfrg: The kind field is initialized in `ColorOutlineCollection::get`. Yes, it is the outline for the associated glyph_id. I added that field (and accessor) to support iteration since the table can be sparse. It might make sense to remove that and change the iterator to yield a tuple `(GlyphId, ColorOutline)`

skrifa/src/provider.rs
dfrg: Thanks. I’m trying to keep the API as consistent as possible. Fonts don’t make it easy 🤷

PR #707, opened 2023-11-16
[skrifa] changes in support of TT hinting

dfrg:
1. Capture required hinting tables (fpgm, prep, cvt)
2. Make hinting specific memory calculations dependent on a parameter at memory request time rather than at _outline request time_

### DIFF
--- a/skrifa/src/scale/glyf/glyph.rs
+++ b/skrifa/src/scale/glyf/glyph.rs
@@ -48,13 +48,13 @@ pub struct ScalerGlyph<'a> {
 impl<'a> ScalerGlyph<'a> {
     /// Returns the minimum size in bytes required to scale an outline based
     /// on the computed sizes.
-    pub fn required_buffer_size(&self) -> usize {
+    pub fn required_buffer_size(&self, with_hinting: bool) -> usize {
         let mut size = 0;
+        let hinting = with_hinting && self.has_hinting;
         // Scaled, unscaled and (for hinting) original scaled points
         size += self.points * size_of::<Point<F26Dot6>>();
         // Unscaled and (if hinted) original scaled points
-        size +=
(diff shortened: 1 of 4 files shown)


skrifa/src/scale/glyf/glyph.rs
```diff
@@ -48,13 +48,13 @@ pub struct ScalerGlyph<'a> {
 impl<'a> ScalerGlyph<'a> {
     /// Returns the minimum size in bytes required to scale an outline based
     /// on the computed sizes.
-    pub fn required_buffer_size(&self) -> usize {
+    pub fn required_buffer_size(&self, with_hinting: bool) -> usize {
```
rsheeter, line 51: I would prefer an enum to a bool; seeing `required_buffer_size(true)` isnt' very intelligible

skrifa/src/scale/glyf/mem.rs
```diff
@@ -190,8 +190,8 @@ mod tests {
         };
         // Required size adds 4 bytes slop to account for internal alignment
         // requirements. So subtract 5 to force a failure.
-        let not_enough = outline_info.required_buffer_size() - 5;
+        let not_enough = outline_info.required_buffer_size(false) - 5;
```
rsheeter, line 193: Hints::Off would be so much clearer :)

rsheeter (approved):

skrifa/src/scale/glyf/mem.rs
dfrg: Agreed :) I'm planning changes to our `Hinting` type which could be used for this purpose. In the meantime, I'd prefer not to conjure a new type just to throw it away tomorrow. Will make sure to address this in a follow up.

dfrg: #708 so I don't forget

Issue #708, opened 2023-11-16
[skrifa/glyf] pass hinting mode instead of bool to buffer size method


dfrg:
Hints::Off would be so much clearer :)

_Originally posted by @rsheeter in https://github.com/googlefonts/fontations/pull/707#discussion_r1396448284_

PR #710, opened 2023-11-17
[skrifa] hinting mode changes/remove feature gate

dfrg:
- removes the hinting feature gate
- adds a `None` variant to `Hinting` enum to use as a replacement for `Option<Hinting>` so call sites taking this as a parameter have greater clarity
- updates the TrueType memory functions to take `Hinting` rather than `bool`

Fixes https://github.com/googlefonts/fontations/issues/708

Depends on #709


cmyr (approved):
This looks like a nice clarifying change :)

PR #712, opened 2023-11-20
[chore] Symlink license files to crate roots

### LINKED ISSUE #711: Publish LICENSE files for read-fonts and font-types
Compare https://bugs.chromium.org/p/chromium/issues/detail?id=1503143
Currently inquiring from @danakj on the issue where these files should become visible for our checker tool to pick them up.

cmyr:
This will ensure that license files are packaged with the crate when it is published, which is required by a downstream user.

- ~didn't bother doing this for `write-fonts`, since I do not anticipate that being used by chrome in the near future.~
- ~QQ: should/can these just be symlinks? I'm not sure how that works with git and searching was inconclusive.~
- fixes #711 

So my first version of this was just copying files, and this version is symlinking. @anthrotype points out that symlinks may [cause some issues on windows](https://www.scivision.dev/git-windows-symlink/)? I'm not sure how much windows use we expect, and there is an obvious fix available, so my instinct is to just go with this.

### DIFF
--- a/font-types/LICENSE-APACHE
+++ b/font-types/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
--- a/font-types/LICENSE-MIT
+++ b/font-types/LICENSE-MIT
@@ -1,0 +1,1 @@
+../LICENSE-MIT
--- a/read-fonts/LICENSE-APACHE
+++ b/read-fonts/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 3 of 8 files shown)


anthrotype:
> should/can these just be symlinks? I'm not sure how that works with git 

maybe this can help? 
https://www.scivision.dev/git-windows-symlink/
https://www.scivision.dev/windows-symbolic-link-permission-enable

anthrotype:
LGTM

+1 for using symlinks, we can add the above instructions for windows users if they have troubles cloning

PR #716, opened 2023-11-21
[read/write-fonts] make STAT offset_to_axis_values nullable

anthrotype:
Fixes https://github.com/googlefonts/fontc/issues/582

OTS gives a warning that our offset_to_axis_values is != 0 even though the count of axis_values == 0. This fixes it.

NOTE: this is a breaking change in that read-fonts offset_to_axis_values now returns an `Option<Result<...>>` instead of just a Result.

### DIFF
--- a/read-fonts/generated/generated_stat.rs
+++ b/read-fonts/generated/generated_stat.rs
@@ -121,13 +121,13 @@ impl<'a> Stat<'a> {
     /// start of the design axes value offsets array. If axisValueCount
     /// is zero, set to zero; if axisValueCount is greater than zero,
     /// must be greater than zero.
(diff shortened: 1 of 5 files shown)


anthrotype:
codegen CI failed.. ok yeah, I cheated. I modified the generated_stat.rs manually and then backfixed the font-codegen stat file 😅 
I'll fix it

anthrotype:
Thanks @cmyr for the quick help getting un-stuck. I just squashed. Also, can confirm that after applying this patch and building GS locally with fontc, OTS is now happy and no longer throws a warning about non-zero offset to axis value array.

cmyr (approved):
Looks correct to me, welcome to codegen!

PR #717, opened 2023-11-22
bump minor version for read-fonts and write-fonts

anthrotype:
bumping minor because STAT.offset_to_axis_values is now nullable after https://github.com/googlefonts/fontations/pull/716

     Changes for read-fonts from read-fonts-v0.13.2 to 0.14.0
             f99b59d [read|write-fonts] make offset_to_axis_values nullable
     Changes for write-fonts from write-fonts-v0.18.0 to 0.19.0
             f99b59d [read|write-fonts] make offset_to_axis_values nullable
             dec4e1b [graph] Move assert to after space check
             e1b9cea [splitting] More careful in handling TableType
             48670bb [write-fonts] add SegmentMaps::is_identity method (#715)
             08d6264 [chore] Bump patch version on read-fonts & font-types
             c745d86 [chore] symlink license files to crate roots
             16daacf Bump read-fonts to 0.13.1
             d04b9bb Enable old gpos spec tests
             0a66324 [write-fonts] Full pack/split test for MarkBase
             edf8bdd [write-fonts] More & better testing of MarkBase splitting
             74bb1b0 [write-fonts] Test for splitting of mark & base arrays
             e4e1c68 [write-fonts] Preliminary impl of MarkBase splitting

### DIFF
--- a/fauntlet/Cargo.toml
+++ b/fauntlet/Cargo.toml
@@ -9,7 +9,7 @@ readme = "README.md"
 publish = false
 
 [dependencies]
-skrifa = { version = "0.13.0", path = "../skrifa" }
+skrifa = { version = "0.14.0", path = "../skrifa" }
 freetype-rs = "0.32.0"
 memmap2 = "0.5.10"
 rayon = "1.8.0"
--- a/otexplorer/Cargo.toml
+++ b/otexplorer/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 
 [dependencies]
 xflags = "0.3.0"
-read-fonts = { path = "../read-fonts",version = "0.13.2" }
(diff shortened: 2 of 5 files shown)


anthrotype:
I did not bump skrifa as well, because I wasn't sure if it was release-able or whether others are working on something that they'd like to wait before releasing.
But now I believe I should do skrifa as well, because fontc requires skrifa (or at least one of its tests does, not the main lib) and if I then bump read-fonts and write-fonts in fontc, there would be two conflicting read-fonts versions (the new one and the old requried by skrifa)... Is my understanding correct?

anthrotype:
Looks like I also have to bump skrifa, minor because read-fonts on which skrifa depends contains a breaking change. However, skrifa itself does not seem to be using the STAT code that contains the breaking change...

PR #719, opened 2023-11-23
Add trait for remapping temporary delta set index ids to their final values

cmyr:
I'm currently reworking the fea-rs API so that it has native types for variable `ValueRecord`s and `AnchorTable`s, which will let actually create the various lookup builders for marks & kerning before feature generation.

Previously fea-rs used the native write-fonts types for value records, and it was responsible for remapping the temporary indices before generating the final lookups. Now it converts to the native types later in the process, and it makes sense to just have a mechanism for walking all the lookups and remapping everything there.

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -42,6 +42,20 @@ pub struct VariationIndexRemapping {
     map: HashMap<TemporaryDeltaSetId, VariationIndex>,
 }
 
+/// Remapping temporary delta set identifiers to the final values.
+///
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -42,6 +42,20 @@ pub struct VariationIndexRemapping {
     map: HashMap<TemporaryDeltaSetId, VariationIndex>,
 }
 
+/// Remapping temporary delta set identifies to the final values.
```
anthrotype, line 45:
```suggestion
/// Remapping temporary delta set identifiers to the final values.
```

anthrotype (approved):

PR #720, opened 2023-11-27
[write-fonts] use codegen tool to generate MVAR modules

anthrotype:
a step towards https://github.com/googlefonts/fontc/issues/531


write-fonts/src/tables/mvar.rs
```diff
@@ -0,0 +1,124 @@
...
+        assert_eq!(read_metric_delta(&read, b"hasc", &[1.0]), 10.0);
+
+        assert_eq!(read.value_records()[1].value_tag(), Tag::new(b"hdsc"));
+        assert_eq!(read.value_records()[1].delta_set_outer_index(), 0);
+        assert_eq!(read.value_records()[1].delta_set_inner_index(), 0);
+        assert_eq!(read_metric_delta(&read, b"hdsc", &[0.0]), 0.0);
+        // this coincides with the peak of intermediate region r2, hence != 30.0/2
+        assert_eq!(read_metric_delta(&read, b"hdsc", &[0.5]), -20.0);
+        assert_eq!(read_metric_delta(&read, b"hdsc", &[1.0]), -30.0);
+    }
```
rsheeter, line 123: Optional, IMHO a helper or reading into a struct so you could have coarser asserts would make this easier to read. For example, `assert_value_record(what it should look like, value_record[0]);`

I find it easier to troubleshoot because when it fails I can see all the values that are wrong not just the first one.

rsheeter (approved):

PR #722, opened 2023-11-27
[skrifa] Callback-based COLRv1 implementation

drott:
Filing a pull request as an initial basis for discussion:

Known issues to be addressed:
- [x] To be hooked up likely similar to paint() method in #702 
- [x] Testing  for serialized paint graph output, idea: serialize as JSON, store expectations in files?
- [x] Design considerations for slice in Brushes, see inline comment. 
- [x] Doc comments missing.
- [x] Agree on a solution for rebaselining: Environment variable choosen as solution.
- [x] Wait for merge of underlying floating point `ColrInstance` API #725 
- [x] Do a last pass of comparing Skia results.
- [x] Check why test baselines do not match on Mac and Linux


skrifa/src/color/color.rs
```diff
@@ -0,0 +1,259 @@
...
+    transforms::Transform,
+};
+
+use crate::prelude::LocationRef;
+
+#[derive(Clone, Debug, Default)]
+// This repr(C) is required so that C-side FFI's
+// are able to cast the ColorStop slice to a C-side array pointer.
+#[repr(C)]
+pub struct ColorStop {
```
dfrg, line 20: Just a random thought: I was thinking about parameterizing this on the (palette_index, alpha) payload so the type can support both paletted and resolved colors. Not really important now but just considering the future and other potential consumers.

skrifa/src/color/color.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub offset: f32,
+    pub palette_index: u16,
+    pub alpha: f32,
+}
+
+// Design considerations: In principle, a local Vec allocation is not required if we're willing
```
dfrg, line 26: Good summary of the trade offs. In my mind, it comes down to how much massaging we want to do with color stops. If we’re doing a substantial amount of useful work here then I think it makes sense to pay the cost of a potential allocation (“potential” assuming we add a smallvec style optimization which is easily doable).

skrifa/src/color/color.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn pop_transform(&mut self);
+
+    fn push_clip_glyph(&mut self, glyph: GlyphId);
+    // Needed for pushing clip box when starting a new root glyph or PaintColrGlyph.
+    fn push_clip_rectangle(&mut self, x_min: f32, x_max: f32, y_min: f32, y_max: f32);
+    fn pop_clip(&mut self);
+
+    fn fill(&mut self, brush: Brush);
+
+    // TODO(drott): Optimization function combining clip and fill.
```
dfrg, line 80: Does this comment apply here? This looks more like a hook to allow a client to reuse already rendered subtrees? I like that idea but I do think the naming might be a bit confusing.

skrifa/src/color/color.rs
```diff
@@ -0,0 +1,259 @@
...
+    instance: LocationRef,
+    painter: &mut impl ColorPainter,
+) -> Result<(), ReadError> {
+    let colr = font.colr()?;
+
+    let color_instance = ColrInstance::new(colr, instance.coords());
+    let (root_paint, root_paint_id) =
+        (*color_instance)
+            .v1_base_glyph(glyph_id)?
+            .ok_or(ReadError::MalformedData(
```
dfrg, line 101: It’s difficult to find good options for high level errors in `ReadError`. We should probably define a new error type specifically for color.

dfrg (commented):
Thanks for taking this on! I left some comments inline but overall this is looking solid and coming together nicely.

drott:
Leaving this as note for discussion:
* TBD as well if the `ColorPainter` calls should have a `Result<>` return type?

skrifa/src/color/colrv1_traversal.rs
```diff
@@ -0,0 +1,533 @@
...
+            match (
+                resolved_stops.first().cloned(),
+                resolved_stops.last().cloned(),
+            ) {
+                (None, _) | (_, None) => {}
+                (Some(first_stop), Some(last_stop)) => {
+                    let mut color_stop_range = last_stop.offset - first_stop.offset;
+
+                    let mut start_angle_scaled =
+                        start_angle + sector_angle * resolved_stops.first().unwrap().offset;
```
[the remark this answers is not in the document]
drott: Yes! Search and replace error.

drott:
For WIP, rebase on #697.

rsheeter:
>  if the ColorPainter calls should have a Result<> return type?

What is the argument for it not to?

drott:
> What is the argument for it not to?

One consideration: They are implemented as callbacks and handled by the client directly so the client can act on errors. On the other hand, we can break if something goes wrong and channel up the error if something goes wrong. Just something I had not fully fleshed out yet.

dfrg:
If want to allow the callbacks to return errors, the error type should be declared by the `ColorPainter` trait and defined by implementors. I don’t think it makes sense for callbacks to return `ReadError` or any other type we may define.

skrifa/Cargo.toml
```diff
@@ -8,9 +8,15 @@ repository = "https://github.com/googlefonts/fontations"
 readme = "README.md"
 categories = ["text-processing", "parsing", "graphics"]
```
dfrg, line 10: I think it’d be better to remove the feature and only add serde to the dev-dependency section. This will make it available for tests but not expose it to the public API.

skrifa/Cargo.toml
```diff
@@ -8,9 +8,15 @@ repository = "https://github.com/googlefonts/fontations"
...
 
+[features]
+serde = ["dep:serde"]
+
 [dependencies]
-read-fonts = { version = "0.14.0", path = "../read-fonts" }
+read-fonts = { version = "0.14.0", path = "../read-fonts", features = ["serde"] }
```
dfrg, line 15: This pulls serde in unconditionally for all builds. Let’s remove this and add the serde feature to the read-fonts dev-dependency.

skrifa/src/color/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        (*color_instance)
+            .v1_base_glyph(glyph_id)?
+            .ok_or(ReadError::MalformedData(
+                "No COLRv1 glyph for requested glyph id.",
+            ))?;
+
+    let clipbox = get_clipbox_font_units(&color_instance, glyph_id)?;
+
+    if let Some(rect) = clipbox {
+        painter.push_clip_rectangle(rect.x_min, rect.y_min, rect.x_max, rect.y_max);
```
dfrg, line 121: Looks like the argument order is incorrect here. I noticed this when looking at the json output where the clip rects seemed wrong, so +1 for generating tests :) Maybe the clip box callback should take a `BoundingBox` parameter to avoid this in the future? In that case, I also think `push_clip_box` might be a better name.

skrifa/src/color/traversal_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+    FillBrush {
+        brush: BrushParams,
+    },
+    PushLayer {
+        composite_mode: u8,
+    },
+    PopLayer,
+}
+
+impl ColorPainter for JsonDumpColorPainter {
```
dfrg, line 135: Optional: we could just implement this directly for PaintDump.

skrifa/src/color/traversal_tests.rs
```diff
@@ -0,0 +1,341 @@
...
+                        dump.glyph_id, serde_json::to_string(&dump).unwrap(), &line.unwrap()
+                    )
+                }
+                None => panic!("Expectation not found for glyph id: {}", dump.glyph_id),
+            }
+        }
+    }
+}
+
+macro_rules! make_ident_string {
```
dfrg, line 269: Any reason for this macro rather than calling stringify! directly in the one below?

skrifa/src/color/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+            paint_colr_glyph_cycle_gid,
+            LocationRef::default(),
+            &mut color_painter,
+        )
+        .is_err());
+    }
+}
+
+#[cfg(all(test, feature = "serde"))]
+pub mod test_glyph_defs;
```
dfrg, line 233: This feels like it should be owned by the traversal_tests module. I’d suggest a new traversal _tests directory under color to group these. 

Also, there’s a general expectation that all mod declarations are grouped near the top of the file.

skrifa/src/color/color.rs
```diff
@@ -0,0 +1,259 @@
...
+        palette_index: u16,
+        alpha: f32,
+    },
+    // Normalized to a straight line between points p0 and p1,
+    // color stops normalized to align with both points.
+    LinearGradient {
+        p0: Point<f32>,
+        p1: Point<f32>,
+        color_stops: &'a [ColorStop],
+        extend: u8,
```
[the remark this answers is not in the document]
drott: That's done.

skrifa/Cargo.toml
drott: Thanks for the help with that, yes, that's better. Had some trouble figuring out the right syntax and what would be the best way to do that here. Updated, should be done.

skrifa/src/color/traversal_tests.rs
drott: At some point I just go `$glyph_set` in the output then - but trying this again, now it worked. 🤷

skrifa/src/color/traversal_tests.rs
drott: Good idea, that looks cleaner, done.

skrifa/Cargo.toml
drott: That's done now.

skrifa/src/color/colrv1_traversal.rs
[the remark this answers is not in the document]
drott: That's done.

skrifa/src/color/colrv1_traversal.rs
```diff
@@ -0,0 +1,533 @@
...
+        ResolvedPaint::Glyph { glyph_id, paint } => {
+            painter.push_clip_glyph(*glyph_id);
+            let result =
+                traverse_with_callbacks(&paint.resolve(instance)?, instance, painter, visited_set);
+            painter.pop_clip();
+            result
+        }
+        ResolvedPaint::ColrGlyph { glyph_id } => match (*instance).v1_base_glyph(*glyph_id)? {
+            Some((base_glyph, base_glyph_paint_id)) => {
+                if visited_set.contains(&base_glyph_paint_id) {
```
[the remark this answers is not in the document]
drott: Hashset insertion change done, removal still TODO.

skrifa/src/color/transforms.rs
```diff
@@ -0,0 +1,130 @@
...
+            yy: 1.0,
+            dx: 0.0,
+            dy: 0.0,
+        }
+    }
+}
+
+pub struct PaintToTransform(pub Result<Transform, ReadError>);
+
+impl From<&ResolvedPaint<'_>> for PaintToTransform {
```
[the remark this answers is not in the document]
drott: How useful :) Thanks for the suggestion, done.

skrifa/src/color/colrv1_traversal.rs
```diff
@@ -0,0 +1,533 @@
...
+                    return Err(ReadError::MalformedData(PAINTCOLRGLYPH_CYCLE_ERROR));
+                } else {
+                    visited_set.insert(paint_id);
+                }
+                traverse_with_callbacks(
+                    &layer_paint.resolve(instance)?,
+                    instance,
+                    painter,
+                    visited_set,
+                )?;
```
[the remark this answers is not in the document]
drott: Good find, now addressed in latest update, added a test glyph for that as well - which indeed failed to render before.

skrifa/src/color/colrv1_traversal.rs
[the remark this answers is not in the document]
drott: Visited set removal done, too.

drott:
Regarding the rebaselining of the tests, I know made this controlled by an environment variable. I went through various options: 
- Using a build.rs step => This becomes awkward as for regenerating everything, one would need to compile the whole color part of the crate as part of `build.rs` and then build it again.
- Using code in `examples/` - this might works, but becomes slightly awkward as the traversal test definitions are test-only and would need to be made available in the general build in my understanding.

dfrg:
I came up with the same problem when thinking about doing this with `build.rs`. Great call on the environment variable solution. This definitely feels like the best option given the constraints.

skrifa/src/color/color.rs
drott: Resolving, as we agreed on using a slice.

skrifa/src/color/color.rs
drott: Comment moved, hope that's clearer now.

dfrg:
I wonder if #692 would fix the advance differences.

drott:
> I wonder if #692 would fix the advance differences.

Good point, thanks, I'll give that a try.

drott:
> I wonder if #692 would fix the advance differences.

Cherry-picking e054febae01ad6c6c1557800d3efe56be0136900 did not fix the advances, I'll see if I can compare those in a Skia unit test against FT for example.

drott:
> 2\. There are differences in those sweep tests, where color stops are made to coincide. FreeType rounded that, but with floating point COLRv1, the stops are not aligning exactly in the same position, which can lead to ordering differences. The tests may need adjustment to achieve that.

Confirmed that there are ordering differences, which disappear if I change the sorting function to ignore small differences (just for debugging). I think one option to keep testing this behavior is to make new sweep tests in the test font that position color stops at coinciding locations instead of relying on the variable parameters for that. 

> There seems to be a small advance difference, 

Analysis on advances in:
https://github.com/googlefonts/fontations/pull/692#issuecomment-1847197599
With clamping the advances to the same as FT, I can further bring down the pixel diff.

In one extend mode tests with color stops shifted using variable axes, there are differences, which I also with high confidence attribute to color stop floating point vs. rounding in FT differences:

image

So I don't think there are any blockers to this change anymore. In chat, @dfrg explained that he'd like to do another review pass, and that he had some concerns on naming and Rust style - I'll stay tuned for those.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+#[repr(C)]
+#[derive(PartialEq)]
+#[cfg_attr(test, derive(Serialize, Deserialize))]
+/// A color stop of a gradient. All gradient callbacks of [`ColorPainter`] normalize color stops to be in the range of 0
+/// to 1. `palette_index` specifies a color from the CPAL table, to be multiplied with `alpha` before use.
+pub struct ColorStop {
+    pub offset: f32,
+    pub palette_index: u16,
+    pub alpha: f32,
+}
```
dfrg, line 105:
```suggestion
/// A color stop of a gradient.
///
/// All gradient callbacks of [`ColorPainter`] normalize color stops to be in the range of 0
/// to 1.
#[derive(Clone, PartialEq, Debug, Default)]
#[cfg_attr(test, derive(Serialize, Deserialize))]
// This repr(C) is required so that C-side FFI's
// are able to cast the ColorStop slice to a C-side array pointer.
#[repr(C)]
pub struct ColorStop {
    pub offset: f32,
    /// Specifies a color from the `CPAL` table.
    pub palette_index: u16,
    /// Additional alpha value, to be multiplied with the color above before use.
    pub alpha: f32,
}
```
Separated the brief comment as above. The general order is doc comment, derives, repr. Also added doc comments to the fields rather than specifying them at the type level.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+// stops to client side and have the client sort the collected items once
+// again. If we do want to pre-ort them, and still use use an
+// `Iterator<Item=ColorStop>`` instead as the `color_stops` field, then we would
+// need a Fontations-side allocations to sort, and an extra allocation on the
+// client side to `.collect()` from the provided iterator before passing it to
+// drawing API.
+//
+/// Used for encapsulating the different types of fills that may occur in a
+/// COLRv1 glyph. The client receives the information about the fill type in the
+/// [`fill``](ColorPainter::fill) callback of the [`ColorPainter`] trait.
```
dfrg, line 122: Same thing with separating first sentence into it's own line.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+    /// [`push_clip_box`](ColorPainter::push_clip_box) call.
+    fn pop_clip(&mut self);
+
+    /// Fill the current clip area with the specified gradient fill.
+    fn fill(&mut self, fill_type: FillType);
+
+    /// Optionally implement this method: Draw an unscaled COLRv1 glyph given
+    /// the current transformation matrix (as accumulated by
+    /// [`push_transform`](ColorPainter::push_transform) calls).
+    fn draw_color_glyph(&mut self, _glyph: GlyphId) -> Result<ColrGlyphDrawResult, ReadError> {
```
dfrg, line 206: As above, I think `paint_cached_color_glyph` better captures the intent. This should also have the same error type as `paint`.

skrifa/src/color/colrv1_traversal.rs
```diff
@@ -0,0 +1,520 @@
...
+                point_normalized *= dot_product(vector, point) / length;
+                point_normalized
+            };
+
+            let mut resolved_stops = make_sorted_resolved_stops(color_stops, instance);
+
+            // If p0p1 or p0p2 are degenerate probably nothing should be drawn.
+            // If p0p1 and p0p2 are parallel then one side is the first color and the other side is
+            // the last color, depending on the direction.
+            // For now, just use the first color. The client should draw a single color if the color stops array only has 1 entry.
```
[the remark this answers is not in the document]
drott: Let's go with that. I'll see if I can invent a test glyph to cover that.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
-//! Color outlines.
+//! Drawing color glyphs.
+//!
+//! TODO(add example, including provider, getcolorpaintables)
+pub mod instance;
+pub mod transforms;
 
-// Temporary until remaining API is filled in.
-#![allow(dead_code)]
+pub(crate) mod colrv1_traversal;
```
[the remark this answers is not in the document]
drott: > I believe these modules can all be private?

Done.

> Nitpick on naming: transforms -> transform and colrv1_traversal -> traversal (I think colrv1 is implied and we'll likely have v2 at some point)

Renamed according to your suggestion.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+
+use self::instance::{resolve_paint, PaintId};
+
+/// An error during drawing a COLR glyph. This covers inconsistencies
+/// in the COLRv1 paint graph as well as downstream
+/// parse errors from read-fonts.
+#[derive(Debug, Clone)]
+pub enum ColorError {
+    ParseError(ReadError),
+    NoColrV1GlyphForId(GlyphId),
```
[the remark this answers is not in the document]
drott: `GlyphNotFound`

skrifa/src/color/mod.rs
drott: Got it, incorporated suggestion.

skrifa/src/color/mod.rs
drott: Renamed to `paint_cached_color_glyph`.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,437 @@
...
+            ColorPaintableRoot::V1Paint(..) => ColorPaintableType::ColrV1,
+        }
+    }
+
+    /// Returns the bounding box. For COLRv1 paintables, this is clipbox of the
+    /// specified COLRv1 glyph, or `None` if there is
+    /// none for the particular glyph.  The `size` argument can optionally be used
+    /// to scale the bounding box to a particular font size. `location` allows
+    /// specifycing a variation instance.
+    pub fn get_bounding_box(
```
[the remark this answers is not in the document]
drott: Renamed to `bounding_box`.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,437 @@
...
+        &self,
+        location: impl Into<LocationRef<'a>>,
+        size: Option<f32>,
```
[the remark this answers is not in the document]
drott: Changed to `Size` - but my criticism on `Size` accepting a 0 and using that to mean unscaled still stands. IMO Size should have a special constructor for expressing Unscaled and a size of 0 should just be one that makes everything be zero. Otherwise we have to special case incoming client size 0 sizes in various places, like I am doing now in this method. Having a client side size 0 become unscaled accidentally is surprising behavior.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,437 @@
...
+    ) -> Result<Option<BoundingBox<f32>>, ColorError> {
```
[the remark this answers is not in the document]
drott: Makes sense. Probably too granular to distinguish reasons for why we can't have a clip box. Simplified to `Option<>`

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,437 @@
...
+                ColorPaintableRoot::V1Paint(paint, paint_id, glyph_id, self.upem.clone())
+            }
+        };
+        Some(ColorPaintable {
+            colr,
+            root_paint_ref,
+        })
+    }
+
+    pub fn get_most_expressive(&self, glyph_id: GlyphId) -> Option<ColorPaintable<'a>> {
```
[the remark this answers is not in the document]
drott: done.

skrifa/src/color/mod.rs
drott: Done.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+// `Iterator<Item=ColorStop>`` instead as the `color_stops` field, then we would
+// need a Fontations-side allocations to sort, and an extra allocation on the
+// client side to `.collect()` from the provided iterator before passing it to
+// drawing API.
+//
+/// Used for encapsulating the different types of fills that may occur in a
+/// COLRv1 glyph. The client receives the information about the fill type in the
+/// [`fill``](ColorPainter::fill) callback of the [`ColorPainter`] trait.
+#[derive(Debug, PartialEq)]
+pub enum FillType<'a> {
```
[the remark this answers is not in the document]
drott: Back to `Brush`, wfm.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+}
+
+/// Result type used in [`draw_color_glyph`](ColorPainter::draw_color_glyph)
+/// through which the client can signal whether a COLRv1 glyph referenced by
+/// another COLRv1 glyph can be drawn from cache or whether the glyphs subgraph
+/// should be traversed.
+pub enum ColrGlyphDrawResult {
+    ColrGlyphDrawn,
+    ColrGlyphDrawingNotImpl,
+}
```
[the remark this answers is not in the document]
drott: Changed according to your suggestion.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,406 @@
...
+    // TODO(drott): Add an optimized callback function combining clip, fill and transforms.
+
+    /// Open a new layer, and merge the layer down using `composite_mode` when
+    /// [`pop_layer`](ColorPainter::pop_layer) is called, signalling that this layer is done drawing.
+    fn push_layer(&mut self, composite_mode: CompositeMode);
+    fn pop_layer(&mut self);
+}
+
+/// Distinguishes available color glyph types.
+pub enum ColorPaintableType {
```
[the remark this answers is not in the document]
drott: `ColorFormat` now.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,437 @@
...
+    /// Creates a new collection of paintable color glyphs for the given font.
+    pub fn new(font: &impl TableProvider<'a>) -> Self {
+        let colr = font.colr().ok();
+        let upem = font.head().map(|h| h.units_per_em());
+
+        Self { colr, upem }
+    }
+
+    /// Returns the paintable color glyph representation for the given glyph identifier.
+    pub fn get_type(
```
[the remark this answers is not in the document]
drott: Done.

skrifa/src/color/mod.rs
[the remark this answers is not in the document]
drott: PS: You explained the visibility considerations a bit to me in our call, and I thought I had understood those. Then, I struggled with some visibility and access for tests, and I think in the process of that I had made those public and then changed vis inside the modules to pub(crate). But now when changing it back I no longer had issues in test compilation 🤷

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,453 @@
-//! Color outlines.
+//! Drawing color glyphs.
+//!
+//! # Examples
+//! ## Retrieve the clip box of a COLRv1 glyph if it has one:
+//!
+//! ```
```
dfrg, line 6: Looks like the examples contain some errors. Usually cargo test catches these. I’m not sure why CI isn’t complaining here.

skrifa/src/color/mod.rs
```diff
@@ -1,6 +1,453 @@
...
+#[cfg(test)]
+mod traversal_tests;
 
-mod instance;
+use raw::tables::colr;
+#[cfg(test)]
+use serde::{Deserialize, Serialize};
+
+use read_fonts::{
+    tables::colr::{CompositeMode, Extend},
```
dfrg, line 47: One QOL improvement I missed: re-exporting `CompositeMode` and `Extend` here would be nice so users don’t have to dig through read-fonts for them.

dfrg (approved):
Two small nits about doc examples and maybe re-exporting types. Otherwise, this looks great to me and I’m happy to see it land. Thanks for getting this done and for your patience with review!

Oh, I assume we’ll do COLRv0 as a follow up?

skrifa/src/color/mod.rs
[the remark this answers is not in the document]
dfrg: Ack, yes. I will submit a fix for Size this week.

skrifa/src/color/mod.rs
drott: Example compile checks should be fixed now and formatted a bit more cleanly.

drott:
\o/

rsheeter:
Woohoo!

skrifa/src/color/color.rs
drott: Turns out we ran into some problems with the cast from `skrifa::color::ColorStop` to `fontations_ffi::ColorStop` (the latter needing to be a CXX type). Rolling the Skia change into the Google3 would have triggered a requirement for an `unsafe` code review whenever we touched the file with the unsafe blocks in it. So we need a wrapper type in FFI, which channels the individuals item over. I implemented that as an iterator now. So we might want to revisit making this into an iterator. 

https://skia-review.googlesource.com/c/skia/+/797756/2..3/src/ports/SkTypeface_fontations_priv.h#b91

drott: CC @bungeman

PR #725, opened 2023-11-29
COLR in floating point (IVS/resolve only)

dfrg:
As requested, this is the floating point refactor extracted from #697.

* add methods to IVS to compute scalars/deltas in f32.
* move `ColrInstance` from read-fonts to a private module in skrifa and update it to produce results in f32.

@drott I believe this captures what you requested as a separate, clean PR.


skrifa/src/color/instance.rs
```diff
@@ -0,0 +1,660 @@
...
+    Composite {
+        source_paint: Paint<'a>,
+        mode: CompositeMode,
+        backdrop_paint: Paint<'a>,
+    },
+}
+
+/// Resolves this paint with the given instance.
+///
+/// Resolving means that all numeric values are converted to 16.16 fixed
```
drott, line 288: This comment could perhaps be rephrased or updated in terms of how and when the conversion to float happens?

skrifa/src/color/instance.rs
```diff
@@ -0,0 +1,660 @@
...
+            }
+        }
+        deltas
+    }
+}
+
+#[derive(Copy, Clone)]
+enum DeltaType {
+    FWord,
+    UfWord,
```
drott, line 95: `UfWord` does not seem to be used in this implementation.

skrifa/src/color/instance.rs
```diff
@@ -0,0 +1,660 @@
...
+    ) -> impl Iterator<Item = ResolvedColorStop> + 'a {
+        self.stops
+            .iter()
+            .map(|stop| ResolvedColorStop {
+                offset: stop.stop_offset().to_f32(),
+                palette_index: stop.palette_index(),
+                alpha: stop.alpha().to_f32(),
+            })
+            .chain(self.var_stops.iter().map(|stop| {
+                let deltas = instance.var_deltas(stop.var_index_base(), [DeltaType::F2Dot14; 2]);
```
drott, line 188: I am investigating, but somehow after applying this and using my COLRv1 implementation on top, I get no fractional offsets anymore.

drott: Correction, I saw a 0.5, but the offsets seam more quantized.

drott:
Scratch my earlier comment on the offsets, I was confused by the test diffs after rebasing. 722 is now rebased on top of this.

skrifa/src/color/instance.rs
```diff
@@ -0,0 +1,660 @@
...
+            stops: &[],
+            var_stops: value.color_stops(),
+        }
+    }
+}
+
+impl<'a> ColorStops<'a> {
+    /// Returns an iterator yielding resolved color stops with variation deltas
+    /// applied.
+    pub fn resolve(
```
drott, line 176: I think given that I see some differences in the offsets in my results before and after this change, it would be useful to add a few tests for this part, just to be able to verify this one level lower down.

skrifa/src/color/instance.rs
dfrg: Oops, good catch.

skrifa/src/color/instance.rs
dfrg: Yes, completely unnecessary here. I’ll remove it.

skrifa/src/color/instance.rs
dfrg: Will do!

dfrg:
Since #722 is already rebased on this, would you like me to address your comments in a follow up after that lands or fix them here? Either works for me.

drott:
> after that lands or fix them here? Either works for me.

I'd say let's fix them here first, shouldn't be hard to rebase again.

dfrg:
> Scratch my earlier comment on the offsets, I was confused by the test diffs after rebasing. #722 is now rebased on top of this.

There actually is an issue here. Some of the intermediate 2.14 deltas in the test font overflow a signed 16-bit integer so they were wrapping negative. The values are such that your stop offset normalization will correct them. I’m not sure if this is in spec but I have a patch that avoids the overflow.

skrifa/src/color/instance.rs
dfrg: Updated the comment to reflect the floating point values.

skrifa/src/color/instance.rs
dfrg: Turns out this is actually used by the radii in radial gradients but the delta types were wrong. I don't think it makes a difference functionally, but this is fixed now to match spec.

skrifa/src/color/instance.rs
dfrg: Fixed the intermediate 2.14 overflow and updated the IVS floating point test to check 2.14 and 16.16 deltas.

drott:
I noticed that my variable clipbox test fails with this (including the latest 2257849 commit).

```
---- color::colrv1_traversal::tests::clipbox_test stdout ----
thread 'color::colrv1_traversal::tests::clipbox_test' panicked at skrifa/src/color/colrv1_traversal.rs:512:13:
Clip boxes do not match. Actual: BoundingBox { x_min: 200.0122, y_min: 500.0, x_max: 500.0, y_max: 1000.0 }, expected: BoundingBox { x_min: 200.0, y_min: 500.0, x_max: 500.0, y_max: 1000.0 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

Clip boxes are 16bit FWORD types with the same size of a delta - represented in float, I think the first shifted `x_min` should not be fractional but amount to 200 precisely.

dfrg:
I’ll dig into this to make sure but I suspect it is working as intended. We’re no longer rounding/truncating after applying deltas to FWORD values and I think this is desired behavior for other properties (specifically, gradient parameters). If we want to special case clip boxes by rounding (or better, floor/ceil the min/max values), that seems reasonable to me. We can also add a `snap_to_grid()` method on `BoundingBox<f32>` that does the appropriate thing and pass this choice to the consumer.

drott:
FYI, rebased #722 on top of this and the JSON colrv1 graph dump tests still pass. I did not observe any differences in values.

PR #727, opened 2023-12-01
[skrifa] new outline API

dfrg:
- adds a new API for scaling glyph outlines
- keeps the scale module temporarily, rebasing it on the new API (remove after Skia makes the transition)

This definitely requires more API documentation with examples but this is blocking work on TrueType hinting so I'd prefer to do that as a follow up.

The API generally looks like this:
```rust
// instance settings
let size = Size::new(18.0);
let var_loc = font.axes().location([("wght", 350.0)]);
// grab a glyph id
let gid = font.charmap().map('G').unwrap();
// get the outline glyph collection from the font
let outline_glyphs = font.outline_glyphs();
// and the outline glyph from the collection
let glyph = outline_glyphs.get(gid).unwrap();

// for an unhinted outline, just call scale directly on the glyph
glyph.scale(size, &var_loc, ScalerMemory::Auto, some_pen);

// if you'd like to control memory allocation...
let buf_size = outline.scaler_memory_size(Hinting::None);
let mut buf = vec![0u8; buf_size];
glyph.scale(size, &var_loc, ScalerMemory::User(&mut buf), some_pen);

// for hinting, we need to initialize the hinter first. This requires
// the outline collection, size, variation info and hinting mode
let hinter = NativeHinter::new(&outline_glyphs, size, &var_loc, Hinting::Full).unwrap();
// then call scale on the hinter, passing the glyph
hinter.scale(&glyph, ScalerMemory::Auto, some_pen);
```


rsheeter:
Glancing at the API:

1. Why do I pass the glyph rather than the glyph id to the hinter?
1. Why do we have to have multiple ways to scale => pen?
   * Can't we just provide the hinter to the scale operation or not

For example, I might naively imagine building up a Scale then executing it:

```rust
// aside: scale is not a very intuitive name for wanting to write something to a pen, in my mind you draw onto a pen

// Can't meaningfully scale without a size and location and some outlines
let scale = Scale::new(Size::new(18.0), font.axes().location([("wght", 350.0)]), font.outline_glyphs())

// at this point scale is ScalerMemory::Auto, no hinter

// you can assign scale.buf, scale.hinter = Some(&Hinter), etc

// and then eventually scale.draw(pen)
```

dfrg:
I should have listed the requirements (as discussed with @drott) which informed the API design decisions:

1. Both native and auto hinting should be explicit. No automatic fallback to auto hinting. This also means they can be independently initialized and stored. Additionally implies a new AutoHinter type, similar to NativeHinter. 
2. Hinter state (native and auto) should be immutable. We want to avoid locking in Skia. This state is configured for a particular set of instance settings. 
3. When possible, we want to determine whether a glyph contains hints up front so we can skip directly to the auto hinter. This requires doing some parsing of the glyf table and we don’t want to redo that work, hence the intermediate glyph object (this is also necessary for custom memory scenarios). 

Due to requirement 2, your proposed API isn’t feasible. An initialized hinter is only usable with the instance settings given at creation time. I could probably come up with a builder pattern for this but I don’t see how it buys us much.

I’m not opposed to changing scale -> draw.

rsheeter:
> Due to requirement 2, your proposed API isn’t feasible. An initialized hinter is only usable with the instance settings given at creation time

Do we have a doc? - these comments that don't thread are a bit suboptimal. I don't immediately understand what about my suggestion precludes a hinter that works as you suggest. Merely the suggestion it be borrowed vs the drawing operation having to own it for the duration?

My experience is a parameters object instead of a bunch of method fields is nice because then you can add new settings w/o breaking all existing clients or growing infinitely long lists of arguments in draw2, drawEx, etc (ala win32).

Is initializing a hinter expensive? If so, how should a client manage that? For example, one be advised to have a hinter pool, threadlocals, ...?

dfrg:
Adjusted the API based on feedback and f2f discussion. 

Reduced to a single `draw` method on `OutlineGlyph` and added a new `DrawSettings` type to capture the various parameters leaving room for potential future expansion (affine transform, embolden, etc).

Now looks like this:
```rust
// instance settings
let size = Size::new(18.0);
let var_loc = font.axes().location([("wght", 350.0)]);
// grab a glyph id
let gid = font.charmap().map('G').unwrap();
// get the outline glyph collection from the font
let outline_glyphs = font.outline_glyphs();
// and the outline glyph from the collection
let glyph = outline_glyphs.get(gid).unwrap();

// drawing unhinted outlines:
glyph.draw(DrawSettings::new(size, &var_loc), some_pen);
// DrawSettings has some conversions for convenience:
glyph.draw((size, &var_loc), some_pen);
// without variations..
glyph.draw(size, some_pen);

// if you'd like to control memory allocation...
let buf_size = outline.draw_memory_size(Hinting::None);
let mut buf = vec![0u8; buf_size];
glyph.draw(DrawSettings::new(size, &var_loc).with_memory(buf), some_pen);

// for hinting, we need to initialize the hinting instance first. This requires
// the outline collection, size, variation info and hinting mode
let hinting_instance = NativeHintingInstance::new(&outline_glyphs, size, &var_loc, NativeHinting::Full).unwrap();
// then the hinting instance can be used to initialize DrawSettings
glyph.draw(DrawSettings::native_hinted(&hinting_instance), some_pen);
// or with a conversion:
glyph.draw(&hinting_instance, some_pen);
// and with memory allocation:
let buf_size = outline.draw_memory_size(Hinting::Native);
let mut buf = vec![0u8; buf_size];
glyph.draw(DrawSettings::native_hinted(&hinting_instance).with_memory(buf), some_pen);
```

dfrg:
> My experience is a parameters object instead of a bunch of method fields is nice because then you can add new settings w/o breaking all existing clients or growing infinitely long lists of arguments in draw2, drawEx, etc (ala win32).

Discussed and agreed on this f2f. Made it so.

> Is initializing a hinter expensive? If so, how should a client manage that? For example, one be advised to have a hinter pool, threadlocals, ...?

Yes, it can be expensive, depending on the size and complexity of the `fpgm` and `prep` tables. Hinting state should definitely be cached and this is designed specifically to allow Skia to do so in `SkScalerContext`. Other users of this library, e.g. vello, will need to come up with their own caching strategy. I plan to implement one for vello and that might inform a design for something general that we can offer in this crate in the future.

skrifa/src/outline/cff/mod.rs
```diff
@@ -87,6 +91,15 @@ impl<'a> Outlines<'a> {
         matches!(self.version, Version::Version2(_))
     }
 
+    /// Returns the number of available glyphs.
+    pub fn len(&self) -> usize {
```
rsheeter, line 95: Optional: I don't find it obvious that this is what len of Outlines would mean, perhaps a more specific name, like subfont_count did, would make sense? glyph_count for example is painfully clear.

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+    Cff2,
+}
+
+/// Specifies the hinting strategy.
+#[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
+pub enum Hinting {
+    /// Hinting is disabled.
+    #[default]
+    None,
+    /// Will apply native embedded hints.
```
rsheeter, line 37: Suggest a wee bit more information to support the reader who is going wtf is a native hint :)

rsheeter: (and what on earth is a non-native hint)

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+    Native,
+}
+
+/// Information and adjusted metrics generated while drawing an outline glyph.
```
rsheeter, line 41: Clarify adjust by what

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+    pub has_overlaps: bool,
+    /// If present, an adjusted left side bearing value generated by the
+    /// scaler.
+    pub lsb: Option<f32>,
+    /// If present, an adjusted advance width value generated by the
+    /// scaler.
+    pub advance_width: Option<f32>,
+}
+
+/// Options that define how a glyph is drawn to a pen.
```
rsheeter, line 55: Link to glyph and pen types if possible

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+pub struct DrawSettings<'a> {
+    instance: DrawInstance<'a>,
+    memory: Option<&'a mut [u8]>,
+}
+
+impl<'a> DrawSettings<'a> {
+    /// Creates settings for an unhinted draw operation with the given size and
+    /// location in variation space.
+    pub fn new(size: Size, location: impl Into<LocationRef<'a>>) -> Self {
```
rsheeter, line 64: Optional: I wonder if this should be more explicit, e.g. unhinted instead of new

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+    }
+}
+
+impl<'a> From<&'a NativeHintingInstance> for DrawSettings<'a> {
+    fn from(value: &'a NativeHintingInstance) -> Self {
+        DrawSettings::native_hinted(value)
+    }
+}
+
+/// A scalable glyph outline.
```
rsheeter, line 121: Suggest spelling out it can come from any of the supported outline tables

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+#[derive(Clone)]
```
rsheeter, line 122: Debug?

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+pub struct OutlineGlyph<'a> {
+    kind: OutlineKind<'a>,
+}
+
+impl<'a> OutlineGlyph<'a> {
+    /// Returns the underlying source format for this outline.
+    pub fn format(&self) -> OutlineGlyphFormat {
+        match &self.kind {
+            OutlineKind::Glyf(..) => OutlineGlyphFormat::Glyf,
+            OutlineKind::Cff(cff, ..) => {
```
rsheeter, line 132: Curiosity, why don't we just have OutlineKind Cff and Cff2 to match the glyph format?

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+    /// instructions.
+    ///
+    /// For CFF outlines, returns `None` since this is unknown prior
+    /// to loading the outline.
+    pub fn has_hinting(&self) -> Option<bool> {
+        match &self.kind {
+            OutlineKind::Glyf(_, outline) => Some(outline.has_hinting),
+            _ => None,
+        }
+    }
```
rsheeter, line 163: Hanging methods that only work for glyf off an explicitly format-agnostic structure feels weird. Why not have these on OutlineKind::Glyf? - user can do an if let and decide for themselves what to do about CFF.

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+
+    /// Returns the size (in bytes) of the temporary memory required to draw
+    /// this outline.
```
rsheeter, line 166: I feel like this should point me to what to do with it? Meant to be used to ... or See pattern at blah.

skrifa/src/outline/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+}
+
+#[derive(Clone)]
+enum OutlineCollectionKind<'a> {
+    None,
+    Glyf(glyf::Outlines<'a>),
+    Cff(cff::Outlines<'a>),
+}
+
+const GLYF_SCALER_STACK_BUFFER_SIZE: usize = 4096;
```
rsheeter, line 317: Comment should explain why this value

skrifa/src/outline/native_hint.rs
```diff
@@ -0,0 +1,147 @@
...
+use super::{
+    cff, AdjustedMetrics, DrawError, Hinting, LocationRef, NormalizedCoord, OutlineCollectionKind,
+    OutlineGlyph, OutlineGlyphCollection, OutlineKind, OutlinePen, Size,
+};
+
+/// Modes for native hinting.
+///
+/// Only the `glyf` format supports all hinting modes.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum NativeHinting {
```
rsheeter, line 12: Do these mirror FreeType hinting modes? If so, we should link to it

skrifa/src/outline/native_hint.rs
```diff
@@ -0,0 +1,147 @@
...
+                    subfonts.push(cff.subfont(i, size, &self.coords)?);
+                }
+                self.kind = HinterKind::Cff(subfonts);
+            }
+            OutlineCollectionKind::None => {}
+        }
+        Ok(())
+    }
+
+    pub(super) fn draw_hinted(
```
rsheeter, line 107: Optional: draw_hinted _on a hinter feels a bit redundant, it's not like you'd ask NativeHintingInstance to draw any other way is it?

rsheeter: Hm. I guess with HinterKind - which feels like maybe it belongs up by the NativeHinting enum - you could have None? I don't find it obvious when you would want to do that though...?

skrifa/src/outline/native_hint.rs
```diff
@@ -0,0 +1,147 @@
...
+                Ok(AdjustedMetrics::default())
+            }
+            _ => Err(DrawError::NoSources),
+        }
+    }
+}
+
+#[derive(Clone)]
+enum HinterKind {
+    None,
```
rsheeter, line 144: I think None deserves a comment, I'm not immediatley seeing when you would want None

skrifa/src/outline/native_hint.rs
```diff
@@ -0,0 +1,147 @@
...
+    Glyf(),
```
rsheeter, line 145: Why `()`?

skrifa/src/scale.rs
```diff
@@ -215,6 +213,196 @@ impl Context {
     }
 }
 
+/// Builder for configuring a glyph scaler.
```
rsheeter, line 216: This description requires a fair bit of font knowledge to parse. Can we describe it in more layperson-friendly terms w/o losing anything?

First attempt:

```Builds [`Scaler`]'s that give access to glyph outlines via [Scaler::draw] to your [`Pen`]s```

rsheeter (approved):
Couple of nits and questions but overall lgtm

skrifa/src/outline/mod.rs
dfrg: I chose Native as the analogue to Auto for when auto hinting is implemented. Maybe Embedded is a better prefix?

skrifa/src/outline/mod.rs
dfrg: I lol’d. My first pass named this unhinted but I thought reviewers might not like that :)

skrifa/src/outline/mod.rs
dfrg: The internal CFF code has these rolled together but Dominik expressed interest in being able to select/identify formats at a more granular level so I just exposed it in this way for the public API.

skrifa/src/outline/mod.rs
dfrg: Agreed that this is not ideal but I couldn’t think of a better API that didn’t require exposing the underlying types. 

With that in mind, I’m not opposed to turning `OutlineGlyph` into an enum and exposing more of the glyf and cff internals. In fact, as part of my cleanup, the new TrueType hinter can now do step debugging which I think might be a valuable feature but would require making public a lot of glyf specific types. This would call for some non-trivial refactoring to make those modules more suitable for public use so my preference would be to defer that if we decide to go this route. 

I admit to feeling a bit of whiplash on this because we used to expose more and we agreed to make the API more agnostic.

skrifa/src/outline/mod.rs
dfrg: Noted. I have actual data now for memory requirements with TT hinting so this is likely to become more complicated with multiple stack allocation sizes. I’ll prep a sheet with this data because it’s rather interesting.

skrifa/src/outline/mod.rs
rsheeter: I missed that those types weren't exposed when I wrote that. Hm. I wonder if it's worth an enum that explicitly indicates something is not supported but ... feels low ROI. I think it's fine then.

skrifa/src/outline/native_hint.rs
dfrg: They do mostly but this type will be changing soon. @drott and I agreed to make the actual functionality of the hinting mode more explicit along with docs mapping to FT load flags. Coming with hinting PR.

skrifa/src/outline/native_hint.rs
dfrg: Happy to drop the _hinted. 

The None variant exists because we can potentially have a font with no outline tables and this state represents that. I can make initialization return an error in this case if you like that better.

skrifa/src/outline/native_hint.rs
dfrg: Oops, answered this above.

skrifa/src/outline/native_hint.rs
dfrg: Useless but harmless? I probably added it as a placeholder for the upcoming TT hinting state 🤷🏻‍♂️

skrifa/src/scale.rs
dfrg: Point taken but the entire scale module effectively dead code at this point. I only kept it here due to our guarantee to keep old APIs during a transition period for Skia. This did remind me that the whole thing should probably be `doc(hidden)` so it doesn’t appear in public docs in the next release.

skrifa/src/outline/cff/mod.rs
dfrg: Changed to glyph_count (and did the same for glyf::Outlines)

skrifa/src/outline/mod.rs
dfrg: Went with the `Embedded` prefix and added more explanation. Happy to bikeshed and rename later if we come up with something better.

skrifa/src/outline/mod.rs
dfrg: Done!

skrifa/src/outline/mod.rs
dfrg: Added links to the doc comment.

skrifa/src/outline/mod.rs
dfrg: Renamed back to `unhinted` :)

skrifa/src/outline/mod.rs
dfrg: Extended the doc comment to talk about source tables with links to spec.

skrifa/src/outline/mod.rs
dfrg: This contains a lot of deeply nested read-fonts types that require traversal for Debug :( Happy to do a custom impl that shows something, but I'd like to defer that.

skrifa/src/outline/mod.rs
dfrg: Added an explanation, link to the with_memory method that provides more and a table showing how the hinting parameter should be mapped to DrawSettings modes

skrifa/src/outline/mod.rs
dfrg: Admitted _arbitrariness_ of this constant. Will make this more clear in the hinting PR.

skrifa/src/outline/native_hint.rs
dfrg: draw_hinted -> draw and added a comment explaining why the `None` variant exists

dfrg:
Ty for the quick review! We might want to iterate a bit more on this in the near future but going to merge so I have a solid base on which to land TT hinting.

PR #729, opened 2023-12-08
[mvar] offset to IVS is actually 16-bit, not 32-bit

anthrotype:
we jumped the gun :)

see https://learn.microsoft.com/en-us/typography/opentype/spec/mvar#table-formats

JMM


cmyr (commented):
oops!

PR #730, opened 2023-12-11
Scale Bounding boxes

drott:
Allowing scaling of bounding boxes, this comes in handily
for example when dealing with clip boxes and scaling them to font size.

### DIFF
--- a/font-types/src/bbox.rs
+++ b/font-types/src/bbox.rs
@@ -1,4 +1,4 @@
-use core::ops::{Mul, MulAssign};
+use core::ops::Mul;
 
 /// Minimum and maximum extents of a rectangular region.
 #[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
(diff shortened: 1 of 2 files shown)


font-types/src/bbox.rs
```diff
@@ -14,3 +16,12 @@ pub struct BoundingBox<T> {
     /// region.
     pub y_max: T,
 }
+
+impl<T> BoundingBox<T> where T: MulAssign + Mul + Copy {
+    pub fn scale(&mut self, factor : T)  {
```
dfrg, line 21: Changing this to take `&self` and return a new copy of the scaled bounding box would make this easier to use when chaining calls, e.g. `map()`. Also lets us remove the `MulAssign` bound.

dfrg (approved):
Seems like a useful addition! Left one suggestion about the signature and this also needs rustfmt, otherwise lgtm.

font-types/src/bbox.rs
drott: Returning a new one, `MulAssign` bound removed.

drott:
Rustfmt formatted, and used in `color/mod.rs`.

PR #734, opened 2023-12-14
[ivs_builder] prune unused regions

### LINKED ISSUE #733: VariationStoreBuilder does not prune unused regions from VarRegionList
fontTools' equivalent has a `prune_regions` method that is called after optimizing the VarStore. Some of the regions that were previously referenced by some VarData.VarRegionIndices may no longer be used after optimization and should be removed.
Otherwise we end up with a VarRegionList that contains Regions that aren't actually referenced and waste bytes.
We should do the same.

anthrotype:
Fixes #733

#735 reproduces the bug, this one fixes it

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -1,7 +1,7 @@
 //! Building the ItemVariationStore
 
 use std::{
-    collections::{BinaryHeap, HashMap},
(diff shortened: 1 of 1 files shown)


cmyr (approved):
Looks good!

my only nit (and it's not blocking) is that I don't love that the 'make_regions_list` method is now also responsible for mutating the subtables, but any other solution (like having it return both a region list and a old->new map) feels a bit overengineered, so this works!

PR #735, opened 2023-12-14
[DO NOT MERGE] add failing test to reproduce #733

anthrotype:
reproduces #733

I'll cherry pick this one from #734


anthrotype:
```
---- tables::variations::ivs_builder::tests::prune_unused_regions stdout ----
thread 'tables::variations::ivs_builder::tests::prune_unused_regions' panicked at write-fonts/src/tables/variations/ivs_builder.rs:1422:9:
  left: 4
 right: 2
```

PR #736, opened 2023-12-15
[COLRv1] Add tests for coincident stops and angles

drott:
Adjust PaintSweepGradient implementation to not reverse color stops when not needed. Matches FireFox behavior. See Skia change under review for implementing the same change [1].

[1] https://skia-review.googlesource.com/c/skia/+/790456


dfrg (approved):
Lgtm!

drott:
Thanks for the quick review.

PR #742, opened 2023-12-15
[write-fonts] Do not autocompute VariationRegion::axis_count

### LINKED ISSUE #741: pruned varstore regions in ivs_builder causing miscompilation
So after updating fontc to use write-fonts 0.20, we're hitting a crash in some of the tests.

This is the result of the changes in #734, but I'm not sure those changes are really to blame; it's possible that they have just uncovered an existing issue?

The problem as I understand it:
- some of our test fonts have no deltas in hvar
- all zeros can be omitted, which means we don't actually store anything in the varstore
- since we have no data in the varstore, it follows that we don't actually reference any regions
- this means we can prune all the regions from the region list
- _however_ when we compile the varstore, we use the number of axes in the first item in the region list to determine the `axis_count` value in the `RegionList`
- but if there are no regions in the region list, we can't compute the correct value.

Writing this all down has, I think, presented the correct solution. We need to stop assuming that we can compute the `axis_count` value, and instead require it to be set explicitly, which we can do in the builder just fine. We also need to add a validation check to ensure that this value matches what is in fvar.

cmyr:
This can not be correctly computed in the case where we have pruned all regions, so we should require it to be set explicitly and then do that in the builder.

Note: this is *a* fix, but may not be the best one. In particular we will still not know the correct axis count in the case where an `ItemVariationStore` is built without having any deltas added to it. I believe this would still cause a miscompilation. If this is a case that we want to allow (instead of just erroring at compile time, which we could consider?) then we will need to change this API to require the user to pass in an explicit `axis_count` when constructing the builder. I'm not sure how much of a hassle that would be in practice.

So basically: I need a second opinion on this. I'm now leaning towards requiring an explicit axis_count, but the deciding factor will be whether or not 'var store with no deltas in it' is a thing we want to allow.

- fixes #741

### DIFF
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -17,11 +17,14 @@ type TemporaryDeltaSetId = u32;
 ///
 /// This handles assigning VariationIndex values to unique sets of deltas and
 /// grouping delta sets into [ItemVariationData] subtables.
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct VariationStoreBuilder {
     // region -> index map
     all_regions: HashMap<VariationRegion, usize>,
     delta_sets: DeltaSetStorage,
+    // must match fvar. We require the user to pass this in because we cannot
+    // infer it in the case where no deltas are added to the builder.
+    axis_count: u16,
 }
 
 /// A collection of delta sets.
@@ -70,8 +73,12 @@ impl VariationStoreBuilder {
     /// data. To use implicit indices, use [`new_with_implicit_indices`] instead.
     ///
     /// [`new_with_implicit_indices`]: VariationStoreBuilder::new_with_implicit_indices
-    pub fn new() -> Self {
-        Default::default()
+    pub fn new(axis_count: u16) -> Self {
+        Self {
+            axis_count,
+            delta_sets: DeltaSetStorage::Deduplicated(Default::default()),
+            all_regions: Default::default(),
+        }
     }
 
     /// Returns `true` if no deltas have been added to this builder
@@ -87,8 +94,9 @@ impl VariationStoreBuilder {
     /// This is used in HVAR, where it is possible to use glyph ids as the
     /// 'inner index', and to generate a single ItemVariationData subtable
     /// with one entry per item.
-    pub fn new_with_implicit_indices() -> Self {
+    pub fn new_with_implicit_indices(axis_count: u16) -> Self {
         VariationStoreBuilder {
+            axis_count,
             all_regions: Default::default(),
             delta_sets: DeltaSetStorage::Direct(Default::default()),
         }
@@ -154,7 +162,7 @@ impl VariationStoreBuilder {
                 .map(|idx| region_map[idx])
                 .collect();
         }
-        VariationRegionList::new(new_regions)
+        VariationRegionList::new(self.axis_count, new_regions)
     }
 
     fn encoder(&self) -> Encoder {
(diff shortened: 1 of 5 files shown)


write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -119,8 +119,27 @@ impl VariationStoreBuilder {
...
+            .map(|reg| reg.region_axes.len());
+        match axis_count {
+            Some(count) => count.try_into().unwrap(),
+            None => {
+                // is it ever reasonable to have an IVS builder to which you add
+                // no items? it seems possible? but maybe zero is valid in that
+                // case, since it would also mean we have an empty region list?
+                // Help me Cosimo!
+                0
+            }
```
rsheeter, line 136: at a glance this looks like reinvention of unwrap or default?

What would it mean if the regions had inconsistent axis count? Do we validate for that?

rsheeter (approved):

write-fonts/src/tables/variations/ivs_builder.rs
rsheeter: We should be able to test this right? Assert the axis count when there are no regions is 0?

cmyr: yes, this is mostly an excuse for me to write this note.


the spec requires the axis count to match between fvar and the RegionList, and so it's possible that other code in the ecosystem is depending on that being the case. In the particular case of this value being zero, if we ever hit this branch I think we are generating an inconsistent font, and we would currently hit the debug assert if we tried to access the region list; however in release we would just receive an empty set of regions, which would be fine.


we don't currently have a mechanism to validate things across table boundaries, so this is not checked. If we are going to merge this patch, we will want to add that check; I don't think we should be compiling this.

cmyr: > We should be able to test this right? Assert the axis count when there are no regions is 0?

the issue is that if there are no region there still may be axes, we just don't have any way of knowing what they are.

rsheeter: I mean literally just to test get_axis_count, e.g. `assert_eq!(0, VariationStoreBuilder::new().get_axis_count())`

rsheeter: > this is mostly an excuse for me to write this note

the note belongs in an issue imho :)

cmyr: Basically we need to decide whether it is ever okay to generate an `ItemVariationStore` with no deltas. If it's not okay, then this patch should go in, and we should add that validation check, or make the builder return a result or something. If it *is* okay than we need to make the user tell us how many axes there are when they create the builder.

rsheeter: Unless there is something in the spec that bans empty IVS I think it's OK. An odd choice certainly but not illegal.

write-fonts/src/tables/variations/ivs_builder.rs
```diff
@@ -17,11 +17,14 @@ type TemporaryDeltaSetId = u32;
...
 /// grouping delta sets into [ItemVariationData] subtables.
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct VariationStoreBuilder {
     // region -> index map
     all_regions: HashMap<VariationRegion, usize>,
     delta_sets: DeltaSetStorage,
+    // must match fvar. We require the user to pass this in because we cannot
+    // infer it in the case where no deltas are added to the builder.
+    axis_count: u16,
```
rsheeter, line 27: If we don't already we should probably validate that the regions match this count since that's possible to check with access to only self?

cmyr: yea let's open an issue for that, I'm going to be done for the day shortly.

write-fonts/src/tables/variations/ivs_builder.rs
anthrotype: > whether it is ever okay to generate an ItemVariationStore with no deltas

no deltas (i.e. ItemVarData without items) is ok even though it's a bit weird, why bother to have the varstore to begin with, most of the time when building one we check if it turns out to be empty, and don't bother storing one (it's usually optional, if it isn't optional then the parent table likely is optional as a whole so both are omitted).

However, having no axes at all is _not_ ok for an allegedly "variable" font. No fvar axes (the var region axis count == len(fvar.axes) by definitino) means the font is not variable! How can something that has "variation" in its very name actually define no axes of variation?!

PR #750, opened 2023-12-20
[write-fonts] reproduce and fix non deterministic gvar serialization

anthrotype:
Meant to reproduce the issue discussed at https://github.com/googlefonts/fontc/issues/647

If we are lucky enough the CI should fail, otherwise flip the coin again and you'll see...

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -273,7 +273,6 @@ impl GlyphVariations {
         // so that we match the behavior of fonttools
         let (pts, _) = max_by_first_key(
             point_number_counts
-                .clone()
                 .into_iter()
                 // no use sharing points if they only occur once
                 .filter(|(_, (_, count))| *count > 1),
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -970,6 +970,143 @@ mod tests {
         );
     }
 
+    #[test]
+    fn compute_shared_points_is_deterministic() {
+        // The deltas for glyph "etatonos.sc.ss06" in GoogleSans-VF are such that the
+        // TupleVariationStore's shared set of point numbers could potentionally be
+        // computed as either PackedPointNumbers::All or PackedPointNumbers::Some([1, 3])
+        // without affecting the size (or correctness) of the serialized data.
+        // However we want to ensure that the result is deterministic, and doesn't
+        // depend on e.g. HashMap random iteration order.
+        // https://github.com/googlefonts/fontc/issues/647
+        let _ = env_logger::builder().is_test(true).try_init();
+        let variations = GlyphVariations::new(
+            GlyphId::NOTDEF,
+            vec![
+                GlyphDeltas::new(
+                    Tuple::new(vec![
+                        F2Dot14::from_f32(-1.0),
+                        F2Dot14::from_f32(0.0),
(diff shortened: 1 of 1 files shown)


anthrotype:
nice, it did fail the first time :)

```
  left: Some(All)
 right: Some(Some([1, 3]))
```

anthrotype:
Finally it took my a few rounds but I managed to also make the CI randomly _pass_ without actually changing anything :)

https://github.com/googlefonts/fontations/actions/runs/7277928604/job/19831276618

write-fonts/src/tables/gvar.rs
```diff
@@ -242,13 +268,17 @@ impl GlyphVariations {
...
-            // no use sharing points if they only occur once
-            .filter(|(_, (_, count))| *count > 1)
-            .max_by_key(|(_, (size, count))| (*count - 1) * *size)?;
+        // find the one that saves the most bytes; if multiple are tied, pick the
+        // first one like python max() does (Rust's max_by_key() would pick the last),
+        // so that we match the behavior of fonttools
+        let (pts, _) = max_by_first_key(
+            point_number_counts
+                .clone()
+                .into_iter()
```
cmyr, line 277: as trivia, there's an interesting distinction between `clone().into_iter()` and `iter().cloned()`. The former clones the whole collection first and then iterates over the moved values, and the former iterates over borrowed values, cloning them one at a time, which is slightly more efficient.


There's probably an even more efficient option, here. Since we only want the max from the collection it would be strictly best to avoid cloning at all during the iteration, and only clone the single winner.


Ultimately this doesn't matter much, but I think it's worth building up that instinct around what the code we're writing is doing, and if it's the most efficient option.

cmyr (approved):
Good find cosimo, and I think that matching the behaviour of fonttools is the right call.

This was interesting, and I'm glad it made me go and look more closely into the rust HashMap implementation. :)

anthrotype:
This particular test glyph is funny: basically, there are 7 tuple variations, 5 of them have PackedPointNumbers::All, 2 of them have PackedPointNumbers::Some([1, 3]). Using either one as the shared point numbers will save the same number of bytes (4 in this particular case).
Now, because of the indeterminacy of hashmap order (which uses a random hash seed that changes at every execution), we would end up using either one at every execution, leading to a gvar with the same total size but different checksum.
So I decided that between the two possible valid shared point numbers, we should use the first one that fits the bill, following the ordering of the tuple variations themselves, i.e. the set of packed point numbers that first occurs when iterating over the tuple variations.
One could also take the last, and Rust max_by_key does exactly that, but since this whole piece of code was meant to port and match fonttools, I think it made more sense to match Python's max() behavior and take the first in case of ties.

cmyr:
Oh one other idea here: you should be able to get closer to ensuring that this test fails consistently by just running it in a loop for some number of iterations.

write-fonts/src/tables/gvar.rs
anthrotype: sorry the clone is a leftover when I was printing the damn thing :)

anthrotype: thanks for spotting!

anthrotype:
> each call to HashMap::new() uses a different seed

oh good to know.

> you should be able to get closer to ensuring that this test fails consistently

good point, but I think now that I fixed it it's no longer going to fail/pass inconsistently so maybe it's not worth doing the loop any more?

PR #754, opened 2024-01-12
[skrifa] Implement an optimized `fill_glyph` operation

### LINKED ISSUE #746: [Skrifa][COLRv1] Add a clip&fill optimized method to paint callbacks
In previous experiments in Skia it became clear that the [`drawPath`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/skia/include/core/SkCanvas.h;l=1456?q=skcanvas) operation is faster than `clipPath` + `drawPaint` operations executed in sequence.

Add an optimized method to the `ColorPainter` trait to capture that operation and signal this combination to the client. This combination can be used near leaf nodes, where we find a PaintGlyph + Transform* + Paint(Solid|LinearGradient|RadialGradient|SweepGradient) in the COLRv1 paint graph.

drott:
In COLRv0/v1 drawing clients may gain a performance benefit from filling
a shape (with a solid or gradient fill) instead of performing a clip +
unbounded fill in sequence.

Allow clients to do that using a fill_glyph() operation that combines
the two.

Implement this using a temporary painter that speculatively traverses
the glyph tree down from a `PaintGlyph` operation collecting potential
transforms along the way.

New behavior shows clearly in new baselines for tests added in #758 that 
nest `PaintGlyph` tables and cover mixed situations in
which the optimization succeeds for a leaf but fails for the
`PaintGlyph` at a layer above, and demonstrate recording transforms
along the way.

Fixes #746.


skrifa/src/color/traversal.rs
```diff
@@ -52,10 +52,76 @@ fn make_sorted_resolved_stops(stops: &ColorStops, instance: &ColrInstance) -> Ve
...
+    fn pop_layer(&mut self) {
+        self.optimizable = false;
+    }
+}
+
+pub(crate) fn traverse_with_callbacks<'a>(
     paint: &ResolvedPaint,
     instance: &ColrInstance,
-    painter: &mut impl ColorPainter,
+    painter: &mut impl ColorPainter<'a>,
```
drott, line 124: Can y'all help me with the lifetimes and tagging them correctly here? Or is what I am doing even possible. 

If I don't put `&'a mut impl ColorPainter<'a>` here, i.e. no explicit borrow lifetime, than the compiler complains with:

```
434 |                 let mut optimizer = CollectFillGlyphPainter::new(painter, *glyph_id);
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ lifetime `'a` required
```

If I add it, I get a lot of errors about not being able to reborrow painter - which previously worked. 

I kind of understand the first error message, but I don't understand how adding an explicit borrow lifetime breaks everything. 

What could be a solution?

cmyr: If you want to take this approach I think what you need to do is have two lifetimes, with an explicit bound. See https://doc.rust-lang.org/1.75.0/reference/trait-bounds.html#lifetime-bounds.

Here that would look like,


```rust
PainterPainter<'inner, 'ourborrow> where 'inner: 'ourborrow {
    painter: &'ourborrow Painter<'inner>,
}
```

drott: Thanks so much, that did the trick. I would not have figured that out so soon. Would a prose way to describe this as follows be correct?
> `PainterPainter` is an object that's constructed with an own lifetime of `ourborrow` that borrows a `Painter` struct for the lifetime of `ourborrow`, but requires that `Painter` struct with own lifetime `inner` to outlive `ourborrow` (`PainterPainter`s own lifetime)?

drott:
I would like to send this for review now. This is still a nested painter-based implementation. I have opted for now not to change signatures of the draw functions to return a Result and thus potentially be able to cancel drawing, as I believe this realizes the performance benefit and we would have yet to discover pathological situations in which there are deeply nested trees behind a `PaintGlyph`. The benefit the nested painter approach I see in avoiding implementing separate imperative traversal logic - instead - the existing `traverse_with_callbacks` can be used.

Please let me know if this works for you or if there are strong opinions about a different approach. I am open to revisiting this, but for now, this definitely meets the mark in terms of benchmark results as shared in the chat. An API change to `Result<>`-returning, cancellable methods would require a roundtrip of deprecation steps and careful chaining it into Skia and then Chrome with a phase of parallel `ColorPainter` functions along the way. I prefer to save that time atm and focus on implementing bitmap fonts support next.

skrifa/src/color/traversal.rs
drott: Thanks both for your help with that - updated change solves the lifetime issues.

skrifa/src/color/mod.rs
```diff
@@ -218,7 +237,27 @@ pub trait ColorPainter {
...
+    fn fill_glyph_supported(&self) -> FillGlyph {
+        FillGlyph::Unimplemented
+    }
+
+    /// Combined clip and fill operation.
+    ///
+    /// Apply the clip path determined by the specified `glyph_id`,
+    /// then fill it with the specified [`brush`](Brush), applying the `_transform`
+    /// transformation matrix to the brush.
+    fn fill_glyph(
```
dfrg, line 253: Since we can provide a default fallback for this using only methods on self, my preference would be to go that route here rather than adding additional methods, types and logic to query support. 

Something like:
```rust
self.push_clip_glyph(glyph_id);
// maybe check for identity?
self.push_transform(transform);
self.fill(brush);
self.pop_transform();
self.pop_clip();
```

This allows the user to override as desired and removes the need for the result return type and the additional error variant.

skrifa/src/color/mod.rs
```diff
@@ -218,7 +237,27 @@ pub trait ColorPainter {
...
+        &mut self,
+        _glyph_id: GlyphId,
+        _transform: Transform,
```
dfrg, line 256: This should probably be called `brush_transform` just to further signify that it applies to the brush and not the clip. 

Also, optional: make the transform optional :) I suspect this will be identity most of the time.

skrifa/src/color/transform.rs
```diff
@@ -25,6 +27,23 @@ pub struct Transform {
     pub dy: f32,
 }
 
+impl MulAssign for Transform {
```
dfrg, line 30: I realize we don’t strictly need it here but this will be exposed publicly and having an implementation of MulAssign without Mul feels a bit weird.

dfrg (approved):
Thanks for this!

I’m happy to land with the nested traversal since it works and seems to hit our performance targets. Upon further thought, I’m not too excited about complicating the API with cancellation. If we do encounter deeply nested graphs under a clip glyph, I’d prefer to implement the procedural matching instead which wouldn’t require changing the API. 

I noted one request for a change and a couple nits. Otherwise lgtm!

skrifa/src/color/mod.rs
drott: I thought about that, and also had this implemented first with just a client callback that can return a failure if it doesn't support the optimization, which then lead to the call sequence you're proposing, so that was semantically similar, but I see a case where this leads to less efficient outcomes.

If the client does not support the operation, we spend extra time for the detection while there is no optimization benefit - so a separate detection query saves time here. 

This also practically happens in the real world: In the no-clipbox situation of the Skia-side `BoundsPainter` where the draw method is used with no-op gradient fills. Here, the detection of the optimizable leaf situation only costs overhead, hence I thought it's better to ask first and avoid the search.

skrifa/src/color/mod.rs
drott: a) Renamed to `brush_transform` and updated arg name in documentation.
b) Transform turned into `Option<Transform>` - I wasn't sure if you prefer to have that be expressed in the test results as an empty string - for now I left testing to dump an identity matrix in that case, instead of special casing the None situation,  but agree this might be helpful for clients. Let me know if you want changes on the testing side. On the Skia side I am also making this a default matrix in the callbacks to C++.

skrifa/src/color/transform.rs
drott: Added a `Mul` implementation, moved the logic into that and made the `MulAssign` implementation based on the former - hope that's what's usually done. Not sure I need the `clone()` there, but I guess I need some sort of temporary value at least?

skrifa/src/color/mod.rs
dfrg: > If the client does not support the operation, we spend extra time for the detection while there is no optimization benefit - so a separate detection query saves time here.

If a client doesn't support the optimization then it's probably not particularly concerned about performance? :)

> This also practically happens in the real world: In the no-clipbox situation of the Skia-side BoundsPainter where the draw method is used with no-op gradient fills. Here, the detection of the optimizable leaf situation only costs overhead, hence I thought it's better to ask first and avoid the search.

If we're worried about performance in the bounds computation case, I'd be more concerned about the color stop evaluation which I suspect is significantly more expensive than leaf detection. This might suggest that we want a traversal designed specifically for bounding boxes that avoids unnecessary work.

Overall, this feels like a case where the cost of using a nested traversal is leaking into the public API and that makes me a bit uneasy. My preference would be to do the idiomatic Rust thing here (default trait method for fallback) and then tune the implementation (which may imply specialized bounds computation) if performance is an issue.

drott: Updated according to yours and Ben's feedback. Providing a default implementation and assuming in traversal that the method is supported.

skrifa/src/color/mod.rs
```diff
@@ -76,6 +76,7 @@ pub enum PaintError {
     ParseError(ReadError),
     GlyphNotFound(GlyphId),
     PaintCycleDetected,
+    FillGlyphOptimizationFailed,
```
dfrg, line 79: This appears to only be used internally. Can we remove this and change back to a boolean flag for the optimization check?

skrifa/src/color/transform.rs
dfrg: We could also derive `Copy` for `Transform` which would remove the need for the clone. I’m good either way.

skrifa/src/color/mod.rs
dfrg: Your approach looks good to me. Thanks for making the change.

skrifa/src/color/transform.rs
drott: I'll add that.

skrifa/src/color/mod.rs
```diff
@@ -218,7 +223,34 @@ pub trait ColorPainter {
...
     /// Fill the current clip area with the specified gradient fill.
-    fn fill(&mut self, brush: Brush);
+    fn fill(&mut self, brush: Brush<'_>);
+
+    /// Combined clip and fill operation.
+    ///
+    /// Apply the clip path determined by the specified `glyph_id`, then fill it
+    /// with the specified [`brush`](Brush), applying the `_brush_transform`
+    /// transformation matrix to the brush. The default implementation works
+    /// based on existing methods in this trait. It is recommend for clients to
```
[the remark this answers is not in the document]
drott: Wording fixed.

skrifa/src/color/mod.rs
drott: Changed to an internal boolean, removed enum value.

PR #756, opened 2024-01-15
[write-fonts/gvar] Fix shared tuples unstable order

anthrotype:
Reproduce and fix https://github.com/googlefonts/fontc/issues/647#issuecomment-1892034218

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -90,7 +90,7 @@ impl Gvar {
 
         fn compute_shared_peak_tuples(glyphs: &[GlyphVariations]) -> Vec<Tuple> {
             const MAX_SHARED_TUPLES: usize = 4095;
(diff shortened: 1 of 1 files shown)


anthrotype:
https://github.com/googlefonts/fontations/actions/runs/7528580955/job/20491085359?pr=756
```
failures:

---- tables::gvar::tests::shared_tuples_stable_order stdout ----
thread 'tables::gvar::tests::shared_tuples_stable_order' panicked at write-fonts/src/tables/gvar.rs:1270:13:
  left: [[-1.0], [1.0]]
 right: [[1.0], [-1.0]]
```

anthrotype:
I can confirm this enables reproducible builds for Oswald. I tested rebuilding a hundred times and fontc always produces exact same bytes for gvar, previously it would fail about 50% of the times.

dfrg (approved):
This looks like a reasonable fix to me, thanks!

PR #758, opened 2024-01-18
[color] Add PaintGlyph nested tests

drott:
In order to test upcoming optimizations that fuse a clip and a fill callback operation in COLRv0/v1 drawing, add tests that have nested PaintGlyph operations that trip up the optimization. In the nested case the optimization cannot be applied an drawing of the outer PaintGlyph has to be performed through the regular, unoptimized callbacks.

As glyph ids in the test font have shifted, rebaseline some existing results.

Preparation for #746.


drott:
For reference, glyphs look like this:

image

dfrg (approved):
Lgtm!

PR #759, opened 2024-01-18
[write-fonts] Add Glyph::Empty

### LINKED ISSUE #561: Consider adding `Glyph::Empty` variant?
This would be used in both read&write-fonts, and would correspond to the situation where we do not write any glyph data, but add a (zero-length) entry in the loca table.

Does this make sense? It feels like it would improve the API, at the cost of a minor deviation from the spec. It might also let us add more invariants in `write-fonts` enforcing that non-empty glyphs have at least one contour/component.

              General curiosity: should we have an `Empty` variant? This is a bit weird on the read-fonts side too. I believe I updated the loca table at some point to return `Result<Option<..` to capture this but it didn't feel good.

_Originally posted by @dfrg in https://github.com/googlefonts/fontations/pull/558#discussion_r1283458667_

cmyr:
This is a cleaner way of representing the idea of a glyph with no outlines, but which should still be added to loca.

I played around with also adding this in read-fonts, but it doesn't really play well with codegen or parsing, there, because we expect a format enum to have a table for each variant, and in the empty case we have no data from which to parse a table.

- closes  #561
- this may require some minor changes in fontc, when it lands

JMM

### DIFF
--- a/write-fonts/src/tables/glyf.rs
+++ b/write-fonts/src/tables/glyf.rs
@@ -36,6 +36,8 @@ impl TopLevelTable for Glyf {
 /// A simple or composite glyph
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Glyph {
+    /// An empty glyph gets an entry in `loca`, but no data is written to `glyf`
+    Empty,
     Simple(SimpleGlyph),
     Composite(CompositeGlyph),
 }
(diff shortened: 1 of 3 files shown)


rsheeter (approved):
LGTM. Please release and update fontc if necessary. This is probably considered a breaking change for the purpose of versioning, correct?

write-fonts/src/tables/glyf.rs
```diff
@@ -57,18 +59,11 @@ impl Glyph {
...
             Glyph::Composite(glyph) => glyph.bbox,
         }
     }
-
-    /// 'true' if the glyph contains no contours or components
-    pub fn is_empty(&self) -> bool {
-        match self {
-            Glyph::Simple(glyph) => glyph.contours().is_empty(),
-            Glyph::Composite(glyph) => glyph.components().is_empty(),
-        }
```
rsheeter, line 70: This still seems valid even in the presence of Empty?

write-fonts/src/tables/glyf.rs
```diff
@@ -57,18 +59,11 @@ impl Glyph {
     /// The bounding box for the glyph
     pub fn bbox(&self) -> Bbox {
         match self {
+            Glyph::Empty => todo!(),
```
rsheeter, line 62: Hmm... this is a big footgun. Un-lgtm!

rsheeter (changes requested):
LGTM *except* the todo! in bbox

write-fonts/src/tables/glyf.rs
cmyr: ahaha oops. This is actually something to discuss, and maybe @dfrg has input? Basically the bounding box of an empty glyph is *not* (0,0), (0,0), because that implies that the point (0,0) is contained. I think for correctness this should probably return an Option? This has been a source of bugs in other geometry code I've written in the past.

dfrg: I think returning an option here makes sense. We can always unwrap_or_default if we want the all zero bbox for some reason.

dfrg (approved):
lgtm!

rsheeter (approved):

Issue #687, opened 2023-11-02
codegen: consider generating wrappers on format enums for shared variant methods


cmyr:
As in, if there are multiple formats for a table, but those formats share some set of methods, we should generate a method on the enum that forwards the call to each of those variants.

We could also get fancier here and generate methods on the wrapper even when only one variant has the method, and make it return `None`? But that feels like it could get noisy. In that case, we could consider adding an annotation that would be added to specific methods on a variant which would cause us to generate a a method call on the enum (e.g. make this opt-in, in the case where not all variants have a method)

see #682 for an example of the code this would save us hand-writing.

rsheeter:
I suggest we start with only the case where all variants support; the user having to drill down to the specific variant seems pretty reasonable when that's not the case.

PR #760, opened 2024-01-19
[codegen] Add wrappers on enums for shared methods

cmyr:
This is for tables that have multiple formats, where all the formats have an identically named method; we will now generate a wrapper method on the enum itself that forwards to the inner methods.

In this patch this is only implemented for raw fields, and we do not generate the typed offset getters. This is quite a bit more complicated, and we can look into that as followup.

-----

This ended up producing less code than I had expected, and most of that code is just exposing the `format` field. The two main 'real' cases are the bbox on glyphs and the coords on anchor tables, both of which we had previously hand-written implementations for.

The main thing that we are missing here (pending support for offset getters) is resolving coverage tables on multi-format layout subtables.

progress on #687

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -685,20 +685,23 @@ impl Field {
         type_tokens
     }
 
-    pub(crate) fn table_getter(&self, generic: Option<&syn::Ident>) -> Option<TokenStream> {
+    pub(crate) fn table_getter_return_type(&self) -> Option<TokenStream> {
         if !self.has_getter() {
             return None;
         }
-
+        let return_type = self.raw_getter_return_type();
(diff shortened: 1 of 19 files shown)


dfrg (approved):
I’m tentatively approving because the implementation is good and this was a requested enhancement. I do have a slight concern that this might lead to suboptimal code… for example, extracting a glyph bbox generating four match expressions instead of one. I’d be curious to see the optimized llvm/asm to determine if these branches will be combined.

cmyr:
Interesting point. This does match the existing implementation, which was just using a macro_rules macro to generate these match statements, but if this ends up being a bottleneck we can always hand-write the more efficient version.

dfrg:
Sounds good to me. 🚢

PR #761, opened 2024-01-19
[skrifa] tthint: graphics state

dfrg:
First of a series of TrueType hinting patches.

This one models the TT "graphics state" which is essentially all of the data involving outlines, measurement and rounding.

Includes a tiny math module that provides support for working with fixed point values as raw i32s. This is not ideal but exists from the initial port from FreeType and will require some time and attention to correct. This will need to be deferred to a later time.


skrifa/src/outline/glyf/hint/graphics_state/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+    /// The value used in place of the control value table distance or the
+    /// actual distance value when the difference between that distance and
+    /// the single width value is less than the single width cut-in.
+    ///
+    /// See <https://developer.apple.com/fonts/TrueType-Reference-Manual/RM04/Chap4.html#single_width_value>
+    pub single_width: i32,
+}
+
+impl Default for RetainedGraphicsState {
+    fn default() -> Self {
```
rsheeter, line 183: Is there a reference for where the defaults came from?

skrifa/src/outline/glyf/hint/math.rs
```diff
@@ -0,0 +1,162 @@
...
+        b += z;
+        if z <= Wrapping(0) {
+            break;
+        }
+    }
+    Point::new(
+        (Wrapping(u.0 as i32) * sx / Wrapping(4)).0,
+        (Wrapping(v.0 as i32) * sy / Wrapping(4)).0,
+    )
+}
```
rsheeter, line 162: Some of these seems complex enough to warrant testing

rsheeter (approved):
Two broad comments:

1. Testing seems light
1. The detailed references are most appreciated

skrifa/src/outline/glyf/hint/graphics_state/mod.rs
dfrg: Added links to graphics state summary table in the spec which lists the default values.

skrifa/src/outline/glyf/hint/math.rs
dfrg: The normalize function is particularly nasty :/ Added test cases for `mul_div_no_round`, `mul14` and `normalize14` with values generated by their respective FT functions. These also include sanity checks that compute the result in floating point and test with a tolerance.

PR #762, opened 2024-01-22
[read-fonts] fix overlapping point flags

dfrg:
At some point I changed the off curve cubic flag from 0x2 to 0x8 to match the proposed cubic glyf spec and this happened to overlap with one of the IUP touched flags which leads to.. bad things. This makes the flags distinct and adds a test to prevent me from repeating the mistake in the future.


cmyr (approved):
looks good!

PR #764, opened 2024-01-23
Fix clippy warning after Transform has Copy trait

drott:


drott:
I'll go ahead and merge this, trivial change IMO.

PR #766, opened 2024-01-26
Improved Lookup API

cmyr:
This adds a new convenience API on top of the `PositionLookup` and `SubstitutionLookup` types, which obviates (in the common case) the need to dig into these enums in order to access their contents. In particular, it provides a uniform interface for accessing lookup subtables, regardless of whether or not the lookup in question is behind an extension table.

The main idea here is that the level lookup types now have a `subtables` method, which returns an enum, with a variant for each non-extension lookup type; this enum provides access to an array-like object that allows access and iteration of the underlying subtables.

In practice, this makes working with lookups *significantly* easier; for instance after updating `layout-normalizer` to use this API, I was able to replace about ~100 lines with ~10.

### DIFF
--- a/font-codegen/src/table.rs
+++ b/font-codegen/src/table.rs
@@ -36,6 +36,8 @@ pub(crate) fn generate(item: &Table) -> syn::Result<TokenStream> {
         }
     });
 
+    let of_unit_docs = " Replace the specific generic type on this implementation with `()`";
+
     // In the presence of a generic param we only impl FontRead for Name<()>,
     // and then use into() to convert it to the concrete generic type.
     let impl_into_generic = generic.as_ref().map(|t| {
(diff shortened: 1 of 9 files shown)


rsheeter (commented):
I really like how Rust usually doesn't feel like inexplicable magic incantations, this feels a little like inexplicable magic. It doesn't help that the IDE can't browse the generated code properly so it's abnormally hard to track anything down.

"to_untyped" is a scary, unintuitive, and seemingly misleading - in that we aren't sacrificing type safety - name. The comments are not particularly informative. If  "to_untyped" is a Rust norm then where might I learn more and could we please provide a reference in the code to guide future noobs? If not I wonder if we might use friendlier and more informative words.

In general, I don't think we're _actually_ doing anything particularly magical so ... how might we make this more accessible?

rsheeter (approved):
LGTM because despite the concern about scary names this does appear to make things significantly better and I don't see any immediate code problems

PR #768, opened 2024-01-30
[skrifa] fix CFF hinting overflow

dfrg:
Fixes a midpoint calculation that overflows at larger sizes.

Effectively replaces `(a + b) / 2` with `a + (b - a) / 2` to avoid the overflow (as FT does). Adds a hint map test with the data from the specific font/glyph/size that triggered this.

### DIFF
--- a/skrifa/src/outline/cff/hint.rs
+++ b/skrifa/src/outline/cff/hint.rs
@@ -573,7 +573,10 @@ impl HintMap {
                 if is_pair {
                     // Preserve stem width: position center of stem with
                     // initial hint map and two edges with nominal scale
-                    let mid = initial.transform(half(second_edge.cs_coord + first_edge.cs_coord));
+                    // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/psaux/pshints.c#L693>
+                    let mid = initial.transform(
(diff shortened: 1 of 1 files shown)


skrifa/src/outline/cff/hint.rs
```diff
@@ -1386,6 +1389,106 @@ mod tests {
         }
     }
 
+    #[test]
+    fn hint_map_midpoint_bug() {
```
rsheeter, line 1393: Can we make a test of just overflow with less setup noise? - this is a slightly stunning amount of setup noise and seemingly tests rather more than just the actual bug.

rsheeter (approved):
LGTM with mild concern about the very long test setup

skrifa/src/outline/cff/hint.rs
dfrg: I can just remove the test. Looking at it now, I think the fix is self evident and there’s no way to poke directly at that code without the setup.

rsheeter: That seems like a yellow flag in it's own right. What prevents us from making this code more readily testable?

dfrg: Your comment got me thinking so I changed the test to synthesize the "initial" hint map (which is required to trigger this code path) rather than the blues and then call the offending `insert` function directly. Reduces the setup noise quite a bit. wdyt?

PR #769, opened 2024-02-05
[skrifa] tthint: managing the graphics state

dfrg:
Add implementations for the "managing the graphics state" group of TrueType instructions (https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#managing-the-graphics-state).

### DIFF
--- a/skrifa/src/outline/glyf/hint/engine/graphics_state.rs
+++ b/skrifa/src/outline/glyf/hint/engine/graphics_state.rs
@@ -514,7 +514,7 @@ impl<'a> Engine<'a> {
         if n < 0 {
             return Err(HintErrorKind::NegativeLoopCounter);
         }
-        // As FreeType, heuristically limit the number of loops to 16 bits.
+        // As in FreeType, heuristically limit the number of loops to 16 bits.
         self.graphics_state.loop_counter = (n as u32).min(0xFFFF);
         Ok(())
     }
@@ -630,13 +630,13 @@ impl<'a> Engine<'a> {
                     *scan_control = false;
                 }
                 if (n & 0x1000) != 0 && is_rotated {
-                    // Bit 12: Set scan_control to FALSE unless the glyph is
-                    // rotated.
+                    // Bit 12: Set scan_control to FALSE based on rotation
+                    // state.
                     *scan_control = false;
                 }
                 if (n & 0x2000) != 0 && is_stretched {
-                    // Bit 13: Set scan_control to FALSE unless the glyph is
-                    // stretched.
+                    // Bit 13: Set scan_control to FALSE based on stretched
+                    // state.
                     *scan_control = false;
                 }
             }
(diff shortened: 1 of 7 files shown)


skrifa/src/outline/glyf/hint/engine/graphics_state.rs
```diff
@@ -0,0 +1,1130 @@
+//! Managing the graphics state.
+//!
+//! See <https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#managing-the-graphics-state>
+
+// 45 instructions
```
cmyr, line 5: maybe this can be part of the mod docs?

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// Pops a zone number, n, from the stack and sets zp0 to the zone with
+    /// that number. If n is 0, zp0 points to zone 0. If n is 1, zp0 points
+    /// to zone 1. Any other value for n is an error.
+    ///
+    /// See <https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#set-zone-pointer-0>
+    /// and <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/truetype/ttinterp.c#L4746>
+    pub(super) fn op_szp0(&mut self) -> OpResult {
+        let n = self.value_stack.pop()?;
+        self.graphics_state.zp0 = n.try_into()?;
+        Ok(())
```
cmyr, line 282: could just,

```suggestion
        self.graphics_state.zp0 = n.try_into()
```

I imagine?

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
```diff
@@ -0,0 +1,1130 @@
...
+    /// and <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/truetype/ttinterp.c#L4871>
+    pub(super) fn op_instctrl(&mut self) -> OpResult {
+        let selector = self.value_stack.pop()? as u32;
+        let value = self.value_stack.pop()? as u32;
+        // Selectors are indices starting with 1; not flags.
+        // Convert index to flag.
+        let selector_flag = 1 << (selector - 1);
+        if !(1..=3).contains(&selector) || (value != 0 && value != selector_flag) {
+            return Ok(());
+        }
```
cmyr, line 562: what does it mean if the selector is out of this range? Is that an expected condition?

cmyr (approved):
🚢

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
dfrg: Makes sense. I'll update the other modules as well.

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
dfrg: This ends up with a return type of `()` while the function expects `Result<(), _>` :(

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
dfrg: Short answer is that we just ignore it.

The long answer is that FreeType has a "pedantic" mode that would fail here but I've never seen it enabled in practice because it breaks real fonts 🤷 

This is something I'd like to spend more time investigating in the future... what I'd like to see is a run of the hinter with full error checking and then fall back to an unhinted glyph if it fails. I've also seen real world font bugs where running FT in non-pedantic mode leads to actual visual artifacts. 

For now though, we stick with the status quo, for better or worse.

PR #770, opened 2024-02-05
[skrifa] support Size::new(0)

### LINKED ISSUE #627: [skrifa] make ppem of 0 representable in Size type
see https://github.com/googlefonts/fontations/issues/590#issuecomment-1711595882

dfrg:
The `Size::new(ppem: f32)` constructor interpreted a ppem of 0.0 as a request for no scaling which might be surprising behavior for users. This changes the internal representation of size to hold an `Option` so that a ppem of 0.0 works as expected.

Fixes #627


drott (approved):
LGTM, thanks for fixing this.

PR #773, opened 2024-02-07
[read-fonts] tt bytecode decoder

dfrg:
Adds an opcode definition, instruction representation and decoder for TrueType bytecode.

Also renames Args -> InlineOperands and moves the type from skrifa to read-fonts.

The bulk of this is the 256 variant Opcode enum + some additional tables of the same size that capture information for each opcode. The actual decoding logic is about 30 lines of code.


read-fonts/src/tables/truetype.rs
```diff
@@ -0,0 +1,3 @@
+//! TrueType (glyf) common code.
```
cmyr, line 1: Is this a table? I might make it a submodule of `glyf` if that is more logical, we do this for a few other tables.

read-fonts/src/tables/truetype/bytecode/decode.rs
```diff
@@ -0,0 +1,237 @@
+//! TrueType bytecode decoder.
+
+use super::{InlineOperands, Instruction, Opcode};
+
+/// An error returned by [`Decoder::decode`] if the end of the bytecode
+/// stream is reached unexpectedly.
+#[derive(Copy, Clone, Debug)]
+pub struct DecodeError(());
```
cmyr, line 8:
```suggestion
pub struct DecodeError;
```

read-fonts/src/tables/truetype/bytecode/decode.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+}
+
+/// Decodes instructions from TrueType bytecode.
+#[derive(Copy, Clone)]
+pub struct Decoder<'a> {
+    /// The bytecode for the program.
+    pub bytecode: &'a [u8],
+    /// The "program counter" or current offset into the bytecode.
+    pub pc: usize,
```
cmyr, line 22: is `pub` necessary here?

read-fonts/src/tables/truetype/bytecode/decode.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    fn decode_inner(&mut self, opcode: Opcode) -> Result<Instruction<'a>, DecodeError> {
+        let mut opcode_len = opcode.len();
+        let mut count_len = 0;
+        // If the opcode length is negative the next byte contains the number
+        // of inline operands and |opcode_len| is the size of each operand.
+        // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/truetype/ttinterp.c#L7046>
+        if opcode_len < 0 {
+            let inline_count = *self.bytecode.get(self.pc + 1).ok_or(DecodeError(()))?;
+            opcode_len = -opcode_len * inline_count as i32 + 2;
```
cmyr, line 47: i would be slightly less confused by `opcode_len.abs()` but nbd

read-fonts/src/tables/truetype/bytecode/opcode.rs
```diff
@@ -0,0 +1,842 @@
...
+    MIRP11111 = 0xFF,
+}
+
+impl Opcode {
+    /// Creates an opcode from the given byte.
+    ///
+    /// There is a 1:1 mapping between bytes and opcodes.
+    #[inline]
+    pub fn from_byte(byte: u8) -> Self {
+        OPCODE_FROM_BYTE[byte as usize]
```
cmyr, line 274: given that each possible u8 value is a valid discriminant for our enum, we could simplify this, at the cost of scaring rod, and just write


```rust
unsafe { std:mem::transmute(byte) }
```

I believe we would also need to mark the enum as #[repr(transparent)].

Depending on how smart the compiler is being, this could also end up eliding a bounds check on each call, which could be meaningful.

Alternatively we could also skip that possible bounds check with `unsafe { OPCODE_FROM_BYTE.get_unchecked(&byte) }`.

read-fonts/src/tables/truetype/bytecode/opcode.rs
```diff
@@ -0,0 +1,842 @@
...
+            || self == Self::NPUSHB
+            || self == Self::NPUSHW
+    }
+
+    pub(super) fn is_push_words(self) -> bool {
+        (self >= Self::PUSHW000 && self <= Self::PUSHW111) || self == Self::NPUSHW
+    }
+
+    pub(super) fn len(self) -> i32 {
+        OPCODE_LENGTHS[self as usize] as i32
```
cmyr, line 296: worth profiling but if we're really concerned about perf and this is called hot this would be a justifiable place to use `get_unchecked` and skip a bounds check.

cmyr (approved):
Looks great, a few little comments inline, mostly about all the fun places this could be unsafe 😈

read-fonts/src/tables/truetype.rs
dfrg: I waffled on this because technically bytecode applies to glyf, fpgm and prep so it is more of a common thing. But I think putting it under glyf makes sense and removes the otherwise empty truetype module.

read-fonts/src/tables/truetype/bytecode/decode.rs
dfrg: The control flow instructions will need to read/write `pc`. Calls need to replace `bytecode` if the target function is in a different program. I could probably add methods for this but that just seems like extra complexity for what essentially becomes full access anyway 🤷🏻‍♂️

read-fonts/src/tables/truetype/bytecode/decode.rs
dfrg: Good call! This was initially `2 - opcode_len * next_byte` which was even more confusing.

read-fonts/src/tables/truetype/bytecode/opcode.rs
dfrg: I’m fairly certain llvm will elide the bounds checks because it knows the size of the array and the range of the index value.

That leaves the memory read. The optimizer certainly has enough information to avoid this as well but I’d be surprised if it does. I’ll test this later. Even if not, this array will be hot in L1 so I don’t see it being a perf factor.

I plan to do some benchmarking and profiling on this at some point so we can revisit then.

dfrg:
Thanks for the reviews. I’m going to merge this and make the requested changes in #774 since it’s closely related and will avoid rebase nonsense.

PR #775, opened 2024-02-07
[read-fonts] Basic closure of glyphs over GSUB

cmyr:
This differs from the fonttools implementation slightly, but I'm not sure if those differences are functional or just a result of how fonttools is organized.

This is a draft because I'd like to figure out tests, but I want to give cosimo a chance to look at it now and see if anything jumps out. In particular I've skipped the fancy logic at https://github.com/fonttools/fonttools/blob/534ceed9db1f5c0355d9ecd73b57db143d79a921/Lib/fontTools/subset/__init__.py#L1832-L1841 because afaict it shouldn't matter? but maybe there's some python magic going on that I'm not aware of..


read-fonts/src/tables/gsub/closure.rs
```diff
@@ -0,0 +1,188 @@
+//! Computing the closure over a set of glyphs
+//!
+//! This means taking a set of glyphs and finding the set of glyphs that are
+//! reachable from those glyphs via substitution.
```
rsheeter, line 4: It's arguably implicit implicit in the signature of closure_glyphs but I suggest you note that this means reachable by a given set of glyphs in any order, including many instances of any member of the set.

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -0,0 +1,188 @@
...
+        // we need to do this iteratively, since any glyph found in one pass
+        // over the lookups could also be the target of substitutions.
+
+        let mut all_glyphs = glyphs;
+        let mut new_glyphs = HashSet::new();
+
+        // we always call this once, and then keep calling if it produces
+        // additional glyphs
+        self.closure_glyphs_once(&all_glyphs, &mut new_glyphs)?;
+        while !new_glyphs.is_subset(&all_glyphs) {
```
dfrg, line 30: Could we use a single hash set here and just iterate as long as the size changes? That seems equivalent to me and might require fewer passes because outputs immediately become inputs that later subtables can process.

anthrotype:
> In particular I've skipped the fancy logic at https://github.com/fonttools/fonttools/blob/534ceed9db1f5c0355d9ecd73b57db143d79a921/Lib/fontTools/subset/__init__.py#L1832-L1841 because afaict it shouldn't matter?

As I said in chat, it may matter when if there are lookups that aren't referenced by any feature so they will never be applied, the subsetter treats these as if they aren't there for the sake of the closure. Even if this situation is arguably not very frequent, I think it makes sense to match that and also skip them in here.

read-fonts/src/tables/gsub/closure.rs
anthrotype: yes, fonttools also uses only one set and breaks the loop when it stops changing

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -0,0 +1,188 @@
...
+
+use super::{
+    AlternateSubstFormat1, Gsub, LigatureSubstFormat1, MultipleSubstFormat1, SingleSubst,
+    SingleSubstFormat1, SingleSubstFormat2, SubstitutionSubtables,
+};
+
+#[cfg(feature = "std")]
+impl<'a> Gsub<'a> {
+    /// Return the set of glyphs reachable from the input set via any substituion.
+    pub fn closure_glyphs(&self, glyphs: HashSet<GlyphId>) -> Result<HashSet<GlyphId>, ReadError> {
```
anthrotype, line 20: this could potentially become part of a trait when we will have to implement closure_glyphs for more tables other than GSUB (cmap, glyf, COLR, MATH, etc.)

read-fonts/src/tables/gsub/closure.rs
anthrotype: actually sorry no, fonttools still uses two sets, a mutable one that accummualtes the union of all the glyphs reachable so far (the output) and a frozenset that is an immutable copy of the former (acts as the input for the substitutions at each pass); the while True loop closing over all the lookups breaks when the ouput set does not change.
Distinguishing what goes in and out feels a bit clearer to me. But maybe Chad is right, taking all from the same cauldron and putting them back there as you co along could shorten the process and produce the same effect.
Though I'm not 100% sure about this.

cmyr: Update: yes it looks like this is no problem.

cmyr:
I've cleaned this up a bit and added tests.

I also now have a clearer sense of what fonttools is doing that we're not, and that is more carefully handling contextual lookups, so that we only apply lookups referenced by contextual lookups if we have all the glyphs to match that context.

I think it's worth doing this, but I also think it makes sense to do it as a followup.

dfrg:
I’m wondering if we can find a nice abstraction for the six contextual subtables that would simplify both closure and shaping. I’ll noodle on this a bit today.

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -0,0 +1,303 @@
...
+
+impl GlyphClosure for ReverseChainSingleSubstFormat1<'_> {
+    fn add_reachable_glyphs(&self, glyphs: &mut HashSet<GlyphId>) -> Result<(), ReadError> {
+        for coverage in self
+            .backtrack_coverages()
+            .iter()
+            .chain(self.lookahead_coverages().iter())
+        {
+            if !coverage?.iter().any(|gid| glyphs.contains(&gid)) {
+                continue;
```
dfrg, line 191: I think this should return?

dfrg (approved):
The fea based tests are pretty slick! Closure logic LGTM with one question on the reverse chain stuff.

read-fonts/src/tables/gsub/closure.rs
cmyr: oops, definitely. :)

PR #783, opened 2024-02-09
[read-fonts] Only inspect reachable lookups for closure

cmyr:
Previously we just looked at every lookup in the lookup list when computing the closure. This sounds okay at first glance, but it overlooks the fact that the lookup list can include lookups that are only reachable via other contextual lookups.

This introduces a much more rigorous approach, where on each run we recalculate the set of reachable lookups, by checking to see if the current set of glyphs matches any (chain) contextual rules, and then adding the appropriate lookups from those rules.


read-fonts/src/tables/gsub/closure.rs
```diff
@@ -49,13 +53,44 @@ impl<'a> Gsub<'a> {
...
+            if !lookups_to_use.contains(&(i as u16)) {
+                continue;
+            }
             let subtables = lookup?.subtables()?;
             subtables.add_reachable_glyphs(glyphs)?;
         }
         Ok(())
     }
+
+    fn find_reachable_lookups(&self, glyphs: &HashSet<GlyphId>) -> Result<HashSet<u16>, ReadError> {
```
dfrg, line 68: Should we also capture lookups that are reachable from feature variations? Since we're not doing instancing (yet) I guess they would all be added unconditionally.

dfrg (approved):
Nice! Just one question about collecting lookups from feature variations.

read-fonts/src/tables/gsub/closure.rs
cmyr: good call... I think I'll do this as followup since it stands alone and doesn't need to clutter this up more.

PR #787, opened 2024-02-13
[read] Consider feature variations in GSUB closure

cmyr:
In the general case we do not know enough to be smarter here, so we are just going to consider any lookups that are referenced by feature variations to be reachable.


dfrg (approved):
Looks good, 🚢!

anthrotype (approved):

PR #791, opened 2024-02-15
[skrifa] tthint: fdef/idef storage

dfrg:
Adds a new `Definition` type that is used to represent function and instruction definitions. Also includes a `DefinitionMap` type that is used to allocate and retrieve these.

This handles the case where `function_number >= max_function_defs` avoiding the question as to whether this is necessary. Well behaved fonts will hit the fast path and broken ones will fall back to a linear search.

There's a bit of churn around the `code_state` module. This was removed because it was meant to contain the decoder logic which was moved to `read-fonts`.

### DIFF
--- a/skrifa/src/outline/glyf/hint/definition.rs
+++ b/skrifa/src/outline/glyf/hint/definition.rs
@@ -14,7 +14,7 @@ use super::{error::HintErrorKind, program::Program};
 pub struct Definition {
     start: u32,
     end: u32,
-    // Either function number or opcode
+    /// The function number for an FDEF or opcode for an IDEF.
     key: i32,
     _pad: u16,
     program: u8,
@@ -168,11 +168,6 @@ impl<'a> DefinitionMap<'a> {
 mod tests {
     use super::*;
 
-    fn alloc(set: &mut DefinitionMap, key: i32) {
-        let def = set.allocate(key).unwrap();
-        *def = Definition::new(Program::Font, 0..0, key);
-    }
-
     #[test]
     fn too_many_and_invalid() {
         let mut buf = vec![Default::default(); 32];
@@ -190,6 +185,8 @@ mod tests {
         ));
     }
 
+    /// Test dense allocation where all keys map directly to indices. This is
+    /// the case for function definitions in well behaved fonts.
     #[test]
     fn allocate_dense() {
         let mut buf = vec![Default::default(); 32];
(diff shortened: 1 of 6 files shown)


skrifa/src/outline/glyf/hint/definition.rs
```diff
@@ -0,0 +1,230 @@
...
+// Note: this type is designed to support allocation from user memory
+// so make sure the fields are all tightly packed and only use integral
+// types.
+// See <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/truetype/ttobjs.h#L158>
+#[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
+#[repr(C)]
+pub struct Definition {
+    start: u32,
+    end: u32,
+    // Either function number or opcode
```
rsheeter, line 17: suggest comment to clarify that, iiuc, type is based on program

skrifa/src/outline/glyf/hint/definition.rs
```diff
@@ -0,0 +1,230 @@
...
+        for i in (0..4).chain(Some(5)) {
+            assert_eq!(slice[i].key, i as i32);
+        }
+        // The rest would be allocated from the end of the array in
+        // reverse order.
+        for (&a, b) in keys[4..7].iter().rev().zip(&slice[slice.len() - 3..]) {
+            assert_eq!(a, b.key());
+        }
+    }
+}
```
rsheeter, line 230: it's hard to tell at a glance if we are testing both forms of matching (key works vs backtrack required), I think I might have expected a test specifically for that

rsheeter (approved):
Testing looks a bit light given the range of paths one can take through some of that code but overall LGTM

skrifa/src/outline/glyf/hint/definition.rs
dfrg: The type is based on whether the definition is generated by FDEF or IDEF instructions. Both can occur in fpgm or prep. Will clarify in the comment.

skrifa/src/outline/glyf/hint/definition.rs
dfrg: This test is meant to capture the mixed case where some keys are used as indices and some aren’t. It explicitly checks the order in the backing slice to make sure allocation works as expected.

It is missing one case where a key would have worked as an index but prior fallback allocation blocks it. I’ll add that case and an additional loop over keys to ensure `get()` works successfully for all.

dfrg: Added comments describing the allocation tests (including a new one for fully sparse allocation). Renamed this one to `allocate_mixed` and simplified the backing store check. Also added the case where direct mapping would work but is blocked by prior fallback. This should cover all code paths.

PR #792, opened 2024-02-15
[skrifa] tthint: call stack

dfrg:
Fairly simple implementation of a stack to track nested function calls. Based on #791 which must land first.

### DIFF
--- a/skrifa/src/outline/glyf/hint/call_stack.rs
+++ b/skrifa/src/outline/glyf/hint/call_stack.rs
@@ -22,39 +22,39 @@ pub struct CallRecord {
 #[derive(Default)]
 pub struct CallStack {
     records: [CallRecord; MAX_DEPTH],
-    top: usize,
+    len: usize,
 }
 
 impl CallStack {
     pub fn len(&self) -> usize {
-        self.top
+        self.len
     }
 
     pub fn is_empty(&self) -> bool {
-        self.top == 0
+        self.len == 0
     }
 
     pub fn records(&self) -> &[CallRecord] {
-        &self.records[..self.top]
+        &self.records[..self.len]
     }
 
     pub fn push(&mut self, record: CallRecord) -> Result<(), HintErrorKind> {
         let top = self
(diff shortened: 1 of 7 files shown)


skrifa/src/outline/glyf/hint/call_stack.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub return_pc: usize,
+    pub current_count: u32,
+    pub definition: Definition,
+}
+
+/// Tracker for nested active function or instruction calls.
+#[derive(Default)]
+pub struct CallStack {
+    records: [CallRecord; MAX_DEPTH],
+    top: usize,
```
rsheeter, line 25: imho the behavior of this would be more intuitive if it was called "len" which more directly implies 0..len (non-inclusive) are the valid indices

skrifa/src/outline/glyf/hint/call_stack.rs
```diff
@@ -0,0 +1,92 @@
...
+        Ok(record)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn stack_ops() {
```
rsheeter, line 67: I suggest use of MAX_DEPTH not hard-coded 32 throughout.

I suggest this is several tests in one, I'd have expected to see a test that CallStack::default().pop() is underflow, a separate test for overflow, yet another test to fill/empty, etc.

rsheeter (approved):

skrifa/src/outline/glyf/hint/call_stack.rs
dfrg: Done

skrifa/src/outline/glyf/hint/call_stack.rs
dfrg: Oops, yes, changed to MAX_DEPTH. Separated the test into separate overflow/underflow/push&pop as suggested.

PR #793, opened 2024-02-16
[skrifa] tthint: function definition and dispatch

dfrg:
Implements the "defining and using functions and instructions" group (https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#defining-and-using-functions-and-instructions).

### DIFF
--- a/skrifa/src/outline/glyf/hint/engine/definition.rs
+++ b/skrifa/src/outline/glyf/hint/engine/definition.rs
@@ -104,7 +104,7 @@ impl<'a> Engine<'a> {
     /// user defined instruction.
     pub(super) fn op_unknown(&mut self, opcode: u8) -> OpResult {
         match self.do_call(DefKind::Instruction, 1, opcode as i32) {
-            // Remap an invalid defintion error to unhandled opcode
+            // Remap an invalid definition error to unhandled opcode
             Err(HintErrorKind::InvalidDefinition(opcode)) => Err(HintErrorKind::UnhandledOpcode(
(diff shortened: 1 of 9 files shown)


dfrg:
There's some unfortunate churn here. The meat is in `engine/definition.rs` and `program.rs`.

skrifa/src/outline/glyf/hint/engine/definition.rs
```diff
@@ -0,0 +1,491 @@
...
+            // Remap an invalid definition error to unhandled opcode
+            Err(HintErrorKind::InvalidDefinition(opcode)) => Err(HintErrorKind::UnhandledOpcode(
+                Opcode::from_byte(opcode as u8),
+            )),
+            result => result,
+        }
+    }
+
+    /// Common code for FDEF and IDEF.
+    fn do_def(&mut self, kind: DefKind, key: i32) -> OpResult {
```
rsheeter, line 116: several things that note they "may not exceed 64K in size" call this but I'm not immediately spotting where we enforce that. If we do, perhaps a comment?

I also don't immediately see a test for an excessively long function or instructin.

rsheeter (approved):

skrifa/src/outline/glyf/hint/engine/definition.rs
dfrg: Good catch. We don’t enforce that limit because FT doesn’t. I’m fine with adding a check if you’d prefer but also don’t see the harm in leaving it out.

PR #800, opened 2024-02-29
Make font-test-data publish

rsheeter:
This is very helpful when importing fontations into internal repositories. If it isn't published I have to either eschew the normal tools or do things like regex transformation to delete the tests.

IMHO publishing this crate is the least bad option.

JMM if happy.


rsheeter:
Proceeding per im

Issue #803, opened 2024-02-29
[skrifa] deferred hinting tests


dfrg:
Tracking issue for deferring hinting instruction tests.

- [ ] GETINFO
- [ ] GETVARIATION

PR #804, opened 2024-02-29
[skrifa] tthint: misc instructions

dfrg:
Implements the "miscellaneous instructions" group (https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#miscellaneous-instructions).

Also updates the hinting mode type to be more specific and adds docs about mapping to/from FT modes.

As discussed f2f, tests will be deferred until the remaining instructions are implemented. Tracked at #803


skrifa/src/outline/embedded_hinting.rs
```diff
@@ -5,23 +5,68 @@ use super::{
...
+        /// If set, enables support for optimized hinting that takes advantage
+        /// of subpixel layouts in LCD displays and corresponds to
+        /// `FT_LOAD_TARGET_LCD` or `FT_LOAD_TARGET_LCD_V` in FreeType.
+        ///
+        /// If unset, corresponds to `FT_LOAD_TARGET_NORMAL` in FreeType.
+        lcd_subpixel: Option<LcdLayout>,
+        /// If true, prevents adjustment of the outline in the horizontal
+        /// direction and preserves inter-glyph spacing.
+        ///
+        /// This has no corresponding setting in FreeType.
```
rsheeter, line 45: optional, indicating when you should use it might be nice

rsheeter (approved):

skrifa/src/outline/embedded_hinting.rs
dfrg: Done

PR #806, opened 2024-02-29
[skrifa] tthint: more outline instructions

dfrg:
Checkpoint of a few more instructions from the "managing outlines" group (https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#managing-outlines).

#803 updated


skrifa/src/outline/glyf/hint/graphics_state/zone.rs
```diff
@@ -401,13 +428,13 @@ impl GraphicsState<'_> {
         point_ix: usize,
         distance: i32,
     ) -> Result<(), HintErrorKind> {
+        // Note: we never adjust x in backward compatibility mode and we never
+        // adjust y in backward compability mode after IUP has been done in
+        // both directions.
```
rsheeter, line 433: It would be nice if the note said why. The inconsistent behavior is a bit weird.

skrifa/src/outline/glyf/hint/graphics_state/zone.rs
```diff
@@ -440,6 +467,70 @@ impl GraphicsState<'_> {
         }
         Ok(())
     }
+
+    pub(crate) fn move_zp2_point(
```
rsheeter, line 471: Not the most informative name or comment. "wtf is a zp2 point" might be a natural reaction. I wonder if a comment and/or informative name is achievable.

rsheeter (approved):

skrifa/src/outline/glyf/hint/graphics_state/zone.rs
dfrg: Added a small comment describing motivation and a link to https://learn.microsoft.com/en-us/typography/cleartype/truetypecleartype that has the (very) long-winded explanation.

skrifa/src/outline/glyf/hint/graphics_state/zone.rs
dfrg: Oops, forgot to link to FreeType code for this one. Added some commentary with a link.

PR #810, opened 2024-03-05
[skrifa] tthint: embrace the 26.6

dfrg:
This changes the `points` and `original` arrays in `Zone` to have an element type of `Point<F26Dot6>` as that is the type used in the outline scaler. The intention is to avoid a bit of unsafe code that was coming in a later PR and just better match the spec in general. Decided to do this now to avoid more pain later.

Lots of churn for the modified types but no functional changes.


rsheeter (approved):
My only concern is that to_bits isn't to_shreds_you_say in honor of Zoidberg

PR #811, opened 2024-03-05
[skrifa] tthint: finish outline instructions

dfrg:
Finishes up the "managing outlines" instruction group. Adds implementations for 70 opcodes.

Also changes a few other graphics state types to `F26Dot6`.

Instructions added to #803

### DIFF
--- a/skrifa/src/outline/glyf/hint/engine/outline.rs
+++ b/skrifa/src/outline/glyf/hint/engine/outline.rs
@@ -1,6 +1,6 @@
 //! Managing outlines.
 //!
-//! Implements 17 instructions.
+//! Implements 87 instructions.
 //!
 //! See <https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#managing-outlines>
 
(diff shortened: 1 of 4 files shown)


skrifa/src/outline/glyf/hint/engine/outline.rs
```diff
@@ -360,6 +360,412 @@ impl<'a> Engine<'a> {
...
+            if original_distance != F26Dot6::ZERO {
+                if old_range != F26Dot6::ZERO {
+                    new_distance = F26Dot6::from_bits(math::mul_div(
+                        original_distance.to_bits(),
+                        cur_range.to_bits(),
+                        old_range.to_bits(),
+                    ));
+                } else {
+                    new_distance = original_distance;
+                }
```
cmyr, line 720: there are a few of these where I would find `let val = if cond` to be slightly preferable to mutation? but 🤷

cmyr (approved):
just one very skippable note inline, looks good!

skrifa/src/outline/glyf/hint/engine/outline.rs
dfrg: Yeah :/ I tried not to stray too far from FreeType code but I can fix some of these up.

Issue #808, opened 2024-03-04
Consider using `bytemuck` as replacement for `read_fonts::FromBytes`


drott:
@anforowicz asks

> Can we remove/delete `read_fonts::FromBytes` and replace it with `bytemuck::AnyBitPattern`?

Reasoning:

> It seems to me that the [`read_fonts::FromBytes`](https://docs.rs/read-fonts/latest/read_fonts/trait.FromBytes.html) unsafe trait is a reimplementation of the [`bytemuck::AnyBitPattern`](https://docs.rs/bytemuck/latest/bytemuck/trait.AnyBitPattern.html) trait (see also [`bytemuck::from_bytes`](https://docs.rs/bytemuck/latest/bytemuck/fn.from_bytes.html) function).

Motivation / Benefits:

- Code reuse: `bytemuck` already provides auto trait impls and casts for various things (slices of T, mut slices of T, values of T, etc).  No need to reimplement this.
- Audit reuse: `bytemuck`'s `unsafe` usage has already been reviewed (e.g. see [go/image-crate-chromium-security-review](https://goto.google.com/image-crate-chromium-security-review)).  Removing some (most? all?) uses of `unsafe` in the `read_fonts` help will help to streamline future audits of this crate.

drott:
@anforowicz - on the second point, could you explain a bit what you include and foresee by "future audits", do you mean Google audits or audits by external parties as well?

For context, this was the original security review of `read_fonts` and fontations in general: [Internal doc](https://docs.google.com/document/d/1Whi4ONnDPdeBqXUViRLMaLTtLkFE_Qs2xRXZl0vDfYo/edit)

Additional context, unsafe / soundness discussion on `read_ref_unchecked` https://github.com/googlefonts/fontations/issues/436

I'll let Fontations folks comment, but my 2 cents:

- `read_fonts::FromBytes` is somewhat foundational code that we expect no churn or much any changes on in the future, so the likeliness of this having to be reviewed again is low. In my understanding means that we can build on the previous security review chain and do not have to revisit this, would you agree?
- We've so far been intentionally conservative with adding dependencies in order to ease adoption of the library in other open-source contexts. `byte_muck` does seem to be widely available and used, but I am not sure adding a dependency to this library for this single use-case might not be the right-tradeoff? 
[...]

PR #814, opened 2024-03-05
Adopt `bytemuck` crate

cmyr:
This replaces our internal `FromBytes` trait with `bytemuck::AnyBitPattern`.

I'm not sure if we're ready to pull the trigger on this yet, but I wanted to at least get a patch ready and see if it presented any unexpected problems. It did not.

- This lets us remove the vast majority of unsafe from read-fonts. The only remaining unsafe here is trivial, and easily replaced with an unwrap: https://github.com/googlefonts/fontations/blob/247c7c35efb9caff55838c61d2882da1dccdf223/read-fonts/src/tables/post.rs#L75-L78
-  By using the derive macro provided by bytemuck, we can be generally confident that we will not end up inadvertently generate unsound trait impls.
- This requires us to add two bits of `unsafe impl trait` to `font-types`, but these are localized and easy to reason about.
- overall I think that this is a good tradeoff, and I am happy to not be responsible for maintaining any fiddly unsafe here.
- see #808

### DIFF
--- a/font-codegen/src/flags_enums.rs
+++ b/font-codegen/src/flags_enums.rs
@@ -24,8 +24,9 @@ pub(crate) fn generate_flags(raw: &BitFlags) -> proc_macro2::TokenStream {
 
     quote! {
         #( #docs )*
-        #[derive(Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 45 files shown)


drott:
Please see https://github.com/googlefonts/fontations/issues/808#issuecomment-1988064026 - ideally let's hold landing this, at least as a mandatory dependency until Chromium adds `bytemuck`.

rsheeter:
Please rig for bytemuck 1.13.1

dfrg (approved):
🚢

PR #816, opened 2024-03-08
[WIP] Avoid cross-crate path assumptions

rsheeter:
Generate compile time consts for colrv1_json because crate paths may differ when imported into other repos.


font-test-data/build.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    for dir_entry in fs::read_dir(json_dir).unwrap() {
+        let path = dir_entry.unwrap().path();
+        let basename = path.file_name().unwrap().to_str().unwrap();
+
+        // finding things under src from OUT_DIR is weird, just copy the file
+        fs::copy(&path, out_dir.join(basename)).unwrap();
+
+        content.push_str(&format!(
+            "       \"{basename}\" => include_str!(\"{basename}\"),\n"
```
dfrg, line 38: Instead of copying, can you just replace `include_str!(..)` with the content of the file here?

dfrg: Something like:
```rust
    let test_data = fs::read_to_string(&path).unwrap();
    content.push_str(&format!(
        "        \"{basename}\" => {test_data:?},\n"
    ));
```

rsheeter: I can. I think that means I have to escape test_data though. From your example I suppose Debug for string does that?

rsheeter: TIL, that's handy.

PR #817, opened 2024-03-08
Avoid path gymnastics that fail when code is imported elsewhere

rsheeter:
The current cross-crate path building fails in google3. I tried using a build script (https://github.com/googlefonts/fontations/pull/816) and figuring out the path. So far only the simplistic approach here works.

### DIFF
--- a/font-test-data/src/lib.rs
+++ b/font-test-data/src/lib.rs
@@ -119,3 +119,68 @@ pub mod cff2 {
         0x06,
     ];
 }
+
+/// This setup to avoid cross-crate path construction and build.rs because both caused problems for google3
+pub mod colrv1_json {
+    /// Gets the expected value for a colrv1 json test
+    pub fn expected(set_name: &str, settings: &[(&str, f32)]) -> &'static str {
+        let mut key = Vec::with_capacity(1 + settings.len());
+        key.push("colrv1_".to_string() + &set_name.to_ascii_lowercase());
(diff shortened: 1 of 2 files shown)


font-test-data/src/lib.rs
```diff
@@ -119,3 +119,68 @@ pub mod cff2 {
...
+pub mod colrv1_json {
+    /// Gets the expected value for a colrv1 json test
+    pub fn expected(set_name: &str, settings: &[(&str, f32)]) -> &'static str {
+        let mut key = Vec::with_capacity(1 + settings.len());
+        key.push("colrv1_".to_string() + &set_name.to_ascii_lowercase());
+        key.extend(settings.iter().map(|(t, v)| format!("{t}_{v}")));
+        let key = key.join("_");
+        // you could generate the cases in bash using something like:
+        // for f in $(ls font-test-data/test_data/colrv1_json); do echo "\"$f\" => include_str!(\"../test_data/colrv1_json/$f\"),"; done
+        match key.as_str() {
```
drott, line 133: Regenerating the results was agreed in an earlier review to be controlled by an environnment variable which then generates the expectations files into the set of folders, compare: https://github.com/googlefonts/fontations/blob/main/skrifa/src/color/traversal_tests/mod.rs#L255

Manually listing them here is prone to breaking when additional tests are added in `traversal_tests`.

dfrg (approved):

PR #818, opened 2024-03-08
Support hb-draw style walking of glyf pointstreams that start with an off-curve

rsheeter:
See problem #2 noted in https://github.com/googlefonts/mimir/pull/3

In making this change I found to_path confusing and attempted to simplify it so this is built on #827.


anthrotype:
FWIW, hb-draw behavior is consistent with fonttools PointToSegmentPen, which similarly finds the first on-curve point and rotates the point list so that the moveTo happens on it, and any leading off-curves are pushed to the last curve segment.

And since you're touching this critical piece of code, please make sure you _keep_ the current behavior in skrifa (which matches FontTools BasePen) for those special TrueType all-quad-off-curve contours (where there aren't any on-curve points to rotate), which is to set the moveTo to the midpoint between the first and last off-curve points. There's probably already a test for that.

rsheeter:
> hb-draw behavior is consistent with fonttools PointToSegmentPen

I'm curious, was there any particular reason to make that choice vs probing the end as FreeType does?

anthrotype:
I have no idea, the code in ufoLib.pointPen is very old, maybe Just or Tal know. It doesn't really matter in the end, both ways are valid.

rsheeter:
> make sure you keep the current behavior in skrifa (which matches FontTools BasePen) for those special TrueType all-quad-off-curve contours (where there aren't any on-curve points to rotate), which is to set the moveTo to the midpoint between the first and last off-curve points

Good point, I'll confirm there is a test or add one.

read-fonts/src/tables/glyf.rs
```diff
@@ -741,15 +753,21 @@ impl<'a> ContourIter<'a> {
             // Establish our first move and the starting index for our waltz over the points
             first_move = contour.points[0];
             if contour.flags[0].is_off_curve_quad() {
-                let maybe_oncurve_ix = contour.points.len() - 1;
+                let maybe_oncurve_ix = match contour.offcurve_first_mode {
+                    OffCurveFirstMode::Back => contour.points.len() - 1,
+                    OffCurveFirstMode::Front => (start_ix + 1).min(contour.points.len() - 1),
```
anthrotype, line 758: why do you only look at the immediately next or previous point and not also the following ones? TrueType curves with implied oncurve points may contain several offcurves preceding an oncurve. The fonttools PointToSegmentPen looks forward until it finds any oncurve point and uses that as the moveTo. If it can't find any oncurves, it takes the midpoint between the first and last off-curves.

skrifa/src/outline/mod.rs
```diff
@@ -554,16 +572,120 @@ mod tests {
...
+        points: Vec<GlyphPoint>,
+    }
+
+    impl PointCaptor {
+        fn new() -> Self {
+            Self { points: Vec::new() }
+        }
+
+        fn into_points(mut self) -> Vec<GlyphPoint> {
+            // if we start and end with the same on-curve we can drop the end
```
anthrotype, line 602: that's ok (the fonttools SegmentToPointPen also does the same) -- however there is an important point to bear in mind when converting from segment to points as you drop the last point overlapping the first. If you want to be able to round-trip back from points to segment, which is an equally important thing to be wishing to do, when you draw a list of points as segments (moveTo, lineTo, quadTo, etc.), you need to take care of explicitly emitting the last lineTo command which would be implied when the contour is closed ('Z' or closePath). The PointToSegmentPen has an option that when True it unconditionally emits these; when False (default) it only emits them for closed contours if the last and first point do overlap (imagine a four points that in some location look like a triangle with duplicate points at the end and then can morph into a rectangle); it does this to disambiguate this occurrence from the implied closing lineTo. If it did not do that, the duplicate points would be lost when converted back to segments.

anthrotype: in write-fonts::tables::glyf::simple, where we have the opposite problem of converting a BezPath (sequence of curve _segments_) into a stream of points, we have this special handling for the duplicate last point in closed contours, see the test here:

https://github.com/googlefonts/fontations/blob/0cf2a550ceb7a6c2830767da9311a66fd1c11ccd/write-fonts/src/tables/glyf/simple.rs#L1285-L1331

it's only natural that the respective read_fonts::tables::glyf::to_path (which converts from point stream to segments or path drawing commands) does the equivalent of this but on the reverse.

read-fonts/src/tables/glyf.rs
rsheeter: Good catch!

I think I should rename the variants to more clearly indicate their goals which are really to match the behavior of existing systems. The back mode matches freetype. Assuming the original source was correct freetype doesn't do the walk until oncurve bit. Front matches hb-draw, if hb-draw does so then I'm missing a test and should add + fix.

rsheeter: For the point stream:

```xml
<pt x="278" y="710" on="0"/>
<pt x="278" y="470" on="0"/>
<pt x="998" y="470" on="1"/>
<pt x="998" y="710" on="0"/>
<pt x="750" y="500" on="0"/>
```

Our internal hb-draw based tool produces:

```
M278,-590
Q278,-470 998,-470
Q998,-710 874,-605
Q750,-500 514,-605
Q278,-710 278,-590
Z
```

Negative is because it just flips y. The starting move to y of 590 seems to support the midpoint between first and second rather than scanning forward multiple steps to find the oncurve at 998,470.

rsheeter: Added a test based on the 2-off, 1-on, 2-off case above.

anthrotype: ok thanks for confirming that.

skrifa/src/outline/mod.rs
rsheeter: Good point, I will rig a test and fix.

rsheeter: Looks to me like pens are always advised of a closing line, even if it's a nop. Added the test `always_get_closing_lines` to prove it and ensure it remains so.

Added `triangle_and_square_retain_compatibility` to prove our triangle and square are both captured as 4 points. PointPen's handling of close now has a more limited drop-last with rationale. Ptal.

read-fonts/src/tables/glyf.rs
```diff
@@ -711,6 +728,7 @@ struct Contour<'a> {
     points: &'a [Point<F26Dot6>],
     /// The flags for points, sliced from a glyphs flag stream
     flags: &'a [PointFlags],
+    offcurve_first_mode: ToPathStyle,
```
dfrg, line 731: This name is used in lots of places but doesn't really match the type. Nothing seems particularly good, but maybe just `path_style` (to avoid the ugly "to" prefix everywhere)?

read-fonts/src/tables/glyf.rs
```diff
@@ -678,6 +682,19 @@ impl fmt::Display for ToPathError {
...
+/// off-curve. Major implementations differ. Which one would you like to match?
+#[derive(Debug, Default, Copy, Clone)]
+pub enum ToPathStyle {
+    /// If the first point is off-curve, check if the last is on-curve
+    /// If it is, start there. If it isn't, start at the implied midpoint between first and last.
+    #[default]
+    FreeType,
+    /// If the first point is off-curve, check if the second is on-curve.
+    /// If it is, start there. If it isn't, start at the implied midpoint between first and second.
+    HbDraw,
```
dfrg, line 695: optional but I feel like `HarfBuzz` is more easily recognizable than `HbDraw` even if it's slightly less specific.

skrifa/src/outline/mod.rs
```diff
@@ -185,6 +189,11 @@ impl<'a> DrawSettings<'a> {
         self.memory = memory;
         self
     }
+
+    pub fn with_offcurve_first_mode(mut self, off_curve_first_mode: ToPathStyle) -> Self {
```
dfrg, line 193: Not super happy with the naming and I feel like this also deserves a doc comment. Alternatively, make it `#[doc(hidden)]` so we don't even confuse users who have no idea why they'd want this

dfrg (approved):
Code LGTM. A few nits about naming and comments.

Also I just merged hinting stuff which probably stepped on some of this code :(

skrifa/src/outline/mod.rs
rsheeter: Oops, let me rename it to match ToPathStyle. I don't think it should be hidden. Will add a comment.

read-fonts/src/tables/glyf.rs
rsheeter: Will fix.

read-fonts/src/tables/glyf.rs
rsheeter: Good call, will rename.

skrifa/src/outline/mod.rs
rsheeter: Fixed both issues.

PR #819, opened 2024-03-08
[write-fonts] Fix bug in ClassDef format selection

cmyr:
For format1 we were just taking the number of glyphs, but as the values stored are u16 it should be number of glyphs * 2.

This was the root cause of an unexpected packing failure.

### DIFF
--- a/write-fonts/src/tables/layout.rs
+++ b/write-fonts/src/tables/layout.rs
@@ -470,7 +470,7 @@ impl ClassDefBuilder {
         // calculate our format2 size:
         let first = self.items.keys().next().map(|g| g.to_u16());
         let last = self.items.keys().next_back().map(|g| g.to_u16());
-        let len_format1 = 3 + (last.unwrap_or_default() - first.unwrap_or_default()) as usize;
+        let len_format1 = 3 + ((last.unwrap_or_default() - first.unwrap_or_default()) * 2) as usize;
         let len_format2 = 4 + iter_class_ranges(&self.items).count() * 6;
 
         len_format1 < len_format2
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/layout.rs
```diff
@@ -706,6 +706,23 @@ mod tests {
             .collect();
 
         assert!(builder.prefer_format_1());
+
+        fn iter_class_items(
```
rsheeter, line 710: This seems like lots of cases in one test. Why isn't this three (one per let builder = ) different tests?

write-fonts/src/tables/layout.rs
```diff
@@ -706,6 +706,23 @@ mod tests {
...
+        }
+
+        // 3 ranges of 4 glyphs at 6 bytes a range should be smaller than writing
+        // out the 3 * 4 classes directly
+        let builder: ClassDefBuilder = iter_class_items(5, 8, 3)
+            .chain(iter_class_items(9, 12, 4))
+            .chain(iter_class_items(13, 16, 5))
+            .collect();
+
+        assert!(!builder.prefer_format_1());
```
rsheeter, line 725: The output when this fails will not be helpful. If it might be useful I suggest printing the builder in all the asserts.

rsheeter (approved):

write-fonts/src/tables/layout.rs
cmyr: no clear advantage to me either way, if you have a preference then happy to change it

rsheeter: As a general pattern I've found more granular tests easier to work with than single tests that encompass many cases. It's very frustrating when you fix the test only to find out you fixed case 1 and that one test is actually 7 unit tests in one.

cmyr: okay I like that rationale.

PR #824, opened 2024-03-08
remove unsafe from skrifa...

dfrg:
and make both read-fonts and skrifa #![forbid(unsafe_code)]

### DIFF
--- a/read-fonts/src/tables/post.rs
+++ b/read-fonts/src/tables/post.rs
@@ -72,10 +72,12 @@ impl<'a> FontRead<'a> for PString<'a> {
             .as_bytes()
             .get(1..len as usize + 1)
             .ok_or(ReadError::OutOfBounds)?;
+
         if pstring.is_ascii() {
-            // SAFETY: `pstring` must be valid UTF-8, which is known to be the
-            // case since ASCII is a subset of UTF-8 and `is_ascii()` is true.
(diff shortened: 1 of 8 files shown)


read-fonts/src/tables/post.rs
```diff
@@ -72,10 +72,12 @@ impl<'a> FontRead<'a> for PString<'a> {
...
             .ok_or(ReadError::OutOfBounds)?;
+
         if pstring.is_ascii() {
-            // SAFETY: `pstring` must be valid UTF-8, which is known to be the
-            // case since ASCII is a subset of UTF-8 and `is_ascii()` is true.
-            Ok(PString(unsafe { std::str::from_utf8_unchecked(pstring) }))
+            let str = std::str::from_utf8(pstring).map_err(|_| {
+                ReadError::MalformedData("we already checked that the string was ASCII!")
+            })?;
+            Ok(PString(str))
```
cmyr, line 80: I think an error here is noise since given the check we know that branch can never be hit. I would just unwrap.

cmyr (approved):
looks good, one note inline.

read-fonts/src/tables/post.rs
dfrg: Sounds good. I’ll change this and merge into your branch so we have a complete version ready to go if we get approval.

PR #825, opened 2024-03-11
Revert "[Skrifa] Avoid initialization of HashSet with system random numbers (#751)

drott:
This reverts commit eef6a26640ace05d48c4e2150e303e5c6ac32201.

See details in [1] and [2]. Rust upstream landed a change switching to a different, sandbox-compatible call for PRNG, except on Windows
7. Windows 7 is no longer supported by Chromium. Chromium is our only known sandboxed environment where this call would fail. Remove the workaround.

[1] https://issues.chromium.org/40277768
[2] https://github.com/rust-lang/rust/commit/08caefbb103d1809113172d16eaf8f66c2edc2f1

### DIFF
--- a/skrifa/src/color/mod.rs
+++ b/skrifa/src/color/mod.rs
@@ -57,9 +57,7 @@ use read_fonts::{
 
 use std::{collections::HashSet, fmt::Debug, ops::Range};
 
-use traversal::{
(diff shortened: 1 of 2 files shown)


dfrg (approved):
LGTM!

PR #827, opened 2024-03-11
Attempt to simplify glyf to_path

rsheeter:
I found the original code, directly ported from FreeType, a bit hard to follow and attempted to break it into smaller more tractable parts. Meant to make #818, which will rebase on top, simpler.

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -775,7 +775,7 @@ impl<'a> ContourIter<'a> {
         &mut self,
         curr: Point<F26Dot6>,
         oncurve: OnCurve,
-        sink: &mut impl Pen,
+        pen: &mut impl Pen,
     ) -> Result<(), ToPathError> {
         let buf_ix = self.off_curve[0];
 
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/glyf.rs
```diff
@@ -665,7 +667,254 @@ impl fmt::Display for ToPathError {
...
+            });
+        }
+        Ok(Self {
+            base_offset,
+            points,
+            flags,
+        })
+    }
+
+    fn drain(self, sink: &mut impl Pen) -> Result<(), ToPathError> {
```
anthrotype, line 901: **¡bikeshed alert!** Draining a sink is not as nice a metaphor as drawing with a pen ... How about draw(pen) ? :)

rsheeter: I was trying to keep the sink naming @dfrg  had going alive. I tend to agree that draw(pen) better describes what's happening, will fix.

cmyr (approved):
No notes, but this is a very nice looking patch, and I think the logic is definitely easier to follow.

read-fonts/src/tables/glyf.rs
rsheeter: Done

rsheeter:
Confirmed good w/@dfrg, merging.

dfrg:
yes, +1 from me. Nice work!

PR #829, opened 2024-03-12
[fauntlet] update for new outline API

dfrg:
This also comments out the advance check (waiting on FT compat metrics), sets us up for hint testing, and removes the old scale module from skrifa.


cmyr (approved):
👍

PR #832, opened 2024-03-13
[skrifa] tthint working checkpoint 

dfrg:
This wires up the TrueType hinting code in a working state that matches FreeType for all of Noto and Google Fonts collections with the following exceptions:

1. Some fonts with buggy bytecode
2. Fonts with control value programs that are not idempotent over multiple runs (these are newish Win11 fonts)
3. Variable fonts with a cvar table (https://github.com/googlefonts/fontations/issues/831)

### DIFF
--- a/skrifa/src/outline/glyf/hint/engine/graphics_state.rs
+++ b/skrifa/src/outline/glyf/hint/engine/graphics_state.rs
@@ -151,7 +151,7 @@ impl<'a> Engine<'a> {
     pub(super) fn op_spvfs(&mut self) -> OpResult {
         let y = self.value_stack.pop()? as i16 as i32;
         let x = self.value_stack.pop()? as i16 as i32;
-        let vector = if x == 0 && y == 0 {
+        let vector = if (x, y) == (0, 0) {
             self.graphics_state.proj_vector
         } else {
             math::normalize14(x, y)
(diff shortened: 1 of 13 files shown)


skrifa/src/outline/glyf/hint/engine/graphics_state.rs
```diff
@@ -151,7 +151,11 @@ impl<'a> Engine<'a> {
     pub(super) fn op_spvfs(&mut self) -> OpResult {
         let y = self.value_stack.pop()? as i16 as i32;
         let x = self.value_stack.pop()? as i16 as i32;
-        let vector = math::normalize14(x, y);
+        let vector = if x == 0 && y == 0 {
```
cmyr, line 154: totally unnecessary but you could write this as, `if (x, y) == (0, 0)`

skrifa/src/outline/glyf/hint/engine/mod.rs
```diff
@@ -47,6 +47,45 @@ pub struct Engine<'a> {
 }
 
 impl<'a> Engine<'a> {
+    #[allow(clippy::too_many_arguments)]
```
cmyr, line 50: I mean, clippy has my sympathies here...

cmyr (approved):
you're wandering in the wilderness; all I can do is stand on a distant hill, holding a torch.

skrifa/src/outline/glyf/hint/engine/graphics_state.rs
dfrg: yes, I like this better

skrifa/src/outline/glyf/hint/engine/mod.rs
dfrg: and mine... I actually wrote the clippy allow before the arguments because I knew what was coming. I'll likely just make all the fields in this thing public in the future anyway. I have aspirations to expose the guts of the hinting code to allow for debugging in which case users will likely want to poke at all the internal state

PR #833, opened 2024-03-13
[read-fonts] parse cvar table

### LINKED ISSUE #831: [read-fonts] add support for the cvar table

dfrg:
A large chunk of common code was moved from `gvar.rs` to `variations.rs` making this look significantly larger than it actually is. Real changes are in the `cvar.rs` files.

Closes #831


read-fonts/src/tables/variations.rs
```diff
@@ -498,6 +502,277 @@ impl<'a> Iterator for TupleVariationHeaderIter<'a> {
...
+        self.header
+            .tuple_index()
+            .tuple_records_index()
+            .and_then(|idx| self.shared_tuples.as_ref()?.tuples().get(idx as usize).ok())
+            .or_else(|| self.header.peak_tuple())
+            .unwrap_or_default()
+    }
+
+    // transcribed from pinot/moscato
+    /// Compute the scalar for a this tuple at a given point in design space.
```
cmyr, line 640:
```suggestion
    /// Compute the scalar for this tuple at a given point in design space.
```

?

read-fonts/src/tables/variations.rs
```diff
@@ -498,6 +502,277 @@ impl<'a> Iterator for TupleVariationHeaderIter<'a> {
...
+pub struct TupleVariationData<'a, T> {
+    pub(crate) axis_count: u16,
+    pub(crate) shared_tuples: Option<SharedTuples<'a>>,
+    pub(crate) shared_point_numbers: Option<PackedPointNumbers<'a>>,
+    pub(crate) tuple_count: TupleVariationCount,
+    // the data for all the tuple variation headers
+    pub(crate) header_data: FontData<'a>,
+    // the data for all the tuple bodies
+    pub(crate) serialized_data: FontData<'a>,
+    pub(crate) _marker: std::marker::PhantomData<T>,
```
cmyr, line 515: This always confuses me but I suspect this is not the param we want? This is telling rustc that we actually own an instance of `T`, which influences things like drop checking. Staring at https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns makes me think we want `PhantomData<fn() -> T>`? 🤮

cmyr (approved):
took me an hour of staring at this cross-eyed to figure out what was going on but it's actually pretty straight forward? Some little things inline..

read-fonts/src/tables/variations.rs
dfrg: Oops. I reworded the comment a bit and removed the "// transcribed.." part. I license my work for this repo :)

read-fonts/src/tables/variations.rs
dfrg: Good catch! I think I actually got caught up on this somewhere and had to add a lifetime bound.

Fixed

dfrg:
> took me an hour of staring at this cross-eyed to figure out what was going on but it's actually pretty straight forward? Some little things inline..

The differences in gvar/cvar are subtle. My first attempt actually just reused almost all of gvar. Second attempt cleaned it up but used a single concrete `TupleDelta` type with x/y coords even for cvar. Final pass changed `TupleDelta` to a trait with distinct delta types for cvar/gvar and added all the generic madness but I think it was ultimately worth it.

PR #836, opened 2024-03-14
Make Cmap::from_mappings fallable.

rictic:
Step one of https://github.com/googlefonts/fontc/issues/457

This checks for collisions as part of sorting, avoiding another walk of the mappings.

As this is a breaking change, is there a changelog I should update or a version number I should bump as part of this?

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -164,6 +164,35 @@ impl CmapSubtable {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MappingError {
+    /// When a char is mapped to multiple distinct GlyphIds.
+    /// gid1 is less than gid2.
+    /// If there are multiple conflicting mappings, one is chosen arbitrarily.
+    ConflictingMappings {
+        ch: char,
+        gid1: GlyphId,
+        gid2: GlyphId,
+    },
+}
+
+impl std::fmt::Display for MappingError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        match self {
+            MappingError::ConflictingMappings { ch, gid1, gid2 } => {
+                let ch32 = *ch as u32;
+                write!(
+                    f,
+                    "Cannot map {ch:?} (U+{ch32:04X}) to two different glyph ids: {gid1} and {gid2}"
+                )
+            }
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for MappingError {}
+
 impl Cmap {
     /// Generates a [cmap](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap) that is expected to work in most modern environments.
     ///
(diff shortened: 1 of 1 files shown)


rictic:
Another nearby opportunity, I think this continue can become a break, as the function documents that the mappings are assumed to be sorted: https://github.com/googlefonts/fontations/blob/main/write-fonts/src/tables/cmap.rs#L45

write-fonts/src/tables/cmap.rs
```diff
@@ -164,6 +164,35 @@ impl CmapSubtable {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum MappingError {
```
cmyr, line 168: if we only have a single error condition it's reasonable to just use a `struct` as the error type, which will simplify this somewhat.


edit: and to bikeshed, I would maybe just call that type `InvalidMapping`? or `CmapConflict`? 🤷

write-fonts/src/tables/cmap.rs
```diff
@@ -164,6 +164,35 @@ impl CmapSubtable {
...
+    /// When a char is mapped to multiple distinct GlyphIds.
+    /// gid1 is less than gid2.
+    /// If there are multiple conflicting mappings, one is chosen arbitrarily.
```
cmyr, line 171: Rustdoc renders doc comments sort of like git renders commits; the first line is expected to be a summary, followed by a newline, and then additional docs follow that.

More concretely, the first line is rendered as a short description when listing items in a module or crate, and the longer docs are only rendered in the specific page for that item.

write-fonts/src/tables/cmap.rs
```diff
@@ -172,9 +201,27 @@ impl Cmap {
...
+                } else {
+                    (*gid1, *gid2)
+                };
+                error = Some(MappingError::ConflictingMappings { ch: *a, gid1, gid2 })
+            }
+            ord
+        });
+        if let Some(err) = error {
+            return Err(err);
+        }
```
cmyr, line 224: this works (and is nice in that we only need a single pass) but it's also a bit tricky to follow, and this code is not particularly hot. I'd personally write this in two passes, like (and now I'm mainly just entertaining myself)

```suggestion
        mappings.sort();
        if let Some((ch, gid1, gid2)) =
            mappings
                .iter()
                .zip(mappings.iter().skip(1))
                .find_map(|((c1, g1), (c2, g2))| {
                    (c1 == c2 && g1 != g2).then(|| (*c1, *g1.min(g2), *g1.max(g2)))
                })
        {
            return Err(MappingError::ConflictingMappings { ch, gid1, gid2 });
        }
```

I'm also wondering how we handle the situation where the mapping just includes duplicates? I think that we should actually be calling `mappings.dedup()` after we sort...

write-fonts/src/tables/cmap.rs
```diff
@@ -360,8 +409,8 @@ mod tests {
             ('\u{1f133}', GlyphId::new(484)),
             // gid 485 skipped, starts third group
             ('\u{1f134}', GlyphId::new(486)),
-            // char 0x1f135 skipped, starts fourth group; 0x1f136 defined twice, last map prevails
-            ('\u{1f136}', GlyphId::new(487)),
+            // char 0x1f135 skipped, starts fourth group. identical duplicate bindings are fine
```
cmyr, line 412: Are we confident this is true?

cmyr (approved):
okay some notes inline but overall this looks good!

write-fonts/src/tables/cmap.rs
rictic: Done

write-fonts/src/tables/cmap.rs
rictic: Yeah, easy to dedup after sorting, assuming this code isn't hot enough to mind doing extra passes.

anthrotype (approved):
LGTM, thanks!

cmyr (approved):
looks good, thanks!

PR #837, opened 2024-03-15
Derive Debug on AxisCollection

rictic:


rictic:
Deferring this for now, this only works with std which we don't want to make mandatory in skrifa

PR #839, opened 2024-03-15
[skrifa] tthint: use cvar table

dfrg:
Applies cvar deltas to the cvt.


skrifa/src/outline/glyf/hint/instance.rs
```diff
@@ -171,13 +171,40 @@ impl HintInstance {
             Definition::default(),
         );
         self.cvt.clear();
+        if let Some(cvar) = outlines.cvar.as_ref() {
```
cmyr, line 174: not sure it's useful here, but something that I have found helpful for the idea of "I need both of these optional things to exist" is,

```rust
if let Some((one, two)) = opt1.zip(opt2) {
...
}
```

skrifa/src/outline/glyf/hint/instance.rs
```diff
@@ -171,13 +171,40 @@ impl HintInstance {
...
+            if let Ok(var_data) = cvar.variation_data(axis_count) {
+                for (tuple, scalar) in var_data.active_tuples_at(coords) {
+                    for delta in tuple.deltas() {
+                        let ix = delta.position as usize;
+                        if let Some(value) = self.cvt.get_mut(ix) {
+                            *value += delta.apply_scalar(scalar).to_bits();
+                        }
+                    }
+                }
+            }
```
cmyr, line 186: it's easy to imagine a little helper method that handles accumulating the deltas?

cmyr (approved):
👍

skrifa/src/outline/glyf/hint/instance.rs
dfrg: ah, hadn't seen this before. I don't see how I'd use it here but will definitely add it to the toolbox!

skrifa/src/outline/glyf/hint/instance.rs
dfrg: good call. I'll add this to the cvar table

PR #840, opened 2024-03-15
[skrifa] post tthint cleanup

dfrg:
This is just a lot of shuffling I wanted to do but avoided while I had chained PRs pending. Makes module structure and naming more consistent.

Also removes the `#![allow(dead_code)]` attribute that was in place while things were incrementally landing.

No functional changes so JMM if you're brave.


cmyr (approved):
👍

PR #842, opened 2024-03-15
[skrifa] no_std support

dfrg:
Brings in the [core_maths](https://github.com/robertbastian/core_maths) as an optional dependency for floating point support. This is relatively new and not currently used much but is a [dependency of icu_segmenter](https://github.com/unicode-org/icu4x/blob/9491b6358f0e59776cc3a1fe2efb2115a5c02506/components/segmenter/Cargo.toml#L33).

Only real functional change is that the visited set of COLRv1 becomes a nop and we depend on the recurse depth check to avoid cycles of the infinite variety.


skrifa/src/color/traversal.rs
```diff
@@ -13,6 +13,33 @@ use super::{
...
+#[allow(unused_imports)]
+use core_maths::*;
+
+#[cfg(any(test, feature = "std"))]
+mod visited_set {
+    pub type VisitedSet = std::collections::HashSet<usize>;
+}
+
+#[cfg(not(any(test, feature = "std")))]
+mod visited_set {
```
rsheeter, line 28: Please add a comment explaining why this implementation actually makes sense as at a glance it looks an awful lot like someone committed a wip by accident

rsheeter (approved):

skrifa/src/color/traversal.rs
dfrg: As opposed to committing a wip on purpose? :)

PR #845, opened 2024-03-18
[font-types] float to fixed point rounding

dfrg:
We had two implementations for float to fixed conversion based on whether std was available and the no_std implementation had a very subtle rounding bug (adding -0.5 for negative values but 0.0 for positive ones) causing some tests to fail.

This fixes the bug, restores the original tests and adds a new one to check rounding behavior in font-types.


cmyr (approved):
Thanks!

PR #848, opened 2024-03-19
Prepare to release 0.25.0 of write-fonts to use fallible Cmap::from_mappings in fontbe

rictic:


rsheeter:
Suggest commenting why, e.g. what is it you need released to use where.

rsheeter (approved):
LGTM. Note that we have an informal no lgtm required for version bumps policy.

PR #850, opened 2024-03-20
[chore] Bump env_logger

cmyr:


write-fonts/Cargo.toml
```diff
@@ -29,7 +29,7 @@ font-test-data = { path = "../font-test-data" }
 read-fonts = { version = "0.17.0", path = "../read-fonts", features = [
   "codegen_test",
 ] }
-env_logger = "0.10.0"
```
rsheeter, line 32: This seems like it should move to the workspace toml?

rsheeter (approved):

write-fonts/Cargo.toml
cmyr: 👍 good idea

PR #852, opened 2024-03-20
Fix font-types docs, and bump patch version

cmyr:
There is now some unsafe in font-types, which lets us remove unsafe in crates with more churn.


rsheeter:
Can we *add* the safety claim to the libraries where it applies post-bytemuck? - e.g. https://crates.io/crates/read-fonts et al

rsheeter (approved):

PR #854, opened 2024-03-21
[clippy] Push the rock back up the hill

cmyr:


font-codegen/src/parsing.rs
```diff
@@ -257,6 +257,9 @@ pub(crate) struct SinceVersion {
 /// ```
 #[derive(Clone, Debug)]
 pub(crate) enum Count {
+    // the field isn't used, but it's nice to hold onto if we want to print errors
+    // in the future
+    #[allow(dead_code)]
```
dfrg, line 262: this was my plan as well 🤷🏻‍♂️

dfrg (approved):
🪨

PR #855, opened 2024-03-21
Add depgraph showing relationships among the crates

rictic:
Split out from https://github.com/googlefonts/fontc/pull/746#pullrequestreview-1950256804

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ robust and performant open tools for a variety of font engineering and
 production tasks. For an overview of the motivations, see
 [googlefonts/oxidize][oxidize].
 
-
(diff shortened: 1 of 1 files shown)


cmyr:
Cool, I think this is useful!

some notes:

- I think we can omit `font-codegen`; it really doesn't make sense as part of the dependency graph, and users don't really need to know it exists.
- I'd also omit `otexplorer`, which is in constant limbo and very likely to be deleted at some point (it was a tool I used to test codegen during development and which I still find helpful but is frail and a minor maintenance burden)

rictic:
Updated, PTAL!

PR #860, opened 2024-03-25
[read-fonts] Properly compute length of delta sets

### LINKED ISSUE #857: read-fonts should not assume ItemVariationData.delta_sets [u8] array contains all remaining table bytes
I discovered this while debugging https://github.com/harfbuzz/harfbuzz/issues/4636 using otexplorer

Apparently when we parse an ItemVariationData table in the ItemVariationStore (e.g. the one found in a GDEF table), we assume that the `delta_sets` array consumes the entirety of the remaining bytes, see:

https://github.com/googlefonts/fontations/blob/db354340f8aa5bb482ee7df07c8662d8231731bd/resources/codegen_inputs/variations.rs#L159-L160

which translates to

https://github.com/googlefonts/fontations/blob/db354340f8aa5bb482ee7df07c8662d8231731bd/read-fonts/generated/generated_variations.rs#L1145

But sometimes HB decides to place the ItemVariationStore at the beginning, instead of at the end of, the GDEF table, and this makes otexplorer believe that the `delta_sets` array is longer than it actually is.

Take this font below, produced by running hb-subset on the font @simoncozens posted in the above linked HB issue:
 
[subset.ttf.zip](https://github.com/googlefonts/fontations/files/14745688/subset.ttf.zip)

If you run `otexplorer subset.ttf -q GDEF`, you'll see that the final `delta_sets` array contain hundreds more entries than the expected 81 (item_count is 81, there is only one region referenced by this VarData table, and each delta takes 1 byte each in this particular example).

It should be fine when using read-fonts's `ItemVariationData.delta_set` method with an `inner_index` that doesn't exceed the stated `item_count`:

https://github.com/googlefonts/fontations/blob/db354340f8aa5bb482ee7df07c8662d8231731bd/read-fonts/src/tables/variations.rs#L982

However otexplorer probably lists all the `delta_sets` that the slice claims to contain and thus you may see additional items that should not be actually there.

cmyr:
We were previously just taking all remaining data here, but we have enough information to compute the expected length exactly. This will let us better handle the case where data is missing.

- closes #857

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -288,6 +288,8 @@ pub(crate) enum CountTransform {
     DeltaSetIndexData,
     /// three args: the axis count, the tuple index, and a constant on that index
     TupleLen,
+    /// only ItemVariationStore: requires item_count, word_delta_count, region_index_count
+    ItemVariationDataLen,
 }
 
 /// Attributes for specifying how to compile a field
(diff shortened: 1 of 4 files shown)


anthrotype:
nice thank you!

PR #861, opened 2024-03-26
Preserve format of value records after reading

### LINKED ISSUE #858: Recompiling GPOS unchanged fails for some fonts
Example font: https://cdn.jsdelivr.net/gh/notofonts/notofonts.github.io/fonts/NotoSansAdlam/full/slim-variable-ttf/NotoSansAdlam[wght].ttf

Test code:

```rust
fn main() {
    use read::{FontRef, TableProvider};
    use write_fonts;
    use write_fonts::read;
    use write_fonts::tables::gpos::{PairPos, PositionLookup};
    use write_fonts::{from_obj::ToOwnedTable, tables::gpos::Gpos, FontBuilder};
    let font_bytes = std::fs::read("NotoSansAdlam[wght].ttf").unwrap();
    let font = FontRef::new(&font_bytes).expect("failed to read font data");
    let gpos: Gpos = font.gpos().expect("missing 'GPOS' table").to_owned_table();
    let new_bytes = FontBuilder::new()
        .add_table(&gpos)
        .expect("cannot compile 'GPOS' table")
        .copy_missing_tables(font)
        .build();
    std::fs::write("mynewfont.ttf", &new_bytes).unwrap();
}
```

Output:

```
cannot compile 'GPOS' table: BuilderError { tag: Tag(GPOS), inner: ValidationFailed(6 validation errors:
#1
"all ValueRecords must have same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[0]
      PairPosFormat1.pair_sets[1].pair_value_records

#2
"all value records should report the same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[1]
      PairPosFormat2.class1_records[0]

#3
"all value records should report the same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[1]
      PairPosFormat2.class1_records[1]

#4
"all value records should report the same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[1]
      PairPosFormat2.class1_records[2]

#5
"all value records should report the same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[1]
      PairPosFormat2.class1_records[3]

#6
"all value records should report the same format"
in: Gpos.lookup_list
  LookupList.lookups[1]
    Lookup.subtables[1]
      PairPosFormat2.class1_records[4]

) }
```

@cmyr says:

> what seems to be happening here is that we are somehow not correctly keeping track of the value formats of value records somewhere

cmyr:
ValueRecords have an explicit format, which is generally stored once in their parent table. In the case of records with device or variation tables, we need to know the format in order to write them correctly, since it's possible for these fields to be legitimately null, and we cannot distinguish between a record which where a field is None because it is not included versus when it is intended to be null.

Previously, we would use the format when reading the table but would then discard it. This led to a bug where we would not know the correct format when we went to write the table back.

With this patch, we hold onto the format after reading, and pass it to write-fonts if we convert to write-fonts types. This ensures that we roundtrip tables correctly in this case.

closes #858

### DIFF
--- a/read-fonts/src/tables/value_record.rs
+++ b/read-fonts/src/tables/value_record.rs
@@ -28,7 +28,7 @@ impl ValueFormat {
 /// NOTE: we create these manually, since parsing is weird and depends on the
 /// associated valueformat. That said, this isn't a great representation?
 /// we could definitely do something much more in the zero-copy mode..
-#[derive(Clone, Default, PartialEq, Eq)]
+#[derive(Clone, Default, Eq)]
 pub struct ValueRecord {
     pub x_placement: Option<BigEndian<i16>>,
     pub y_placement: Option<BigEndian<i16>>,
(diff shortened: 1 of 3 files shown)


anthrotype (approved):

read-fonts/src/tables/value_record.rs
```diff
@@ -38,16 +38,31 @@ pub struct ValueRecord {
...
+    #[doc(hidden)]
+    // exposed so that we can preserve format when we round-trip a value record
+    pub format: ValueFormat,
 }
 
-impl ValueRecord {
-    pub fn read_old(data: &[u8], format: ValueFormat) -> Result<Self, ReadError> {
-        let data = FontData::new(data);
-        Self::read(data, format)
+// we ignore the format for the purpose of equality testing, it's redundant
```
rsheeter, line 46: if it's redundant why do we need to store it? - seems slightly contradictory. Perhaps further explanation of how this makes sense?

rsheeter (approved):

read-fonts/src/tables/value_record.rs
cmyr: sorry, it is redundant for the purposes of comparing two value records. In that case (testing, e.g.) I don't think we care.

PR #862, opened 2024-03-27
[skrifa] reduce pedantry

dfrg:
Propagate the `is_pedantic` flag to a few places and reduce the strictness in various parts of the hinter based on that flag.

With these changes we match FT hinting for all of Google Fonts and Noto.


cmyr (approved):
Although I oppose this on a philosophical level I guess we can all make concessions.

dfrg:
I really tried to avoid this but the data wasn’t on our side :/

PR #863, opened 2024-03-28
[read-fonts] capture full range of image data in cblc

dfrg:
These can be sparse which requires reading offset and offset +1 to compute the full range.

ref https://skia-review.googlesource.com/c/skia/+/809056


read-fonts/src/tables/cbdt.rs
```diff
@@ -56,4 +56,23 @@ mod tests {
...
+        let size = &cblc.bitmap_sizes()[0];
+        // this font has a sparse set with gid 1 missing
+        for gid in 0..=3 {
+            let location = size
+                .location(cblc.offset_data(), GlyphId::new(gid))
+                .unwrap();
+            if gid == 1 {
+                assert!(cbdt.data(&location).is_err());
+            } else {
+                cbdt.data(&location).unwrap();
```
rsheeter, line 74: nit: if this ever fails you won't know which gid. Suggest unwrap_or_else panic with a helpful message.

rsheeter (approved):

read-fonts/src/tables/cbdt.rs
dfrg: Good call. Changed both to asserts with messages describing the failure.

drott:
👍  thanks for the fix! I guess it's time for me to roll some versions, into Chromium as well, with bytemuck and all.

PR #864, opened 2024-03-30
[skrifa] tthint: rounding phantom points

dfrg:
This fixes a weird edge case where the user requests a strong hinting mode (FT_LOAD_TARGET_MONO) but no instructions are present in a simple glyph. FreeType still rounds the phantom points in this case but we weren't doing so causing some glyphs to be offset a tiny bit in the x direction.

With this fix we now have 100% matching fauntlet runs for all of Google Fonts and Noto with all four hinting modes.

### DIFF
--- a/skrifa/src/outline/glyf/mod.rs
+++ b/skrifa/src/outline/glyf/mod.rs
@@ -49,6 +49,7 @@ pub struct Outlines<'a> {
     max_storage: u16,
     glyph_count: u16,
     units_per_em: u16,
+    os2_vmetrics: [i16; 2],
     has_var_lsb: bool,
 }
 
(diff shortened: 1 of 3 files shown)


skrifa/src/outline/glyf/mod.rs
```diff
@@ -741,26 +758,26 @@ where
         if have_deltas {
             self.component_delta_count = delta_base;
         }
-        if self.is_hinted {
+        if let (Some(hinter), true) = (self.hinter.as_ref(), self.is_hinted) {
```
cmyr, line 761: let chains will be nice for this stuff, e.g. `if let Some(hinter) = self.hinter && self.is_hinted { .. }`

cmyr (approved):
sorry to have missed this, lgtm :)

skrifa/src/outline/glyf/mod.rs
dfrg: yeah, I occasionally attempt it and am still somehow surprised when it doesn’t work

drott:
> With this fix we now have 100% matching fauntlet runs for all of Google Fonts and Noto with all four hinting modes.

💯 🥇

PR #868, opened 2024-04-02
Unlock bytemuck version

dfrg:
Switch bytemuck version from `=1.31.1` to `1.31.1`. Also moves bytemuck to a workspace dependency to make it easier to update the version in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ members = [
 ]
 
 [workspace.dependencies]
+# note: bytemuck version must be available in all deployment environments
+bytemuck = "1.13.1"
 # dev dependencies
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -13,6 +13,8 @@ members = [
 ]
 
 [workspace.dependencies]
+# note: bytemuck version must be available in all deployment environments
```
rsheeter, line 16: Let's go one step more detailed and say it should be the floor of google3 and Chrome?

rsheeter (approved):

Cargo.toml
dfrg: sg, done

PR #880, opened 2024-04-17
Support hb-draw style rounding

### LINKED ISSUE #874: Make ToPathStyle::HarfBuzz match hb-draw rounding
Rounding doesn't currently match. This complicates swapping in Rust for icon svg productions. See https://github.com/googlefonts/sleipnir/pull/15 for an example.

If, as per IM, exactly matching hb-draw is feasible with modest (say < 1 week) effort it would be worthwhile.

@dfrg suggested we would need to duplicate the code in `glyf/deltas.rs` for FP and then handle the branching in the scaler. read-fonts would have to compute an f32 scalar.

rsheeter:
Taken from @dfrg work on https://github.com/googlefonts/sleipnir/pull/15, specifically https://github.com/googlefonts/sleipnir/blob/opsz48/src/glyf.rs.

Fixes #874. 

Notes:

* Makes FreeTypeScalar vs HarfBuzzScalar explicit in the code
   * Initially attempted to just make the scalar generic over coord and delta but this proved painful
   * Generic coord and delta types are still used at some levels
* Everything in sight being public makes this scarier than it would be with narrow public interfaces
* Expected FreeType draw_transformed_component values obtained by running the test on main

TODO: review load simple/composite for sharing opportunities, currently each scaler has it's own implementation
   * Might wish to leave this to a follow-on


font-types/src/fixed.rs
```diff
@@ -7,7 +7,7 @@ macro_rules! fixed_impl {
     ($name:ident, $bits:literal, $fract_bits:literal, $ty:ty) => {
         #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
         #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-        #[cfg_attr(feature = "bytemuck", derive(bytemuck::AnyBitPattern, bytemuck::NoUninit))]
+        #[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
```
dfrg, line 10: Is there any particular reason for changing these traits?

font-types/src/point.rs
```diff
@@ -29,6 +29,8 @@ pub struct Point<T> {
 /// it. This condition only exists because the bytemuck derive
 /// macro cannot guarantee the first three conditions in a type
 /// with generic parameters.
+// #[cfg(feature = "bytemuck")]
+// unsafe impl<T> bytemuck::NoUninit for Point<T> where T: bytemuck::NoUninit + Pod {}
```
dfrg, line 33: Looks like some leftovers.

skrifa/src/outline/error.rs
```diff
@@ -31,6 +31,10 @@ pub enum DrawError {
     ToPath(ToPathError),
     /// Error occured when reading font data.
     Read(ReadError),
+    /// HarfBuzz style drawing with hints is not supported
+    ///
+    /// Error rather than silently returning unhinted per f2f discussion.
```
dfrg, line 36: Context seems useful but probably shouldn't be part of the doc comment.

skrifa/src/outline/glyf/mod.rs
```diff
@@ -276,9 +248,116 @@ impl<'a> Outlines<'a> {
...
+        let glyph = match &glyph {
+            Some(glyph) => glyph,
+            // This is a valid empty glyph
+            None => return Ok(()),
+        };
+        let bounds = [glyph.x_min(), glyph.x_max(), glyph.y_min(), glyph.y_max()];
+        let outlines = self.outlines();
+        let coords: &[F2Dot14] = self.coords();
+        let lsb = outlines.lsb(glyph_id, coords);
+        let advance = outlines.advance_width(glyph_id, coords);
```
dfrg, line 288: HB diff: these have variations applied... HB always uses base values and computes phantom point variations with gvar

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+        self.contour_count += contour_count;
+        // Append phantom points to the outline.
+        for (i, phantom) in self.phantom.iter().enumerate() {
+            points[phantom_start + i] = *phantom;
+            flags[phantom_start + i] = Default::default();
+        }
+        // Acquire deltas
+        if self.outlines.gvar.is_some() && !self.coords.is_empty() {
+            let gvar = self.outlines.gvar.as_ref().unwrap();
+            let glyph = deltas::SimpleGlyph {
```
dfrg, line 1015: HB diff: this requires different a full floating point delta computation as well as application of deltas to points after processing each tuple

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+                }
+            }
+        }
+        // Apply scaling
+        if self.is_scaled {
+            let scale = self.scale.to_f32();
+            for point in points.iter_mut() {
+                *point = point.map(|c| c * scale);
+            }
+        }
```
dfrg, line 1052: I didn't really look into how HB applies scale because it wasn't necessary for icons. This probably needs to be slotted in at the correct place to maintain compatibility.

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+        // The base indices of the points for the current glyph.
+        let point_base = self.point_count;
+        // Compute the per component deltas. Since composites can be nested, we
+        // use a stack and keep track of the base.
+        let mut have_deltas = false;
+        let delta_base = self.component_delta_count;
+        if self.outlines.gvar.is_some() && !self.coords.is_empty() {
+            let gvar = self.outlines.gvar.as_ref().unwrap();
+            let count = glyph.components().count() + PHANTOM_POINT_COUNT;
+            let deltas = self
```
dfrg, line 1080: HB diff: same as simple-- floating point delta computation

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+                }
+                have_deltas = true;
+            }
+            self.component_delta_count += count;
+        }
+        if self.is_scaled {
+            for point in self.phantom.iter_mut() {
+                *point *= scale;
+            }
+        }
```
dfrg, line 1099: Same as simple. I'm not sure where this goes

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+                        && component.flags
+                            & (CompositeGlyphFlags::SCALED_COMPONENT_OFFSET
+                                | CompositeGlyphFlags::UNSCALED_COMPONENT_OFFSET)
+                            == CompositeGlyphFlags::SCALED_COMPONENT_OFFSET
+                    {
+                        // Scale x by the magnitude of the x-basis, y by the y-basis
+                        // FreeType implements hypot, we can just use the provided implementation
+                        x *= hypot(transform[0], transform[2]);
+                        y *= hypot(transform[1], transform[3]);
+                    }
```
dfrg, line 1153: HB diff: this should use the code scaled component offset code from sleipnir

skrifa/src/outline/glyf/mod.rs
```diff
@@ -827,25 +937,267 @@ impl<'a> Scaler<'a> {
...
+                        .memory
+                        .points
+                        .get(point_base + base_offset)
+                        .ok_or(DrawError::InvalidAnchorPoint(glyph_id, base))?;
+                    let component_point = self
+                        .memory
+                        .points
+                        .get(start_point + component_offset)
+                        .ok_or(DrawError::InvalidAnchorPoint(glyph_id, component))?;
+                    *base_point - *component_point
```
dfrg, line 1174: .. but this was missing from the sleipnir implant and should be okay

dfrg (approved):
As discussed f2f, we should make sure we actually offer HB compatibility beyond 2 decimal places if we're going to promise it. Otherwise, code generally LGTM.

font-types/src/fixed.rs
rsheeter: Hm, I don't think I need that anymore, vestiges of trying to make a more generic version.

font-types/src/point.rs
rsheeter: Fixed, ty.

font-types/src/fixed.rs
rsheeter: Fixed, ty

skrifa/src/outline/error.rs
rsheeter: Now with significantly fewer /'s

skrifa/src/outline/glyf/mod.rs
rsheeter: Deferred to https://github.com/googlefonts/fontc/issues/784

skrifa/src/outline/glyf/mod.rs
rsheeter: Deferred to https://github.com/googlefonts/fontc/issues/784

skrifa/src/outline/glyf/mod.rs
rsheeter: Deferred to https://github.com/googlefonts/fontc/issues/784

skrifa/src/outline/glyf/mod.rs
rsheeter: Deferred to https://github.com/googlefonts/fontc/issues/784

PR #881, opened 2024-04-19
setup klippa crate, try to make most basic subset for hmtx

qxliu76:


klippa/src/hmtx.rs
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            let mut new_metrics = Vec::<LongMetric>::with_capacity(num_long_metrics);
+            let mut new_side_bearings = Vec::<i16>::new();
+            for gid in gids {
+                let glyph_id = gid.to_u16() as usize;
+                // skip invalid gids
+                if glyph_id >= num_glyphs {
+                    break;
+                }
```
cmyr, line 38: what do we know about the `gids` array? we're doing a lot of checking for invalid values, but would it be possible to validate ahead of time that all the gids in the plan exist in the font?

klippa/src/hmtx.rs
```diff
@@ -0,0 +1,79 @@
...
+                    });
+                } else {
+                    new_side_bearings.push(side_bearing);
+                }
+            }
+
+            // update num_of_hmetrics in the plan so we can update hhea table
+            plan.num_h_metrics = new_metrics.len() as u16;
+            // temporary: update num_output_glyphs in the plan for maxp table, will do this during planning phase
+            plan.num_output_glyphs = new_metrics.len() as u16 + new_side_bearings.len() as u16;
```
cmyr, line 56: Do we need to have the plan be mutable? Maybe we could just have a final pass at the end that sets values in tables that depend on other tables, but looking at the tables directly?

More concretely, my concern is that the mutable plan means we can't subset tables in parallel, which we can do trivially otherwise. This probably doesn't matter right now, but it might be worth thinking about..

klippa/src/hmtx.rs
```diff
@@ -0,0 +1,79 @@
...
+fn get_gid_advance(metrics: &[LongMetric], gid: usize) -> u16 {
+    if gid >= metrics.len() {
+        metrics.last().unwrap().advance
+    } else {
+        metrics[gid].advance
+    }
+}
+
+fn get_gid_side_bearing(metrics: &[LongMetric], side_bearings: &[i16], gid: usize) -> i16 {
+    if gid < metrics.len() {
```
cmyr, line 74: use `metrics.get(gid)`, it performs the bounds check for you and returns an `Option`.

klippa/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+use std::collections::BTreeSet;
+use thiserror::Error;
+use write_fonts::{
+    from_obj::FromTableRef, tables::hhea::Hhea, tables::hmtx::Hmtx, tables::maxp::Maxp, FontBuilder,
+};
+
+#[derive(Error, Debug)]
+pub enum InvalidInputError {
+    #[error("Invalid input gid {0}")]
+    InvalidGidError(String),
```
cmyr, line 19: no need to convert the gid to a string, just keep it as a u16 or a `GlyphId` or whatever it was, you can still print it.


Also we can just call this `InvalidGid`, the 'Error' bit is redundant I think.

klippa/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    #[error("Invalid gid range {start}-{end}")]
+    InvalidGidRangeError { start: u16, end: u16 },
+}
+
+fn main() {
+    let args = match flags::Args::from_env() {
+        Ok(args) => args,
+        Err(_e) => {
```
cmyr, line 28: in both of these branches we're discarding the error, which means we'll just exit silently. We probably want to at least print the error beforehand?

klippa/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "klippa"
+version = "0.1.0"
+edition = "2021"
```
cmyr, line 4: We should include the same fields we include in the other crates in this repo, take a look at those `Cargo.toml` as an example.

klippa/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+font-types = { path = "../font-types" }
+read-fonts = { path = "../read-fonts" }
+thiserror = "1.0.58"
+write-fonts = { path = "../write-fonts" }
```
cmyr, line 12: Add the `read` feature here to get the `write_fonts::read` module I mention above:


```suggestion
write-fonts = { path = "../write-fonts", features = ["read"] }
```

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "write-fonts",
     "otexplorer",
     "skrifa",
-    "fauntlet",
+    "fauntlet", "klippa",
```
cmyr, line 12: I'd put this on a newline just to match the formatting in the rest of the file.

cmyr (commented):
Okay I have a bunch of notes inline; some fixups for `Cargo.toml` and then the rest of it is just me giving examples of idiomatic alternatives for some patterns.


In any case this looks like a good start!

klippa/src/hmtx.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::{Plan, Subset, SubsetError};
+use read_fonts::TopLevelTable;
+use std::cmp;
+use write_fonts::tables::{hmtx::Hmtx, hmtx::LongMetric};
+
+impl Subset for Hmtx {
+    fn subset(&mut self, plan: &mut Plan) -> Result<bool, SubsetError> {
+        let gids = &plan.glyph_ids;
+        if gids.is_empty() {
+            Err(SubsetError::SubsetTableError(Hmtx::TAG))
```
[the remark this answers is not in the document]
qxliu76: oh yeah, this is purely because VS code has been pushing me to complete the if{} else{} pattern

klippa/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+xflags = "0.3.2"
```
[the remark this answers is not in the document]
qxliu76: I'll try to make this change

klippa/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "klippa"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+font-types = { path = "../font-types" }
+read-fonts = { path = "../read-fonts" }
```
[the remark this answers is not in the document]
qxliu76: Thanks! Seems that `write_fonts::types` works but `read` doesn't, I got this error:
```
error[E0432]: unresolved import `write_fonts::read`
 --> klippa/src/hmtx.rs:4:18
  |
4 | use write_fonts::read::TopLevelTable;
  |                  ^^^^ could not find `read` in `write_fonts`

```

klippa/src/hmtx.rs
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            let num_glyphs = self.h_metrics.len() + self.left_side_bearings.len();
+            let mut num_long_metrics = cmp::min(num_glyphs, 0xFFFF_usize);
+            let last_advance = get_gid_advance(&self.h_metrics, num_glyphs - 1);
+            for gid in gids.iter().rev() {
+                let glyph_id = gid.to_u16() as usize;
+                // skip invalid gids
+                if glyph_id >= num_glyphs {
```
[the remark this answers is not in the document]
qxliu76: I think we might need to continue in this loop, which is in reverse order, because we skip invalid gids and count the number of glyphs with identical advance for valid gids.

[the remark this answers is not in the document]
garretrieger: Oh sorry I didn't notice this was rerversed order, in which case yeah no change is needed.

klippa/src/hmtx.rs
qxliu76: check and filter out invalid gids and then do old gid->new gid mapping, all these should happen during planning phase. So this is just temporary code for setting up the crate and will be revisited later.

klippa/src/main.rs
qxliu76: oh I have a question: this error is emitted when parsing a `&str` to `u16`, so I assume at this time we only have `&str` available?  Can I just change `String` to `&str`?

klippa/Cargo.toml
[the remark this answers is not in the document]
qxliu76: thank you!

klippa/src/main.rs
cmyr: oh I'm sorry, I misread. Your initial version was good, having a lifetime in the error will complicate things way too much. :)

LaurenzV:
(Sorry, not sure if you want/take comments from "outsiders". 😅)

Just wanted to say it's great to see that subsetting is being worked on now! I just have a small question, is this crate intended to really just be a binary crate, or are there also plans for it to be able to be used as a library? If the latter, maybe it makes sense to put `clap` behind a `cli` feature gate (and I've heard that `thiserror` is also a pretty heavy dependecy since it uses proc macros and there are more lightweight solutions out there like `quickerror`, but I don't know the details here)? 

Because `clap` is a pretty heavy dependency which would not be required when using it as a library... But I'm not sure what the plans for this crate are. But I think there are many people interested in having a pure-Rust subsetting solution available as a library crate, or at least I am, so it would be great if this is within the scope of the project!

cmyr:
Thanks for the thoughts @LaurenzV, I think you raise valid concerns. I think it makes sense for us to punt on these issues for now, though; it will be easy in the future to make sure clap is not compiled when we're just used as a library, and if `thiserror` ends up being bloaty we can always replace it fairly easily in the future, and using it now will speed us up in the immediate term. :)

LaurenzV:
Makes sense, thanks!

rsheeter (approved):
LG to merge, we can always iterate in followons

PR #889, opened 2024-05-01
Add the beginnings of a port of harfbuzz hb_set_t as a new crate, int-set.

garretrieger:
This implements a fast bit set, that uses pages internally to minimize memory usage. Additionally, the set is invertible. This allows it to express patterns like all integers except x, y, z. Port of [hb-set](https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-set.h). Currently only very basic functionality is implemented (add, remove, contains, length, invert).


int-set/src/bitpage.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+fn iter_bit_indices(val: Element) -> impl Iterator<Item = u32> {
+    let mut idx = 0;
+
+    std::iter::from_fn(move || {
+        if idx >= ELEM_BITS {
+            return None;
+        }
+        let mask = 1u64.checked_shl(idx).unwrap_or(0) - 1;
```
cmyr, line 140: since we know that `idx < ELEM_BITS` we I don't think we need the `checked_shl` (or the `unwrap_or`)?

int-set/src/bitpage.rs
```diff
@@ -0,0 +1,353 @@
...
+
+#[cfg(test)]
+mod test {
+    use std::collections::HashSet;
+
+    use super::*;
+
+    impl BitPage {
+        /// Create a new page with all bits set.
+        pub fn new_ones() -> Self {
```
cmyr, line 186: is this down here because you were getting a dead code warning?

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
+//! A fast & efficient unsigned integer bitset that keeps it's members ordered.
```
cmyr, line 1:
```suggestion
//! A fast & efficient ordered set for unsigned integers
```

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
...
+#[derive(Clone, Debug, Default)]
+pub(super) struct BitSet<T> {
+    // TODO(garretrieger): consider a "small array" type instead of Vec.
+    pages: Vec<BitPage>,
+    page_map: Vec<PageInfo>,
+    len: Cell<usize>, // TODO(garretrieger): use an option instead of a sentinel.
+    phantom: std::marker::PhantomData<T>,
+}
+
+impl<T: Into<u32> + Copy + Default> BitSet<T> {
```
cmyr, line 22: Do we need this `Default` trait bound?

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// Return the major value (top 23 bits) of the page associated with value.
+    fn get_major_value(&self, value: u32) -> u32 {
+        return value >> PAGE_BITS_LOG_2;
+    }
+
+    fn major_start(&self, major: u32) -> u32 {
+        return major << PAGE_BITS_LOG_2;
+    }
+
+    /// Return a reference to the that 'value' resides in.
```
cmyr, line 117:
```suggestion
    /// Return a reference to the page that 'value' resides in.
```

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
...
+            .binary_search_by(|probe| probe.major_value.cmp(&major_value))
+            .ok()
+            .and_then(|info_idx| {
+                let real_idx = self.page_map[info_idx].index as usize;
+                self.pages.get(real_idx)
+            })
+    }
+
+    /// Return a mutable reference to the that 'value' resides in. Insert a new
+    /// page if it doesn't exist.
```
cmyr, line 130:
```suggestion
    /// Return a mutable reference to the page that 'value' resides in. 
    ///
    /// Inserts a new page if it doesn't exist.
```

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
...
+    fn page_for_mut(&mut self, value: u32) -> &mut BitPage {
+        let major_value = self.get_major_value(value);
+        self.page_for_major_mut(major_value)
+    }
+
+    // Return a mutable reference to the page with major value equal to major_value.
+    // Inserts a new page if it doesn't exist.
```
cmyr, line 137: might as well make these comments into doc comments, even if the method is private; it means that at in our own use we will see these comments on hover in an IDE, for example


```suggestion
    /// Return a mutable reference to the page with the provided major value.
    ///
    /// Inserts a new page if it doesn't exist.
```

int-set/src/bitset.rs
```diff
@@ -0,0 +1,402 @@
...
+mod test {
+    use super::*;
+    use std::collections::HashSet;
+
+    impl<T> BitSet<T> {
+        /// Return true if there are no members in this set.
+        pub fn is_empty(&self) -> bool {
+            self.len() == 0
+        }
+    }
```
cmyr, line 213: for less churn I would just keep this method in the main impl, but you can annotate it with either `#[allow(dead_code)]` or `#[cfg(test)]` there.

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Records a set of integers which are members of the set.
+    Standard(BitSet<T>),
+
+    /// Records the set of integers which are not members of the set.
+    Inverted(BitSet<T>),
+}
+
+impl<T: Into<u32> + Copy + Default> IntSet<T> {
+    /// Create a new empty set.
+    pub fn default() -> IntSet<T> {
```
cmyr, line 19: we should implement `default` through the `Default` trait, and here we should just have a `new` constructor.

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Clone, Debug)]
+pub enum IntSet<T> {
+    /// Records a set of integers which are members of the set.
+    Standard(BitSet<T>),
+
+    /// Records the set of integers which are not members of the set.
+    Inverted(BitSet<T>),
+}
+
+impl<T: Into<u32> + Copy + Default> IntSet<T> {
```
cmyr, line 17: here too we can probably get rid of the `Default` bound?

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
+mod bitpage;
```
cmyr, line 1: we should have some `//!` top level comment here, this is what would show up on the main page when rendering the docs.

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+mod bitset;
+
+use bitset::BitSet;
+use std::ops::RangeInclusive;
+
+/// A fast & efficient unsigned integer (u32) bit set which is invertible.
```
cmyr, line 7:
```suggestion
/// A fast & efficient invertible ordered set for small (up to 32-bit) unsigned integer types.
```

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    Standard(BitSet<T>),
+
+    /// Records the set of integers which are not members of the set.
+    Inverted(BitSet<T>),
+}
+
+impl<T: Into<u32> + Copy + Default> IntSet<T> {
+    /// Create a new empty set.
+    pub fn default() -> IntSet<T> {
+        IntSet::Standard(BitSet::<T>::default())
```
cmyr, line 20: the compiler can infer the generics, so you can just write `BitSet::default()`. (you could even write `Default::default()` and it will figure out what the type is)

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Create a new set which contains all integers.
+    pub fn all() -> IntSet<T> {
+        IntSet::Inverted(BitSet::<T>::default())
+    }
+
+    /// Return the inverted version of this set.
+    pub fn inverted(self) -> IntSet<T> {
+        match self {
+            IntSet::<T>::Standard(s) => IntSet::<T>::Inverted(s),
```
cmyr, line 31: here also we don't need to write out the `<T>::` stuff

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Return a new version of this set with all members removed.
+    pub fn clear(mut self) -> IntSet<T> {
+        self.clear_internal_set();
+        match self {
+            IntSet::<T>::Standard(s) => IntSet::<T>::Standard(s),
+            IntSet::<T>::Inverted(s) => IntSet::<T>::Standard(s),
+        }
+    }
+
+    /// Add val as a member of this set.
```
cmyr, line 45:
```suggestion
    /// Adds a value to the set.
    ///
    /// Returns `true` if the value was newly inserted.
```

(I'm just copying/adapting the docs from `HashSet::insert`)

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Return the inverted version of this set.
+    pub fn inverted(self) -> IntSet<T> {
+        match self {
+            IntSet::<T>::Standard(s) => IntSet::<T>::Inverted(s),
+            IntSet::<T>::Inverted(s) => IntSet::<T>::Standard(s),
+        }
+    }
+
+    /// Return a new version of this set with all members removed.
```
cmyr, line 36: from `HashSet::clear`:

```suggestion
    /// Clears the set, removing all values.
```

(By convention clear is a mutating method on self, and does not return anything.)

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    /// Add all values in range as members of this set.
+    pub fn insert_range(&mut self, range: RangeInclusive<T>) {
+        match self {
+            IntSet::<T>::Standard(s) => s.insert_range(range),
+            IntSet::<T>::Inverted(_) => todo!("implement bitset::remove_range and call here."),
+        }
+    }
+
+    /// Remove val from this set.
```
cmyr, line 61:
```suggestion
    /// Removes a value from the set. Returns whether the value was present in the set.
```

(`HashSet::remove`)

int-set/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub fn remove(&mut self, val: T) -> bool {
+        match self {
+            IntSet::<T>::Standard(s) => s.remove(val),
+            IntSet::<T>::Inverted(s) => s.insert(val),
+        }
+    }
+
+    /// Returns true if val is a member of this set.
```
cmyr, line 69:
```suggestion
    /// Returns `true` if the set contains a value.
```

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "write-fonts",
     "otexplorer",
     "skrifa",
-    "fauntlet",
+    "fauntlet", "int-set",
```
cmyr, line 12: might as well add a newline to match everything else?

cmyr (approved):
Looks good! A few little cleanups and doc suggestions, but overall I'm looking forward to using this in a few places :)

int-set/src/bitpage.rs
garretrieger: Yes, we currently only use it in tests. We may eventually need it for code in the public api, at that time we can move it back up.

int-set/src/lib.rs
garretrieger: Because to properly clear the set we have to change the enum type I believe this requires returning a new IntSet (similar to the inverted() method).

nicoburns: You should be able to do:

```rust
pub fn clear(&mut self) {
    self.clear_internal_set();
    *self.0 = match self.0 {
        Membership::Inclusive(s) => IntSet(Membership::Inclusive(s)),
        Membership::Exclusive(s) => IntSet(Membership::Inclusive(s)),
    };
}
```

if you want.

garretrieger: That doesn't work unfortunately, I've played around with it a bit and the closest I could get to making it work is this:

pub fn clear(&mut self) {
        self.clear_internal_set();
        self.0 = match &mut self.0 {
            Membership::Inclusive(s) => Membership::Inclusive(*s),
            Membership::Exclusive(s) => Membership::Inclusive(*s),
        };
    }
    
 But that still has problems:
 
 error[E0507]: cannot move out of `*s` which is behind a mutable reference

cmyr: Ah okay yea, that's a funny little hiccup.

There are two basic options:

- if we want to forget about reusing the storage we can just replace the bitset, like:
   ```rust
    *s = Membership::Inclusive(BitSet::empty());
   ```
- and if we want to preserve the existing set, we can do a little dance to get around borrowck:
```rust
/// Return a new version of this set with all members removed.
    pub fn clear(&mut self) {
        let reuse_storage = match &mut self.0 {
            // if we're inclusive, we just clear the storage
            Membership::Inclusive(s) => {
                s.clear();
                return;
            }
            // otherwise take the existing storage to reuse in a new
            // inclusive set:
            Membership::Exclusive(s) => std::mem::replace(s, BitSet::empty()),
        };
        // reuse the now empty storage and mark us as inclusive
        self.0 = Membership::Inclusive(reuse_storage);
    }
```

garretrieger: I was hoping to keep the allocated memory in the set around, so looks like I'll go with the second version.

int-set/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+        IntSet(Membership::Inclusive(BitSet::empty()))
+    }
+
+    /// Create a new set which contains all integers (exclusive).
+    pub fn all() -> IntSet<T> {
+        IntSet(Membership::Exclusive(BitSet::empty()))
+    }
+
+    /// Return the inverted version of this set.
+    pub fn invert(&mut self) {
```
cmyr, line 53: these methods (and any others that don't have the `T` param? can probably go in the impl block that has no type constraints?

int-set/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+        }
+    }
+
+    /// Return true if there are no members in this set.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+}
+
+impl<T> std::hash::Hash for IntSet<T> {
```
cmyr, line 136: any reason not to derive Hash? it should generate basically the same code..

int-set/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+                s.hash(state);
+                1u32.hash(state);
+            }
+        }
+    }
+}
+
+impl<T> std::cmp::Eq for IntSet<T> {}
+
+impl<T> std::cmp::PartialEq for IntSet<T> {
```
cmyr, line 153: this also looks derivable

cmyr (approved):
okay two last little notes, nothing major. looks good!

int-set/src/lib.rs
garretrieger: BitSet::empty() requires that T hash Into and Copy

garretrieger: Nm, got it working.

PR #895, opened 2024-05-06
Add extend() and iter() to IntSet public API.

garretrieger:


rsheeter (approved):

int-set/src/bitset.rs
```diff
@@ -124,51 +144,88 @@ impl<T> BitSet<T> {
...
+                page_index
+            }
+        }
+    }
+
+    /// Return a reference to the page that 'value' resides in.
+    fn page_for(&self, value: u32) -> Option<&BitPage> {
+        let major_value = self.get_major_value(value);
+        let pages_index = self.page_index_for_major(major_value)?;
+        Some(&self.pages[pages_index])
```
cmyr, line 187: I'd just write this as `self.pages.get(page_index)`; in general I think it's a good habit to avoid direct indexing, which often generates worse code (specifically the compiler has to generate a panic path for when indexing fails) 🤷

int-set/src/lib.rs
```diff
@@ -86,6 +102,32 @@ impl<T> IntSet<T> {
...
+        }
+    }
+
+    /// If this is an inclusive membership set then returns an iterator over the members, otherwise returns None.
+    pub fn inclusive_iter(&self) -> Option<impl Iterator<Item = u32> + '_> {
+        match &self.0 {
+            Membership::Inclusive(s) => Some(s.iter()),
+            Membership::Exclusive(_) => None,
+        }
+    }
```
cmyr, line 121: not worth changing, but personally I don't think I would add this to public API until I had really felt like I needed it somewhere?

int-set/src/bitset.rs
```diff
@@ -43,21 +43,34 @@ impl<T: Into<u32> + Copy> BitSet<T> {
...
-            self.page_for_major_mut(major)
+            self.ensure_page_for_major_mut(major)
                 .insert_range(page_start, page_end);
         }
         self.mark_dirty();
     }
 
+    /// An alternate version of extend() which is optimized for inserting an unsorted
+    /// iterator of values.
+    pub fn extend_unsorted<U: IntoIterator<Item = T>>(&mut self, iter: U) {
```
cmyr, line 54: this is slightly confusing, since we wouldn't want `extend` to ever assume that input was sorted.


My gut then would be to invert this, and have `extend` be naive, and have `extend_sorted` be the special additional function.


At a higher level, I would really prefer if we could have a single implementation that just switched behaviour after seeing the first out-of-order item?

int-set/src/bitset.rs
```diff
@@ -104,6 +117,13 @@ impl<T> BitSet<T> {
         self.len.get()
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = u32> + '_ {
```
cmyr, line 120: something I'm noticing now that I'm doing code review is that outside of the broader context of the file (e.g. just seeing this change in the diff) it's not clear that this is defined on a non-public type, which makes this look like public API. Maybe that's a reason to prefer `pub(crate)` etc even if the visibility of the containing item is non-`pub`?

cmyr (commented):
A few little notes inline, overall looks good!

int-set/src/bitset.rs
garretrieger: Sure I'll give that a try and see if it effects the benchmarks in either direction.

int-set/src/bitset.rs
garretrieger: The current extend() implementation works fine for unsorted or partial sorted inputs, but will be slightly slower (~7%) then this version. When input is sorted or partially sorted however there will be a massive speedup (84% in the benchmark), so I think the right tradeoff is to have that as the default and leave extend_unsorted() as a niche optimization. The way extend() is currently written there's no need to bail out to an alternate implementation when encountering unsorted data. This allows is to provide speedups for partially sorted inputs as well.

int-set/src/bitset.rs
garretrieger: https://github.com/googlefonts/fontations/pull/896

garretrieger: Bench times were neutral

int-set/src/bitset.rs
garretrieger: Sounds good

PR #896, opened 2024-05-08
Avoid using pages[...] in BitSet implementation.

garretrieger:


int-set/src/lib.rs
```diff
@@ -80,7 +82,21 @@ impl<T: Into<u32> + Copy> IntSet<T> {
         }
     }
 
-    // TODO(garrretrieger): add remove_range.
+    // Removes all values in iter from the set.
+    pub fn remove_all<U: IntoIterator<Item = T>>(&mut self, iter: U) {
```
cmyr, line 86: Do you anticipate this being useful as public API? I would imagine us eventually wanting various set operations (union, difference etc) which would overlap with this in specific cases, but it's not clear (as in, I actually don't know) to me how often we would be working with different sorts of iterators?

int-set/src/bitset.rs
```diff
@@ -184,30 +235,35 @@ impl<T> BitSet<T> {
...
         self.ensure_page_for_major_mut(self.get_major_value(value))
     }
 
     // Return a mutable reference to the page with major value equal to major_value.
     // Inserts a new page if it doesn't exist.
-    fn ensure_page_for_major_mut(&mut self, major_value: u32) -> &mut BitPage {
+    fn ensure_page_for_major_mut(&mut self, major_value: u32) -> Option<&mut BitPage> {
         let page_index = self.ensure_page_index_for_major(major_value);
-        &mut self.pages[page_index]
+        self.pages.get_mut(page_index)
```
cmyr, line 266: I have led you astray in this case: since it is an invariant that the page is always present, we should just be returning `&mut BitPage` and we should handle the unwrapping instead of requiring the caller to do it.

int-set/src/lib.rs
```diff
@@ -40,6 +40,13 @@ impl<T: Into<u32> + Copy> Default for IntSet<T> {
 }
 
 impl<T: Into<u32> + Copy> IntSet<T> {
+    // TODO(garretrieger): add additional functionality that the harfbuzz version has:
+    // - Iteration in reverse.
```
cmyr, line 44: There is the trait in `std` called [`DoubleEndedIterator`](https://doc.rust-lang.org/1.78.0/std/iter/trait.DoubleEndedIterator.html); if you implement this, then you also get a `rev()` method, which reverses iteration order.

cmyr (approved):
looks good, a couple of notes inline :)

int-set/src/lib.rs
garretrieger: I believe so, it provides an optimized alternative to just calling .remove() multiple times so useful any time you're doing a sequence of removals that aren't listed in another IntSet. Also since we need to implement this to support extend() on exclusive sets we might as well make it available in the public api.

int-set/src/bitset.rs
garretrieger: no problem, I'll change it back. I prefer the panic because it will provide a clear signal that the invariants were violated instead of being silently ignored by the optional.

int-set/src/lib.rs
garretrieger: ah nice, yes that looks like the way to go.

garretrieger: I'll do this in a separate PR.

PR #899, opened 2024-05-12
[read-fonts] cmap4 iterator fix

### LINKED ISSUE #897: Skrifa charmap doesn't produce the same codepoints as hb_face_collect_unicodes
In testing migration from C++ to Rust codepoint enumeration in google3 I found some fonts where the result doesn't match:

```
FAIL ofl/gentiumbookbasic/GenBkBasR.ttf
FAIL ofl/gentiumbookbasic/GenBkBasB.ttf
FAIL ofl/gentiumbookbasic/GenBkBasBI.ttf
FAIL ofl/gentiumbookbasic/GenBkBasI.ttf
FAIL ofl/gentiumbasic/GenBasB.ttf
FAIL ofl/gentiumbasic/GenBasR.ttf
FAIL ofl/gentiumbasic/GenBasBI.ttf
FAIL ofl/gentiumbasic/GenBasI.ttf
4788/4796 match
```

Failure means the set of codepoints from `hb_face_collect_unicodes` differs from the set obtained via Skrifa's [`Charmap.mappings`](https://docs.rs/skrifa/latest/skrifa/charmap/struct.Charmap.html#method.mappings)

dfrg:
The cmap4 subtable iterator would give up when encountering an out of bounds range offset. The equivalent HB code continues to the next character in this case so now we do the same.

Should fix #897

### DIFF
--- a/read-fonts/src/tables/cmap.rs
+++ b/read-fonts/src/tables/cmap.rs
@@ -460,7 +460,7 @@ impl<'a> Iterator for NonDefaultUvsIter<'a> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{FontRef, GlyphId, TableProvider};
+    use crate::{test_helpers::BeBuffer, FontRef, GlyphId, TableProvider};
 
     #[test]
     fn map_codepoints() {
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
LGTM subject to addition of a test that proves it was broken and is now fixed

dfrg:
Added a test containing a subset of the failing sparse range from Gentium Basic. #903 contains just the test proving that it fails before the fix.

PR #900, opened 2024-05-12
extract_glyphs.py should ignore non-scalable fonts

### LINKED ISSUE #878: font-test-data/test_data/rebuild.sh fails
```
Extracting glyphs from "fontations/font-test-data/test_data/ttf/cbdt.ttf" to "fontations/font-test-data/test_data/extracted/cbdt-glyphs.txt"...
Traceback (most recent call last):
  File "fontations/font-test-data/test_data/extract_glyphs.py", line 143, in <module>
    glyphs.add_glyph(face, size, glyph_id)
  File "fontations/font-test-data/test_data/extract_glyphs.py", line 68, in add_glyph
    face.set_pixel_sizes(size, size)
  File "fontations/.venv/lib/python3.11/site-packages/freetype/__init__.py", line 1309, in set_pixel_sizes
    if error: raise FT_Exception(error)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
freetype.ft_errors.FT_Exception: FT_Exception:  (invalid pixel size)
```

dfrg:
Probably a good idea to bail on trying to extract outlines from fonts that don't have outlines

Fixes #878


font-test-data/test_data/extract_glyphs.py
```diff
@@ -115,6 +115,8 @@ def add_glyph(self, face: freetype.Face, size, glyph_id, coords=[], hinting="non
 
 try:
     face = freetype.Face(font_path)
+    # make sure we have scalable outlines
+    face.load_glyph(0, freetype.FT_LOAD_NO_BITMAPS)
 except:
```
anthrotype, line 120: `except: exit(0)` looks a bit too broad, catch-all to my taste, you can't even distinguish success from failure, but maybe you really don't care in your case how/why it failed.
Alternatively you could still handle all exceptions with exit(0) but at least print the error and traceback so at least one could see in the console that something may have gone wrong.
Up to you

anthrotype (approved):

font-test-data/test_data/extract_glyphs.py
dfrg: Yeah, I’m not particularly happy with this either but some of the fonts are just not complete enough for FreeType to read (but skrifa can, for testing purposes). I wanted to avoid spam during rebuild but I’ll add a brief note here that we’re skipping the font for outline testing so that we can catch this if it’s not intended. 

I might just remove this entirely once I get fauntlet hooked up in CI.

PR #901, opened 2024-05-14
[subset] Add --unicodes command line option

qxliu76:
Also replace BTreeSet with IntSet.
--unicodes option is needed for setting up test suite because we use harfbuzz tests. I only added most basic support for this option.

### DIFF
--- a/klippa/src/main.rs
+++ b/klippa/src/main.rs
@@ -5,8 +5,8 @@
 //!
 
 use clap::Parser;
+use int_set::IntSet;
 use klippa::{Plan, Subset};
-use std::collections::BTreeSet;
 use thiserror::Error;
 use write_fonts::read::{FontRef, TableProvider, TopLevelTable};
 use write_fonts::types::GlyphId;
@@ -21,6 +21,12 @@ pub enum InvalidInputError {
 
     #[error("Invalid gid range {start}-{end}")]
     InvalidGidRange { start: u16, end: u16 },
+
+    #[error("Invalid input unicode {0}")]
+    InvalidUnicode(String),
+
+    #[error("Invalid unicode range {start}-{end}")]
+    InvalidUnicodeRange { start: u32, end: u32 },
 }
 
 #[derive(Parser, Debug)]
@@ -32,7 +38,11 @@ struct Args {
 
     /// List of glyph ids
     #[arg(short, long)]
-    gids: String,
+    gids: Option<String>,
+
+    /// List of unicode codepoints
+    #[arg(short, long)]
+    unicodes: Option<String>,
 
     /// The output font file
     #[arg(short, long)]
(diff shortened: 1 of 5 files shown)


klippa/src/main.rs
```diff
@@ -100,6 +121,34 @@ fn populate_gids(gid_str: &str) -> Result<BTreeSet<GlyphId>, InvalidInputError>
...
+    if unicode_str.is_none() {
+        return Ok(result);
+    }
+    let re = regex::Regex::new(r"[<+->{},;&#\\xXuUnNiI\n\t\v\f\r]").unwrap();
+    let s = re.replace_all(unicode_str.as_deref().unwrap(), " ");
+    for cp in s.split_whitespace() {
+        if let Some((start, end)) = cp.split_once('-') {
+            let start: u32 = start
+                .parse::<u32>()
+                .map_err(|_| InvalidInputError::InvalidUnicode(start.to_owned().clone()))?;
```
cmyr, line 138: either one of `to_owned()` or `clone()` should be enough? if going from `&str` to `String` use `to_owned()`.

klippa/src/main.rs
```diff
@@ -79,9 +97,12 @@ fn main() {
...
 
-fn populate_gids(gid_str: &str) -> Result<BTreeSet<GlyphId>, InvalidInputError> {
-    let mut result = BTreeSet::new();
-    for gid in gid_str.split(',') {
+fn populate_gids(gid_str: &Option<String>) -> Result<IntSet<GlyphId>, InvalidInputError> {
+    let mut result = IntSet::empty();
+    if gid_str.is_none() {
+        return Ok(result);
+    }
+    for gid in gid_str.as_deref().unwrap().split(',') {
```
cmyr, line 105: try,

```rust
let Some(gid_str) = gid_str else {
    return Ok(result);
};
```


then you won't need to unwrap later.

klippa/src/main.rs
```diff
@@ -100,6 +121,34 @@ fn populate_gids(gid_str: &str) -> Result<BTreeSet<GlyphId>, InvalidInputError>
...
+                return Err(InvalidInputError::InvalidUnicodeRange { start, end });
+            }
+            result.extend(start..=end);
+        } else {
+            let unicode: u32 = cp
+                .parse::<u32>()
+                .map_err(|_| InvalidInputError::InvalidUnicode(cp.to_owned().clone()))?;
+            result.insert(unicode);
+        }
+    }
```
cmyr, line 152: do we care if the input is a valid `u32` but not a valid codepoint? Do we handle that later? it might make sense to also do that validation here..

cmyr (approved):
Looks good, a few notes inline!

klippa/src/main.rs
qxliu76: yeah we can probably skip here cause we will handle that later during planning phase

PR #902, opened 2024-05-15
In the public api of IntSet always use the generic type T.

garretrieger:
- This primarily effects the iterator which previously iterated over u32 values instead of the generic type T. To make this work I've introduced a new trait that set member types must implement. It provides functions to convert to and from u32's that are used internally.
- Additionally I've updated BitSet (the internal set) to only use u32's all conversions to and from T are done at the public API boundary in IntSet.
- Added tests which check that a custom type with limited domain (eg. doesn't map all u32 values) works correctly.


int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+
+/// Members of IntSet must implement this trait. Members of IntSet's must meet the following
+/// conditions to be used in an IntSet:
+///
+/// 1. Every possible unique value of T must be able map to and from a unique u32
+///    integer.
+///
+/// 2. The mapped u32 values must retain the same ordering as the values in T.
+///
+/// 3. all_values_iter() must iterate over all values in T in sorted order (ascending).
```
rsheeter, line 40: Nit: `.sorted()` that returns an iter would more obviously convey the nature of this fn

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+///
+/// from_mapped() will only ever be called with u32 values that were generated by a
+/// call to to_mapped() with a valid value of T. So it doesn't need to correctly handle values
+/// that are outside the domain of T.
+pub trait IntSetMember<T> {
+    fn to_mapped(&self) -> u32;
+    fn from_mapped(member: InDomain) -> T;
+    fn all_values_sorted_iter() -> impl Iterator<Item = u32>;
```
rsheeter, line 48: I think I might have expected Item to be T, add a comment to explain why not or make it T?

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+
+impl<T: IntSetMember<T>> Default for IntSet<T> {
     fn default() -> IntSet<T> {
         IntSet::empty()
     }
 }
 
-impl<T: Into<u32> + Copy> IntSet<T> {
+impl<T: IntSetMember<T>> IntSet<T> {
     // TODO(garretrieger): add additional functionality that the harfbuzz version has:
```
rsheeter, line 79: Suggest preference for issues over code todos. Issue is far more visible.

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
     // - Iteration in reverse.
     // - Iteration starting from some value (and before some value for reverse).
     // - Set operations (union, subtract, intersect, sym diff).
     // - Intersects range and intersects iter.
     // - min()/max()
 
+    /// Returns an iterator over all members of the set.
+    /// Note: iteration of inverted sets can be extremely slow due to the very large number of members in the set
+    /// care should be taken when using .iter() in combination with an inverted set.
```
rsheeter, line 88: Do you mean it's slower per item or merely that because the inversion probably has a lot of entries it'll be slow?

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+///
+/// 2. The mapped u32 values must retain the same ordering as the values in T.
+///
+/// 3. all_values_iter() must iterate over all values in T in sorted order (ascending).
+///
+/// from_mapped() will only ever be called with u32 values that were generated by a
+/// call to to_mapped() with a valid value of T. So it doesn't need to correctly handle values
+/// that are outside the domain of T.
+pub trait IntSetMember<T> {
+    fn to_mapped(&self) -> u32;
```
rsheeter, line 46: After reading some of the subsequent code I suggest `to_u32` might make things clearer. `to_mapped` is quite obtuse.

rsheeter (approved):

int-set/src/bitset.rs
```diff
@@ -11,28 +11,26 @@ const PAGE_BITS_LOG_2: u32 = PAGE_BITS.ilog2();
...
+pub(crate) struct BitSet {
     // TODO(garretrieger): consider a "small array" type instead of Vec.
     pages: Vec<BitPage>,
     page_map: Vec<PageInfo>,
     len: Cell<usize>, // TODO(garretrieger): use an option instead of a sentinel.
-    phantom: std::marker::PhantomData<T>,
 }
 
-impl<T: Into<u32> + Copy> BitSet<T> {
+impl BitSet {
```
cmyr, line 21: agree it makes sense to not have any trait bounds here, and let that live up above 👍

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+    fn from_mapped(member: InDomain) -> T;
+    fn all_values_sorted_iter() -> impl Iterator<Item = u32>;
+}
+
+/// Marks a mapped value as being in the domain of T for IntSetMember<T>. See IntSetMember
+/// for more information.
+pub struct InDomain {
```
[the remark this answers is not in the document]
cmyr: fwiw I generally follow the following convention for tuples:
- zero items: unit struct (`struct Foo`);
- exactly one item: tuple struct
- more than one item: named fields


🤷

int-set/src/lib.rs
```diff
@@ -192,64 +235,98 @@ impl<T> IntSet<T> {
...
         }
 
-        let next_index = index;
-        index = next_index + 1;
-        Some(next_index as u32)
+        None
     })
 }
 
+impl IntSetMember<u32> for u32 {
```
cmyr, line 302: we should also provide an impl for `GlyphId`.

int-set/src/lib.rs
```diff
@@ -110,22 +163,12 @@ impl<T: Into<u32> + Copy> IntSet<T> {
...
-
-    /// Returns an iterator over all members of the set.
-    /// Note: iteration of inverted sets can be extremely slow due to the very large number of members in the set
-    /// care should be taken when using .iter() in combination with an inverted set.
-    pub fn iter(&self) -> impl Iterator<Item = u32> + '_ {
-        match &self.0 {
-            Membership::Inclusive(s) => int_set_iter(s.iter(), false),
-            Membership::Exclusive(s) => int_set_iter(s.iter(), true),
-        }
+        IntSet(Membership::Exclusive(BitSet::empty()), PhantomData::<T>)
```
cmyr, line 171: Just out of curiousity, is this constructor used much in harfbuzz?

int-set/src/lib.rs
cmyr, line 88:
```suggestion
    /// Returns an iterator over all members of the set.
    ///
    /// Note: iteration of inverted sets can be extremely slow due to the very large number of members in the set
    /// care should be taken when using .iter() in combination with an inverted set.
```

maybe mentioned this at some point, but the first line of doc comments gets special treatment (as the summary in rendered docs) and you need an empty line for rustdoc to figure this out.

int-set/src/lib.rs
cmyr: if we're bikeshedding I might go with `ordered_values`.


Also an alternative way to implement this would be as a fn, `in_domain` that returns a bool?

int-set/src/lib.rs
cmyr, line 53: By convention if you're referencing types or identifiers that are used in the code you put them in backticks.

Also it's very easy to add links between items in your docs, and this has an additional benefit: during CI we check that these links actually reference things that exist, which means if you rename something your docs will break if you don't update them.

```suggestion
/// A value in the domain of `T` for some `IntSetMember<T>`.
///
/// See [`IntSetMember`] for more information.
pub struct InDomain {
```

(I didn't make the first `IntSetMember<T>` a link only because I'm not sure the autolinking works with generics? It probably does)

int-set/src/lib.rs
```diff
@@ -12,45 +12,92 @@
...
+///    integer.
+///
+/// 2. The mapped u32 values must retain the same ordering as the values in T.
+///
+/// 3. all_values_iter() must iterate over all values in T in sorted order (ascending).
+///
+/// from_mapped() will only ever be called with u32 values that were generated by a
+/// call to to_mapped() with a valid value of T. So it doesn't need to correctly handle values
+/// that are outside the domain of T.
+pub trait IntSetMember<T> {
```
cmyr, line 45: Looking at code below, I'd be tempted to call this trait `Domain`?

cmyr (approved):
some bike-shedding in the comments 🤓

int-set/src/lib.rs
garretrieger: The iterator is probably better since you'll avoid having to process large unused parts of the u32 domain for smaller types like u16. ordered_values() sounds good to me, I'll change to that.

int-set/src/lib.rs
garretrieger: Minor optimization so we don't have to list all of the T's and then immediately convert them all to u32 which is what's actually used under the hood. I'll add a comment.

int-set/src/lib.rs
garretrieger: I'm going to add some more of the missing stuff that's needed immediately in a follow up PR after this one, then I'll remove this TODO and file an issue for anything else that remains.

int-set/src/lib.rs
garretrieger: The latter, if the domain of T is all u32's iterating through them all takes a pretty long time even if the individual next() operations are fast. In harfbuzz we found if you're trying to iterate through an inverted set there's usually a better way to set up the algorithm that avoids the iteration in the inverted case.

int-set/src/lib.rs
garretrieger: Sounds good.

int-set/src/lib.rs
garretrieger: Sounds good, any other "integer" types that we use inside of fonttations that we might want to define?

int-set/src/lib.rs
garretrieger: Yes, mostly in parsing the command line flags/public api. It's used when asking for something like --unicodes="*"

int-set/src/lib.rs
dfrg: +1. I was going to suggest dropping the prefix but I really like `Domain`.

garretrieger: Domain makes sense, I'll rename.

PR #904, opened 2024-05-16
Make IntSet iterator double ended.

garretrieger:
To allow for reverse iteration.

### DIFF
--- a/int-set/src/lib.rs
+++ b/int-set/src/lib.rs
@@ -112,8 +112,8 @@ impl<T: Domain<T>> IntSet<T> {
     /// care should be taken when using .iter() in combination with an inverted set.
     pub fn iter(&self) -> impl DoubleEndedIterator<Item = T> + '_ {
         let u32_iter = match &self.0 {
-            Membership::Inclusive(s) => Iter::iter(s.iter(), None),
-            Membership::Exclusive(s) => Iter::iter(s.iter(), Some(T::ordered_values())),
+            Membership::Inclusive(s) => Iter::new(s.iter(), None),
+            Membership::Exclusive(s) => Iter::new(s.iter(), Some(T::ordered_values())),
         };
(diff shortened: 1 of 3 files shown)


int-set/src/lib.rs
```diff
@@ -189,6 +190,16 @@ impl<T: Domain<T>> IntSet<T> {
         }
     }
 
+    /// Returns the smallest member of this set if there is one.
+    pub fn min(&self) -> Option<T> {
```
cmyr, line 194: `BTreeSet` calls these fns `first` and `last`, maybe that would make sense? See https://doc.rust-lang.org/1.78.0/std/collections/struct.BTreeSet.html#method.last

int-set/src/lib.rs
```diff
@@ -374,11 +471,33 @@ impl Domain<u16> for u16 {
         true
     }
 
-    fn ordered_values() -> impl Iterator<Item = u32> {
+    fn ordered_values() -> impl DoubleEndedIterator<Item = u32> {
         (u16::MIN as u32)..=(u16::MAX as u32)
     }
 
-    fn ordered_values_range(range: RangeInclusive<u16>) -> impl Iterator<Item = u32> {
+    fn ordered_values_range(range: RangeInclusive<u16>) -> impl DoubleEndedIterator<Item = u32> {
```
cmyr, line 478: at some point I would probably write a macro do to these impls, since they're all so similar.

cmyr (approved):
looks great, a couple comments inline but nothing blocking :)

int-set/src/lib.rs
garretrieger: sounds good.

int-set/src/lib.rs
garretrieger: Yeah, sounds good

PR #907, opened 2024-05-23
[read] Add support for the 'gasp' table

### LINKED ISSUE #906: `gasp` table support needed in `read_fonts`
In order to read some hinting settings preferences from the font and configure hinting accordingly, we need access to the `gasp` table in `read_fonts`.

https://learn.microsoft.com/en-us/typography/opentype/spec/gasp

cmyr:
@drott is this fit for your purposes?

- closes #906


read-fonts/src/tables/gasp.rs
```diff
@@ -0,0 +1,3 @@
+//! The [gasp](https://learn.microsoft.com/en-us/typography/opentype/spec/gasp) table
+
+include!("../../generated/generated_cpal.rs");
```
rsheeter, line 3: s/cpal/gasp/

rsheeter (approved):
LGTM subject to addition of a test proving we can read at least one actual gasp as expected. Might need a new testdata item if we don't have any gasps?

drott:
Thank you @cmyr, yes, this is exactly what I was looking for.

read-fonts/src/tables/gasp.rs
cmyr: eek

PR #908, opened 2024-05-23
[IntSet] Add benchmark suite for int set.

garretrieger:
These benchmark functions are setup up to mirror the ones we have in harfbuzz for hb-set.


cmyr (approved):
looks good modulo the clippy complaints.

PR #911, opened 2024-05-24
[skrifa] Fix COLRv1 painting for font without ClipList

### LINKED ISSUE #909: Color glyph paint fails for colrv1 font without `ClipList`
Originally filed as: 

https://issues.skia.org/342377938 [Fontations] Honk and other COLRv1 fonts not displaying in Chrome Canary

drott:
A missing ClipList was passed up as a PaintError resulting in a paint failure. 
Remove confusion about how to handle a missing ClipList vs. a missing entry in a ClipList
by not returning a nested Result<Option<>> but a simpler Option<>.

Fixes #909.


dfrg (approved):
Ultimately, we only care about whether a clip box is available, and not why it might be missing so this simplification seems like a solid improvement. Thanks for the fix!

PR #913, opened 2024-05-24
[IntSet] Implement the set processing framework.

garretrieger:
This processing framework allows applying arbitrary bitwise operations between two bit sets. Ported fairly directly from the harfbuzz implementation in https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-bit-set.hh#L483. In the public api of IntSet I've added support for union() and intersect().

### DIFF
--- a/int-set/src/bitpage.rs
+++ b/int-set/src/bitpage.rs
@@ -136,25 +136,25 @@ impl BitPage {
     }
 
     pub(crate) fn union(a: &BitPage, b: &BitPage) -> BitPage {
-        BitPage::process(a, b, |a, b| a | b)
+        a.process(b, |a, b| a | b)
     }
 
     pub(crate) fn intersect(a: &BitPage, b: &BitPage) -> BitPage {
-        BitPage::process(a, b, |a, b| a & b)
+        a.process(b, |a, b| a & b)
     }
 
     pub(crate) fn subtract(a: &BitPage, b: &BitPage) -> BitPage {
-        BitPage::process(a, b, |a, b| a & !b)
+        a.process(b, |a, b| a & !b)
     }
 
-    fn process<Op>(a: &BitPage, b: &BitPage, op: Op) -> BitPage
+    fn process<Op>(&self, other: &BitPage, op: Op) -> BitPage
     where
-        Op: Fn(&Element, &Element) -> Element,
+        Op: Fn(Element, Element) -> Element,
     {
         let mut out = BitPage::new_zeroes();
         out.mark_dirty();
         for i in 0usize..(PAGE_SIZE as usize) {
-            out.storage[i] = op(&a.storage[i], &b.storage[i]);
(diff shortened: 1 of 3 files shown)


int-set/src/bitset.rs
```diff
@@ -325,6 +553,14 @@ mod test {
     use super::*;
     use std::collections::HashSet;
 
+    impl BitSet {
+        fn from<U: IntoIterator<Item = u32>>(iter: U) -> BitSet {
```
cmyr, line 557: ditto, I'd just implement `FromIterator`?

int-set/src/bitset.rs
```diff
@@ -172,6 +193,196 @@ impl BitSet {
...
+        one.insert(0);
+        let zero: BitPage = BitPage::new_zeroes();
+
+        // Determine the passthrough behaviour of the operator. The passthrough behaviour
+        // is what happens to a page on one side of the operation if the other side is 0.
+        // For example union passes through both left and right sides since it preserves
+        // the left or right side when the other side is 0. Knowing this lets us optimize
+        // some cases when only one page is present on one side.
+        let passthrough_left: bool = op(&one, &zero).contains(0);
+        let passthrough_right: bool = op(&zero, &one).contains(0);
```
cmyr, line 210: I'd be tempted to put this in a separate little helper?

int-set/src/bitset.rs
```diff
@@ -252,6 +463,23 @@ impl BitSet {
...
         self.pages.get_mut(page_index).unwrap()
     }
+
+    // Return the mutable page at a given index
+    fn page_for_index_mut(&mut self, index: usize) -> Option<&mut BitPage> {
+        if let Some(page_info) = self.page_map.get(index) {
+            self.pages.get_mut(page_info.index as usize)
+        } else {
+            None
+        }
```
cmyr, line 473: could just write this as,

```suggestion
        self.page_map
            .get(index)
            .and_then(|info| self.pages.get_mut(info.index as usize))
```

cmyr (approved):
Looks good, a few little observations inline!

int-set/src/bitpage.rs
```diff
@@ -135,6 +135,30 @@ impl BitPage {
...
+    pub(crate) fn subtract(a: &BitPage, b: &BitPage) -> BitPage {
+        BitPage::process(a, b, |a, b| a & !b)
+    }
+
+    fn process<Op>(a: &BitPage, b: &BitPage, op: Op) -> BitPage
+    where
+        Op: Fn(&Element, &Element) -> Element,
+    {
+        let mut out = BitPage::new_zeroes();
+        out.mark_dirty();
```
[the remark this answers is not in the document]
garretrieger: I don't think we've decided which way to go yet, but it's definitely worth prototyping and checking in benchmarks to see which approach is faster.

Issue #420, opened 2023-05-22
oss-fuzz for _at least_ font-types, read-fonts and skrifa


rsheeter:
Bonus points for write-fonts. https://github.com/google/oss-fuzz/tree/master/projects/woff2 might be a good starting point.

rsheeter:
https://google.github.io/oss-fuzz/getting-started/new-project-guide/rust-lang/ has instructions

rsheeter:
For skrifa https://github.com/google/oss-fuzz/tree/master/projects/freetype2 might be useful too, should be testing somewhat similar ops

rsheeter:
Per discussion with @garretrieger timeouts, very slow reads due to exploding graphs, seem likely to apply to Rust. Hopefully fewer use after free and out of bounds reads :)

rsheeter:
We'll need a font corpus to feed it. After a quick glance at https://github.com/google/oss-fuzz/blob/master/projects/freetype2/Dockerfile, perhaps `*.[ot]t[cf]` from:

1. https://github.com/unicode-org/text-rendering-tests
1. https://github.com/googlefonts/color-fonts 
1. https://github.com/harfbuzz/harfbuzz/tree/main/test

drott:
https://github.com/freetype/freetype2-testing/tree/master/fuzzing/corpora - in particular the COLRv1 corpus is something I derived for FT from https://github.com/googlefonts/color-fonts. The fonts in the corpora should be small and probably make sense to subset from actual font collections.

PR #917, opened 2024-05-29
Outline of a Skrifa fuzzer

rsheeter:
First step toward #420. Fuzzer is currently nop. Meant to enable oss-fuzz setup and establish a pattern we can use to add additional fuzzers for subsetting, etc.

For context, first swing at oss-fuzz rigging in https://github.com/google/oss-fuzz/pull/12011.

JMM

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -7,6 +7,10 @@ edition = "2021"
 [package.metadata]
 cargo-fuzz = true
 
+# cargo-release settings
+[package.metadata.release]
+release = false
+
 [dependencies]
 libfuzzer-sys = "0.4"
 skrifa = { path="../skrifa" }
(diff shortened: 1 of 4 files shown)


fuzz/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "fuzz"
+version = "0.0.0"
+publish = false
+edition = "2021"
+
```
cmyr, line 6: I think we might also want,

```toml
# cargo-release settings
[package.metadata.release]
release = false
```

cmyr (approved):
one note, otherwise lgtm

fuzz/Cargo.toml
rsheeter: Will add, ty

rsheeter: Fixed, ty

PR #919, opened 2024-05-29
[write-fonts/gvar] add axis_count param to Gvar::new

anthrotype:
this ensures that we (e.g. fontc) can make valid empty gvar tables where the axis_count matches the expected (fvar's) axis_count as per OT spec. 

Will eventually fix https://github.com/googlefonts/fontc/issues/815 once fontc is updated to use the new API.

This is a breaking changes which warrants minor write-fonts version bump.

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -52,8 +52,12 @@ pub struct GlyphDelta {
 /// An error representing invalid input when building a gvar table
 #[derive(Clone, Debug)]
 pub enum GvarInputError {
-    /// Glyphs do not have a consistent axis count
-    InconsistentAxisCount,
+    /// Glyph variations do not have the expected axis count
+    UnexpectedAxisCount {
+        gid: GlyphId,
+        expected: u16,
+        got: u16,
+    },
     /// A single glyph contains variations with inconsistent axis counts
     InconsistentGlyphAxisCount(GlyphId),
     /// A single glyph contains variations with different delta counts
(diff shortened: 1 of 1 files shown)


cmyr:
Okay so to understand the situation here: sometimes we would like to have an empty gvar table, but set an explicit axis_count?

If yes: is it true that there are basically two ways we expect the new API to be called:
- previously, where the axis_count _could_ be inferred from the input
- the empty case, where there are no variations but we want a non-zero axis count?


If yes, my one nit with this PR is that it introduces a new error and possible error condition that we might be able to avoid. For instance, we could have a separate constructor `Gvar::empty(axis_count: u16)` that would handle this case, and since there would be no provided variations we wouldn't need to worry about a possible error?

That does push the responsibility down to the caller, to ensure that they call a different constructor to get an empty table, so there is that downside.


What do you think?

anthrotype:
> sometimes we would like to have an empty gvar table, but set an explicit axis_count?

yes, because some old apps expect gvar to always be present in a VF even if empty, and the spec wants gvar's axis_count == fvar.axis_count no matter what (e.g. fonttools fails to decompile an empty gvar with axis_count of 0 where fvar.axis_count != 0)

> previously, where the axis_count could be inferred from the input...

technically it would still be possible for the caller to construct a gvar table where all the provided variations have the same, consistent axis_count, but the latter turns out *not* to match the global fvar's axis_count, but you could argue that's not write-font's fault.

But I'm not opposed to the proposed additional `Gvar::empty(axis_count)` constructor either. I think we are splitting hairs for an arguably-very-limited use case so whatever makes you guys happy. /cc @rsheeter

cmyr:
My preference is just for simpler code and fewer error conditions, but agree it's not a big deal.

cmyr (approved):
minor preference for the alternative version I think but merge this if you prefer!

write-fonts/src/tables/gvar.rs
```diff
@@ -106,7 +95,21 @@ impl Gvar {
...
         }
 
-        let axis_count = validate_variations(&variations)?;
+        for var in &variations {
+            var.validate()?;
+        }
+
+        for var in &variations {
+            if let Some(x) = var.axis_count() {
+                if x != axis_count {
```
rsheeter, line 104: Could be done with less nesting I think? Untested, but something like:

```rust
if let Some(bad_var) = variations.iter()
  .filter(|v| v.axis_count().is_some())
  .find(|v| Some(axis_count) != v.axis_count()) {
  return Err(...)
}

File: write-fonts/src/tables/gvar.rs [hunk 2]
```diff
@@ -52,8 +52,12 @@ pub struct GlyphDelta {
 /// An error representing invalid input when building a gvar table
 #[derive(Clone, Debug)]
 pub enum GvarInputError {
-    /// Glyphs do not have a consistent axis count
-    InconsistentAxisCount,
+    /// Glyph variations do not have the expected axis count
+    UnexpectedAxisCount {
+        gid: GlyphId,
+        expected: u16,
+        got: u16,
```
rsheeter, line 59: Suggest s/got/actual/; pairing expected and actual is pretty standard.

rsheeter (approved):

PR #924, opened 2024-05-30
[write-fonts] Fix incorrect classdef format selection

### LINKED ISSUE #923: incorrect format selection when building empty classdef tables
Currently if we attempt to build an empty classdef table we will select format 1, which requires at least one item, so we end up with a table that has the `0` gid in the `0` class, which is incorrect.

Ran into this while trying to figure out why fontc and fontmake are generating slightly different contextual lookups in some cases.

cmyr:
- fixes #923

### DIFF
--- a/write-fonts/src/tables/layout.rs
+++ b/write-fonts/src/tables/layout.rs
@@ -467,15 +467,21 @@ impl FromIterator<(GlyphId, u16)> for ClassDef {
 
 impl ClassDefBuilder {
     fn prefer_format_1(&self) -> bool {
+        const U16_LEN: usize = std::mem::size_of::<u16>();
+        const FORMAT1_HEADER_LEN: usize = U16_LEN * 3;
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/layout.rs
```diff
@@ -467,6 +467,10 @@ impl FromIterator<(GlyphId, u16)> for ClassDef {
 
 impl ClassDefBuilder {
     fn prefer_format_1(&self) -> bool {
+        // empty table must always use format 2
```
behdad, line 470: I would change the comment saying it's more compact to use format 2. Or am I missing something?

cmyr: oops yes I was confused by the fact that format 1 has a `startGlyphId` field, and was thinking that that meant it had to have at least one item in it.

In addition in the process of going and looking through this again I noticed another miscalculation here, and have pushed an additional commit to fix that.

anthrotype (approved):

PR #925, opened 2024-05-30
[kilppa] initial setup for integration tests

qxliu76:
Try to set up testing framework for klippa(most basic version).
- Moved some functions to lib.rs so they're visible for tests as well.
- I start by making a copy of the test data from harfbuzz, which contains only 1 basic test. Later on we can update to sync with harfbuzz repo.


klippa/src/main.rs
```diff
@@ -52,15 +31,15 @@ struct Args {
 fn main() {
     let args = Args::parse();
 
-    let gids = match populate_gids(&args.gids) {
+    let gids = match populate_gids(&args.gids.unwrap_or(String::new())) {
```
cmyr, line 34: - as written, you can do `unwrap_or_default()`.
- it seems a bit funny to call this fn with an empty string when there's no argument, so I might prefer something like,

```rust
let gids = args.gids.as_ref().map(populate_gids).transpose()?.unwrap_or_default();
```
this is imagining that you're in a function that returns a result, and so can use `?`. (`.transpose()` is a method that turns `Option<Result<T, E>>` into `Result<Option<T>, E>`)

Also I would suggest a minor refactor here: instead of having a bunch of different places where you catch an error, you can have a single `fn run() -> Result<(), SubsetError>` and in your `main()` all you do is call that function and print any errors; that function would have all of your actual logic.

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
+//! Test subsetter output against expected results.
```
cmyr, line 1: I think the name of this file is slightly vague, and I would prefer to call it something like `harfbuzz_tests` or `hb_subset_tests` or something?

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+//!
+//! This reads test configs from harfbuzz subset test suite,
+//! generate a group of tests to perform, run and then compare the output against the stored expected result
+//!
+//! To generate the expected output files, pass GEN_EXPECTED_OUTPUTS=1 as an
+//! environment variable.
```
cmyr, line 7: is this something we want to do? I would expect that the expected outputs in harfbuzz to be canonical?

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+                other => panic!("unexpected heading '{other}'"),
+            }
+        }
+        self.case
+    }
+
+    fn parse_fonts(&mut self, lines: &mut LinesIter) {
+        while !lines.is_end() {
+            if let Some(next) = lines.next() {
+                self.case.fonts.push(next.to_owned());
```
cmyr, line 120: I would probably make all of these `next.trim().to_owned()`, in case there happens to be any unexpected whitespace.

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+            for subset in &self.subsets {
+                //TODO: add support for instances/iup_options
+                self.run_one_test(font, subset, output_dir);
+            }
+        }
+    }
+
+    pub fn gen_expected_output(&self) {
+        //let output_dir = PathBuf::from("./qxliu_test");
+        //fs::create_dir("./qxliu_test").unwrap();
```
cmyr, line 213: do we intend to commit these?

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+    }
+}
+
+impl SubsetTestCase {
+    pub fn new(path: &Path) -> Self {
+        let parser = TestCaseParser::new();
+        parser.parse(path)
+    }
+
+    pub fn run(&self) {
```
cmyr, line 197: I suspect nothing here actually needs to be `pub`.

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+        //TODO: re-enable subset="*" once populate_unicodes_to_retain supports *
+        if subset == "*" {
+            return;
+        }
+
+        let subset_font_name = gen_subset_font_name(font, subset);
+        let output_file = output_dir.join(&subset_font_name);
+        gen_subset_font_file(font, subset, &output_file);
+
+        assert_has_ttx_exec();
```
cmyr, line 257: I might just move this into the top of the main test body, where we just run it once before we do anything else?

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+        let mut unicodes_option = String::from("--unicodes=");
+        unicodes_option.push_str(subset);
+
+        let mut output_option = String::from("--output-file=");
+        output_option.push_str(expected_file.to_str().unwrap());
+
+        let mut org_font_file = String::new();
+        org_font_file.push_str(TEST_DATA_DIR);
+        org_font_file.push_str("/fonts/");
+        org_font_file.push_str(font);
```
cmyr, line 271: I would just work with paths here? like,

```rust
let org_font_file = Path::new(TEST_DATA_DIR).join(fonts).join(font);
```

which isn't perfectly efficient (both `join` calls allocate) but is nice and concise at least.

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+        } else {
+            write!(&mut out, ",{:x}", c).unwrap();
+        }
+    }
+    out
+}
+
+fn strip_unicode_prefix(text: &str) -> String {
+    text.replace("U+", "")
+}
```
cmyr, line 346: just `text.trim_start_matches("U+")` should be fine, and it doesn't reallocate.

klippa/tests/integration_test.rs
```diff
@@ -0,0 +1,538 @@
...
+    };
+    let mut subset_font_name = String::new();
+    let (font_base_name, font_extension) = font.rsplit_once('.').unwrap();
+
+    //TODO: add profiles/instances later
+    subset_font_name.push_str(font_base_name);
+    subset_font_name.push('.');
+    subset_font_name.push_str(subset_name);
+    subset_font_name.push('.');
+    subset_font_name.push_str(font_extension);
```
cmyr, line 362: you can do this all with the `format!` macro I think? like,


```rust
let subset_font_name = format!("{font_base_name}.{subset_name}.{font_extension}");
```

cmyr (approved):
looks like a good start!


- I would add a README in the `test-data` directory explaining where the data comes from, and I would also encourage you to commit a script that can be used to update this data, if we ever expect it to change.


Otherwise a bunch of little notes and suggestions inline, but overall looks reasonable!

klippa/src/lib.rs
```diff
@@ -103,3 +118,81 @@ pub trait Subset {
     /// Subset this object. Returns `true` if the object should be retained.
     fn subset(&mut self, plan: &Plan) -> Result<bool, SubsetError>;
 }
+
+pub fn parse_unicodes(unicode_str: &str) -> Result<IntSet<u32>, SubsetError> {
```
garretrieger, line 122: These two parsing functions should probably be pub(crate) since we don't need to use them outside of this crate. Also suggest moving them into their own module ("cli", "parsing", or "util" maybe?). Could you also add some unit tests for the two parsing functions.

klippa/src/main.rs
qxliu76: I changed to `unwrap_or_default()`. And I'll refactor the code as you suggested later in another PR.
I tried your preferred format but I got an error because `args.gids.as_ref() `is an `Option<&str>`, while the input of `populate_gids` is `&str`.

klippa/tests/integration_test.rs
qxliu76: yes, we'd like to be able to generate our own golden files. Because currently we only have the most basic subset support and nothing matches harfbuzz results.

klippa/tests/integration_test.rs
qxliu76: we cannot, because we generate multiple subset font files in a nested loop for one test

klippa/tests/integration_test.rs
qxliu76: seems `trim_start_matches()` only trims the prefix once, I need to remove all "U+", because the input is like "U+1,U+2,U+3", any better ideas?

klippa/tests/integration_test.rs
qxliu76: oh actually this is our own integration test, we'll store our expected files as well

garretrieger (approved):

PR #927, opened 2024-06-03
Make it possible to draw cubics w/glyf in Skrifa

rsheeter:
We were destroying the cubic flag, resulting in cubic off curves being interpreted as quad with an implicit oncurve.


dfrg (approved):
Looks good. 

There’s an interesting ambiguity here in how we should treat flags where both the on curve and cubic bits are set. I suppose the on curve bit should have precedence otherwise the TT flip region instructions would not work with cubic glyphs.

@behdad wdyt?

behdad:
> There’s an interesting ambiguity here in how we should treat flags where both the on curve and cubic bits are set. I suppose the on curve bit should have precedence otherwise the TT flip region instructions would not work with cubic glyphs.

We specifically made the cubic bit on on-curve points being a no-op, exactly because TT flip region instructions can use it. So yes, that combination is well-defined as far as I was involved in the design.

font-test-data/test_data/ttx/cubic_glyf.ttx
```diff
@@ -0,0 +1,184 @@
...
+      <instructions/>
+    </TTGlyph>
+
+    <TTGlyph name="space"/><!-- contains no outline data -->
+
+    <TTGlyph name="uni000E" xMin="278" yMin="470" xMax="998" yMax="710">
+      <contour>
+        <pt x="278" y="710" on="1"/>
+        <pt x="278" y="470" on="1"/>
+        <!-- cubic, not implicit on-curve cubic -->
```
anthrotype, line 173: > cubic, not implicit on-curve cubic

I think you meant to write "cubic, not implicit on-curve quadratic"?

rsheeter: Erk yes indeed

behdad:
Did you see the tests in fonttools? Those are good.

rsheeter:
> Did you see the tests in fonttools?

I did not, ty for the reference!

drott:
@rsheeter I was looking into release skrifa and due to how my repo was set up, I came across this commit. Can we please put extensions that are not in OpenType or in OFF yet behind a flag? It's risky to start shipping something like cubic glyf unofficially in Chrome if we have to edit the spec later.

PR #930, opened 2024-06-04
Add the ability to read the VARC condition list

rsheeter:
I love a headache from hand editing ttx in the evening


read-fonts/src/tables/varc.rs
```diff
@@ -1,17 +1,139 @@
...
     }
+
+    #[test]
+    fn identify_all_conditional_types() {
+        let font = FontRef::new(font_test_data::varc::CONDITIONALS).unwrap();
+        let table = font.varc().unwrap();
+
+        // We should have all 5 condition types in order
+        assert_eq!(
+            (1..=5).collect::<Vec<_>>(),
```
cmyr, line 38: iirc you can compare slices/arrays with vecs (there are `PartialEq` impls) so I think you can write, `assert_eq!([1, 2, 3, 4, 5], blah.collect::<Vec<_>>())`

resources/codegen_inputs/varc.rs
```diff
@@ -5,9 +5,32 @@
...
+
+}
+
+table ConditionList {
+    condition_count: u32,
+    #[count($condition_count)]
+    condition_offsets: [Offset32<Condition>],
+}
+
+table AxisIndicesList {
```
cmyr, line 30: maybe some docs explaining why these empty tables exist?

resources/codegen_inputs/layout.rs
```diff
@@ -611,14 +611,17 @@ table ConditionSet {
...
+    condition_offsets: [Offset32<Condition>],
 }
 
-///// [Condition Table](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#condition-table)
-//Condition {
-    ///// FIXME: make an enum
-    //no_field: fake,
-//}
+/// [Condition Table](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#condition-table)
+format u16 Condition {
```
cmyr, line 618: I would expand on the docs a bit to mention that we are including items that are not yet part of the spec, and then link to whatever the best current reference is?

read-fonts/src/tables/varc.rs
rsheeter: I'm getting ```can't compare `[{integer}; 5]` with `Vec<i32>``` so I will leave it for the time being

resources/codegen_inputs/varc.rs
rsheeter: Added a terse explanation; I'm not planning to leave them blank very long (subject to meeting load)

resources/codegen_inputs/layout.rs
rsheeter: Good idea, done

read-fonts/src/tables/varc.rs
```diff
@@ -1,17 +1,139 @@
...
 
 include!("../../generated/generated_varc.rs");
 
 #[cfg(test)]
 mod tests {
     use crate::{FontRef, TableProvider};
 
+    use super::{Condition, Varc};
+
+    fn read_conditions<'a>(table: &'a Varc) -> Vec<Condition<'a>> {
```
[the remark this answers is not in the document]
rsheeter: Good point, I think I underutilize this. The code then reads more clearly (LTR) which is nice. So updated.

read-fonts/src/tables/varc.rs
cmyr: ah okay the issue there might be that you need to be explicit about the integer type, like `[1i32, 2, 3, 4, 5]`

PR #932, opened 2024-06-04
Box write-fonts offset markers

rsheeter:
lest recursive data structures like COLR Paint (not yet generating a write-fonts impl) or Conditions in #930 get upset about infinite recursion to get size.

```
// This is OK, presumably because we don't generate compile (write-fonts) for colr
format u8 Paint {
   Glyph(PaintGlyph),
}
table PaintGlyph {
   paint_offset: Offset24<Paint>,
}

// This is NOT OK; reading is fine but writing fails
// error[E0072]: recursive types `tables::layout::Condition` and `ConditionFormat5` have infinite size
format u16 Condition {
    Format5Negate(ConditionFormat5),
}
table ConditionFormat5 {
    condition_offset: Offset24<Condition>,
}
```

Fix suggested by @dfrg , errors all by me. 

Meant to unblock https://github.com/googlefonts/fontations/pull/930.

### DIFF
--- a/write-fonts/src/offsets.rs
+++ b/write-fonts/src/offsets.rs
@@ -18,7 +18,7 @@ pub const WIDTH_32: usize = 4;
 #[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct OffsetMarker<T, const N: usize = 2> {
-    obj: T,
+    obj: Box<T>,
 }
 
 /// An offset subtable which may be null.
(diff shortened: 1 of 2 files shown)


write-fonts/src/offsets.rs
```diff
@@ -44,7 +44,7 @@ impl<T, const N: usize> std::ops::DerefMut for OffsetMarker<T, N> {
 }
 
 impl<T, const N: usize> std::ops::Deref for NullableOffsetMarker<T, N> {
-    type Target = Option<T>;
+    type Target = Option<Box<T>>;
```
dfrg, line 47: This is a bit unfortunate but I don’t see a better… option. 

I believe adding `as_ref(&self) -> Option<&T>` to this type would avoid the changes outside this file if we want to reduce the diff.

dfrg (approved):
LGTM with one suggestion.

write-fonts/src/offsets.rs
cmyr: to elaborate on this a tiny bit:


if you add the following fns to `NullableOffsetMarker`:

```rust
/// Convert from `&NullableOffsetMarker<T>` to `Option<&T>`.
    pub fn as_ref(&self) -> Option<&T> {
        match &self.obj {
            Some(obj) => Some(&*obj),
            None => None,
        }
    }

    /// Convert from `&NullableOffsetMarker<T>` to `Option<&T>`.
    pub fn as_mut(&mut self) -> Option<&mut T> {
        match &mut self.obj {
            Some(obj) => Some(&mut *obj),
            None => None,
        }
    }
```

Then the only other file you need to change is `validate.rs`, and there it's a couple of lines (basically make the body of the `Validate` impl for this type look like the one for `Option<T>`, directly below it)

PR #937, opened 2024-06-06
[codegen] Add #[if_flag] annotation

cmyr:
This adds the annotation and tests, but does not use it in any real tables.

One limitation of this implementation is that no non-conditional fields can follow the first conditional field; I'm not sure if this is an issue in practice since in the examples I'm aware of conditional fields are always after all non-conditional fields.

I have a followup PR that will use this in GPOS/GSUB. @dfrg can you think of anywhere else where this would be useful immediately? And can you point me to the other example of a conditional that you had mentioned?

One thing I could now consider is using this mechanism for `ValueRecord`.


dfrg:
The CFF/CFF2 INDEX tables are the only others that come to mind. These are codegen’d as Index1 and Index2 here: https://github.com/googlefonts/fontations/blob/975287b3ee1f39a81c313d9e59a24dc38583637e/resources/codegen_inputs/postscript.rs#L4

font-codegen/src/fields.rs
```diff
@@ -1447,6 +1521,14 @@ fn width_for_offset(offset: &syn::Ident) -> Option<syn::Ident> {
     }
 }
 
+/// turn a syn path like 'std :: hmm :: Thing' into "Thing"
+fn stringify_path(path: &syn::Path) -> String {
```
rsheeter, line 1525: I would have expected stringify path to return the whole path as a string but this seemingly doesn't do that at all. I also don't love the parsing of token stream to string.

Wdyt about the alternative proposed in https://github.com/googlefonts/fontations/pull/946?

rsheeter (approved):
General comment: I wish we had more unit testing for codegen, for example fields.rs has no tests at all and yet contains testable functionality. We have tests, but they tend to be larger blocks, more like integration tests perhaps. `stringify_path` is a great example, it's definitely testable on it's own.

font-codegen/src/fields.rs
cmyr: so this is very narrowly scoped, and is specifically about getting a shorter name that we can use in certain diagnostic messages, for which we need a string; `syn::Path` does not implement `Display`.

cmyr:
So my general approach to testing codegen is to focus testing on integration testing, and ensuring that we are producing the code that we want to produce for a given input.

Given that codegen is only ever run by us, and that we control the inputs, I am definitely less concerned with handling certain sorts of edge cases; for instance with this patch you could totally write,

```rust
table Hmm {
    flags: MyFlags,
    #[if_flag($flags, ::std::option::Option<isize>)
    other_field: i16,
}
```

(that is, having a nonsensical value for the constant in this attribute) and codegen will happily run, and generate some code that won't compile. We _could_ have some unit test here, and have codegen fail with some slightly more ergonomic error message, but realistically will that ever actually be helpful?

I do write unit tests for any parsing code that is modestly complicated, but in general I find the integration tests to be much more helpful for actually ensuring that the code is doing what we want.

rsheeter:
> I am definitely less concerned with handling certain sorts of edge cases; for instance with this patch you could totally write,

I agree edge cases are less value than something faced with arbitrary input. IMHO that doesn't eliminate the value of unit tests, just confirming the happy path works on a fine grained basis can be quite nice.

PR #939, opened 2024-06-06
[read] Remove methods from LookupFlag

cmyr:
now that we have a `contains` method (or will as of #938) and the `BitOr` and `BitOrAssign` impls for this type, these methods are redundant and unnecessarily verbose.


rsheeter (approved):
Good change but I'm mildly alarmed that we seemingly neither use nor test these methods

cmyr:
they're used when we compile features in fea-rs, but we'll replace that with using the constants.

PR #940, opened 2024-06-06
Add support for 4 byte deltas

rsheeter:
https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#tuplevalues makes the previously invalid combination of DELTAS_ARE_ZERO and DELTAS_ARE_WORDS mean the value is an i32.

### DIFF
--- a/read-fonts/src/tables/variations.rs
+++ b/read-fonts/src/tables/variations.rs
@@ -411,12 +411,45 @@ impl<'a> PackedDeltas<'a> {
     }
 }
 
+/// Flag indicating that this run contains no data,
+/// and that the deltas for this run are all zero.
+const DELTAS_ARE_ZERO: u8 = 0x80;
+/// Flag indicating the data type for delta values in the run.
+const DELTAS_ARE_WORDS: u8 = 0x40;
+/// Mask for the low 6 bits to provide the number of delta values in the run, minus one.
+const DELTA_RUN_COUNT_MASK: u8 = 0x3F;
+
+/// The type of values for a given delta run (influences the number of bytes per delta)
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum DeltaRunType {
+    Zero,
+    I8,
+    I16,
+    I32,
+}
+
+impl DeltaRunType {
+    /// The run type for a given control byte
+    pub fn new(control: u8) -> Self {
+        // if the top two bits of the control byte (DELTAS_ARE_ZERO and DELTAS_ARE_WORDS) are both set,
+        // then the following values are 32-bit.
+        // <https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#tuplevalues>
+        let are_zero = (control & DELTAS_ARE_ZERO) != 0;
+        let are_words = (control & DELTAS_ARE_WORDS) != 0;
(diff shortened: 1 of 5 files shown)


read-fonts/src/tables/variations.rs
```diff
@@ -449,18 +456,26 @@ impl Iterator for DeltaRunIter<'_> {
...
+            // then the following values are 32-bit.
+            // <https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#tuplevalues>
+            let are_zero = (control & DELTAS_ARE_ZERO) != 0;
+            let are_words = (control & DELTAS_ARE_WORDS) != 0;
+            self.value_type = match (are_zero, are_words) {
+                (false, false) => DeltaRunValueType::I8,
+                (false, true) => DeltaRunValueType::I16,
+                (true, false) => DeltaRunValueType::Zero,
+                (true, true) => DeltaRunValueType::I32,
+            };
```
cmyr, line 469: maybe this can be a constructor on `DeltaRunValueType`?

read-fonts/src/tables/variations.rs
```diff
@@ -411,28 +411,35 @@ impl<'a> PackedDeltas<'a> {
     }
 }
 
+/// The type, notably influence how many bytes per entry, of value in a delta run
```
cmyr, line 414: I found this hard to grok, maybe just,

```suggestion
/// The type of values for a given delta run (influences the number of bytes per delta)
```

or something?

read-fonts/src/tables/variations.rs
```diff
@@ -411,28 +411,35 @@ impl<'a> PackedDeltas<'a> {
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum DeltaRunValueType {
```
cmyr, line 416: `RunType` also feels like a reasonable name? Or `RunValueType`? 🤷

read-fonts/src/tables/variations.rs
```diff
@@ -449,18 +456,26 @@ impl Iterator for DeltaRunIter<'_> {
...
         }
 
         self.remaining -= 1;
-        if self.are_zero {
-            Some(0)
-        } else if self.two_bytes {
-            self.cursor.read().ok()
-        } else {
-            self.cursor.read::<i8>().ok().map(|v| v as i16)
+        match self.value_type {
```
cmyr, line 474: aren't enums lovely 🥲

cmyr (approved):
looks clean, using the enum to represent the different value lengths is definitely the right approach. A few little things inline, at your discretion.

read-fonts/src/tables/variations.rs
rsheeter: I had to write not-Rust recently and I really really missed them!

read-fonts/src/tables/variations.rs
rsheeter: That would be much nicer! Will do

read-fonts/src/tables/variations.rs
rsheeter: Everything else had the Delta prefix so it feels weird without that so ... DeltaRunType perhaps. Will try that.

PR #943, opened 2024-06-07
Begin to parse VARC glyphs

rsheeter:
Not done but a significant enough step I'd like to land it.

### DIFF
--- a/read-fonts/src/font_data.rs
+++ b/read-fonts/src/font_data.rs
@@ -179,6 +179,25 @@ impl<'a> Cursor<'a> {
         self.pos += n_bytes;
     }
 
+    /// Read a variable length u32 and advance the cursor
+    pub(crate) fn read_u32_var(&mut self) -> Result<u32, ReadError> {
+        let mut next = || self.read::<u8>().map(|v| v as u32);
+        let b0 = next()?;
+        // TODO this feels possible to simplify, e.g. compute length, loop taking one and shifting and or'ing
+        let result = match b0 {
+            _ if b0 < 0x80 => b0,
+            _ if b0 < 0xC0 => (b0 - 0x80) << 8 | next()?,
+            _ if b0 < 0xE0 => (b0 - 0xC0) << 16 | next()? << 8 | next()?,
+            _ if b0 < 0xF0 => (b0 - 0xE0) << 24 | next()? << 16 | next()? << 8 | next()?,
+            _ => {
+                // TODO: << 32 doesn't make sense. (b0 - 0xF0) << 32
+                next()? << 24 | next()? << 16 | next()? << 8 | next()?
+            }
+        };
+
+        Ok(result)
+    }
+
     /// Read a scalar and advance the cursor.
     pub(crate) fn read<T: Scalar>(&mut self) -> Result<T, ReadError> {
         let temp = self.data.read_at(self.pos);
@@ -261,6 +280,14 @@ impl<'a> Cursor<'a> {
     // end of a table.
     pub(crate) fn remaining_bytes(&self) -> usize {
         self.data.len().saturating_sub(self.pos)
+    }
+
+    pub(crate) fn remaining(self) -> Option<FontData<'a>> {
+        self.data.split_off(self.pos)
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.pos >= self.data.len()
     }
 
     pub(crate) fn finish<T>(self, shape: T) -> Result<TableRef<'a, T>, ReadError> {
--- a/read-fonts/src/tables/varc.rs
+++ b/read-fonts/src/tables/varc.rs
@@ -8,21 +8,230 @@ pub use super::{
 
 include!("../../generated/generated_varc.rs");
 
+/// Let's us call self.something().get(i) instead of get(self.something(), i)
+trait Get<'a> {
+    fn get(self, nth: usize) -> Result<&'a [u8], ReadError>;
+}
+
+impl<'a> Get<'a> for Option<Result<Index2<'a>, ReadError>> {
+    fn get(self, nth: usize) -> Result<&'a [u8], ReadError> {
+        self.transpose()?
+            .ok_or(ReadError::NullOffset)
+            .and_then(|index| index.get(nth).map_err(|_| ReadError::OutOfBounds))
+    }
+}
+
 impl<'a> Varc<'a> {
+    /// Friendlier accessor than directly using raw data via [Index2]
     pub fn axis_indices(&self, nth: usize) -> Result<PackedDeltas, ReadError> {
-        let Some(axis_indices_list) = self.axis_indices_list() else {
-            return Err(ReadError::InvalidCollectionIndex(nth as u32));
+        let raw = self.axis_indices_list().get(nth)?;
+        Ok(PackedDeltas::consume_all(raw.into()))
+    }
+
+    /// Friendlier accessor than directly using raw data via [Index2]
+    ///
+    /// nth would typically be obtained by looking up a [GlyphId] in [Self::coverage].
+    pub fn glyph(&self, nth: usize) -> Result<VarcGlyph<'_>, ReadError> {
+        let raw = Some(self.var_composite_glyphs()).get(nth)?;
+        Ok(VarcGlyph {
+            table: self,
+            data: raw.into(),
+        })
+    }
+}
+
+/// A VARC glyph doesn't have any root level attributes, it's just a list of components
+///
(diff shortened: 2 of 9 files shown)


read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
...
+    ///  * HarfBuzz [VarComponent::get_path_at](https://github.com/harfbuzz/harfbuzz/blob/0c2f5ecd51d11e32836ee136a1bc765d650a4ec0/src/OT/Var/VARC/VARC.cc#L132)
+    // TODO: do we want to be able to parse into an existing glyph to avoid allocation?
+    fn parse(table: &Varc, cursor: &mut Cursor<'a>) -> Result<Self, ReadError> {
+        let raw_flags = cursor.read_u32_var()?;
+        let flags = VarcFlags::from_bits_truncate(raw_flags);
+        // Ref https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#variable-component-record
+
+        // This is a GlyphID16 if GID_IS_24BIT bit of flags is clear, else GlyphID24.
+        let gid = if flags.contains(VarcFlags::GID_IS_24BIT) {
+            let gid = cursor.read_be::<Uint24>()?.get().to_u32();
```
cmyr, line 108: you can use `read` instead of `read_be`, the only difference is that `read_be` returns `BigEndian<T>` and `read` returns `T` (handling the conversion). `read_be` only exists for codegening certain records that we can't zerocopy because they have variable size, but the types expect the `BigEndian` wrapper.

can open an issue to clarify this on the docs of `read_be`.

read-fonts/src/font_data.rs
```diff
@@ -179,6 +179,34 @@ impl<'a> Cursor<'a> {
         self.pos += n_bytes;
     }
 
+    /// Read a variable length u32 and advance the cursor
+    pub(crate) fn read_u32_var(&mut self) -> Result<u32, ReadError> {
```
cmyr, line 183: Is this particular variable-length u32 encoding going to be used in more than just Varc? If not, I think this logic would make more sense as a function in the varc module... I think it's higher-level than the cursor itself, and the logic is easy to implement without really needing any access to cursor internals (anything with access to  `&mut cursor` can handle it.)


not a big deal in any case, but it feels _slightly_ out of place.

read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
...
+            };
+            (Some(axis_indices_index), deltas)
+        } else {
+            (None, None)
+        };
+
+        let axis_values_var_index = if flags.contains(VarcFlags::AXIS_VALUES_HAVE_VARIATION) {
+            Some(cursor.read_u32_var()?)
+        } else {
+            None
```
cmyr, line 146: not huge savings, but these can also be written as,
```suggestion
let axis_values_var_index = flags
            .contains(VarcFlags::AXIS_VALUES_HAVE_VARIATION)
            .then(|| cursor.read_u32_var())
            .transpose()?;
```

read-fonts/src/tables/variations.rs
```diff
@@ -390,24 +390,28 @@ impl<'a> ExactSizeIterator for PackedPointNumbersIter<'a> {}
 #[derive(Clone, Debug)]
 pub struct PackedDeltas<'a> {
     data: FontData<'a>,
+    // How many values we expect
     count: usize,
 }
 
 impl<'a> PackedDeltas<'a> {
+    pub fn new(data: FontData<'a>, count: usize) -> Self {
```
cmyr, line 398: does this need to be pub-pub? could it be pub(crate)? If pub, are there any invariants we should document?

read-fonts/src/tables/variations.rs
```diff
@@ -448,36 +452,53 @@ impl DeltaRunType {
...
+    }
 }
 
 impl Iterator for DeltaRunIter<'_> {
     type Item = i32;
 
     fn next(&mut self) -> Option<Self::Item> {
+        if let Some(limit) = self.limit {
+            if limit == 0 {
+                debug_assert!(
```
cmyr, line 483: a `debug_assert!`???

cmyr (approved):
some notes inline but overall looks sane!

read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
 
 include!("../../generated/generated_varc.rs");
 
+trait RawNth<'a> {
```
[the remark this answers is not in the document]
rsheeter: I'm of the opinion left to right is more legible, first_step.second_step instead of reading inside-out, second(first()).

read-fonts/src/font_data.rs
rsheeter: Interesting. To me it's just like any of the cursor ops and fits more naturally there than anywhere else but I guess it's somewhat subjective. Fwiw I had it in varc and had _exactly_ the same feeling: it felt slightly out of place.

read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
...
+    fn nth(self, nth: usize) -> Result<&'a [u8], ReadError>;
+}
+
+impl<'a> RawNth<'a> for Option<Result<Index2<'a>, ReadError>> {
+    fn nth(self, nth: usize) -> Result<&'a [u8], ReadError> {
+        let Some(index) = self else {
+            return Err(ReadError::InvalidCollectionIndex(nth as u32));
+        };
+        let index = index?;
+        index.get(nth).map_err(|_| ReadError::OutOfBounds)
```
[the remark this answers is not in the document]
rsheeter: I use transpose so rarely I evidently forgot it exists. I like that better, so updated.

read-fonts/src/tables/varc.rs
rsheeter: Can confirm I was initially unclear whether I could safely use read or not. I think I concluded I could and updated to do so. Updated _almost_ everything I suppose. Will update.

read-fonts/src/font_data.rs
```diff
@@ -179,6 +179,34 @@ impl<'a> Cursor<'a> {
...
+
+    /// Read the next byte and advance the cursor
+    fn read_u8(&mut self) -> Result<u8, ReadError> {
+        let Some(v) = self.data.bytes.get(self.pos) else {
+            return Err(ReadError::OutOfBounds);
+        };
+        self.pos += 1;
+        Ok(*v)
+    }
+
```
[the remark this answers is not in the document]
rsheeter: Nope! This is a copy/paste error from a point in editing when I had reinvented some of cursor. Deleted.

read-fonts/src/tables/varc.rs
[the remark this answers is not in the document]
rsheeter: Also added a comment

read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
...
 include!("../../generated/generated_varc.rs");
 
+trait RawNth<'a> {
+    fn nth(self, nth: usize) -> Result<&'a [u8], ReadError>;
+}
+
+impl<'a> RawNth<'a> for Option<Result<Index2<'a>, ReadError>> {
+    fn nth(self, nth: usize) -> Result<&'a [u8], ReadError> {
+        let Some(index) = self else {
+            return Err(ReadError::InvalidCollectionIndex(nth as u32));
```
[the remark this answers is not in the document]
rsheeter: Hm. No, not really. We didn't even get to use nth. Will update the error.

read-fonts/src/tables/varc.rs
```diff
@@ -8,21 +8,230 @@ pub use super::{
...
+    pub fn glyph(&self, nth: usize) -> Result<VarcGlyph<'_>, ReadError> {
+        let raw = Some(self.var_composite_glyphs()).nth(nth)?;
+        Ok(VarcGlyph {
+            table: self,
+            data: raw.into(),
+        })
+    }
+}
+
+/// I'm just a happy little blob full of components
```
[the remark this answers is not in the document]
rsheeter: Onit!

read-fonts/src/tables/variations.rs
rsheeter: Good catch, updated to `pub(crate)`

read-fonts/src/tables/variations.rs
rsheeter: I miss having log, I want the equivalent of warn. Maybe debug_assert is a little more firepower than warn ... removed it .

PR #945, opened 2024-06-08
Read the VARC transform

rsheeter:
Reading the decomposed transform is relatively useless so I added a method to get affine matrix. That took me on a wee bit of a core_maths adventure.

Builds on #943.

### DIFF
--- a/read-fonts/src/tables/varc.rs
+++ b/read-fonts/src/tables/varc.rs
@@ -8,23 +8,24 @@ pub use super::{
 
 include!("../../generated/generated_varc.rs");
 
-/// Let's us call self.something().get(i) instead of get(self.something(), i)
-trait Get<'a> {
-    fn get(self, nth: usize) -> Result<&'a [u8], ReadError>;
+trait RawNth<'a> {
+    fn nth(self, nth: usize) -> Result<&'a [u8], ReadError>;
 }
 
-impl<'a> Get<'a> for Option<Result<Index2<'a>, ReadError>> {
-    fn get(self, nth: usize) -> Result<&'a [u8], ReadError> {
-        self.transpose()?
-            .ok_or(ReadError::NullOffset)
-            .and_then(|index| index.get(nth).map_err(|_| ReadError::OutOfBounds))
+impl<'a> RawNth<'a> for Option<Result<Index2<'a>, ReadError>> {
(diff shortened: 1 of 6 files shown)


read-fonts/src/tables/varc.rs
```diff
@@ -152,55 +148,36 @@ impl<'a> VarcComponent<'a> {
...
-        };
-
-        let center_x = if flags.contains(VarcFlags::HAVE_TCENTER_X) {
-            cursor.read::<FWord>()?.to_i16()
-        } else {
-            0
-        };
-        let center_y = if flags.contains(VarcFlags::HAVE_TCENTER_Y) {
-            cursor.read::<FWord>()?.to_i16()
+        let mut transform = DecomposedTransform::default();
```
cmyr, line 151: does it make sense to now just have a constructor on this type that takes the cursor and the flags and parses itself?

read-fonts/src/tables/varc.rs
```diff
@@ -215,26 +192,115 @@ impl<'a> VarcComponent<'a> {
...
+            skew_x: 0.0,
+            skew_y: 0.0,
+            center_x: 0.0,
+            center_y: 0.0,
+        }
+    }
+}
+
+impl DecomposedTransform {
+    /// Convert decomposed form to standard 2x3 matrix form
```
cmyr, line 230: iirc there is some variation in the world in terms of how the fields of an affine matrix are ordered, and I think it's worth being explicit about the field order in the documentation (see [kurbo, for example](https://docs.rs/kurbo/latest/kurbo/struct.Affine.html#method.new))

cmyr (approved):
A couple ideas, nothing blocking

read-fonts/src/tables/varc.rs
rsheeter: Oh good point. I tend to think of kurbo-style (a b are x-basis, c d are y-basis, e f are translation) as standard but it is not at all true.

read-fonts/src/tables/varc.rs
rsheeter: My rationale for the current structure was to keep the bytestream parsing together. I would like to keep that unless you feel strongly about it. If so I will address in a follow-on.

PR #946, opened 2024-06-08
Sketch of alternative to stringify_path

rsheeter:


cmyr (commented):
responded in #937; if we were actually using this type in the generated code I think something like this would make sense (in which case I would propose just erroring at parse time if the provided path did not end in an ident) I think for our purposes it's overkill, since we really just want a string we can include in a diagnostic message.

PR #949, opened 2024-06-08
Read the fields of the multitem variation store

rsheeter:
Read the fields of the MultiItemVarStore. This gives us raw and fairly unfriendly access to VARC and hopefully positions us to start thinking providing friendlier access, such as by loading paths in Skrifa.

### DIFF
--- a/resources/codegen_inputs/varc.rs
+++ b/resources/codegen_inputs/varc.rs
@@ -42,10 +42,10 @@ table SparseVariationRegionList {
 table SparseVariationRegion {
     region_axis_count: u16,
     #[count($region_axis_count)]
-    region_axis_offsets: [Offset32<SparseRegionAxisCoordinates>],
+    region_axis_offsets: [SparseRegionAxisCoordinates],
 }
 
-table SparseRegionAxisCoordinates
+record SparseRegionAxisCoordinates
 {
   axis_index: u16,
   start: F2Dot14,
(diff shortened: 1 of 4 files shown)


resources/codegen_inputs/varc.rs
```diff
@@ -20,8 +22,45 @@ table Varc {
     var_composite_glyphs_offset: Offset32<Index2>,
 }
 
+/// * <https://github.com/fonttools/fonttools/blob/5e6b12d12fa08abafbeb7570f47707fbedf69a45/Lib/fontTools/ttLib/tables/otData.py#L3451-L3457>
+/// * <https://github.com/harfbuzz/harfbuzz/blob/7be12b33e3f07067c159d8f516eb31df58c75876/src/hb-ot-layout-common.hh#L3517-L3520C3>
 table MultiItemVariationStore {
-    // TODO(rsheeter) Doing VARC incrementally, haven't got here yet.
+    #[format = 1]
```
cmyr, line 28: if there's only one format defined you can omit this attribute, since we won't use it anywhere (you can leave it in if  you want, it just generates some extra code we don't need)

rsheeter: I feel like it shoudl have it, it's a format. If it doesn't cause any particular harm perhaps we can keep it.

cmyr: so the only thing that this annotation does, currently, is let this table be used in a 'format enum', which we don't use if there's only one format.


Thinking about this more, though, I think we should be doing more: I think that for any table with a defined format we should actually be checking the format at parse time and returning an error if it isn't the expected value; this would mostly be future-proofing.

cmyr (approved):
looks good! my one comment here is a response about the`#[format = N]` annotation, I'm realizing we should be doing more with this (I'll open an issue)

PR #951, opened 2024-06-11
Fix integer overflow in Cmap12::group()

dfrg:
As discovered by fuzzing.
ref https://oss-fuzz.com/testcase-detail/5141969742397440 and  https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69547

First commit contains the failing test.


rsheeter (approved):
Nit: in both issue and comments I suggest citing https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69547.

PR #953, opened 2024-06-11
A little more fuzzing

rsheeter:
Contributes to #420

### DIFF
--- a/fuzz/fuzz_targets/fuzz_name.rs
+++ b/fuzz/fuzz_targets/fuzz_name.rs
@@ -1,0 +1,22 @@
+#![no_main]
+use std::error::Error;
+
+use libfuzzer_sys::fuzz_target;
+use skrifa::{raw::TableProvider, FontRef};
+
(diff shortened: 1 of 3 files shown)


fuzz/fuzz_targets/fuzz_name.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use libfuzzer_sys::fuzz_target;
+use skrifa::{raw::TableProvider, FontRef};
+
+fn do_name_things(data: &[u8]) -> Result<(), Box<dyn Error>> {
+    let font = FontRef::new(data)?;
+    let name = font.name()?;
+
+    for nr in name.name_record() {
+        let _ = nr.string(name.string_data());
```
dfrg, line 12: Suggest adding `.map(|ns| ns.to_string())` to exercise the string data parsing.

dfrg (approved):

fuzz/fuzz_targets/fuzz_name.rs
rsheeter: Fixed, ty

PR #955, opened 2024-06-16
First crack at fuzzing for Skrifa outlines

rsheeter:
Let's boost https://storage.googleapis.com/oss-fuzz-coverage/fontations/reports/20240615/linux/src/fontations/report.html a little :)

Part of #420.

JMM if happy.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_skrifa_outline.rs
+++ b/fuzz/fuzz_targets/fuzz_skrifa_outline.rs
@@ -1,0 +1,127 @@
+#![no_main]
+use std::{error::Error, fmt::Display};
+
+use libfuzzer_sys::{
+    arbitrary::{self, Arbitrary, Unstructured},
+    fuzz_target,
+};
+use skrifa::{
+    instance::Size,
+    outline::{DrawError, DrawSettings, HintingInstance, HintingMode, OutlinePen},
+    raw::tables::glyf::ToPathStyle,
+    FontRef, MetadataProvider,
+};
+
+/// The pen for when you don't really care what gets drawn
+struct NopPen;
+
+impl OutlinePen for NopPen {
+    fn move_to(&mut self, _x: f32, _y: f32) {
+        // nop
+    }
+
+    fn line_to(&mut self, _x: f32, _y: f32) {
+        // nop
+    }
+
+    fn quad_to(&mut self, _cx0: f32, _cy0: f32, _x: f32, _y: f32) {
+        // nop
+    }
+
+    fn curve_to(&mut self, _cx0: f32, _cy0: f32, _cx1: f32, _cy1: f32, _x: f32, _y: f32) {
+        // nop
+    }
+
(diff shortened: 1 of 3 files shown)


dfrg (approved):
Nice! Looking forward to seeing what this reveals

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+use skrifa::{
+    instance::Size,
+    outline::{DrawError, DrawSettings, HintingInstance, HintingMode, OutlinePen},
+    raw::tables::glyf::ToPathStyle,
+    FontRef, MetadataProvider,
+};
+
+/// The pen for when you don't really care what gets drawn
+struct NopPen;
```
drott, line 16: This might be at some point useful to have in Skrifa itself. I have a NopPen in Skia for some cases as well.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
```diff
@@ -0,0 +1,127 @@
...
+    let location = font.axes().location(raw_location);
+    let mut buf: Vec<u8> = Vec::with_capacity(
+        outline_request
+            .with_memory
+            .then_some(outline_request.memory_size)
+            .unwrap_or_default() as usize,
+    );
+    let hinting_instance = if outline_request.hinted {
+        Some(
+            HintingInstance::new(&outlines, size, &location, HintingMode::default())
```
drott, line 90: Okay with me to have this in a follow-up but HintingMode enum could be cycled by OutlineRequest as well?

drott (commented):
LGTM with one suggestion for hinting mode.

Out of curiosity: How does it work with the different executables? How does the fuzzer iterate over these? Or does it spawn parallel tasks for each of them? Is the same corpus applied to all of them or does each one use/generate/store-back a different traversal corpus?

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
rsheeter: Good catch, will do HintingMode in a followup.

rsheeter:
> How does it work with the different executables? How does the fuzzer iterate over these? Or does it spawn parallel tasks for each of them? Is the same corpus applied to all of them or does each one use/generate/store-back a different traversal corpus?

It uses the same initial corpus (https://github.com/google/oss-fuzz/blob/master/projects/fontations/build.sh), I believe it then evolves (makes guided random derivatives) separately from there but my understanding of the nuances is a bit hand wavy.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
rsheeter: https://github.com/googlefonts/fontations/pull/957

PR #960, opened 2024-06-20
[read-fonts/codegen] Move reading cursor implementation to saturating math

### LINKED ISSUE #959: Panic in sbix header parsing due to overflow when computing end of strikes array
Spun off from [crbug.com/347835680](https://crbug.com/347835680):

https://clusterfuzz.com/testcase-detail/5420281400066048 has more details:

Relevant stack context:

```
#16 0xf2397321 in core::panicking::panic_fmt::h2dc742118ebea367 [third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/panicking.rs:72](https://chromium.googlesource.com/chromium/src/+/2055f9b7d19a9b1d2342eb5818b80d3fab334169/third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/panicking.rs#72):14
    #17 0xf2364317 in core::panicking::panic_const::panic_const_mul_overflow::h11989d665f9cc022 [third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/panicking.rs:179](https://chromium.googlesource.com/chromium/src/+/2055f9b7d19a9b1d2342eb5818b80d3fab334169/third_party/rust-toolchain/lib/rustlib/src/rust/library/core/src/panicking.rs#179):21
    #18 0xee6415f2 in read_fonts::tables::sbix::_$LT$impl$u20$read_fonts..read..FontReadWithArgs$u20$for$u20$read_fonts..table_ref..TableRef$LT$read_fonts..tables..sbix..SbixMarker$GT$$GT$::read_with_args::h41422c524b7432b0 [third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/generated/generated_sbix.rs:359](https://chromium.googlesource.com/chromium/src/+/2055f9b7d19a9b1d2342eb5818b80d3fab334169/third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/generated/generated_sbix.rs#359):39
    #19 0xee64170d in read_fonts::tables::sbix::_$LT$impl$u20$read_fonts..table_ref..TableRef$LT$read_fonts..tables..sbix..SbixMarker$GT$$GT$::read::hd98b1281c207fad4 [third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/generated/generated_sbix.rs:375](https://chromium.googlesource.com/chromium/src/+/2055f9b7d19a9b1d2342eb5818b80d3fab334169/third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/generated/generated_sbix.rs#375):9
    #20 0xee4d5618 in read_fonts::table_provider::TableProvider::sbix::h5a74189d0aa1cb60 [third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/src/table_provider.rs:165](https://chromium.googlesource.com/chromium/src/+/2055f9b7d19a9b1d2342eb5818b80d3fab334169/third_party/rust/chromium_crates_io/vendor/read-fonts-0.19.3/src/table_provider.rs#165):9
```

And the bits of generated code here look like this:

```
impl<'a> FontReadWithArgs<'a> for Sbix<'a> {
    fn read_with_args(data: FontData<'a>, args: &u16) -> Result<Self, ReadError> {
        let num_glyphs = *args;
        let mut cursor = data.cursor();
        cursor.advance::<u16>();
        cursor.advance::<HeaderFlags>();
        let num_strikes: u32 = cursor.read()?;
        ➡️ let strike_offsets_byte_len = num_strikes as usize * Offset32::RAW_BYTE_LEN;
        cursor.advance_by(strike_offsets_byte_len);
        cursor.finish(SbixMarker {
            num_glyphs,
            strike_offsets_byte_len,
        })
    }
}
```

So  `num_strikes as usize * Offset32::RAW_BYTE_LEN;` overflows and is only then passed to `advance_by()` - which has a range check for whether we stay inside the sbix table. 

I see two ways of fixing this. Using `saturating_mul` here in which case the `cursor.advance_by()` would catch the out of bounds cursor, or adding an `advance_by_n_times()` function or something like that which would then catch this as well?

drott:
Fixes #959 and related issues. 

Move the cursor position to a Saturating::<usize> as cursor overflows are real font bugs.
Using wrapping math here would hide out of bounds failures. 

In code generation, using saturating multiplication to avoid an undetected 
32bit build overflow.


dfrg:
Hooray for codegen to swat many of these at once.

I haven’t thought through the full implications but using saturating multiplication seems like a reasonable choice here.

This lgtm but I’d like @cmyr to take a look as well.

read-fonts/src/font_data.rs
```diff
@@ -28,7 +29,7 @@ pub struct FontData<'a> {
 /// call `finish` when you're done to ensure you're in bounds
 #[derive(Debug, Default, Clone, Copy)]
 pub struct Cursor<'a> {
-    pos: usize,
+    pos: Saturating<usize>,
```
cmyr, line 32: First I heard of this type!

font-codegen/src/fields.rs
```diff
@@ -1132,7 +1132,7 @@ impl Field {
                         // Prevent identity-op clippy error with `1 * size`
                         size_expr
                     }
-                    _ => quote!(  #count_expr * #size_expr ),
+                    _ => quote!( (#count_expr).saturating_mul(#size_expr) ),
```
cmyr, line 1135: so the one alternative here that I think we should at least discuss is `checked_mul`, where we will just error immediately on overflow.

The possible issue with `saturating_mul` is that we can end up with an array where the last item is truncated; in this case we will end up failing later anyway (see here) https://github.com/googlefonts/fontations/blob/4e5fde8a3da573467a29ca3d46ca00fddc317e1c/read-fonts/src/font_data.rs#L154

but maybe it makes sense to just fail immediately?

read-fonts/src/font_data.rs
```diff
@@ -232,7 +235,9 @@ impl<'a> Cursor<'a> {
         T: FontReadWithArgs<'a> + ComputeSize,
     {
         let len = len * T::compute_size(args);
-        let temp = self.data.read_with_args(self.pos..self.pos + len, args);
+        let end = self.pos + Saturating(len);
```
cmyr, line 238: do we need to explicitly wrap `len` in `Saturating`? just curious because up above in `advance` we seem to be doing arithmetic without the wrapper?

read-fonts/src/font_data.rs
```diff
@@ -217,7 +218,9 @@ impl<'a> Cursor<'a> {
         T: FontReadWithArgs<'a> + ComputeSize,
     {
         let len = T::compute_size(args);
-        let temp = self.data.read_with_args(self.pos..self.pos + len, args);
+        let end = self.pos + Saturating(len);
```
cmyr, line 221: when we're in these methods that return `Result` my inclination would be to use checked operations and return `Errror::OutOfBounds` directly, instead of continuing with a saturated value.

cmyr (commented):
So the only alternative I could convincingly argue for here would be to use checked operations in the cursor, and convert overflows to errors earlier, where possible.

Some other thoughts:

- would it make sense to store the cursor position as a `u32`, making it explicit that we're working a 32-bit space?
- would it make sense to introduce the invariant that a cursor position of `u32::MAX` indicates invalid data? This would make me more comfortable with the saturating arithmetic, because we could know that anytime we saturated it would be considered an error later. I'm not sure if any potentially 'valid' font file could ever cause the cursor to finish at `u32::MAX` but if so it is an extremely pathological case...

read-fonts/src/font_data.rs
rsheeter: +1

rsheeter:
Is it plausible to use CheckedMul/Add/etc and return errors instead of saturating?

font-codegen/src/fields.rs
drott: Trying this, I found that in many situations doing
`(#count_expr).checked_mul(#size_expr).ok_or(ReadError::OutOfBounds)?` compiles ok, because the embedding function has a signature returning `Result<..., ReadError>` but when `computed_len_expr()` is embedded in `record::ComputeSize::compute_size` this one cannot return a `ReadError` and would need some other type of wrapping. 

I am afraid I am not familiar enough with codegen to upgrade embedding sites of `computed_len_expr()` to consistently handle `OutOfBounds` errors. So perhaps we can file that as an issue as potential future improvement and perhaps you or Chad could consider upgrading codegen sites to handle that better?

dfrg: Good catch. I haven’t dug into it but it doesn’t seem unreasonable to change `compute_size()` to return a `Result`. It would likely require fixing up a few manual impls.

read-fonts/src/font_data.rs
drott: I think sometimes it's required or the alternative is to make sure len is typed `u32` explicitly. AFAICS it is usually a question of type inference and only `u32` -> `Saturating<u32>` is inferred.

read-fonts/src/font_data.rs
drott: Agree.

drott:
> * would it make sense to store the cursor position as a `u32`, making it explicit that we're working a 32-bit space?

Factored that out into a separate change: #965 before updating this PR.

font-codegen/src/fields.rs
cmyr: I've addressed this in #966.

drott:
Closing in favor of #967 and #966.

PR #962, opened 2024-06-20
[skrifa] autohinting scripts/styles

dfrg:
This handles assignment of script styles to each glyph that is covered by cmap.

Unmapped glyphs must be detected with some analysis of the layout tables which will come later.

Basically contains a bunch of script info scraped from FreeType and translated into what is hopefully a nice python data structure for easy updates (see skrifa/generated/autohint_scripts.py). The bottom of that file has the code for generating rust data.


skrifa/src/outline/autohint/script.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use super::style::GlyphStyle;
+
+/// Defines the basic properties for each script supported by the
+/// autohinter.
+#[derive(Clone, Debug)]
+pub(super) struct ScriptClass {
+    pub name: &'static str,
+    /// Unicode tag for the script.
+    pub tag: [u8; 4],
```
cmyr, line 11: `Tag` not good enough for you? 😢

skrifa/generated/autohint_scripts.py
```diff
@@ -0,0 +1,1385 @@
+# Generates Rust tables that define Unicode "script classes" for the purposes
```
cmyr, line 1: I find it confusing that this lives in the `generated` folder; I'd generally expect that to only contain generated files (e.g. it should be safe to blow it away when rebuilding).

Is this hand-written?

skrifa/src/outline/autohint/style.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl GlyphStyle {
+    // The following flags roughly correspond to those defined in FreeType
+    // here: https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/afglobal.h#L76
+    // but with different values because we intend to store "meta style"
+    // information differently.
+    pub const SCRIPT_INDEX_MASK: u16 = 0xFF;
+    pub const UNASSIGNED: u16 = Self::SCRIPT_INDEX_MASK;
+    pub const NON_BASE: u16 = 0x100;
+    pub const DIGIT: u16 = 0x200;
```
cmyr, line 19: are these actually used anywhere outside of this impl block?

cmyr (approved):
looks good!

skrifa/generated/autohint_scripts.py
dfrg: This is a combined of stuff that was very painfully scraped from FreeType along with some hand-written code. Itis intended to be our source of truth forevermore so should hopefully not be blown away.

Moved to `skrifa/scripts/gen_autohint_scripts.py`.

skrifa/src/outline/autohint/style.rs
dfrg: `NON_BASE` is currently used in `script.rs` and there are other passes that may modify the values. This type is very likely to change as I port more code so I'd prefer to keep it as is for now and do some cleanup after I've covered all potential uses.

PR #964, opened 2024-06-22
[read-fonts] common AAT tables

dfrg:
Adds support for lookups and state tables. These are required to handle `morx`, `kerx`, `kern`, and friends.

There are a couple TODOs for binary searching segment lists as well as a missing test for the legacy state machine. I'll fix these before merging.

Sending this now to get the ball rolling and have others check this for sanity. Codegen is somewhat nasty for these types.


font-types/src/raw.rs
```diff
@@ -159,6 +159,10 @@ impl<T: Scalar> FixedSize for BigEndian<T> {
     const RAW_BYTE_LEN: usize = T::RAW_BYTE_LEN;
 }
 
+impl FixedSize for () {
+    const RAW_BYTE_LEN: usize = 0;
```
cmyr, line 163: I'm worried about this; there are various places where we do `bytes.len() / T::RAW_BYTE_LEN`, and we really don't expect this to be zero, even if we're doing checked ops.

I'm sure there will be an interesting explanation coming up, but I want to note my concern up front.


_much later:_

okay, coming back to this cliffhanger after getting through everything... I have a general sense of what you're trying to do, and although I don't have a good enough sense of what the allowed values of a `StateEntry` are, _assuming_ we go with basically the approach you've taken here, I'd like to do it with a custom sealed type instead of with `()`.

Basically: this is okay as a hack as long as it is an invariant that the zero-sized `FixedSize` type is only ever used as a component of some other type, and has no public constructors.

as a first sketch, I might try:

```rust
/// An empty type for use exclusively in `StateEntry`.
// NOTE: the inner field is private, and this type cannot be constructed outside of this module
pub struct NoPayload(());
```

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+include!("../../generated/generated_aat.rs");
+
+/// Predefined classes.
+///
+/// See <https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6Tables.html>
+pub mod class {
+    pub const END_OF_TEXT: u8 = 0;
+    pub const OUT_OF_BOUNDS: u8 = 1;
+    pub const DELETED_GLYPH: u8 = 2;
+}
```
cmyr, line 14: Would an enum make more sense? Are we going to handle the user class codes?

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+    lookup: Lookup<'a>,
+    _marker: std::marker::PhantomData<fn() -> T>,
+}
+
+impl<'a, T: LookupValue> TypedLookup<'a, T> {
+    /// Returns the value associated with the given index.
+    pub fn value(&self, index: u16) -> Result<T, ReadError> {
+        match &self.lookup {
+            Lookup::Format0(lookup) => {
+                let data = lookup.values_data();
```
cmyr, line 78: I'd be tempted to have a separate `value` method for each lookup type, and then this method would just forward calls? a bit more readable, and I often find that for some test or something I end up instantiating a specific format and want to be able to call methods on it..

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+impl<'a> Lookup6<'a> {
+    fn entries<T: LookupValue>(&self) -> Result<&[LookupSingle<T>], ReadError> {
+        FontData::new(self.entries_data())
+            .cursor()
+            .read_array(self.n_units() as usize)
+    }
+}
+
+impl<T: LookupValue> FixedSize for LookupSingle<T> {
+    const RAW_BYTE_LEN: usize = std::mem::size_of::<Self>();
```
cmyr, line 65: this is fine as long as we're always `packed`, so I might document somewhere that that's important? Or maybe it's enforced by `bytemuck::AnyBitPattern`?

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+pub type LookupGlyphId<'a> = TypedLookup<'a, GlyphId>;
+
+/// Entry in an (extended) state table.
+pub struct StateEntry<T = ()> {
+    /// Index of the next state.
+    pub new_state: u16,
+    /// Flag values are table specific.
+    pub flags: u16,
+    /// Payload is table specific.
+    pub payload: T,
```
cmyr, line 228: I'd like to know more around what the allowed values are, what the invariants are, etc. e.g,

- for a given table, is `T` always identical? At what point do we know the value?
- what are the possible values for `T`, overall?


This would help me make a more informed assessment of whether or not this is the best way of expressing what we want.

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+        })
+    }
+}
+
+impl<T> FixedSize for StateEntry<T>
+where
+    T: FixedSize,
+{
+    // Two u16 fields + payload
+    const RAW_BYTE_LEN: usize = 4 + T::RAW_BYTE_LEN;
```
cmyr, line 251: just for maximum explicitness I'd write,

```suggestion
    const RAW_BYTE_LEN: usize = u16::RAW_BYTE_LEN + u16::RAW_BYTE_LEN + T::RAW_BYTE_LEN;
```

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+            .entry_table()?
+            .data()
+            .get(entry_offset..)
+            .ok_or(ReadError::OutOfBounds)?;
+        let mut entry = StateEntry::read(FontData::new(entry_data))?;
+        // For legacy state tables, the newState is a byte offset into
+        // the state array. Convert this to an index for consistency.
+        let new_state = ((entry.new_state as i32)
+            - (self.header.state_array_offset().to_u32() as i32))
+            / n_classes as i32;
```
cmyr, line 297: I'd be defensive with these operations, looks like fuzzer bait.

cmyr (commented):
I see what you mean, this is definitely a bit hairy. I've noted some possible concerns inline, but I think we can get a version of this in without too much noodling.

font-types/src/raw.rs
dfrg: I was wondering if an impl of FixedSize for a zero-sized type might bite us. I’ll try the private type as you suggest.

The payload is going to be one of _nothing_, a u16 or a pair of u16s (I think there’s an example of this one in the tests). 

Looking forward, state tables will always be embedded directly in other tables so we won’t be able to use codegen anyway. I might just eschew the FontRead impl and add a private method so we don’t have to worry about the implications of these in the public API.

read-fonts/src/tables/aat.rs
dfrg: These are predefined classes that are meant to be provided by the shaping engine. Each AAT (sub)table can provide its own glyph classes (through lookup tables) in addition to these so the set is unbounded.

read-fonts/src/tables/aat.rs
dfrg: I waffled a bit on this because the type of the lookup is dependent on the containing table so the value methods on each variant would need a type parameter.

I don’t have strong feelings on this though and agree with your points so I’ll split them out.

dfrg: Done

read-fonts/src/tables/aat.rs
dfrg: `AnyBitPattern` doesn't guarantee this but these types are marked `repr(packed)` and I added a note about the requirement

read-fonts/src/tables/aat.rs
dfrg: Good call. Changed to `checked_sub` and added an additional check for divide by zero

read-fonts/src/tables/aat.rs
```diff
@@ -0,0 +1,619 @@
...
+
+    /// Returns the entry for the given state and class.
+    pub fn entry(&self, state: u16, class: u16) -> Result<StateEntry<T>, ReadError> {
+        let n_classes = self.header.n_classes() as usize;
+        let mut class = class as usize;
+        if class >= n_classes {
+            class = class::OUT_OF_BOUNDS as usize;
+        }
+        let state_array = self.header.state_array()?.data();
+        let state_ix = state as usize * n_classes + class;
```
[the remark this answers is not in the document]
dfrg: Might as well be defensive. Changed to `checked_mul`. If the compiler is smart, this is probably a nop on 64-bit

read-fonts/src/tables/aat.rs
dfrg: I think I answered this above but I can elaborate if not

read-fonts/src/tables/aat.rs
```diff
@@ -218,8 +250,15 @@ pub type LookupU16<'a> = TypedLookup<'a, u16>;
 pub type LookupU32<'a> = TypedLookup<'a, u32>;
 pub type LookupGlyphId<'a> = TypedLookup<'a, GlyphId>;
 
+#[derive(Copy, Clone, bytemuck::AnyBitPattern)]
+pub struct NoPayload(());
```
cmyr, line 254: I'd like to at least document why this exists, we're going to have at least one confused reader somewhere down the road

cmyr (approved):
looks good, one last little nit but overall happy to see this merged!

PR #965, opened 2024-06-24
[read-fonts] Move font data cursor implementation to u32

drott:
Make sure that we can't advance the cursor or read further than u32 boundaries. Preparation for #960.

### DIFF
--- a/read-fonts/src/font_data.rs
+++ b/read-fonts/src/font_data.rs
@@ -1,5 +1,6 @@
 //! raw font bytes
 
+use core::num::Saturating;
 use std::ops::{Range, RangeBounds};
 
 use bytemuck::AnyBitPattern;
@@ -28,7 +29,7 @@ pub struct FontData<'a> {
 /// call `finish` when you're done to ensure you're in bounds
 #[derive(Debug, Default, Clone, Copy)]
 pub struct Cursor<'a> {
-    pos: usize,
+    pos: Saturating<u32>,
     data: FontData<'a>,
 }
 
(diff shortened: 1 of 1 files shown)


read-fonts/src/font_data.rs
```diff
@@ -172,11 +183,13 @@ impl<'a> FontData<'a> {
...
     pub(crate) fn advance<T: Scalar>(&mut self) {
-        self.pos += T::RAW_BYTE_LEN
+        let advance_bytes = Saturating(T::RAW_BYTE_LEN.try_into().unwrap_or(u32::MAX));
+        self.pos += advance_bytes;
     }
 
     pub(crate) fn advance_by(&mut self, n_bytes: usize) {
-        self.pos += n_bytes;
+        let advance_bytes = Saturating(n_bytes.try_into().unwrap_or(u32::MAX));
+        self.pos += advance_bytes;
```
cmyr, line 192: hmm, okay... the original motivation for this design was that I wanted to reduce the number of bounds checks and branching during the initial parsing phase, so if this is actually generating code that includes a bounds check that doesn't feel great.

- That said, we don't have actual data to inform this decision, which would be nice.
- in the `advance` case, given that `RAW_BYTE_LEN` is `const` and the fn is generic, I would expect the compiler to elide the bounds check.
- we could probably make `RAW_BYTE_LEN` a u32 as well, if that is helpful?
- overall my main concern with this patch is that it makes it harder to understand the code that `rustc`/`llvm` are going to end up generating for us here..

drott:
Closing in favor of #967 and #966.

PR #966, opened 2024-06-24
[read] Make ComputeSize fallible

cmyr:
This will ensure we do not panic on overflow when handling pathological inputs.


rsheeter (approved):

rsheeter:
Testable?

cmyr:
> Testable?

This is only changing a method signature, there's no implementation to test. On top of this we'll want to update codegen to use checked operations in the generated impls, which would be more amenable to testing.

PR #967, opened 2024-06-24
[read] Remove & deny overflow in key modules

cmyr:
based on #966.

This enables the `clippy::arithmetic_side_effects` lint in the `read`, `font_data` and `array` modules, which are the most obvious places in read-fonts where overflow could occur, and it reworks all arithmetic in those modules to be checked.

This collides a bit with #960, but I think this is a reasonably clean and unobtrusive way to ensure that we're handling arithmetic edge cases in these files.

@drott does this work for you?

### DIFF
--- a/read-fonts/src/array.rs
+++ b/read-fonts/src/array.rs
@@ -1,5 +1,7 @@
 //! Custom array types
 
+#![deny(clippy::arithmetic_side_effects)]
+
 use bytemuck::AnyBitPattern;
 use font_types::FixedSize;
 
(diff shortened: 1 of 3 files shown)


rsheeter (approved):
This LGTM, @drott wdyt?

drott (approved):
LGTM, thanks. I tested that with the sbix fuzzer test and it produces an OutOfBounds error, which is good.

drott:
Just to confirm - we're no longer aspiring to move cursor to u32?

drott:
> Just to confirm - we're no longer aspiring to move cursor to u32?

Ok, saw your comment in the chat - which answered this question.

drott (commented):
You might want to add a reference to #959 in the commit message.

drott:
#969 adds the sbix test case from fuzzing which passes with this here landed.

drott:
From working on #970 it seems that the clippy deny directive does not necessarily capture overflows for our `fixed_impl` types. When I add that directive, the fvar normalization still compiles and the directive does not capture the potential overflow.

cmyr:
> From working on #970 it seems that the clippy deny directive does not necessarily capture overflows for our `fixed_impl` types. When I add that directive, the fvar normalization still compiles and the directive does not capture the potential overflow.

good note, I'll fix that in a followup (opened #971 to track)

PR #968, opened 2024-06-25
[int-set] Update the sparse bit set encoding to detect and output encoding with filled nodes.

garretrieger:


int-set/src/sparse_bit_set.rs
```diff
@@ -361,13 +452,72 @@ mod test {
...
+    #[test]
+    fn encode_two_level_filled() {
+        let actual_bytes = to_sparse_bit_set_with_bf::<8>(&(3..=21).collect());
+        let expected_bytes = [0b0_00010_10, 0b00000111, 0b11111000, 0b00000000, 0b00111111];
+        assert_eq!(actual_bytes, expected_bytes);
+    }
+
+    #[test]
+    fn encode_two_level_not_filled() {
+        let actual_bytes = to_sparse_bit_set_with_bf::<4>(&[0, 4, 8, 12].iter().copied().collect());
```
cmyr, line 490: totally academic but you can do `into_iter()` instead of `iter().copied()`, here.

cmyr (approved):
looks good!

int-set/src/sparse_bit_set.rs
garretrieger: Sounds good, sending another set PR later I'll fix it there.

PR #969, opened 2024-06-25
[read-fonts] Add test case for SBIX overflow

drott:
Add test case from crbug.com/347835680 which triggered overflows in font data reading, which are fixed after #966.

Test for #959. 

Test fails only on 32 bit, for example building and testing with: `--target=i686-unknown-linux-gnu`:
`cargo test --target=i686-unknown-linux-gnu "sbix_strikes_count_overflow"`

### DIFF
--- a/read-fonts/src/tables/sbix.rs
+++ b/read-fonts/src/tables/sbix.rs
@@ -22,3 +22,25 @@ impl<'a> Strike<'a> {
         Ok(Some(GlyphData::read(data)?))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::tables::sbix::Sbix;
+    use crate::test_helpers::BeBuffer;
+
+    #[test]
+    fn sbix_strikes_count_overflow_table() {
+        // Contains an invalid `num_strikes` values which would move the cursor outside the able.
+        // See https://issues.chromium.org/issues/347835680 for the ClusterFuzz report.
+        // Failure only reproduces on 32-bit, for example, run with:
+        // cargo test --target=i686-unknown-linux-gnu "sbix_strikes_count_overflow_table"
+        let sbix = BeBuffer::new()
+            .push(1u16) // version
+            .push(0u16) // flags
+            .push(u32::MAX); // num_strikes
+
+        let table = Sbix::read(sbix.font_data(), 5);
+        // Must not panic with "attempt to multiply with overflow".
+        assert!(table.is_err());
+    }
+}


cmyr:
Shouldn't this be failing right now, since it doesn't include #966?

drott:
> edit: oh, will this only overflow on a 32-bit arch?

Yes, should have mentioned that here, too.
`cargo test --target=i686-unknown-linux-gnu "sbix_strikes_count_overflow"`
as I mentioned in #959.

rsheeter (approved):

read-fonts/src/tables/sbix.rs
```diff
@@ -22,3 +22,41 @@ impl<'a> Strike<'a> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::{FontRef, TableProvider};
+
+    #[test]
+    fn sbix_strikes_count_overflow() {
+        // Contains an invalid `num_strikes` values which would move the cursor outside the able.
+        let test_case = &[
```
cmyr, line 33: just for informational purposes, there is a `test_helpers` module in this crate that contains a `BeBuffer` type that is intended for creating intelligible test cases for this sort of scenario; using that we could write a version of this that looked like,

```rust
#[test]
fn sbix_strikes_count_overflow() {
    // Contains an invalid `num_strikes` values which would move the cursor outside the able.
    let sbix = BeBuffer::new()
        .push(1u16) // version
        .push(0u16) // flags
        .push(u32::MAX);

    let table = Sbix::read(sbix.font_data(), 5);
    table.unwrap();
}
```

which makes it easier for the reader to understand the problem.

(I haven't checked that this example actually fails)

cmyr (approved):
mildly annoying that this doesn't currently fail in CI, but agree it's worth capturing and we can try to figure out an appropriate CI runner later.

read-fonts/src/tables/sbix.rs
drott: Thanks for the suggestion and distilling the test case - indeed, this works (same, on 32-bit) only, but is indeed a smaller and more legible way to express that. I was expecting to run into trouble trying to make other tables align with sbix for this to trigger but using the `read_with_args` entry point makes this easier, thanks for demoing that.

drott:
I am going to go ahead and merge the test case in the form that Colin suggested, more compact. Hope that is not a too generous interpretation of your LGTMs. Otherwise please speak up.

PR #970, opened 2024-06-25
[read-fonts] Avoid overflow in axis normalization

### LINKED ISSUE #958: Panic in `fixed_impl` triggered by `VariationAxisRecord::normalize`
https://oss-fuzz.com/testcase-detail/5278123998904320 found a panic in `VariationAxisRecord::normalize` - but this issue got marked as fixed because presumably a subsequent fuzzer logic change used the entropy differently and did not reach this code anymore.

It is however still reproducible with:
```
$ git revert 4e5fde8a3da573467a29ca3d46ca00fddc317e1c
$ target/debug/fuzz_skrifa_outline clusterfuzz-testcase-minimized-fuzz_skrifa_outline-5278123998904320
```

```
thread '<unnamed>' panicked at /Users/drott/dev/fontations/font-types/src/fixed.rs:279:1:
attempt to subtract with overflow
```

Not sure if and how we want to fix this? Would it make sense to use saturating arithmetic here?

drott:
Fixes #958.

Do not overflow in scaling computations in axis normalization by using saturating math.

### DIFF
--- a/read-fonts/src/tables/fvar.rs
+++ b/read-fonts/src/tables/fvar.rs
@@ -29,8 +29,12 @@ impl VariationAxisRecord {
         let max_value = self.max_value().max(min_value);
         value = value.clamp(min_value, max_value);
         value = match value.cmp(&default_value) {
-            Less => -((default_value - value) / (default_value - min_value)),
-            Greater => (value - default_value) / (max_value - default_value),
+            Less => {
+                -((default_value.saturating_sub(value)) / (default_value.saturating_sub(min_value)))
+            }
+            Greater => {
+                (value.saturating_sub(default_value)) / (max_value.saturating_sub(default_value))
+            }
             Equal => Fixed::ZERO,
         };
         value.clamp(-Fixed::ONE, Fixed::ONE)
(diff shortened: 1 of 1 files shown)


drott:
Following up from the conversation in:
https://github.com/googlefonts/fontations/issues/958#issuecomment-2178836916

I don't think we need to match FreeType in what I would call inappropriate error handling. If FT wraps around I see that mostly as their failure to detect the overflow and their lack of tools for saturating math in  their codebase. Skia would usually use saturating functions for this sort of situation, I believe.

The font in question, as produced by the fuzzer, has the following:
```
min: -26335.87451171875 
def 8224.12548828125 
max 8224.12548828125
```

The value to normalize is: `0.0`using the formula:
`-((default_value - value) / (default_value - min_value))`

In which `default-min` then overflows:
`default - min: saturated sub 32767.99998474121 wrapped sub: -30976.0`

I don't think that axis position would be meaningful and IMO the code would not look better structured if we introduce additional clamping after wrapping compared to using saturating math. Opinions?

read-fonts/src/tables/fvar.rs
```diff
@@ -92,4 +96,23 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn normalize_overflow() {
```
dfrg, line 101: We’ve been adding a comment with links to the oss-fuzz and chromium bugs to tests designed to capture them.

dfrg (approved):
When the values are not ordered or are defined such that the normalization calculation will overflow, the font is effectively broken so agreed that we shouldn’t be obligated to match FreeType in these cases. I’m good with using saturating arithmetic here.

read-fonts/src/tables/fvar.rs
drott: Link added.

PR #972, opened 2024-06-25
[read/codegen] Use checked ops in length calculations

cmyr:
following on from #966, this updates all of the codegen'd length calculations to use checked math.

So to the best of my understanding these overflows would only be possible on 32-bit architectures, so I'm not sure how best to test this? @drott added a test case in #969, but that test passes everywhere we run it. If we wanted to test this in CI we would need to either use a custom build machine, or _maybe_ we could have `wasm32` as a target?

### DIFF
--- a/font-codegen/src/record.rs
+++ b/font-codegen/src/record.rs
@@ -79,10 +79,22 @@ fn generate_read_with_args(item: &Record) -> TokenStream {
     let args = item.attrs.read_args.as_ref().unwrap();
     let args_type = args.args_type();
     let destructure_pattern = args.destructure_pattern();
-    let field_size_expr = item.fields.iter().map(Field::record_len_expr);
+    let field_size_expr: Vec<_> = item.fields.iter().map(Field::record_len_expr).collect();
     let field_inits = item.fields.iter().map(Field::record_init_stmt);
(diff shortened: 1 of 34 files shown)


font-codegen/src/record.rs
```diff
@@ -79,20 +79,33 @@ fn generate_read_with_args(item: &Record) -> TokenStream {
...
     let constructor_args = args.constructor_args();
     let args_from_constructor_args = args.read_args_from_constructor_args();
 
+    let comp_size_expr = match field_size_expr.as_slice() {
+        [] => panic!("should never be empty"),
+        [one_expr] => quote!( Ok(#one_expr) ),
+        exprs => {
+            quote! {
+                let mut result = 0usize;
+                #( result = result.checked_add(#exprs).ok_or(ReadError::OutOfBounds)?; )*
```
drott, line 93: 👍  Looking at this change, indeed, I think _this_ is the spot I wasn't sure about how to rephrase due to the repetition operator at the end. Thanks.

drott (approved):
LGTM. FWIW the fuzzer will also verify the issue as fixed with this change, which might serve as a confirmation this works.

PR #973, opened 2024-06-26
Try to play nicer with the seed corpus

rsheeter:
The `data` provided to the fuzz target is likely to look like a font. If some fool cut the front off it would become unlikely to do so. I suspect this is why fuzzing outlines has near zero coverage:

image

https://storage.googleapis.com/oss-fuzz-coverage/fontations/reports/20240625/linux/src/fontations/skrifa/src/outline/report.html.

What if instead we try a crossproduct of likely and random options against the fuzzer provided data?

Contributes to #420.

JMM if happy

### DIFF
--- a/fuzz/fuzz_targets/fuzz_skrifa_outline.rs
+++ b/fuzz/fuzz_targets/fuzz_skrifa_outline.rs
@@ -162,23 +162,17 @@ fn hinting_modes(hinted: bool) -> Vec<HintingMode> {
     ]
 }
 
-fuzz_target!(|data: &[u8]| {
-    // data from corpus is likely to be a font. If we chope off the head to make an outline request
-    // it is likely data is no longer a font. So, take the cross product of likely values for various options
-    // If a lot of values are possible choose randomly with rng seeded from data to ensure reproducible results.
-    let Ok(font) = FontRef::new(data) else {
-        return;
-    };
-    let mut requests = vec![OutlineRequest::default()];
-    requests = requests
+// To avoid consuming corpus data that is likely a font just build the cross product of likely values
+// for various options
+fn create_request_scenarios(font: &FontRef) -> Vec<OutlineRequest> {
+    fuzz_sizes()
         .into_iter()
-        .flat_map(|r| {
-            fuzz_sizes()
-                .into_iter()
-                .map(move |size| OutlineRequest { size, ..r.clone() })
+        .map(move |size| OutlineRequest {
+            size,
+            ..Default::default()
         })
         .flat_map(|r| {
-            fuzz_locations(&font)
+            fuzz_locations(font)
                 .into_iter()
                 .map(move |location| OutlineRequest {
                     location,
@@ -219,9 +213,14 @@ fuzz_target!(|data: &[u8]| {
                 ..r.clone()
             })
         })
-        .collect();
+        .collect::<Vec<_>>()
+}
 
-    for request in requests {
+fuzz_target!(|data: &[u8]| {
+    let Ok(font) = FontRef::new(data) else {
+        return;
+    };
+    for request in create_request_scenarios(&font) {
         let _ = do_glyf_things(request, data);
     }
 });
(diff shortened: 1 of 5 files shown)


drott:
I am definitely in favor of interpreting the fuzzer data as a font immediately and not doing the truncation at the beginning. This is also useful for debugging. But I think generating random data from the fuzzer input data very likely messes with the coverage guiding of the fuzzer. The fuzzing library can no longer associate a sample/corpus mutation operation with a code coverage change, which probably slows down fuzzing coverage growth. How about taking the input for mutation OutlineRequest etc. from somewhere later in the file?

cmyr:
Thinking out loud: what if our corpus included explicit 'header' bytes in this case that will always correspond to the `OutlineRequest`? Then the fuzzer could mutate those bytes to explore the space of possible requests in isolation?

It makes sense to me that having changes in the input data having side effects will confuse the coverage guidance, since a small change in the input data will cause a completely different set of options to be selected, which could drastically effect what code gets run..

rsheeter:
Per IM I will try guing extra bytes on the end to feed the outline request.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
```diff
@@ -37,55 +38,39 @@ impl OutlinePen for NopPen {
...
+    hinting_mode: HintingMode,
     with_memory: bool, // ~half tests should be with memory, half not
     memory_size: u16, // if we do test with_memory, how much of it? u16 to avoid asking for huge chunks.
     harfbuzz_pathstyle: bool,
 }
 
+impl Default for OutlineRequest {
+    fn default() -> Self {
+        Self {
+            size: Size::unscaled(),
```
cmyr, line 62: not really for this PR, but should `Size` derive `Debug`?

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
```diff
@@ -154,11 +129,99 @@ fn do_glyf_things(outline_request: OutlineRequest, data: &[u8]) -> Result<(), Bo
...
-    let data = unstructured.take_rest();
-    let _ = do_glyf_things(outline_request, data);
+    let mut requests = vec![OutlineRequest::default()];
+    requests = requests
+        .into_iter()
+        .flat_map(|r| {
+            fuzz_sizes()
+                .into_iter()
+                .map(move |size| OutlineRequest { size, ..r.clone() })
+        })
```
cmyr, line 179: I find this all very hard to follow, but I think at the very least you could start with,
```suggestion
    let requests = fuzz_sizes()
        .into_iter().map(move |size| OutlineRequest { size, ..Default::default() })
    })
```

and save a few lines?


it would also be a bit easier to read if it was in its own little fn.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
```diff
@@ -154,11 +129,99 @@ fn do_glyf_things(outline_request: OutlineRequest, data: &[u8]) -> Result<(), Bo
...
+            }
+            .into_iter()
+            .map(move |memory_size| OutlineRequest {
+                memory_size,
+                ..r.clone()
+            })
+        })
+        .collect();
+
+    for request in requests {
```
cmyr, line 224: I've lost track, but how many different requests are we testing? I don't suppose we're worried about long runtime?

cmyr (approved):
looks like forward progress, curious to see if this gets the coverage ticking upwards.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
rsheeter: I believe 96 * the location cross. I suppose when the fuzzer develops very large fvar tables that might be a problem lol.

fuzz/fuzz_targets/fuzz_skrifa_outline.rs
rsheeter: Migrated to fn with a comment explaining wht's happening

PR #981, opened 2024-06-28
Adjust charmap fuzzer based on coverage

rsheeter:
Ref https://storage.googleapis.com/oss-fuzz-coverage/fontations/reports/20240627/linux/src/fontations/skrifa/src/charmap.rs.html (from https://storage.googleapis.com/oss-fuzz-coverage/fontations/reports/20240627/linux/src/fontations/skrifa/src/report.html)

Contributes to #420. JMM.


dfrg (approved):
I expect this one to produce some overflows…

PR #983, opened 2024-06-28
[klippa] populate_unicodes and closure glyphs for cmap, composite glyphs

qxliu76:
populate_gids is incomplete yet, we need to do closure for COLR and layout tables later.

### DIFF
--- a/klippa/src/lib.rs
+++ b/klippa/src/lib.rs
@@ -7,67 +7,204 @@ mod parsing_util;
 pub use parsing_util::{parse_unicodes, populate_gids};
 
 use int_set::IntSet;
-use std::collections::BTreeSet;
+use skrifa::MetadataProvider;
+use std::collections::{BTreeSet, HashMap};
 use std::path::PathBuf;
 use thiserror::Error;
-use write_fonts::read::{FontRef, TableProvider, TopLevelTable};
+use write_fonts::read::{
+    tables::glyf::{Glyf, Glyph},
+    tables::loca::Loca,
+    FontRef, TableProvider, TopLevelTable,
+};
 use write_fonts::types::GlyphId;
 use write_fonts::types::Tag;
 use write_fonts::{
     from_obj::FromTableRef, tables::hhea::Hhea, tables::hmtx::Hmtx, tables::maxp::Maxp, FontBuilder,
 };
+
+const MAX_COMPOSITE_OPERATIONS_PER_GLYPH: u8 = 64;
+const MAX_NESTING_LEVEL: u8 = 64;
+const MAX_GID: GlyphId = GlyphId::new(u16::MAX);
+
+#[allow(dead_code)]
+#[derive(Default)]
 pub struct Plan {
-    glyph_ids: IntSet<GlyphId>,
+    unicodes: IntSet<u32>,
+    glyphset_gsub: IntSet<GlyphId>,
+    glyphset: IntSet<GlyphId>,
     num_h_metrics: u16,
     num_output_glyphs: u16,
+    font_num_glyphs: usize,
+    unicode_to_new_gid_list: Vec<(u32, GlyphId)>,
+    codepoint_to_glyph: HashMap<u32, GlyphId>,
 }
 
 impl Plan {
     pub fn new(input_gids: &IntSet<GlyphId>, input_unicodes: &IntSet<u32>, font: &FontRef) -> Self {
-        let gids = populate_unicodes_to_retain(input_gids, input_unicodes, font);
-        // remove invalid gids
-        let maxp = font.maxp().expect("Error reading maxp table");
-        let maxp = Maxp::from_table_ref(&maxp);
-        let gids: IntSet<GlyphId> = gids
-            .iter()
-            .filter(|gid| gid.to_u16() < maxp.num_glyphs)
-            .collect();
-        let num_glyphs = gids.len() as u16;
+        let mut this = Plan {
+            font_num_glyphs: get_font_num_glyphs(font),
+            ..Default::default()
+        };
+
+        this.populate_unicodes_to_retain(input_gids, input_unicodes, font);
+        this.populate_gids_to_retain(font);
+        this.num_output_glyphs = this.glyphset.len() as u16;
 
         // compute new h_metrics
         let hmtx = font.hmtx().expect("Error reading hmtx table");
         let hmtx = Hmtx::from_table_ref(&hmtx);
-        let new_h_metrics = compute_new_num_h_metrics(&hmtx, &gids);
+        this.num_h_metrics = compute_new_num_h_metrics(&hmtx, &this.glyphset);
 
-        Self {
-            glyph_ids: gids,
-            num_h_metrics: new_h_metrics,
-            num_output_glyphs: num_glyphs,
+        this
+    }
+
+    pub fn populate_unicodes_to_retain(
+        &mut self,
+        input_gids: &IntSet<GlyphId>,
+        input_unicodes: &IntSet<u32>,
+        font: &FontRef,
+    ) {
+        let charmap = font.charmap();
+        if input_gids.is_empty() && input_unicodes.len() < self.font_num_glyphs {
+            let cap = input_unicodes.len();
+            self.unicode_to_new_gid_list.reserve(cap);
+            self.codepoint_to_glyph.reserve(cap);
+            //TODO: add support for subset accelerator?
+
+            for cp in input_unicodes.iter() {
+                match charmap.map(cp) {
+                    Some(gid) => {
+                        self.codepoint_to_glyph.insert(cp, gid);
+                        self.unicode_to_new_gid_list.push((cp, gid));
+                    }
+                    None => {
+                        continue;
+                    }
+                }
+            }
+        } else {
+            //TODO: add support for subset accelerator?
+            let cmap_unicodes = charmap.mappings().map(|t| t.0).collect::<IntSet<u32>>();
+            let unicode_gid_map = charmap.mappings().collect::<HashMap<u32, GlyphId>>();
+
+            let vec_cap = input_gids.len() + input_unicodes.len();
+            let vec_cap = vec_cap.min(cmap_unicodes.len());
+            self.codepoint_to_glyph.reserve(vec_cap);
+            self.unicode_to_new_gid_list.reserve(vec_cap);
+            //TODO: possible micro-optimize: set iteration over ranges next_range()? getting ranges is faster in Harfbuzz int set
+            for cp in cmap_unicodes.iter() {
+                match unicode_gid_map.get(&cp) {
+                    Some(gid) => {
+                        if !input_gids.contains(*gid) && !input_unicodes.contains(cp) {
+                            continue;
+                        }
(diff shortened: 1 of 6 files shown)


klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+    glyphset: IntSet<GlyphId>,
     num_h_metrics: u16,
     num_output_glyphs: u16,
+    font_num_glyphs: usize,
+    unicode_to_new_gid_list: Vec<(u32, GlyphId)>,
+    codepoint_to_glyph: HashMap<u32, GlyphId>,
 }
 
 impl Plan {
+    pub fn empty() -> Self {
```
cmyr, line 42: can we derive `Default` for this type?

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+    operation_count
+}
+
+fn remove_invalid_gids(gids: &mut IntSet<GlyphId>, num_glyphs: usize) {
+    gids.remove_range(GlyphId::new(num_glyphs as u16)..=GlyphId::new(CODEPOINT_INVALID as u16));
+}
+
+fn get_font_num_glyphs(font: &FontRef) -> usize {
+    let mut ret = 0;
+    let loca = font.loca(None).expect("Error reading loca table");
```
cmyr, line 215: do we want to panic if these tables are missing, or should we return an error? (here and elsewhere)

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
-    gids
+    operation_count
+}
+
+fn remove_invalid_gids(gids: &mut IntSet<GlyphId>, num_glyphs: usize) {
+    gids.remove_range(GlyphId::new(num_glyphs as u16)..=GlyphId::new(CODEPOINT_INVALID as u16));
+}
+
+fn get_font_num_glyphs(font: &FontRef) -> usize {
+    let mut ret = 0;
```
cmyr, line 214: can we just remove this, and do `let ret = local.len()` below?

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
-    let mut gids = IntSet::empty();
-    gids.insert(GlyphId::new(0_u16));
-    gids.extend(input_gids.iter());
-
-    let cmap = font.cmap().expect("Error reading cmap table");
-    for cp in input_unicodes.iter() {
-        match cmap.map_codepoint(cp) {
-            Some(gid) => {
-                gids.insert(gid);
+    pub fn populate_unicodes_to_retain(
```
cmyr, line 72: this shouldn't need to be public?

Also my instinct is that this would make more sense as a free function that returns `(IntSet<GlyphId>, IntSet<u32>)`; that should be easier to write tests against in isolation, and is easier to follow the control flow of (since it's a pure function instead of relying on side effects)

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+        &mut self,
+        input_gids: &IntSet<GlyphId>,
+        input_unicodes: &IntSet<u32>,
+        font: &FontRef,
+    ) {
+        let charmap = font.charmap();
+        if input_gids.is_empty() && input_unicodes.len() < self.font_num_glyphs {
```
cmyr, line 79: Just curious, but do we generally expect to get both gids and unicodes as input, or is it normally one or the other?

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+            for cp in input_unicodes.iter() {
+                match charmap.map(cp) {
+                    Some(gid) => {
+                        self.codepoint_to_glyph.insert(cp, gid);
+                        self.unicode_to_new_gid_list.push((cp, gid));
+                    }
+                    None => {
+                        continue;
+                    }
+                }
```
cmyr, line 94: This is fine, but you could also write it as,

```suggestion
            for cp in input_unicodes.iter().flat_map(|cp| charmap.map(cp)) {
                self.codepoint_to_glyph.insert(cp, gid);
                self.unicode_to_new_gid_list.push((cp, gid));
```

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
             }
-            None => {
-                continue;
+        } else {
+            //TODO: add support for subset accelerator?
+            let charmap = font.charmap();
```
cmyr, line 98: I think this is redundant? You declare the same binding at the start of this function and I think it's still in scope.

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+            let cmap_unicodes = charmap.mappings().map(|t| t.0).collect::<IntSet<u32>>();
+            let unicode_gid_map = charmap.mappings().collect::<HashMap<u32, GlyphId>>();
+
+            let mut vec_cap = input_gids.len() + input_unicodes.len();
+            vec_cap = vec_cap.min(cmap_unicodes.len());
```
cmyr, line 103: this is a huge nit from me, so please forgive me, but: I find in general if you can write something in such a way as to avoid needing mutation it's worth doing so, unless it's much more awkward.

The basic rationale is that when reading code, anytime you see `mut` you suddenly need to keep track of the fact that this value might be changed at any point from there forwards, which is an extra burden to understanding.

In this particular case, there's a very simple alternative which is common in rust (but is sometimes discouraged in other languages) and that is to reuse a binding. Specifically, you can write

```suggestion
            let vec_cap = input_gids.len() + input_unicodes.len();
            let vec_cap = vec_cap.min(cmap_unicodes.len());
```

where the second binding is shadowing the first.

There's a good section on this [in the Rust docs](https://doc.rust-lang.org/1.79.0/book/ch03-01-variables-and-mutability.html#shadowing).

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+                    }
+                }
+            }
+
+            /* Add gids which where requested, but not mapped in cmap */
+            //TODO: possible micro-optimize: set iteration over ranges next_range()? getting ranges is faster in Harfbuzz int set
+            for gid in input_gids.iter() {
+                if gid.to_u32() >= self.font_num_glyphs as u32 {
+                    break;
+                }
```
cmyr, line 127: this is fine, but just for illustration you could do,

```suggestion
            for gid in input_gids.iter().take_while(|gid| gid.to_u32() < self.font_num_glyphs as u32) {
```

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
         }
+        self.glyphset_gsub
+            .extend(self.unicode_to_new_gid_list.iter().map(|t| t.1));
+        self.unicodes
+            .extend(self.unicode_to_new_gid_list.iter().map(|t| t.0));
+    }
+
+    pub fn populate_gids_to_retain(&mut self, font: &FontRef) {
+        //not-def
+        self.glyphset_gsub.insert(GlyphId::from(0));
```
cmyr, line 139: you can use the `GlyphId::NOTDEF` constant here, which is a bit more self-documenting

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+                operation_count as i32,
+                0,
+            );
+        }
+        remove_invalid_gids(&mut self.glyphset, self.font_num_glyphs);
+    }
+}
+
+/// glyph closure for Composite glyphs in glyf table
+fn glyf_closure_glyphs(
```
cmyr, line 171: I would add a bit more documentation of how this is expected to work, in particular how it uses the return value.

klippa/src/lib.rs
```diff
@@ -7,67 +7,216 @@ mod parsing_util;
...
+};
 use write_fonts::types::GlyphId;
 use write_fonts::types::Tag;
 use write_fonts::{
     from_obj::FromTableRef, tables::hhea::Hhea, tables::hmtx::Hmtx, tables::maxp::Maxp, FontBuilder,
 };
+
+const MAX_COMPOSITE_OPERATIONS_PER_GLYPH: u8 = 64;
+const MAX_NESTING_LEVEL: u8 = 64;
+const CODEPOINT_INVALID: u32 = 0xFFFFFFFF;
```
cmyr, line 27: I find this constant a bit confusing; there are many invalid codepoints before this one, and it looks like it is only used to construct a `GlyphId`? It might be simpler to just do,

```rust
const MAX_GID: GlyphId = GlyphId::new(u16::MAX);
```

read-fonts/src/tables/cmap.rs
```diff
@@ -39,6 +40,22 @@ impl<'a> Cmap<'a> {
         }
         None
     }
+
```
cmyr, line 43: So there is already a trait defined for this in GSUB, and we should make the signatures match. That implementation is older than `IntSet`, though, so let's just open an issue to fix this later?

cmyr (approved):
Looks good! I've left some comments inline but they're mostly small things, and overall this all seems very reasonable.

klippa/src/lib.rs
qxliu76: These 2 options are not exclusive, so we can have only gids or only unicodes or both as input

klippa/src/lib.rs
qxliu76: oh this function will change 4 members of the subset plan, if I make it a free function, I would get a this is a warning that a function has too many arguments.

klippa/src/lib.rs
qxliu76: thanks! This is very useful

read-fonts/src/tables/cmap.rs
qxliu76: Done, added a new issue.

klippa/src/lib.rs
qxliu76: uh.. I'll work on this later in another PR, but I think we should not panic here.

garretrieger (approved):

read-fonts/Cargo.toml
```diff
@@ -22,6 +22,7 @@ font-types = { version = "0.5.5", path = "../font-types", features = ["bytemuck"
 serde = { version = "1.0", features = ["derive"], optional = true }
 core_maths = { workspace = true, optional = true }
 bytemuck = { workspace = true }
+int-set = { path = "../int-set" }
```
dfrg, line 25: Just a note that we can’t currently take this dependency in read-fonts. This will prevent releasing and break Chrome builds. 

The cmap closure code that makes use of it is relatively simple and only uses public APIs so I’d recommend just moving it to klippa.

qxliu76: oh, thanks! I will move it to klippa. I'm taking next Mon and Tue off, if it's not urgent, I will try to do this next Wed.

dfrg: No worries! I don’t think we have any need to release soon but if we do, this is a simple patch so we can take care of it. Enjoy your time off!

PR #984, opened 2024-06-30
[skrifa] internal compact unscaled outline repr

dfrg:
Autohinting requires loading a large number of outlines and the algorithms generally expect these structures to match that of `FT_Outline`. This PR includes a new internal API for doing so with a simple compact representation that should meet all of our autohinting needs.

Notably, we do not want to run TrueType outlines through an `OutlinePen` because that materializes implied oncurves which should not affect correctness but will likely prevent us from generating results that exactly match FreeType. It's also just generally wasteful.

Based on #962 since the next PR will need both.


skrifa/src/outline/unscaled.rs
```diff
@@ -0,0 +1,437 @@
...
+                self.use_heap = true;
+                self.heap.extend(&self.inline);
+                self.inline_len = 0;
+            }
+        }
+        self.heap.push(point);
+    }
+
+    pub fn clear(&mut self) {
+        self.inline_len = 0;
```
cmyr, line 106: oops mentioned this in the wrong PR I guess? but just want to check you intend to not set `use_heap = false`

cmyr (approved):

skrifa/src/outline/unscaled.rs
dfrg: heh, yes, I intended to keep the heap allocation if it already exists

PR #987, opened 2024-07-03
[int-set] Optimize the sparse set decoding operation.

garretrieger:
- Switch to reading and processing a node at a time instead of bit by bit.
- Using count trailing zeroes to find the '1' bits instead of checking all bits.
- Make branch factor during decode a compile time constant to allow the compiler to generate specialized versions for each of the four branch factors.

Yields a ~45% speedup versus the previous implementation.

### DIFF
--- a/int-set/src/input_bit_stream.rs
+++ b/int-set/src/input_bit_stream.rs
@@ -46,7 +46,7 @@ impl<'a, const BF: u8> InputBitStream<'a, BF> {
     ///
     /// See: <https://w3c.github.io/IFT/Overview.html#sparse-bit-set-decoding>
     /// Returns None if the stream does not have enough remaining bits.
-    #[allow(clippy::unusual_byte_groupings)]
+    #[allow(clippy::unusual_byte_groupings)] // Used to separate bit values into units used in the set encoding.
     pub(crate) fn decode_header(data: &'a [u8]) -> Option<(BranchFactor, u8)> {
         let first_byte = data.first()?;
         let bf_bits = 0b0_00000_11 & first_byte;
--- a/int-set/src/input_bit_stream.rs
+++ b/int-set/src/input_bit_stream.rs
@@ -1,137 +1,161 @@
 //! Reads individual bits from a array of bytes.
 
-pub(crate) struct InputBitStream<'a> {
+use crate::sparse_bit_set::BranchFactor;
+
+pub(crate) struct InputBitStream<'a, const BF: u8> {
     data: &'a [u8],
     byte_index: usize,
-    bit_index: u8,
+    sub_index: u32,
 }
 
-impl<'a> InputBitStream<'a> {
-    pub fn from(data: &'a [u8]) -> InputBitStream {
+impl<'a, const BF: u8> Iterator for InputBitStream<'a, BF> {
+    type Item = u32;
+    fn next(&mut self) -> Option<Self::Item> {
+        match BF {
+            2 | 4 => {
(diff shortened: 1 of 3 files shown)


int-set/src/input_bit_stream.rs
```diff
@@ -1,137 +1,161 @@
...
-
-        self.move_to_next();
-
-        Some(bit_value)
-    }
-
-    /// Skips the next bit in the stream.
-    pub fn skip_bit(&mut self) {
-        self.move_to_next();
+    #[allow(clippy::unusual_byte_groupings)]
```
cmyr, line 49: when opting out of a clippy lint I find it helpful to include a little comment explaining why.

int-set/src/sparse_bit_set.rs
```diff
@@ -41,66 +41,71 @@ impl IntSet<u32> {
...
-            for index in 0..branch_factor as u32 {
-                let Some(bit) = bits.read_bit() else {
-                    return Err(DecodingError);
-                };
-
-                if !bit {
-                    continue;
+            let Some(mut bits) = bits.next() else {
+                return Err(DecodingError);
+            };
```
cmyr, line 73: alternatively,

```suggestion
            let mut bits = bits.next().ok_or(DecodingError)?;
```

int-set/src/input_bit_stream.rs
```diff
@@ -1,137 +1,161 @@
...
-    pub fn from(data: &'a [u8]) -> InputBitStream {
-        InputBitStream {
-            data,
-            byte_index: 0,
-            bit_index: 0,
-        }
-    }
+impl<'a, const BF: u8> Iterator for InputBitStream<'a, BF> {
+    type Item = u32;
+    fn next(&mut self) -> Option<Self::Item> {
```
cmyr, line 13: This is good, but I wanted to point out for fun that in the off chance that we were really concerned about the possible branching below (which in practice shouldn't matter, the branch predictor will figure it out) we could write separate impls for the different possible values, like:

```rust
 impl<'a> Iterator for InputBitStream<'a, 2> { .. }
 impl<'a> Iterator for InputBitStream<'a, 4> { .. }
```

in any case don't do this, I just thought you might find it interesting.

int-set/src/sparse_bit_set.rs
```diff
@@ -134,7 +139,14 @@ impl IntSet<u32> {
...
 }
 
-fn to_sparse_bit_set_with_bf<const BF: u8>(set: &IntSet<u32>) -> Vec<u8> {
+/// Encode this set as a sparse bit set byte encoding with a specified branch factor.
+///
+/// Branch factor can be 2, 4, 8 or 32. It's a compile time constant so that optimized decoding implementations
+/// can be generated by the compiler.
+///
+/// Sparse bit sets are a specialized, compact encoding of bit sets defined in the IFT specification:
+/// <https://w3c.github.io/IFT/Overview.html#sparse-bit-set-decoding>
```
cmyr, line 148: great, I always appreciate more docs.

cmyr (approved):
looks good!

int-set/src/input_bit_stream.rs
garretrieger: Since BF is a compile time constant my expectation is that the compiler will generate specialized versions for each branch factor where the branching is compiled out.

cmyr: hmm interesting that hadn't occurred to me but I suspect you're right!

PR #989, opened 2024-07-03
[int-set] Add a iter_ranges() method to int set

garretrieger:
Provides an iterator which iterates over ranges present in the set. Uses a specialized implementation where possible to quickly locate ranges in the underlying bit sets.


cmyr (approved):
looks good!

PR #990, opened 2024-07-04
[int-set] Add benchmark suite for sparse set decoding.

garretrieger:


int-set/benches/bench_helper.rs
```diff
@@ -0,0 +1,17 @@
+use int_set::IntSet;
+
+use rand::Rng;
+
+pub fn random_set(size: u32, max_value: u32) -> IntSet<u32> {
```
cmyr, line 5: if this is used to generate input to a benchmark fn, do we want to seed the RNG or otherwise do something to ensure our inputs are identical between runs? I imagine different distributions of numbers could noticeably impact performance?

cmyr (approved):
looks good, one note inline.

int-set/benches/bench_helper.rs
garretrieger: I've so far found that even with random inputs the bench times are very consistent. My worry with a fixed seed is that you might inadvertently make optimizations that actually only work well on the specific set that's being used to test. An alternative might be to generate a small number (eg. 10) of fixed input sets and rotate through them on each iteration.

PR #991, opened 2024-07-04
[int-set] Optimize set building time during sparse set decoding.

garretrieger:
- Make specialized sorted insert available and use it during the decoding operation.
- Results in speedups up to 60% for the decoding of large sets (32k elements).

### DIFF
--- a/int-set/src/bitset.rs
+++ b/int-set/src/bitset.rs
@@ -489,22 +489,45 @@ impl BitSet {
 
 impl Extend<u32> for BitSet {
     fn extend<U: IntoIterator<Item = u32>>(&mut self, iter: U) {
+        let mut builder = BitSetBuilder::start(self);
+        for val in iter {
+            builder.insert(val);
+        }
+        builder.finish();
+    }
+}
+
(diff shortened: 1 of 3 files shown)


int-set/src/bitset.rs
```diff
@@ -489,22 +489,49 @@ impl BitSet {
...
+            self.last_page_index = self.set.ensure_page_index_for_major(major_value);
+            self.last_major_value = major_value;
+        };
+        if let Some(page) = self.set.pages.get_mut(self.last_page_index) {
+            page.insert_no_return(val);
         }
-        self.mark_dirty();
+    }
+
+    pub(crate) fn finish(&mut self) {
```
cmyr, line 533: this just reminds me that we'd discussed getting rid of the 'dirty' tracking, any further thoughts on that?

cmyr (approved):
one unrelated qq inline, but looks good!

int-set/src/bitset.rs
garretrieger: I haven't had a chance to look into it further yet.

PR #992, opened 2024-07-10
Add read-fonts support for IFT Patch Map Format 1

garretrieger:
See: https://w3c.github.io/IFT/Overview.html#patch-map-format-1

Adds support for parsing format 1 patch maps in an "IFT " or "IFTX" table. Adds Skrifa code to convert the patch map format 1 into the logical set of entries that it represents.

Couple of things that are left for future PRs:
1. Implementation of URI template substitution.
2. Spec compliant error handling.


rsheeter:
General comment: would prefer to see int-set merge into an existing crate. To me font-types is the most obvious choice, followed by read-fonts.

read-fonts/src/tables/ift.rs
```diff
@@ -0,0 +1,344 @@
...
+                .unwrap_or(0),
+        }
+        .filter(|(_, entry_index)| *entry_index > 0)
+    }
+
+    pub fn entry_count(&self) -> u32 {
+        self.max_entry_index() as u32 + 1
+    }
+
+    pub fn uri_template_as_string(&self) -> Result<&str, Utf8Error> {
```
dfrg, line 64: I'm fine with leaving this as `Utf8Error` but mapping it to a `ReadError` variant (maybe `MalformedData("URI template not valid UTF-8")`) might make this API easier to use.

read-fonts/src/tables/ift.rs
```diff
@@ -0,0 +1,344 @@
...
+            self.feature_record_it = None;
+            return None;
+        };
+
+        let new_entry_index = feature_record.first_new_entry_index.get()
+            + (feature_record.entry_map_count().get() - self.remaining);
+        self.remaining -= 1;
+
+        let size = U8Or16::compute_size(&self.max_entry_index).ok()? * 2;
+        self.data = &self.data[size..];
```
dfrg, line 188: Is this guaranteed to be in bounds?

Changing to `self.data.get(size..)?` would prevent a panic just in case.

resources/codegen_inputs/ift.rs
```diff
@@ -0,0 +1,126 @@
...
+#[skip_font_write]
+#[skip_from_obj]
+table IFT {
+}
+
+#[tag = "IFTX"]
+#[skip_font_write]
+#[skip_from_obj]
+table IFTX {
+}
```
dfrg, line 15: As mentioned, I think we can just remove these.

skrifa/src/patchmap.rs
```diff
@@ -0,0 +1,393 @@
+//! Loads incremental font transfer <https://w3c.github.io/IFT/Overview.html> patch mappings.
```
dfrg, line 1: Given that this is somewhat low level and we've started adding `std` dependent code in `read-fonts`, I think this should be moved there under the `ift` module and behind a `std` feature gate.

The only complication I can see is the use of the `Charmap` selection and iterator logic but my preference would be to move that to read-fonts as well (it'll be useful for shaping anyway).

dfrg (commented):
I'd like to move the patchmap functionality to read-fonts but other than that and a few nits, this looks good to me.

I can work on extending the read-fonts cmap functionality to support the move.

If @cmyr wants to do a quick pass, particularly over the codegen bits, that might be helpful.

skrifa/src/patchmap.rs
garretrieger: Since Charmap hasn't been moved yet I'll leave this here for this PR and move it into read-fonts after Charmap is moved.

dfrg: Sounds good. Let’s land this and I’ll ping you when the necessary functionality is available in read-fonts

dfrg (approved):
oops, forgot to approve earlier

PR #994, opened 2024-07-10
Big glyph ids

### LINKED ISSUE #784: Support big glyph ids (beyond 64k)
in accordance with the [spec proposal](https://github.com/harfbuzz/boring-expansion-spec/blob/main/beyond-64k.md). I recommend we do this in a few phases to limit breakage:

1. Rename `font_types::GlyphId` to `GlyphId16` and update codegen and resource files accordingly. Add a `pub type GlyphId = GlyphId16` to `font-types` which should make this transparent to dependencies. 
2. Update `fontc` projects and `skrifa` to use `GlyphId16` explicitly.
3. Remove the type alias and add a new `GlyphId` type backed by a `u32` with appropriate conversions. Some of the utility methods in `read-fonts` should probably change to take `impl Into<GlyphId>`. 
4. Update `fontc` projects and `skrifa` to use the new `GlyphId` where appropriate. To avoid breaking Skia, we'll likely need to keep a truncating `to_u16()` method on `GlyphId` for a transitional period. Alternatively, we can add a `to_u32()` method on the current type now and push that change through (this might be the easier route).
5. Add a new `GlyphId24` type to support the actual table extensions.

@drott are there any complications I might be missing on the Skia side?

dfrg:
Renames current `GlyphId` to `GlyphId16` and adds a new 32-bit `GlyphId`.

All codegen inputs changed to use explicit 16-bit ids.

Changes all skrifa public APIs to accept/return the new type. Some parts of read-fonts and write-fonts have been updated to work with the larger type but not all. We might need to work this out as we go.

Fixes #784


write-fonts/src/tables/cmap.rs
```diff
@@ -40,7 +40,12 @@ impl CmapSubtable {
 
         let mut prev = (u16::MAX - 1, u16::MAX - 1);
         for (cp, gid) in mappings {
-            let gid = gid.to_u16();
+            let gid = gid.to_u32();
+            if gid > 0xFFFF {
+                // Should we just fail here?
+                continue;
+            }
+            let gid = gid as u16;
```
dfrg, line 48: (generating a cmap4 subtable)

I'm not sure what to do here. I assume if we're generating a font with more than 64k glyphs that we just shouldn't emit a cmap4 subtable.

int-set/src/lib.rs
```diff
@@ -1593,11 +1593,11 @@ mod test {
...
 
-        let Some(max) = set.iter().max() else {
-            panic!("should have a max");
-        };
+        // let Some(max) = set.iter().max() else {
+        //     panic!("should have a max");
+        // };
 
-        assert_eq!(max, GlyphId::new(u16::MAX));
+        // assert_eq!(max, GlyphId::new(u32::MAX as _));
```
dfrg, line 1600: This test now takes a substantial amount of time with 32-bit glyph ids...

garretrieger: I think we'll want to implement Domain for the glyphid 16 type, use that here in this test, and then add a new test that checks on the new GlyphId type but doesn't do max()

dfrg: Good idea. Made it so.

font-types/src/glyph_id.rs
```diff
@@ -58,4 +112,55 @@ impl From<GlyphId> for u32 {
...
+                .0
+                .try_into()
+                .map_err(|_| TryFromGlyphIdError(value.0))?,
+        ))
+    }
+}
+
+/// The error type returned when a glyph identifier conversion fails.
+#[derive(Debug)]
+pub struct TryFromGlyphIdError(u32);
```
cmyr, line 160: should we impl `Error` for this? gated on `std`?

cmyr (approved):
Looks good!

PR #995, opened 2024-07-10
Move IntSet into read_fonts.

garretrieger:


LaurenzV:
Just wondering, what's the motivation behind this? Couldn't this crate be useful outside of fontations, too?

garretrieger:
> Just wondering, what's the motivation behind this? Couldn't this crate be useful outside of fontations, too?

Yes, it is not tied to fonts in anyway and could be used as a general int set. There is concern that having it as a separate crate will make it more difficult to use read_fonts/skrifa by consumers of the library. We're not fully decided on moving this yet, so if there is interest in using the int set as a standalone thing that would be helpful input into the decision.

cmyr (approved):
So I agree that all things being equal this definitely makes the most sense as a standalone crate.

The arguments for merging it into read-fonts are (to the best of my knowledge)

- it's a hassle to add a crate to g3
- not needing to manage an additional dependency

I don't find the second of these very compelling, and I don't know enough about the first to have an informed opinion; so if there is consensus amongst other people that merging this makes sense than I defer to that.

PR #998, opened 2024-07-11
Use explicit features for `serde`, `bytemuck`, `dot2`

waywardmonkeys:
Provide explicit features for optional dependencies rather than implicit ones. This is slated to be required in Rust 2024 Edition.


cmyr (approved):
Thanks!

PR #999, opened 2024-07-11
[int-set] Fix subtraction overflow in remove_range.

### LINKED ISSUE #997: IntSet::remove_range() subtract with overflow
Minimal test case:

```rust
let mut set = IntSet::<u32>::empty();
set.remove_range(0..=u32::MAX);
```

Using the range `0..=u32::MAX - 512` works but any greater end range fails. Initializing the set with `all()` succeeds.

The overflow occurs here: https://github.com/googlefonts/fontations/blob/a52dbf59cda0aa16909d0af93fe74c518278a085/int-set/src/bitset.rs#L110

@garretrieger

garretrieger:
Fixes #997

### DIFF
--- a/int-set/src/bitset.rs
+++ b/int-set/src/bitset.rs
@@ -107,7 +107,7 @@ impl BitSet {
 
         for major in major_start..=major_end {
             let page_start = start.max(self.major_start(major));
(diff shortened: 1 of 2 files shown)


dfrg (approved):
LGTM!

PR #1001, opened 2024-07-12
fix: make Weight and Stretch getters const

jcdickinson:
I want to convert standard ratios to those used by fontconfig, so I would want to construct these constants:

``` fs
const WEIGHT_RATIO: f32 = FC_WEIGHT_EXTRABLACK as f32 / Weight::EXTRA_BLACK.value();
const STRETCH_RATIO: f32 = FC_WIDTH_ULTRAEXPANDED as f32 / Stretch::ULTRA_EXPANDED.ratio();
```

There doesn't seem to be any reason why these fns (value, ratio) can't be const. I also changed the new fns to const, for the same reason.

### DIFF
--- a/skrifa/src/attribute.rs
+++ b/skrifa/src/attribute.rs
@@ -127,7 +127,7 @@ impl Stretch {
 
 impl Stretch {
     /// Creates a new stretch attribute with the given ratio.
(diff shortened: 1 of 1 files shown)


dfrg (approved):
This seems like a useful change, thanks!

PR #1003, opened 2024-07-15
[int-set] Add iter_after() and intersects(range) to IntSet

garretrieger:
iter_after() creates an iterator of values in the set that come after a given value.
intersects(range) checks if any values in the set intersect a range.

### DIFF
--- a/int-set/src/lib.rs
+++ b/int-set/src/lib.rs
@@ -102,13 +102,6 @@ impl<T: Domain<T>> Default for IntSet<T> {
 }
 
 impl<T: Domain<T>> IntSet<T> {
-    // TODO(garretrieger): add additional functionality that the harfbuzz version has:
-    // - Iteration in reverse.
-    // - Iteration starting from some value (and before some value for reverse).
-    // - Set operations (union, subtract, intersect, sym diff).
-    // - Intersects range and intersects iter.
-    // - min()/max()
-
     /// Returns an iterator over all members of the set in sorted ascending order.
     ///
     /// Note: iteration of inverted sets can be extremely slow due to the very large number of members in the set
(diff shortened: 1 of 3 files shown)


int-set/src/lib.rs
```diff
@@ -121,6 +114,39 @@ impl<T: Domain<T>> IntSet<T> {
         u32_iter.map(|v| T::from_u32(InDomain(v)))
     }
 
+    /// Returns and iterator over the members of this set that come after 'value' in ascending order.
```
cmyr, line 117:
```suggestion
    /// Returns an iterator over the members of this set that come after 'value' in ascending order.
```

int-set/src/lib.rs
```diff
@@ -246,6 +272,34 @@ impl<T: Domain<T>> IntSet<T> {
         }
     }
 
+    /// Returns true if this set contains at least one element in 'range'.
+    pub fn intersects(&mut self, range: RangeInclusive<T>) -> bool {
```
cmyr, line 276: I would expect an `intersects` method to take another set, and would call this method `intersects_range` 🤷

int-set/src/lib.rs
```diff
@@ -121,6 +114,39 @@ impl<T: Domain<T>> IntSet<T> {
         u32_iter.map(|v| T::from_u32(InDomain(v)))
     }
 
+    /// Returns and iterator over the members of this set that come after 'value' in ascending order.
+    ///
+    /// Note: iteration of inverted sets can be extremely slow due to the very large number of members in the set
+    /// care should be taken when using .iter() in combination with an inverted set.
+    pub(crate) fn iter_after(&self, value: T) -> impl Iterator<Item = T> + '_ {
```
cmyr, line 121: is this expected to be public API?

int-set/src/lib.rs
```diff
@@ -498,6 +552,79 @@ where
     }
 }
 
+struct IterAfter<SetIter, AllValuesIter>
```
cmyr, line 555: naively I'm surprised we need to define a new type here? I would think that 'iter after' could be implemented with the same type as normal iteration, and it just being a matter of setting up the correct starting state.

int-set/src/lib.rs
```diff
@@ -498,6 +552,79 @@ where
...
+where
+    SetIter: Iterator<Item = u32>,
+    AllValuesIter: Iterator<Item = u32>,
+{
```
cmyr, line 559: this is a kind of funny point, but there's a general rust idiom towards separating data and implementation, and specifically around avoiding trait bounds on data types unless necessary, saving them for the actual impl blocks.

This both saves typing but also avoids the possible 'bounds contagion' where any type that contains another type that has bounds has to also declare the appropriate bounds.

This isn't particularly relevant here, but I think it's a good habit to avoid the bounds on the type decls where possible.

cmyr (approved):
looks good, left some little notes inline but nothing blocking. It feels like you're pretty much up to speed, rust-wise :)

int-set/src/lib.rs
garretrieger: oops yes, fixed.

int-set/src/lib.rs
garretrieger: The problem is the existing Iter type works specifically with DoubleEndedIterator's. For simplicity I've made iter_after() not double ended, but unfortunately that means with need a new iter impl to deal with it.

garretrieger: Actually thanks to your comment below about not bounding the types on the struct I think I can make this work reusing Iter

int-set/src/lib.rs
garretrieger: Thanks, used this approach to merge IterAfter into Iter.

PR #1004, opened 2024-07-17
[read-fonts] Reduce typo count.

waywardmonkeys:


dfrg (approved):
Looks like many of these are mine 🫣. Thanks Bruce!

PR #1007, opened 2024-07-17
[int-set] fix bug in iter_ranges() of IntSet.

garretrieger:
It was incorrectly using page.index in place of page.major_value when calculating the page end.


cmyr (approved):
nice catch :)

PR #1008, opened 2024-07-17
[int-set] Add an IntSet fuzzer that checks for correctness.

garretrieger:
The fuzzer input is treated as a series of op codes which can invoke any of the int sets public api methods. The operations are performed on both and IntSet and BTreeSet in parallel and then method results and set contents are checked for equivalence.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_int_set.rs
+++ b/fuzz/fuzz_targets/fuzz_int_set.rs
@@ -8,35 +8,20 @@ use libfuzzer_sys::fuzz_target;
 // TODO allow inverted sets to be accessed.
 // TODO allow a limited domain set to be accessed.
 
-// TODO ops for most public api methods (have operations for iter() be what checks for
-//      iter() equality alongside the check at end):
-// - iter
-// - iter ranges
-// - extend / extend_unsorted
-// - remove_all
-// - union
-// - intersect
-// - first
-// - last
-// - contains
-// - is_empty
-// - len
-// - clear
-
 trait Operation {
     fn size(&self) -> u64;
     fn operate(&self, int_set: &mut IntSet<u32>, btree_set: &mut BTreeSet<u32>);
 }
 
+/* ### Insert ### */
 struct InsertOp(u32);
 
 impl InsertOp {
     fn new(data: &[u8]) -> (Option<Box<dyn Operation>>, &[u8]) {
-        if data.len() < 4 {
+        let (Some(val), data) = read_u32(data) else {
             return (None, data);
-        }
-
-        (Some(Box::new(InsertOp(read_u32(data)))), &data[4..])
+        };
+        (Some(Box::new(Self(val))), data)
     }
 }
 
@@ -51,31 +36,29 @@ impl Operation for InsertOp {
     }
 }
 
+/* ### Insert Range ### */
+
 struct InsertRangeOp(u32, u32);
 
 impl InsertRangeOp {
     fn new(data: &[u8]) -> (Option<Box<dyn Operation>>, &[u8]) {
-        if data.len() < 8 {
+        let (Some(min), data) = read_u32(data) else {
             return (None, data);
-        }
+        };
+        let (Some(max), data) = read_u32(data) else {
+            return (None, data);
+        };
 
-        (
-            Some(Box::new(InsertRangeOp(
-                read_u32(data),
-                read_u32(&data[4..]),
-            ))),
-            &data[8..],
-        )
+        (Some(Box::new(Self(min, max))), data)
     }
 }
 
 impl Operation for InsertRangeOp {
     fn size(&self) -> u64 {
-        if self.1 > self.0 {
-            self.1 as u64 - self.0 as u64
-        } else {
-            1
+        if self.1 < self.0 {
+            return 1;
         }
+        (self.1 as u64 - self.0 as u64) + 1
     }
 
     fn operate(&self, int_set: &mut IntSet<u32>, btree_set: &mut BTreeSet<u32>) {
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+        if self.1 < self.0 {
+            return 1;
+        }
+        ((self.1 as usize - self.0 as usize) + 1) * (length.ilog2() as usize)
+    }
+}
+
+/* ### First  ### */
+
+struct FirstOp();
```
cmyr, line 282: I wonder if these might not be more cleanly expressed as an enum? but probably not worth changing at this point.

cmyr (approved):
looks good!

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: I originally did this as an enum, but there ends up being more boiler plate (eg. match statements) needed then versus using dyn + a trait.

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
+#![no_main]
+//! A correctness fuzzer that checks all of the public API methods of IntSet.
```
rsheeter, line 2: Suggest making this first ref a link

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+//!
+//! This fuzzer exercises the public API of IntSet and compares the results to the same operations run
+//! on a BTreeSet. Any differences in behaviour and/or set contents triggers a panic.
+//!
+//! The fuzzer input data is interpretted as a series of op codes which map to the public api methods of IntSet.
```
rsheeter, line 7: spelling

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+//!
+//! Note: currently only inclusive mode IntSet's are tested.
```
rsheeter, line 9: Can we file and reference an issue for all the unfuzzed work? - I immediately see both this and a todo for inverted sets

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+        Input { int_set, btree_set }
+    }
+}
+
+trait Operation {
+    fn size(&self, set_len: usize) -> usize;
+    fn operate(&self, input: Input, other: Input);
+}
+
+/* ### Insert ### */
```
rsheeter, line 42: Seems self-evident, no comment required?

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+    btree_set: &'a mut BTreeSet<u32>,
+}
+
+impl Input<'_> {
+    fn from<'a>(int_set: &'a mut IntSet<u32>, btree_set: &'a mut BTreeSet<u32>) -> Input<'a> {
+        Input { int_set, btree_set }
+    }
+}
+
+trait Operation {
```
rsheeter, line 37: It seems to me your life would be easier if Operation was an enum that defined size and operation in it's impl. It would then have a fixed, small, size and you wouldn't need the whole mess of boxed dyn traits.

fuzz/fuzz_targets/fuzz_int_set.rs
rsheeter: I'm reading this and thinking the same thing.

fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -0,0 +1,673 @@
...
+    (
+        Some(
+            ((data[0] as u32) << 24)
+                | ((data[1] as u32) << 16)
+                | ((data[2] as u32) << 8)
+                | (data[3] as u32),
+        ),
+        &data[4..],
+    )
+}
```
rsheeter, line 550: I think you are re-inventing FontData + Cursor and should likely use them?

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: If this is done as an enum then every time a new type is added 3 match statements will need to be updated, versus the current setup where only one (op code -> constructor) needs to be updated. The mild performance improvement isn't important in this context (the performed set operations are going to significantly outweigh the cost of dyn here) so the advantages of enums reducing size aren't relevant.

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: I found these useful when working on this to visually break up the blocks of op specific code.

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: I've got a PR coming soon with the missing functionality.

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: My initial write of this was using enum's but I switched to dyn because I found enums to be cumbersome when adding new ops.

fuzz/fuzz_targets/fuzz_int_set.rs
garretrieger: using cursor sounds good to me.

PR #1010, opened 2024-07-18
[skrifa] autohinting: the Latin blues

dfrg:
Adds support for computing "blue alignment zones" for scripts that use the Latin writing system (which is most of them).

Based on #984 which needs to land first.


skrifa/src/outline/autohint/latin.rs
```diff
@@ -0,0 +1,513 @@
...
+                // Reject glyphs that don't produce any rendering
+                if outline.points.len() <= 2 {
+                    continue;
+                }
+                let mut best_y: Option<i16> = None;
+                let mut best_y_extremum = if is_top { i32::MIN } else { i32::MAX };
+                let mut best_is_round = false;
+                // Find the extreme point depending on whether this is a top or
+                // bottom blue
+                let best_contour_and_point = if is_top_like {
```
cmyr, line 75: do you think there's any room to move any of this stuff into some little helpers or something? The diff gets very hard to read with all the right-drift...

skrifa/src/outline/unscaled.rs
```diff
@@ -0,0 +1,437 @@
...
+                self.heap.extend(&self.inline);
+                self.inline_len = 0;
+            }
+        }
+        self.heap.push(point);
+    }
+
+    pub fn clear(&mut self) {
+        self.inline_len = 0;
+        self.heap.clear();
```
cmyr, line 107: should this set `self.use_heap = false`? I guess if you've already allocated you might as well reuse that?

cmyr (approved):
I wouldn't change a thing (and great PR name)

skrifa/src/outline/unscaled.rs
dfrg: yeah, that was my thinking.. we've already allocated so might as well reuse it. Per IM, I'm going to extend the functionality of the already existing internal `SmallArray` type to cover these cases so I can delete this code.

PR #1011, opened 2024-07-18
[int-set] Extend the IntSet fuzzer to exercise remaining public api methods.

garretrieger:
Further extends the IntSet fuzzer to allow it to exercise the remaining parts of the public API that the current implementation does not cover:
- Inverted sets (invert(), is_inverted())
- Limited domain types (including discontinous domain types).
- Inclusive Iteration.

Additionally address some comments from the review of #1008 including switching to use "Cursor".

### DIFF
--- a/fuzz/fuzz_targets/fuzz_int_set.rs
+++ b/fuzz/fuzz_targets/fuzz_int_set.rs
@@ -26,17 +26,20 @@ use libfuzzer_sys::fuzz_target;
 const OPERATION_COUNT: usize = 7_500;
 
 // TODO(garretrieger): use "Cursor" to manage the input buffer.
-// TODO(garretrieger): allow inverted sets to be accessed.
-// TODO(garretrieger): allow a limited domain set to be accessed.
 
 trait SetMember<T>: Domain<T> + Ord + Copy + Add<u32, Output = T> + Debug {
     fn create(val: u32) -> Option<T>;
+    fn can_be_inverted() -> bool;
     fn increment(&mut self);
 }
 
 impl SetMember<u32> for u32 {
     fn create(val: u32) -> Option<u32> {
         Some(val)
+    }
+
+    fn can_be_inverted() -> bool {
+        false
     }
 
     fn increment(&mut self) {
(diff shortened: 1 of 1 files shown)


fuzz/fuzz_targets/fuzz_int_set.rs
```diff
@@ -525,124 +832,175 @@ impl Operation for IntersectOp {
     }
 }
 
-/* ### End of Ops ### */
+/* ### Invert ### */
 
-fn read_u8(data: &[u8]) -> (Option<u8>, &[u8]) {
-    if data.is_empty() {
-        return (None, data);
+struct InvertOp();
```
cmyr, line 837: totally academic but you can declare a 'unit struct' that has no fields with the syntax, `struct InvertOp;`.

int-set/src/lib.rs
```diff
@@ -152,19 +163,29 @@ impl<T: Domain<T>> IntSet<T> {
     /// Returns an iterator over all disjoint ranges of values within the set in sorted ascending order.
     pub fn iter_ranges(&self) -> impl Iterator<Item = RangeInclusive<T>> + '_ {
         let u32_iter = match &self.0 {
-            Membership::Inclusive(s) => RangeIter::Inclusive {
-                ranges: s.iter_ranges(),
-            },
+            Membership::Inclusive(s) => {
```
cmyr, line 166: what's the relationship between these changes and the nominal goals of the patch?

cmyr (approved):
Looks good, my only question is whether or not this really wants to be two PRs, or whether the  `RangeIter` changes and the fuzzing changes are directly related in some way?

garretrieger:
> Looks good, my only question is whether or not this really wants to be two PRs, or whether the `RangeIter` changes and the fuzzing changes are directly related in some way?

The RangeIter changes fixes a few bugs that the additional fuzzer functional exposed. If it's helpful I can split it off into it's own PR.

int-set/src/lib.rs
garretrieger: This is a fix for incorrect behaviour found by the fuzzer additions. In this case the iteration over ranges discontinous was incorrect: ranges that were actually adjacent (but not in u32 space) were being returned as disjoint.

garretrieger:
> > Looks good, my only question is whether or not this really wants to be two PRs, or whether the `RangeIter` changes and the fuzzing changes are directly related in some way?
> 
> The RangeIter changes fixes a few bugs that the additional fuzzer functional exposed. If it's helpful I can split it off into it's own PR.

Moved the IntSet changes to #1014

PR #1013, opened 2024-07-19
Reduce typo count further.

waywardmonkeys:


waywardmonkeys:
This leaves a small number of known typos (which are in actual code identifiers). After that, if there's interest, we could enable the same typo check in CI that we use in LineBender repos.

waywardmonkeys:
Rebased to fix merge conflicts.

cmyr (approved):
thanks!

cmyr:
I think a typo check in CI would be reasonable, if it's simple enough?

waywardmonkeys:
@cmyr Very simple! There's one in `norad` now. I have another PR to file and then I'll do that one and you can see what you think.

PR #1014, opened 2024-07-19
[int-set] Fix several bugs in IntSet found by the fuzzer:

garretrieger:
- Fix IntSet::len() for inverted sets with domains smaller than u32.
- Fix overflow bug in BitPage iter: next() was casting an i32 (backwards_index) to u32 prior to comparison. If backwards_index is -1 this leads to unexpected behaviour.
- Fix iter_ranges() on discontinous domains: consider two values adjacent if they are adjancent in the ordered_values() iterator instead of requiring them to be adjacent in the u32 domain.
- Fix iter_ranges() on inclusive sets w/ discontinous domains. Adjacent ranges need to be merged together.

### DIFF
--- a/int-set/src/lib.rs
+++ b/int-set/src/lib.rs
@@ -73,6 +73,9 @@ pub trait Domain<T> {
     /// Values should be converted to `u32`'s according to the mapping defined in
     /// `to_u32`/`from_u32`.
     fn ordered_values_range(range: RangeInclusive<T>) -> impl DoubleEndedIterator<Item = u32>;
+
+    /// Returns the number of members in the domain.
+    fn count() -> usize;
 }
 
 /// Marks a mapped value as being in the domain of `T` for [`Domain<T>`].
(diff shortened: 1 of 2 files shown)


int-set/src/lib.rs
```diff
@@ -576,14 +589,20 @@ where
...
     AllValuesIter: Iterator<Item = u32>,
+    T: Domain<T>,
 {
     Inclusive {
         ranges: InclusiveRangeIter,
     },
+    InclusiveDiscontinous {
+        ranges: InclusiveRangeIter,
+        current_range: Option<RangeInclusive<u32>>,
+        phantom: PhantomData<T>,
```
cmyr, line 604: why is this necessary?

cmyr (approved):
A couple notes online but thanks for splitting this out?

int-set/src/lib.rs
garretrieger: The InclusiveDiscontinous arm needs access to the T type inside of the Iterator::next() implementation. The only way I could find to get it the T type was to make it part of RangeIter, from there the phantom data is needed to avoid an unused type parameter error.

PR #1015, opened 2024-07-19
Update sparse bit set decode to avoid numerical overflows.

garretrieger:
Given that we decode to a u32 int set, added two guards:
1. Enforce a maximum height per branch factor which will allow values up to u32 to be encoded, reject decoding sets with heights beyond this. Since the maximum value possible at each height in some cases still exceeds u32, the second guard is needed as well.
2. Do computations during decode using u64's and check for overflow before populating the output set. Reject decoding sets that exceed the u32 maximum.

### DIFF
--- a/int-set/src/output_bit_stream.rs
+++ b/int-set/src/output_bit_stream.rs
@@ -9,7 +9,7 @@ pub(crate) struct OutputBitStream {
 }
 
 impl OutputBitStream {
-    pub(crate) const MAX_HEIGHT: u8 = 32;
+    pub(crate) const MAX_HEIGHT: u8 = 31;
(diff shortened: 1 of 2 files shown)


cmyr (approved):
looks good!

one thing I've been thinking about is that if we really wanted `int-set` to be general purpose, we might want to support larger integer types, probably by being generic over the inner type?

garretrieger:
> looks good!
> 
> one thing I've been thinking about is that if we really wanted `int-set` to be general purpose, we might want to support larger integer types, probably by being generic over the inner type?

Yeah, I believe this should be possible without causing too many issues.

PR #1016, opened 2024-07-19
[int-set] Add a fuzzer for sparse bit set encoding and decoding.

garretrieger:
Re-uses part of the op processing code from the int set fuzzer, which has been moved to a common file.


fuzz/fuzz_targets/int_set_op_processor.rs
```diff
@@ -0,0 +1,1045 @@
...
+        }
+        Some(SmallInt::new(val))
+    }
+
+    fn can_be_inverted() -> bool {
+        true
+    }
+
+    fn increment(&mut self) {
+        self.0 = min(self.0 + 1, Self::MAX_VALUE);
```
cmyr, line 73: in general there's no need to import `min` for this, you can just do `Self::MAX_VALUE.min(self.0 + 1)`.

fuzz/fuzz_targets/int_set_op_processor.rs
```diff
@@ -0,0 +1,1045 @@
...
+    fn ordered_values() -> impl DoubleEndedIterator<Item = u32> {
+        0..=Self::MAX_VALUE
+    }
+
+    fn ordered_values_range(
+        range: RangeInclusive<SmallInt>,
+    ) -> impl DoubleEndedIterator<Item = u32> {
+        if range.start().0 > Self::MAX_VALUE || range.end().0 > Self::MAX_VALUE {
+            panic!("Invalid range of the SmallInt set.");
+        }
```
cmyr, line 99: this feels like it would be slightly simpler expressed as an assert?

fuzz/fuzz_targets/int_set_op_processor.rs
```diff
@@ -0,0 +1,1045 @@
...
+    }
+
+    fn count() -> usize {
+        ((Self::MAX_VALUE / 2) + 1) as usize
+    }
+}
+
+struct Input<'a, T>
+where
+    T: SetMember<T>,
```
cmyr, line 177: mentioned elsewhere but you shouldn't need bounds on the types themselves, unless there's something tricky going on?

fuzz/fuzz_targets/int_set_op_processor.rs
```diff
@@ -0,0 +1,1045 @@
...
+{
+    // The state includes 2 of each type of sets to allow us to test out binary set operations (eg. union)
+    int_set: &'a mut IntSet<T>,
+    btree_set: &'a mut BTreeSet<T>,
+}
+
+impl<T> Input<'_, T>
+where
+    T: SetMember<T>,
```
cmyr, line 186: Also just a style thing but generally when there's only a single (or a small number) of generic bounds it's easier to express them inline, like:

```rust
impl<T: SetMember<T>> Input<'_, T> { .. }
```

cmyr (approved):
looks good, I have a couple comments on the file that you just moved but all just little style things.

PR #1018, opened 2024-07-22
[int-set] Optimize remove_range.

garretrieger:
Iterate over only the existing pages contained within the range instead of all possible pages in the range.


int-set/src/bitset.rs
```diff
@@ -462,13 +484,13 @@ impl BitSet {
         value >> PAGE_BITS_LOG_2
     }
 
-    fn major_start(&self, major: u32) -> u32 {
+    fn major_start(major: u32) -> u32 {
```
cmyr, line 487: not paritcularly important, but these can all be marked `const` now, which means they can be called & evaluated at compile time.

cmyr (approved):
looks good!

PR #1022, opened 2024-07-23
[int-set] Add a second version of sparse bit set decode with max set size.

garretrieger:
It's very easy to construct a sparse bit set which contains all values in u32, which can be quite slow to decode. Add an alternate decoding method which can be given a maximum set size for callers who want to avoid long decoding times. Use the bounded version in the fuzzer.

### DIFF
--- a/int-set/src/sparse_bit_set.rs
+++ b/int-set/src/sparse_bit_set.rs
@@ -13,7 +13,7 @@ use crate::output_bit_stream::OutputBitStream;
 use crate::BitSet;
 use crate::IntSet;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct DecodingError;
 
 impl Error for DecodingError {}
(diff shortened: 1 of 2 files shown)


int-set/src/sparse_bit_set.rs
```diff
@@ -41,6 +41,19 @@ impl IntSet<u32> {
...
+        Self::from_sparse_bit_set_bounded(data, u32::MAX as u64 + 1)
+    }
+
+    /// Populate this set with the values obtained from decoding the provided sparse bit set bytes.
+    /// The size (in number of members) of the output set is bounded to be equal to or below the provided max size.
+    /// If the decoded set would contain more members than the limit an error will be returned.
+    ///
+    /// Sparse bit sets are a specialized, compact encoding of bit sets defined in the IFT specification:
+    /// <https://w3c.github.io/IFT/Overview.html#sparse-bit-set-decoding>
+    pub fn from_sparse_bit_set_bounded(
```
cmyr, line 53: do we desire for this to actually be public API? my hunch is no, and it's really just useful for the fuzzer, in which case I would add `#[doc(hidden)]` and include a note that it is not considered public for the purposes of semver.

cmyr (approved):
looks good!

int-set/src/sparse_bit_set.rs
garretrieger: This, or something like this will be needed in the public api. For example in the IFT case we know we are decoding sets that contain unicode values so we can restrict the maximum set size accordingly to put an upper bound on sparse set decode times.

We can introduce a bound in two different ways:
1. Maximum set size.
2. Maximum value allowed to be decoded.

For now I choose 1. as that allows the fuzzer to still probe edge cases (eg. u32::MAX), but ultimately 2. might be more useful for things like IFT. My plan for now was to leave 1. public and then potentially revisit this as I work through the IFT stuff.

PR #1023, opened 2024-07-24
[skrifa] unify custom collections on one SmallVec

dfrg:
The initial autohinting code introduced some custom collections for various things. This renames our `SmallArray` to `SmallVec` (it's now growable) and adds functionality to cover all of our current use cases.


skrifa/src/collections.rs
```diff
@@ -0,0 +1,290 @@
...
+}
+
+impl<T, const N: usize> PartialEq for SmallVec<T, N>
+where
+    T: PartialEq,
+{
+    fn eq(&self, other: &Self) -> bool {
+        self.as_slice() == other.as_slice()
+    }
+}
```
cmyr, line 122: maybe also `PartialEq`/`Eq` for `[T]`?

cmyr (approved):
I imagine this will be helpful in a variety of places where we're currently using vecs...

skrifa/src/collections.rs
dfrg: done! also added a Default impl

PR #1029, opened 2024-07-25
Add typo ci

rsheeter:
Shamelessly copied from @waywardmonkeys  https://github.com/linebender/kurbo/pull/355, updated `.typos.toml` to get it passing.

It can be a nuisance but I think it's a net win.

JMM

PR #1030, opened 2024-07-25
ci: Add a check for common typos.

waywardmonkeys:
False positives can be addressed by adding them into `.typos.toml`.


waywardmonkeys:
This is an alternative to #1029.

rsheeter:
LGTM, thank you!

PR #1032, opened 2024-07-25
Make `font-types` an optional dependenct in `int-set`

LaurenzV:
So that it can be used without pulling in a whole new dependency. I think in theory it would be nicer if the default were to not include it, but since many of the tests rely on it I've made it opt out instead. Perhaps it would be nice to also check it in CI.


cmyr:
we've finally gone ahead and chose expediency here, and made `int-set` part of `read-fonts`, so this is moot. Sorry!

PR #1033, opened 2024-07-25
[skrifa] tthint: don't base execution budget on glyph count

### LINKED ISSUE #936: Question regarding discrepancies when using hinted drawing mode
Hello,

I'm trying to use `skrifa` as a fuzzing tool to verify that my subsetter is working correctly. Everything works fine when I use the unhinted drawing mode, however, I'm facing some problems when using the hinted drawing mode. I'm facing the same problem when using subsets produced by `fonttools` instead of my own subsetter, so I don't think the issue is on my side.

In particular, the problem is that there seem to be slight variations in the drawn outlines depending on which glyphs are added to the subset (even though the glyphs shouldn't be related to each other at all! I verified this by comparing the `ttx` output of fonttools). So for example, I have a font where I'm drawing the glyph 1 using hinted mode. If I create a subset with `fonttools` using the glyphs 1-5 and I redraw glyph 1 in hinted mode, I get the same result. However, if the subset includes the glyphs 1-10 I suddenly get slightly different results for the same glyph!

So I wanted to ask whether this is expected behavior, i.e. whether it's possible that the number of glyphs somehow affects the result of hinting. If not, I can try to get a small example to working to demonstrate the issue.

dfrg:
The execution budget for the TrueType interpreter includes the glyph count as a limiting factor. This means that a font subset can fail to execute hinting instructions in cases where the full font succeeds, which seems entirely undesirable.

This patch simply removes the `min(100 * glyph_count)` limit. Note that this is a deviation from FreeType. We will succeed in cases where FT would fail but the reverse will never happen. We could keep a limiting factor that isn't dependent on glyph count but anything less than `100 * u16::MAX` _might_ cause us to fail when FT succeeds.

First commit contains a failing test, showcasing the problem.

edit: Fixes #936


dfrg:
Failing test in CI:
```
---- outline::tests::tthint_with_subset stdout ----
thread 'outline::tests::tthint_with_subset' panicked at skrifa/src/outline/mod.rs:1336:14:
called `Result::unwrap()` on an `Err` value: HintingFailed(HintError { program: Font, glyph_id: Some(GlyphId(1)), pc: 1011, opcode: Some(LOOPCALL), kind: ExceededExecutionBudget })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

cmyr (approved):
👍

PR #1034, opened 2024-07-25
Move int-set to read_fonts::collections::intset

cmyr:
This is gated on `std`. Chad keeps mentioning that we really only need alloc, but that's trickier, and we're already using `std` as a feature for closure in GSUB.

This also requires some changes to make the colr/cmap closure code compile; all that is now also gated on `std`.


read-fonts/src/collections.rs
```diff
@@ -0,0 +1,4 @@
+//! data structures useful for font work
```
dfrg, line 1:
```suggestion
//! Data structures useful for font work.
```
Hurts my eyes when public docs are missing capitalization and punctuation :)

read-fonts/src/collections.rs
```diff
@@ -0,0 +1,4 @@
...
+
+pub mod intset;
+pub use intset::IntSet;
```
dfrg, line 4: `std` uses `hash_set` and `btree_set` so I think we should follow with `int_set` just for consistency :shrug:

dfrg (approved):
Thanks for getting this done!

Per IM, we may want to take on an unconditional dependency on `alloc` and just provide this functionality without a feature gate but I think this is perfectly fine for now.

Left the two tiniest nits below.

read-fonts/src/collections.rs
cmyr: oh I went with `intset` _specifically_ because I thought that's what `std` does. 🙄

PR #1036, opened 2024-07-25
[int-set] Fix equality and hash for intset.

garretrieger:
Consider two sets with different membership types, but equivalent effective member sets to be equal (and update hash to match).


cmyr (approved):
👍

PR #1040, opened 2024-07-26
[int-set] Remove unnecessary generic param on Domain

cmyr:
This is redundant, since it is always equal to `Self`.


garretrieger (approved):

dfrg (approved):
Somehow we all missed this

PR #1041, opened 2024-07-26
[int-set] Impl From<[T; N]> for IntSet<T>

cmyr:
This is a common way to initialize construct sets (particularly for testing), and is implemented for both HashSet and BTreeSet; having it here will make it easier to use IntSet as a drop-in replacement for those types.


dfrg (approved):
👍

PR #1042, opened 2024-07-26
[int-set] Impl PartialOrd & Ord

cmyr:


read-fonts/src/collections/int_set/mod.rs
```diff
@@ -468,6 +468,27 @@ impl<T: Domain> Hash for IntSet<T> {
     }
 }
 
+impl<T: Domain + Ord> Ord for IntSet<T> {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        // there is probably room for optimization here, but it gets us working
```
garretrieger, line 473: Don't need to do this now but probably worth capturing in a comment, this is what a more efficient implementation probably entails:
1. Implement Ord on BitSet, then there we can implement it by comparing page element values to each other and avoid having to produce the set members which is comparatively much slower.
2. If membership types are the same we can use the BitSet ord impl.
3. If membership types are different then use iter_ranges() which will avoid excessive slowness for exclusive sets.

garretrieger (commented):
Can you add some tests, and if you're up for it add a new operation to the fuzzer which invokes cmp().

dfrg (approved):

cmyr:
oh funny this did have some basic tests and they got lost in a rebase somehow? Will take a look at the fuzzer stuff..

cmyr:
okay, added some basic tests. @garretrieger I tried to add a fuzzer operation, but the obvious implementation would basically be identical to the `Equals` op? Since if we're comparing the behaviour of intset agaisnt btreeset, if they're equal than `cmp` will also be identical? to test other orderings we would need to do something where we are constructing two different sets at once?

garretrieger:
> okay, added some basic tests. @garretrieger I tried to add a fuzzer operation, but the obvious implementation would basically be identical to the `Equals` op? Since if we're comparing the behaviour of intset agaisnt btreeset, if they're equal than `cmp` will also be identical? to test other orderings we would need to do something where we are constructing two different sets at once?

The two input sets 'a' and 'b' given to the op can be different from each other. The fuzzer input applies modification operations (eg. insert, remove) to one or the other based on the msb of the op code.

The CmpOp should look just like equals but call cmp on a and b and then check that the result of cmp is that same for the btree sets.

cmyr:
oh gotcha I had misread the signatures, added the fuzzing opt.

garretrieger (approved):

PR #1044, opened 2024-07-26
[skrifa] autohint: axis types

dfrg:
Adds some new types to describe hints along a single axis.

Also does some shuffling of modules to make room for more Latin specific code.


cmyr (approved):
👍

PR #1045, opened 2024-07-26
[skrifa] autohint: Latin segments

dfrg:
Implements the Latin segment computation and linking algorithms.

A segment is a "series of at least two consecutive points that are appropriately aligned along a coordinate axis."

The linking stage associates pairs of segments to form stems and identifies serifs with a post-process pass.


skrifa/src/outline/autohint/latin/segments.rs
```diff
@@ -0,0 +1,624 @@
+//! Latin segment computation and linking.
```
cmyr, line 1: "a segment is ...", 😇 (probably helpful for future readers of this file)

cmyr (approved):
one little idea inline :)

PR #1046, opened 2024-07-27
Add support for the `SVG` table

### LINKED ISSUE #1039: Support for SVG table
Would you be open for a PR that adds support for the `SVG` table? I would need it for a use case of mine. I've never worked with the codegen code so not sure how hard it is, but the table is very short and simple, so I could at least give it a shot.

LaurenzV:
I added a simple test case and also manually tried it with Noto Color Emoji and Twitter Color Emoji and it seems to work as expected, so I hope it's fine like that. Not sure if I got all the data types in the codegen template right, though.

Fixes #1039


read-fonts/src/tables/svg.rs
```diff
@@ -0,0 +1,135 @@
+//! The [SVG](https://learn.microsoft.com/en-us/typography/opentype/spec/svg) table
+
+use core::cmp::Ordering;
+
+include!("../../generated/generated_svg.rs");
+
+/// An [SVG document](https://learn.microsoft.com/en-us/typography/opentype/spec/svg). Is not
+/// guaranteed to be valid and might be compressed.
+pub struct SVGDocument<'a>(&'a [u8]);
```
dfrg, line 9: This type seems unnecessary to me since we don’t expect to add any additional functionality here. I suggest just returning the slice directly from `glyph_data` and adding a comment there describing the content.

resources/codegen_inputs/svg.rs
```diff
@@ -0,0 +1,39 @@
+#![parse_module(read_fonts::tables::svg)]
+
+/// [SVG](https://learn.microsoft.com/en-us/typography/opentype/spec/svg)
+#[tag = "SVG "]
+table SVG {
```
dfrg, line 5: We’ve discussed naming table types with the same capitalization as the tag but haven’t made that change yet so this should be `Svg` to match the current naming convention. The SVG in the other codegen’d types is fine. 

Also, just for consistency, the doc comment here should match the one on the module, e.g. “The SVG table” with the link on SVG.

dfrg (commented):
Happy to see new contributors working with codegen. Requested a few small changes but otherwise this looks great, thanks!

resources/codegen_inputs/svg.rs
```diff
@@ -0,0 +1,39 @@
+#![parse_module(read_fonts::tables::svg)]
+
+/// The [SVG](https://learn.microsoft.com/en-us/typography/opentype/spec/svg) table
+#[tag = "SVG "]
+table SVG {
```
dfrg, line 5:
```suggestion
table Svg {
```

Please change this and rerun codegen and then we'll get this merged.

LaurenzV: Sorry! I read the comment wrong. 😅

dfrg: No worries :)

LaurenzV:
I hope it's good now.

dfrg:
All good, thanks!

LaurenzV:
Thanks to you!

PR #1047, opened 2024-07-27
[skrifa] autohint: compute Latin widths

dfrg:
Computes horizontal and vertical stem widths and metrics (edge distance threshold and standard width) for the Latin writing system.

Based on #1045 which should land first.


skrifa/src/outline/autohint/axis.rs
```diff
@@ -22,23 +27,43 @@ pub struct Axis {
...
+        } else {
+            self.major_dir = Direction::Left;
+        }
+        if orientation == Some(Orientation::Clockwise) {
+            if dim == Self::HORIZONTAL {
+                self.major_dir = Direction::Down;
+            } else {
+                self.major_dir = Direction::Right;
+            }
+        }
```
cmyr, line 57: totally academic but I'd be tempted to write this as, `let self.major_dir = match (dim, orientation) { .. }`

cmyr (approved):

skrifa/src/outline/autohint/axis.rs
dfrg: I didn't even try a match because of the multiple catch-all conditions but it does look much better. Fixed

PR #1048, opened 2024-07-28
[font-types] Improve docs.

waywardmonkeys:
This adds more cross-linking, some formatting fixes, as well as bits of new content.

Also, `clippy::doc_markdown` is now enabled for this crate to keep it cleaner. This can't be enabled in the workspace yet.

A clippy configuration is provided that ignores `OpenType` and `PostScript`.


dfrg (approved):
Looks good, thanks!

PR #1049, opened 2024-07-28
ci: More complete check for `no_std` builds

### LINKED ISSUE #1006: no-std builds are not working and are checked incorrectly
CI checks `no-std` builds for `font-types`, `read-fonts` and `skrifa`. However, it uses the default target which has an `std`, so this isn't a comprehensive check.

These jobs need to specify a target and that needs to not have an `std` implementation. I tend to use one of `thumbv7m-none-eabi`, `riscv64imac-unknown-none-elf`, or `x86_64-unknown-none`.

The `rust-toolchain` action will have to install this target:

```
  check-no-std:
    name: cargo check no std
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: install stable toolchain
        uses: dtolnay/rust-toolchain@stable
            with:
                targets: riscv64imac-unknown-none-elf

      - name: cargo check font-types
        run: cargo check --manifest-path=font-types/Cargo.toml --no-default-features --target riscv64imac-unknown-none-elf

      - name: cargo check read-fonts
        run: cargo check --manifest-path=read-fonts/Cargo.toml --no-default-features --features=libm --target riscv64imac-unknown-none-elf

      - name: cargo check skrifa
        run: cargo check --manifest-path=skrifa/Cargo.toml --no-default-features --features=libm --target riscv64imac-unknown-none-elf
```

Once you do that, you'll see that `int-set` doesn't build for `no-std` and therefore `read-fonts` doesn't build for `no-std`.

waywardmonkeys:
* Use a `no_std` target to be sure that nothing is linking with `std`.
* Use `cargo build` rather than `cargo check` to go all the way through the linking stage.
* Use `-p` rather than `--manifest-path` as it is shorter and should still work here.

Fixes #1006.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -101,15 +101,18 @@ jobs:
 
       - name: install stable toolchain
         uses: dtolnay/rust-toolchain@stable
+        with:
+          # Use a target without `std` to make sure we don't link to `std`
+          target: thumbv7em-none-eabihf
 
-      - name: cargo check font-types
-        run: cargo check --manifest-path=font-types/Cargo.toml --no-default-features
+      - name: cargo build font-types
+        run: cargo build -p font-types --target thumbv7em-none-eabihf --no-default-features
 
-      - name: cargo check read-fonts
-        run: cargo check --manifest-path=read-fonts/Cargo.toml --no-default-features --features=libm
+      - name: cargo build read-fonts
+        run: cargo build -p read-fonts --target thumbv7em-none-eabihf --no-default-features --features=libm
 
-      - name: cargo check skrifa
-        run: cargo check --manifest-path=skrifa/Cargo.toml --no-default-features --features=libm
+      - name: cargo build skrifa
+        run: cargo build -p skrifa --target thumbv7em-none-eabihf --no-default-features --features=libm
 
   check-wasm:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -101,15 +101,18 @@ jobs:
...
 
-      - name: cargo check read-fonts
-        run: cargo check --manifest-path=read-fonts/Cargo.toml --no-default-features --features=libm
+      - name: cargo build read-fonts
+        run: cargo build -p read-fonts --target thumbv7em-none-eabihf --no-default-features --features=libm
 
-      - name: cargo check skrifa
-        run: cargo check --manifest-path=skrifa/Cargo.toml --no-default-features --features=libm
+      - name: cargo build skrifa
+        run: cargo build -p --target thumbv7em-none-eabihf --no-default-features --features=libm
```
dfrg, line 115: Looks like we're missing a "skrifa" here

dfrg (approved):
Glad to have a more robust test for this. Looks good to me once the checks are all green.

.github/workflows/rust.yml
waywardmonkeys: That's an embarrassing omission! Fixed now.

PR #1052, opened 2024-07-29
[read-fonts] Fvar::user_to_normalized()

dfrg:
Moves coordinate conversion logic from skrifa to read-fonts to support this functionality in harfruzz.

### DIFF
--- a/read-fonts/src/tables/fvar.rs
+++ b/read-fonts/src/tables/fvar.rs
@@ -32,8 +32,13 @@ impl<'a> Fvar<'a> {
     /// * If no user coordinate for an axis is provided, the associated
     ///   coordinate is set to the normalized value 0.0, representing the
     ///   default location in variation space.
+    /// * The length of `normalized_coords` should equal the number of axes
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/fvar.rs
```diff
@@ -17,6 +18,55 @@ impl<'a> Fvar<'a> {
...
+    /// * User coordinate values are clamped to the range of their associated
+    ///   axis before normalization.
+    /// * If more than one user coordinate is provided for the same tag, the
+    ///   last one is used.
+    /// * If no user coordinate for an axis is provided, the associated
+    ///   coordinate is set to the normalized value 0.0, representing the
+    ///   default location in variation space.
+    ///
+    /// If the `avar` table is provided, applies remapping of coordinates.
+    pub fn user_to_normalized(
```
cmyr, line 37: we should document the expectations of the `normalized_coords` argument. I assume we expect it to be of a given length, and that we will overwrite the entire contents?

cmyr (approved):

PR #1053, opened 2024-07-29
Use workspace inheritance for package properties.

waywardmonkeys:
This adds `[workspace.package]` to the root `Cargo.toml` with (initially) `edition`, `license`, and `repository` properties. It then inherits them into each package in the repository.

This gives `fuzz` a license that wasn't previously stated in the Cargo.toml and `font-codegen`, `fuzz`, and `otexplorer` get a `repository` property that they didn't have before.

The license in the `Cargo.toml` is now a valid SPDX license expression rather than using a form deprecated by cargo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,11 @@ members = [
     "fuzz",
 ]
 
+[workspace.package]
+edition = "2021"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/googlefonts/fontations"
+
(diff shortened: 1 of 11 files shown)


waywardmonkeys:
This will make it easier when adding and updating data like an MSRV since we can express it once in the root `Cargo.toml` and then inherit it like these properties.

cmyr (approved):
This is a nice improvement, thanks!

PR #1059, opened 2024-07-31
[skrifa] autohint: Latin metrics

dfrg:
Computes unscaled (and then scaled) style-specific metrics for Latin-like hinting.

This basically ties all the previous work together then applies a scaling factor and adjusts and selects active blue zones.


cmyr (approved):
nice, feels like we're getting close to something here!

PR #1060, opened 2024-07-31
[klippa] subset glyf/head/loca/hmtx/hhea

qxliu76:


klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,298 @@
...
+        return out;
+    };
+    let num_coords = num_coords.get() + 1;
+    let mut coord_bytes: usize = 0;
+    let mut coords_with_flags: u16 = 0;
+
+    let glyph_data = g.glyph_data();
+    let length = glyph_data.len();
+    let mut i: usize = 0;
+    while i < length {
```
garretrieger, line 135: Can you move the code for trimming padding to a helper function. That will make it clear what this loop is doing and also should help keep this function a bit more manageable once additional hint stripping and overlap flag setting is added.

garretrieger (approved):

klippa/src/glyf_loca.rs
qxliu76: Done for Simple Glyph. For composite glyph, it seems more efficient to do trim padding/hints removal/overlap flag together within one loop.

garretrieger: Sounds good.

garretrieger (approved):

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+};
+
+pub fn subset_glyf_loca(
+    plan: &Plan,
+    font: &FontRef,
+    builder: &mut FontBuilder,
+) -> Result<(), SubsetError> {
+    let Ok(glyf) = font.glyf() else {
+        return Err(SubsetTableError(Glyf::TAG));
+    };
```
cmyr, line 27: not necessary but you could write this as, 
```rust
let glyf = font.glyf().ok_or(SubsetTableError(Glyf::TAG)?;
```

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+        }
+        let instruction_len = u16::from_be_bytes([out[i], out[i + 1]]);
+        i += 2 + instruction_len as usize;
+    }
+
+    out.truncate(i);
+    out
+}
+
+fn trim_simple_glyph_padding(glyph_data: &[u8], num_coords: u16) -> usize {
```
cmyr, line 202: I think this deserves a bit of explanation, it isn't clear to me what padding we're expecting to find? is it only the padding required to make the offset an even number?

I'd also document the meaning of the return value.

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+    i
+}
+
+fn subset_head(head: &Head, loca_format: u8) -> Result<Vec<u8>, SubsetError> {
+    let mut out = Vec::new();
+    out.extend_from_slice(head.offset_data().as_bytes());
+
+    let Some(index_loca_format) = out.get_mut(50..52) else {
+        return Err(SubsetTableError(Head::TAG));
+    };
```
cmyr, line 254: if we have a `Head` table at this point it mens that we've already validated that it is the correct length, so this branch is unreachable.

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+    if num_coords != coords_with_flags {
+        return 0;
+    }
+    i += coord_bytes;
+    i
+}
+
+fn subset_head(head: &Head, loca_format: u8) -> Result<Vec<u8>, SubsetError> {
+    let mut out = Vec::new();
+    out.extend_from_slice(head.offset_data().as_bytes());
```
cmyr, line 250: just `let mut out = head.offset_data().as_bytes().to_owned()` should do it?

klippa/src/glyf_loca.rs
cmyr: and given that, this can all be written as,

`out.get_mut(50..52).unwrap().copy_from_slice(&[0, loca_format]);`

klippa/src/head.rs
```diff
@@ -0,0 +1,22 @@
...
+    font: &FontRef,
+    _plan: &Plan,
+    builder: &mut FontBuilder,
+) -> Result<(), SubsetError> {
+    let Ok(head) = font.head() else {
+        return Err(SubsetTableError(Head::TAG));
+    };
+
+    let mut out = Vec::with_capacity(head.offset_data().len());
+    out.extend_from_slice(head.offset_data().as_bytes());
```
cmyr, line 18: ditto, I'd just write `let mut out = head.offset_data().as_bytes().to_owned()`.

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+fn subset_glyph(glyph: &Glyph, plan: &Plan) -> Vec<u8> {
+    //TODO: support set_overlaps_flag and drop_hints
+    match glyph {
+        Composite(comp_g) => subset_composite_glyph(comp_g, plan),
+        Simple(simple_g) => subset_simple_glyph(simple_g),
+    }
+}
+
+// TODO: drop_hints and set_overlaps_flag
+fn subset_simple_glyph(g: &SimpleGlyph) -> Vec<u8> {
```
cmyr, line 122: it feels to me like there could be some modest gains to be had here by not doing as much intermediate allocation... for instance maybe we could just be writing all the glyphs into a buffer directly? we'd need to make sure we were padding, and we might need to rewrite that buffer to remove the padding if we ended up choosing the long loca format, but it doesn't feel that tricky, although we'd need to profile to know if it was meaningful.

klippa/src/glyf_loca.rs
```diff
@@ -0,0 +1,304 @@
...
+                    subset_glyphs.push(Vec::new());
+                    continue;
+                };
+                let subset_glyph = subset_glyph(&glyph, plan);
+                let trimmed_len = subset_glyph.len();
+                max_offset += padded_size(trimmed_len) as u32;
+                subset_glyphs.push(subset_glyph);
+            }
+            _ => {
+                return Err(SubsetTableError(Glyf::TAG));
```
cmyr, line 55: I suspect at some point you are going to be frustrated by how general these errors are. In this case I would be tempted to have a `MissingGlyph(GylphId)`  error, or something?

klippa/src/hmtx.rs
```diff
@@ -1,48 +1,79 @@
...
+    let Ok(hhea) = font.hhea() else {
+        builder.add_raw(Hmtx::TAG, hmtx_out);
+        return Ok(());
+    };
+
+    let mut hhea_out = Vec::with_capacity(hhea.offset_data().len());
+    hhea_out.extend_from_slice(hhea.offset_data().as_bytes());
+    let Some(index_num_h_metrics) = hhea_out.get_mut(34..36) else {
+        return Err(SubsetTableError(Hhea::TAG));
+    };
```
cmyr, line 59: this should not be reachable if hhea exists.

klippa/src/lib.rs
```diff
@@ -283,33 +280,85 @@ pub enum SubsetError {
...
+}
 
-    builder.copy_missing_tables(font);
+pub fn estimate_subset_table_size(font: &FontRef, table_tag: Tag, plan: &Plan) -> usize {
+    let Some(table_data) = font.data_for_tag(table_tag) else {
+        return 0;
+    };
+
+    let table_len = table_data.len();
+    let bulk: usize = 8192;
```
cmyr, line 347: this looks like a `const`?

cmyr (approved):
looks good, some little notes inline!

klippa/src/glyf_loca.rs
qxliu76: thanks! This is very useful

klippa/src/glyf_loca.rs
qxliu76: These allocations are not hot in profiling and we're fine with current performance(so will explore ways to improve it in the future). Also when drop_hints option is enabled, it will become tricky if we write all glyphs together.

PR #1058, opened 2024-07-30
Change IntSet::len() to return u64 instead of usize.

garretrieger:
A u32 containing intset can have up to u32::MAX + 1 elements in it which can overflow a usize when usize is a u32. Fixes #1057

PR #1061, opened 2024-08-01
ci: Improve wasm32 checks

waywardmonkeys:
Switch to using `cargo build` instead of `cargo check` as that will pick up important errors, such was `arithemetic_overflow`.

Additionally, test all 4 library crates. This makes sure that `skrifa` is also checked. We do all 4 separately so that an error in one is more clear.


waywardmonkeys:
This will fail until #1058 is landed. I will rebase this on top of that once it lands.

waywardmonkeys:
This is now rebased on top of the fix from #1058, so it should pass now once it is run.

cmyr:
Looks like a solid improvement, thanks!

PR #1066, opened 2024-08-01
[IFT] Rework Patch Map Format 1 API to allow for partial reading of the table.

garretrieger:
Instead of providing an API which lists all entries encoded in a format 1 table instead provide an API which provides all entries which intersect a subset definition.  This allows for a more optimized parsing implementation of format 1 to be used which only looks at the data from relevant parts of the tables.

Also this updates error handling in the format 1 parsing to be compliant with https://github.com/w3c/IFT/pull/199. In many cases invalid data is ignored instead of failing with an error. This allows consistency of the optimized implementation with an implementation that parses the whole table.


read-fonts/src/tables/ift.rs
```diff
@@ -64,27 +64,28 @@ impl<'a> PatchMapFormat1<'a> {
...
+impl<'a> FeatureMap<'a> {
+    pub fn entry_records_size(&self, max_entry_index: u16) -> Result<usize, ReadError> {
+        let field_width = if max_entry_index < 256 { 1 } else { 2 };
+        let mut num_bytes = 0usize;
+        for record in self.feature_records().iter() {
+            let record = match record {
+                Ok(record) => record,
+                Err(err) => return Err(err.clone()),
+            };
+            num_bytes += record.entry_map_count().get() as usize * field_width * 2;
```
dfrg, line 86:
```suggestion
            num_bytes += record?.entry_map_count().get() as usize * field_width * 2;
```

skrifa/src/patchmap.rs
```diff
@@ -3,223 +3,295 @@
...
+    let mut cumulative_entry_map_count = 0;
+    let mut largest_tag: Option<Tag> = None;
+    loop {
+        let (Some(tag), Some(record)) = (next_tag, next_record.as_ref()) else {
+            break;
+        };
+        let record = match record {
+            Ok(record) => record,
+            Err(err) => return Err(err.clone()),
+        };
```
dfrg, line 197:
```suggestion
        let Some((tag, record)) = next_tag.zip(next_record.clone()) else {
            break;
        };
        let record = record?;
```

skrifa/src/patchmap.rs
```diff
@@ -3,223 +3,295 @@
...
-    fn apply_uri_template(uri_template: &str, _entry_index: usize) -> String {
-        // TODO(garretrieger): properly implement this, may deserve to go into it's own module.
-        uri_template.to_string()
-    }
+    Ok(())
+}
 
-    // TODO(garretrieger): add template application method that takes a string entry id (will be needed for format 2)
+#[derive(Clone, Eq, PartialEq, Debug, Hash, Copy)]
+pub enum PatchEncoding {
```
dfrg, line 250: A doc comment with a brief description and a link to the spec would be nice here.

skrifa/src/patchmap.rs
```diff
@@ -3,223 +3,295 @@
...
 }
 
-#[derive(Clone, Eq, PartialEq, Hash)]
-pub struct Entry {
-    pub patch_uri: PatchUri,
-    pub codepoints: IntSet<u32>,
-    pub feature_tags: BTreeSet<Tag>,
-    pub compatibility_id: [u32; 4],
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct PatchUri {
```
dfrg, line 274: Same here wrt doc comment. Template seems to imply that the string contains a pattern that is replaced with the index?

dfrg (approved):
LGTM with a few suggestions to clean up some of the error checking and a request for some additional doc comments.

skrifa/src/patchmap.rs
garretrieger: That's right, added a comment and linked to the relevant parts of the spec.

PR #1067, opened 2024-08-02
[skrifa] autohint: Latin edge hinting

dfrg:
Here we actually align edges to the pixel grid. This is the final step before applying the adjusted edges to the original outline points.

Based on #1063


cmyr (approved):
Nice to see this coming together!

PR #1068, opened 2024-08-02
[skrifa] autohint: apply hints to outline

dfrg:
Apply edge hints to an outline.

This happens in three passes:
1. Align points that are directly attached to edges.
2. Interpolate non-weak points that were not touched by the previous pass.
3. Interpolate remaining (weak) points.

The final result is a fully hinted outline.

Builds on #1067


cmyr (approved):
I didn't go super super in depth with this but I it looks sane and I'm reassured by our testing framework.

PR #1070, opened 2024-08-06
Move Pen from font-types to skrifa

dfrg:
This adds a new `contour_to_path` function that emits a single contour to a pen. The signature for this function is a bit unfortunate because it requires both an iterator of all points _and_ the last point as a separate argument. This is necessary to support FreeType style path conversion where we start with the last point if the first is an off-curve.

The intention is to allow for conversion of autohinted outlines to paths without allocating and copying new buffers.

Moves all the path conversion code into a new module under `glyf` because the size of that file was becoming unwieldy.

Adds a couple new unit tests to capture an additional edge case and also tested with fauntlet against all of gfonts and noto to ensure it doesn't break anything.

~~We should consider moving the pen stuff to skrifa and making this code internal to that crate.~~ Now does this, see below.


dfrg:
This still does the changes to the path conversion above but now also:

* moves `font_types::Pen` to `skrifa::outline::OutlinePen` (which was an existing alias).
* moves `font_types::PenCommand` to `skrifa::outline::pen::PathElement`.
* adds `OutlinePen` impls for `NullPen`, `SvgPen`, and `Vec<PathElement>` so we can stop writing these all over the place.

... and contains a lot more churn than expected.

cmyr (approved):
Looks good, I agree this is a better separation of concerns. Shame it's another breaking change in font-types, but hopefully the last for quite a while?

dfrg:
> Shame it's another breaking change in font-types, but hopefully the last for quite a while?
Let's hope so? I don't foresee any near term changes after this.

PR #1071, opened 2024-08-06
[write-fonts] remove font_types::Pen usage

dfrg:
Deletes the entire `pen` module which is unused after we added the required functionality to `kurbo`.

Reworks two tests that used `to_path()` to generate a `BezPath` for round-tripping. I just inlined the SVG style path data for these.


cmyr (approved):
this always felt a bit funny to me, happy to see it nix'd!

PR #1073, opened 2024-08-07
[skrifa] Do not issue COLRv1 draw command for empty sweep

### LINKED ISSUE #1017: Possible optimization for COLR glyphs for sweep gradients with same start/end angle
From the COLR spec on sweep gradients:

```
If the color line's extend mode is reflect or repeat and start and end angle are equal, nothing shall be drawn.
```

From what I can see, this seems to apply for glyph 183 from the COLRv1 test glyphs font:

```
<BaseGlyphPaintRecord index="161">
  <BaseGlyph value="sweep_coincident_angles_forward_blue_red_reflect"/>
  <Paint Format="10"><!-- PaintGlyph -->
    <Paint Format="8"><!-- PaintSweepGradient -->
      <ColorLine>
        <Extend value="reflect"/>
        <!-- StopCount=4 -->
        <ColorStop index="0">
          <StopOffset value="0.0"/>
          <PaletteIndex value="4"/>
          <Alpha value="1.0"/>
        </ColorStop>
        <ColorStop index="1">
          <StopOffset value="0.3333"/>
          <PaletteIndex value="7"/>
          <Alpha value="1.0"/>
        </ColorStop>
        <ColorStop index="2">
          <StopOffset value="0.6667"/>
          <PaletteIndex value="8"/>
          <Alpha value="1.0"/>
        </ColorStop>
        <ColorStop index="3">
          <StopOffset value="1.0"/>
          <PaletteIndex value="0"/>
          <Alpha value="1.0"/>
        </ColorStop>
      </ColorLine>
      <centerX value="500"/>
      <centerY value="600"/>
      <startAngle value="90.0"/>
      <endAngle value="90.0"/>
    </Paint>
    <Glyph value="circle_r350"/>
  </Paint>
</BaseGlyphPaintRecord>
```

Nevertheless, when I draw that glyph using `skrifa`, I still get the command to draw that gradient. Would it be possible to implement this edge case in skrifa, so that if the sweep gradient has the same angles and repeat/reflect, the command isn't emitted in the first place? Or is there a reason why this isn't handled there (or am I missing something else)?

P.S.: For context, I'm trying to use `skrifa` to convert COLR glyphs into PDF objects, so I was trying to use the [Skia reference image from fontations](https://chromium-review.googlesource.com/c/chromium/src/+/5614312/16/third_party/blink/web_tests/platform/linux/virtual/text-antialias/colrv1-expected.png) for comparison. There, it seems like with spread mode pad it's drawn as expected, while with repeat/reflect it instead turns into a solid purple? 

image

Do you know why this is the case, since, at least according to my understanding of the spec, nothing should be drawn at all?

drott:
If even after normalization the angles match and the sector is empty, do not issue a draw command. The spec has an extra sentence describing that situation. Fixes a difference between FreeType and Fontations COLRv1 backends in Skia as well, see GM_colrv1_sweep_coincident gm test. Covered by that test and the sweep-coincident glyphs in the COLRv1 test font.

Fixes #1017.


dfrg (approved):
Thanks for the fix and the analysis in #1017. LGTM

PR #1075, opened 2024-08-08
[skrifa] autohint: style classes and script groups

dfrg:
Lots of unfortunate churn to generate the extended set of style classes on top of script classes. These will be used in cases where OpenType layout is processed to categorize glyphs.

Also assigns each script to a group which will be used to select different behaviors in the various autohinting functions.

No real functional changes here aside from the python script to generate these things.

This is in preparation for building the actual glyph style map so we can determine how to process a glyph without manually calling all the functions as we currently do in tests.


cmyr (approved):
🫡

PR #1076, opened 2024-08-08
[skrifa] autohint glyph style map

dfrg:
Replaces the `compute_mapped_styles` function with a new `GlyphStyleMap` type that computes glyph styles and also keeps track of the styles that are actually used so we can allocate a minimally sized style metrics array.

Based on #1075

### DIFF
--- a/skrifa/src/outline/autohint/style.rs
+++ b/skrifa/src/outline/autohint/style.rs
@@ -1,7 +1,7 @@
 //! Styles, scripts and glyph style mapping.
 
-use crate::GlyphId;
-use alloc::boxed::Box;
+use crate::{charmap::Charmap, GlyphId};
+use alloc::vec::Vec;
 use raw::types::Tag;
 
 /// Defines the script and style associated with a single glyph.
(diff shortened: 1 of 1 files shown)


skrifa/src/outline/autohint/style.rs
```diff
@@ -51,6 +51,121 @@ impl Default for GlyphStyle {
...
+                        // This the first time we've seen this style so record
+                        // it in the metrics map
+                        map.metrics_map[style_ix] = map.metrics_count;
+                        map.metrics_count += 1;
+                    }
+                }
+                last_range = Some((ix, range));
+            }
+        }
+        // TODO: Step 3: compute script based coverage
```
cmyr, line 124: are these actual todos? or were they once?

cmyr (approved):
Looks like a useful little refactoring.

skrifa/src/outline/autohint/style.rs
dfrg: These are new and just markers for where I need to add "shaping" support, hopefully next week.

PR #1077, opened 2024-08-08
[skrifa] autohint: unscaled style metrics set

dfrg:
This adds an `UnscaledStyleMetricsSet` type that manages the collection of style metrics that are necessary based on the glyph set of a single font.

As discussed, this has two flavors: precomputed and lazy.

The lazy variant guards the whole collection with a single `RwLock` so that our state can remain nominally immutable while still being threadsafe. We could use a lock per entry, but we only hold the lock long enough to either load or store the data (the code chooses to race on a miss so the expensive computation is done outside the lock).

The laziness is limited to `std` builds since we need access to `std::sync::RwLock`.

Based on #1076


skrifa/src/outline/autohint/metrics.rs
```diff
@@ -68,6 +72,71 @@ pub(crate) struct UnscaledStyleMetrics {
...
+
+impl UnscaledStyleMetricsSet {
+    /// Creates a precomputed style metrics set containing all metrics
+    /// required by the glyph map.
+    pub fn precomputed(font: &FontRef, coords: &[F2Dot14], style_map: &GlyphStyleMap) -> Self {
+        let mut vec = Vec::with_capacity(style_map.metrics_count());
+        for style in style_map.metrics_styles() {
+            let metrics = super::latin::compute_unscaled_style_metrics(font, coords, style);
+            vec.push(metrics);
+        }
```
cmyr, line 93: any reason to prefer this to map/collect?

cmyr (approved):
looks good!

skrifa/src/outline/autohint/metrics.rs
dfrg: `GlyphStyleMap::metrics_styles()` doesn't produce an exact size iterator and I'm trying to avoid over-allocating memory whenever possible in this code.

rsheeter: That is non-obvious enough to merit a comment

dfrg: Will change to extend/map and add a comment noting why we don't collect

PR #1078, opened 2024-08-09
[klippa] add subset_flags options and support retain-gids/notdef-outline/drop_hints/set_overlaps_flags

qxliu76:


klippa/src/glyf_loca.rs
```diff
@@ -51,33 +66,90 @@ pub fn subset_glyf_loca(
...
+            last += 1;
         }
     } else {
-        loca_out.extend_from_slice(&0_u32.to_be_bytes());
         let mut offset: u32 = 0;
-        for g in &subset_glyphs {
-            offset += g.len() as u32;
-            loca_out.extend_from_slice(&offset.to_be_bytes());
+        let mut value = 0_u32.to_be_bytes();
+        for ((new_gid, _), i) in zip(&plan.new_to_old_gid_list, 0u16..) {
```
cmyr, line 132: conventionally you wouldn't import this, and would just call `iter1.zip(iter2)`.

klippa/src/glyf_loca.rs
```diff
@@ -30,10 +36,19 @@ pub fn subset_glyf_loca(
     let mut subset_glyphs = Vec::with_capacity(num_output_glyphs);
     let mut max_offset: u32 = 0;
 
-    //TODO: support not_def_outline and drop_hints
-    for (_new_gid, old_gid) in &plan.new_to_old_gid_list {
+    for (new_gid, old_gid) in &plan.new_to_old_gid_list {
         match loca.get_glyf(*old_gid, &glyf) {
             Ok(g) => {
+                if *old_gid == GlyphId::from(0u32)
+                    && *new_gid == GlyphId::from(0u32)
```
cmyr, line 43:
```suggestion
                if *old_gid == GlyphId::NOTDEF
                    && *new_gid == GlyphId::NOTDEF
```

klippa/src/glyf_loca.rs
```diff
@@ -121,12 +184,39 @@ fn subset_simple_glyph(g: &SimpleGlyph) -> Vec<u8> {
...
         return out;
     };
+    out.extend_from_slice(header_slice);
+
+    if plan
+        .subset_flags
+        .contains(SubsetFlags::SUBSET_FLAGS_NO_HINTING)
+    {
+        out[header_len - 2] = 0;
+        out[header_len - 1] = 0;
```
cmyr, line 200: I think this would benefit from a comment like, 'set instructionLength field to 0'

klippa/src/glyf_loca.rs
```diff
@@ -121,12 +184,39 @@ fn subset_simple_glyph(g: &SimpleGlyph) -> Vec<u8> {
...
+        let instruction_end = header_len + g.instruction_length() as usize;
+        let Some(instruction_slice) = glyph_bytes.get(header_len..instruction_end) else {
+            return Vec::new();
+        };
+        out.extend_from_slice(instruction_slice);
+    }
+
+    let Some(trimmed_slice) = glyph_data.get(0..i) else {
+        return Vec::new();
+    };
```
cmyr, line 211: again I don't think this is reachable if we got this far? But it does raise the question of whether or not we want to be reporting errors...

klippa/src/glyf_loca.rs
```diff
@@ -143,10 +233,31 @@ fn subset_composite_glyph(g: &CompositeGlyph, plan: &Plan) -> Vec<u8> {
...
         if flags.contains(CompositeGlyphFlags::WE_HAVE_INSTRUCTIONS) {
             we_have_instructions = true;
+            if plan
+                .subset_flags
+                .contains(SubsetFlags::SUBSET_FLAGS_NO_HINTING)
+            {
+                flags.remove(CompositeGlyphFlags::WE_HAVE_INSTRUCTIONS);
+                out.get_mut(i..i + 2)
+                    .unwrap()
+                    .clone_from_slice(&flags.bits().to_be_bytes());
```
cmyr, line 247: this should be `copy_from_slice`?

klippa/src/hmtx.rs
```diff
@@ -60,15 +72,25 @@ pub fn subset_hmtx_hhea(
...
-    let num_long_metrics = gids.len().min(0xFFFF) as usize;
-    let last_gid = gids.last().unwrap();
-    let last_advance = hmtx.advance(last_gid).unwrap();
+fn compute_new_num_h_metrics(
+    hmtx: &Hmtx,
+    gid_set: &IntSet<GlyphId>,
+    num_output_glyphs: usize,
+) -> usize {
+    let mut num_long_metrics = num_output_glyphs.min(0xFFFF) as u32;
+    let last_gid = num_long_metrics - 1;
```
cmyr, line 81: is it possible to get this far and have `num_output_glyphs` be zero?

klippa/src/hmtx.rs
```diff
@@ -60,15 +72,25 @@ pub fn subset_hmtx_hhea(
...
-        .rev()
-        .take_while(|gid| hmtx.advance(*gid).unwrap() == last_advance)
-        .count();
-    (num_long_metrics - num_skippable_glyphs + 1).max(1)
+    while num_long_metrics > 1 {
+        let gid = GlyphId::from(num_long_metrics - 2);
+        let mut advance = 0;
+        if gid_set.contains(gid) {
+            advance = hmtx.advance(gid).unwrap();
+        }
```
cmyr, line 89: why mutate advance? Can't we just set it directly, like

```rust
let advance = gid_set.contains(gid)
    .then(|| hmtx.advance(gid).unwrap())
    .unwrap_or(0);
```

klippa/src/lib.rs
```diff
@@ -356,7 +467,19 @@ pub fn estimate_subset_table_size(font: &FontRef, table_tag: Tag, plan: &Plan) -
...
+    if plan
+        .subset_flags
+        .contains(SubsetFlags::SUBSET_FLAGS_RETAIN_GIDS)
+    {
+        if table_tag == Cff::TAG {
+            //Add some extra room for the CFF charset
+            bulk += src_glyphs * 16;
+        } else if table_tag == Cff2::TAG {
+            // Just extra CharString offsets
+            bulk += src_glyphs * 4;
```
cmyr, line 479: I think it's useful to define named constants for these magic numbers, as a form of documentation. What are 16 and 4 the lengths of?

cmyr (approved):
looks good, left some notes inline. :)

klippa/src/lib.rs
qxliu76: unfortunately I'm also not aware of what these numbers are for, I just ported them from HB.:)

klippa/src/hmtx.rs
qxliu76: no, we make sure to at least insert .notdef glyph during planning.

klippa/src/glyf_loca.rs
qxliu76: thanks!

klippa/src/hmtx.rs
qxliu76: new learning:) Thanks!

klippa/src/glyf_loca.rs
qxliu76: It might be reachable if something wrong in calculating trimmed length

klippa/src/lib.rs
rsheeter: I think it's worth the investment to figure out the answer and document in the form of constants. We presumably have code that puts something of the indicated size in place somewhere, no?

rsheeter: Per f2f, ok to file an issue and land this. @behdad might know off the top of his head.

klippa/src/lib.rs
```diff
@@ -37,6 +39,91 @@ const MAX_NESTING_LEVEL: u8 = 64;
...
+
+    //If set hinting instructions will be dropped in the produced subset.
+    //Otherwise hinting instructions will be retained.
+    pub const SUBSET_FLAGS_NO_HINTING: Self = Self(0x0001);
+
+    //If set glyph indices will not be modified in the produced subset.
+    //If glyphs are dropped their indices will be retained as an empty glyph.
+    pub const SUBSET_FLAGS_RETAIN_GIDS: Self = Self(0x0002);
+
+    //If set and subsetting a CFF font the subsetter will attempt to remove subroutines from the CFF glyphs.
```
garretrieger, line 57: Can you add a note to all of the unimplemented flags that that are not implemented yet.

garretrieger (approved):

klippa/src/lib.rs
behdad: The CharStrings are stored in an INDEX object. While empty CharStrings take no room, their offset is still stored. The offsets each take at most 4 bytes, hence the value here.

For CFF1, we add more, because the CharSet stores glyph names... These are all heuristics to reduce the need for rework with a larger buffer.

qxliu76: Thank you!

PR #1079, opened 2024-08-10
[skrifa] FontRef instead of TableProvider

dfrg:
Per IM, changes skrifa to use `FontRef` directly rather than `TableProvider`.

Extracts some shared functionality into a `BaseOutlines` type for outlines in preparation for more autohinting work but no real functional changes.

### DIFF
--- a/read-fonts/src/tables/postscript/index.rs
+++ b/read-fonts/src/tables/postscript/index.rs
@@ -8,6 +8,7 @@ use crate::codegen_prelude::*;
 /// Common type for uniform access to CFF and CFF2 index formats.
 #[derive(Clone)]
 pub enum Index<'a> {
+    Empty,
     Format1(Index1<'a>),
     Format2(Index2<'a>),
 }
@@ -28,6 +29,7 @@ impl<'a> Index<'a> {
     /// Returns the number of objects in the index.
     pub fn count(&self) -> u32 {
         match self {
+            Self::Empty => 0,
             Self::Format1(ix) => ix.count() as u32,
             Self::Format2(ix) => ix.count(),
         }
@@ -51,6 +53,7 @@ impl<'a> Index<'a> {
     /// Returns the total size in bytes of the index table.
     pub fn size_in_bytes(&self) -> Result<usize, ReadError> {
         match self {
+            Self::Empty => Ok(0),
             Self::Format1(ix) => ix.size_in_bytes(),
             Self::Format2(ix) => ix.size_in_bytes(),
         }
@@ -59,6 +62,7 @@ impl<'a> Index<'a> {
     /// Returns the offset at the given index.
     pub fn get_offset(&self, index: usize) -> Result<usize, Error> {
         match self {
+            Self::Empty => Err(ReadError::OutOfBounds.into()),
             Self::Format1(ix) => ix.get_offset(index),
             Self::Format2(ix) => ix.get_offset(index),
         }
@@ -67,6 +71,7 @@ impl<'a> Index<'a> {
     /// Returns the data for the object at the given index.
     pub fn get(&self, index: usize) -> Result<&'a [u8], Error> {
         match self {
+            Self::Empty => Err(ReadError::OutOfBounds.into()),
             Self::Format1(ix) => ix.get(index),
             Self::Format2(ix) => ix.get(index),
         }
@@ -82,6 +87,12 @@ impl<'a> From<Index1<'a>> for Index<'a> {
 impl<'a> From<Index2<'a>> for Index<'a> {
     fn from(value: Index2<'a>) -> Self {
         Self::Format2(value)
+    }
+}
+
+impl Default for Index<'_> {
+    fn default() -> Self {
+        Self::Empty
     }
 }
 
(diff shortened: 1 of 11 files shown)


read-fonts/src/tables/postscript/index.rs
rsheeter, line 1: Adding Empty seems like it might have been a nice standalone, very small and easy to review

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
+//! Common functionality for glyf, cff and autohinting scalers.
```
rsheeter, line 1: Elaborating on what is/isn't common functionality wouldn't hurt

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
...
+pub(crate) struct BaseOutlines<'a> {
+    pub font: FontRef<'a>,
+    pub hmtx: Hmtx<'a>,
+    pub hvar: Option<Hvar<'a>>,
+}
+
+impl<'a> BaseOutlines<'a> {
+    pub fn new(font: &FontRef<'a>) -> Option<Self> {
+        let hmtx = font.hmtx().ok()?;
+        let hvar = font.hvar().ok();
```
rsheeter, line 20: perhaps worth a comment that we need hmtx, it's ok if there is no hvar

when the types don't popup it's very easy to miss that only one has a ?

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
...
+            hvar,
+        })
+    }
+
+    pub fn advance_width(&self, gid: GlyphId, coords: &'a [F2Dot14]) -> i32 {
+        let mut advance = self.hmtx.advance(gid).unwrap_or_default() as i32;
+        if let Some(hvar) = &self.hvar {
+            advance += hvar
+                .advance_width_delta(gid, coords)
+                // FreeType truncates metric deltas...
```
rsheeter, line 33: bonus points for reference

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
...
+        let hmtx = font.hmtx().ok()?;
+        let hvar = font.hvar().ok();
+        Some(Self {
+            font: font.clone(),
+            hmtx,
+            hvar,
+        })
+    }
+
+    pub fn advance_width(&self, gid: GlyphId, coords: &'a [F2Dot14]) -> i32 {
```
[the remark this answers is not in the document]
rsheeter: (here and below)

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
...
+            lsb += hvar
+                .lsb_delta(gid, coords)
+                // FreeType truncates metric deltas...
+                .map(|delta| delta.to_f64() as i32)
+                .unwrap_or(0);
+        }
+        lsb
+    }
+
+    pub fn cvt(&self) -> &[BigEndian<i16>] {
```
rsheeter, line 52: Both lsb and cvt merit a comment clarifying what that even is

skrifa/src/outline/base.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use raw::{
+    tables::{hmtx::Hmtx, hvar::Hvar},
+    types::{BigEndian, F2Dot14, GlyphId, Tag},
+    FontRef, TableProvider,
+};
+
+/// Common functionality for glyf, cff and autohinting scalers.
+#[derive(Clone)]
+pub(crate) struct BaseOutlines<'a> {
```
rsheeter, line 11: When I see cff or glyf structs holding base: BaseOutlines I'm left unclear what that does. I definitely wouldn't have anticipated the FontRef hides here.

Base is strongly associated with something you derive from for many of us and doesn't really conveying the common bits concept. 

To ask the obvious, why not write a trait and implement it for FontRef? - cff and glyf holding `font: FontRef` and calling e.g. `font.lsb` would read more clearly to me. The main service this seems to serve is to avoid calling `font.hmtx` but if that's too slow ... maybe we could make it not be?

skrifa/src/outline/base.rs
rsheeter, line 1: suggest `s/base/common/`

rsheeter (approved):
LGTM subject to BaseOutlines concerns being addressed

skrifa/src/outline/base.rs
dfrg: Perf measurements suggest that looking up the tables on demand roughly doubles the cost of metrics retrieval. This may be acceptable for static text, but the animated variable case is one we like to highlight so wasting cycles for the sake of purity seems ill advised.

I renamed base -> common and BaseOutlines to OutlinesCommon for now. I agree this isn't ideal but I'd like to make progress on autohinting so I filed #1085 to revisit this in the future.

skrifa/src/outline/base.rs
dfrg: This comment is actually stale so I removed it. We added rounding to the IVS itself in an earlier patch to match FT.

skrifa/src/outline/base.rs
dfrg: Done

skrifa/src/outline/base.rs
dfrg: Done

PR #1081, opened 2024-08-12
Begin Implementing IFT Format 2 Patch Map parsing

garretrieger:
https://w3c.github.io/IFT/Overview.html#patch-map-format-2

Work in progress. Creating draft PR to get feedback on codegen modifications.

### DIFF
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -246,7 +246,26 @@ pub(crate) struct FieldReadArgs {
 #[derive(Clone, Debug)]
 pub(crate) enum Condition {
     SinceVersion(VersionSpec),
-    IfFlag { field: syn::Ident, flag: syn::Path },
+    IfFlag {
+        field: syn::Ident,
+        flag: syn::Path,
+    },
+    IfCond {
+        xform: IfTransform,
+        args: Vec<IfArg>,
+    },
+}
+
+#[derive(Clone, Debug)]
+pub(crate) enum IfArg {
+    Field(syn::Ident),
+    Path(syn::Path),
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum IfTransform {
+    FlagAndNonZero,
+    AnyFlag,
 }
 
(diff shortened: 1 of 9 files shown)


font-codegen/src/parsing.rs
```diff
@@ -246,7 +246,26 @@ pub(crate) struct FieldReadArgs {
...
+#[derive(Clone, Debug)]
+pub(crate) enum IfArg {
+    Field(syn::Ident),
+    Path(syn::Path),
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum IfTransform {
+    FlagAndNonZero,
+    AnyFlag,
```
cmyr, line 268: somewhere in here we should be documenting what this is all used for, since it's getting a bit obscure.

font-codegen/src/parsing.rs
```diff
@@ -1794,6 +1817,81 @@ fn parse_if_flag(attr: &syn::Attribute) -> syn::Result<Condition> {
...
+
+    attr.parse_args::<If>()
+        .map(|If(xform, args)| Condition::IfCond { xform, args })
+        .map_err(|e| {
+            syn::Error::new(
+                e.span(),
+                format!("expected #[if_cond(condition_function(...))]: '{e}'"),
+            )
+        })
+}
```
cmyr, line 1893: good stuff, syn can be very confusing at first but this all looks right.

cmyr (commented):
this all seems reasonable to me, was there anything in particular you're unsure about?

one slightly annoying request before we get too far, but it would be nice to PR the codegen changes in isolation, and then have a followup PR for the code that actually uses it.

garretrieger:
> this all seems reasonable to me, was there anything in particular you're unsure about?
> 
> one slightly annoying request before we get too far, but it would be nice to PR the codegen changes in isolation, and then have a followup PR for the code that actually uses it.

Here's the PR for just the codegen changes: https://github.com/googlefonts/fontations/pull/1088

PR #1082, opened 2024-08-13
[klippa] add --drop-tables option

qxliu76:


klippa/src/lib.rs
```diff
@@ -157,10 +158,12 @@ impl Plan {
         input_unicodes: &IntSet<u32>,
         font: &FontRef,
         flags: SubsetFlags,
+        drop_tables: &IntSet<u32>,
```
cmyr, line 161: we should either make `Tag` work with `IntSet`, or just use a HashMap? I'd be very surprised if this showed up on a profile..

klippa/src/parsing_util.rs
```diff
@@ -65,6 +65,27 @@ pub fn parse_unicodes(unicode_str: &str) -> Result<IntSet<u32>, SubsetError> {
...
+        return Ok(out);
+    }
+
+    for tag in input_str.split(|c| c == ',' || c == ' ') {
+        if tag.is_empty() {
+            continue;
+        }
+        let Ok(table_tag) = Tag::new_checked(tag.as_bytes()) else {
+            return Err(SubsetError::InvalidTag(tag.to_owned()));
+        };
```
cmyr, line 83: this function returns an `InvalidTag` error describing the problem if the tag can't be parsed. Maybe we should hold on to that? (as in, have `SubsetError::InvalidTag { tag: String, error: InvalidTag }`)

cmyr (approved):
looks good, a couple little ideas below :)

klippa/src/lib.rs
garretrieger: I think for non-performance critical parts BTreeSet<Tag> is slightly preferred as int set's + tags can consume excessive memory due to needing approximately one 512 bit page per tag. Using a btree will keep tag order sorted which is generally what we want.

garretrieger: Looking at the usage here we don't depend on ordering anywhere so HashSet<Tag> would also be totally fine.

qxliu76: I used IntSet for convenience and would happily change it to HashSet. Though I have one question: what is the best way to use HashSet when we have `--drop-tables=*, --drop-tables-=some_table`

qxliu76: HashSet doesn't support `all()`

klippa/src/parsing_util.rs
qxliu76: SubsetError::InvalidTag defined in lib.rs:
```
#[error("Invalid tag {0}")]
InvalidTag(String),
```

klippa/src/lib.rs
qxliu76: Thanks! I'll work on this

qxliu76: Done, can you help review again?

read-fonts/src/collections/int_set/mod.rs
```diff
@@ -968,6 +969,33 @@ impl Domain for GlyphId {
...
+        Tag::from_u32(member.value())
+    }
+
+    fn is_continuous() -> bool {
+        true
+    }
+
+    // Tag is 4 bytes, each in the ascii range (`0x20..=0x7E`)
+    fn ordered_values() -> impl DoubleEndedIterator<Item = u32> {
+        0x20202020..=0x7E7E7E7E
```
garretrieger, line 987: It looks like it's possible to construct Tag's with values outside of ASCII so ordered_values() and count should just use u32::MIN and u32::MAX.

PR #1084, opened 2024-08-14
[int-set] special case IntSet::cmp to compare ranges when comparing mixed inclusive/exclusive sets.

garretrieger:
For exclusive sets this avoids iterating over all of the individual members of the set which can be extremely large.


cmyr (approved):
looks like a nice improvement.

Issue #1085, opened 2024-08-15
Refactor skrifa::outlines module


dfrg:
Some recent restructuring (ref #1079) to support autohinting leaves us in an unfortunate state with a new "common" object for glyf and cff. After autohinting is done and all requirements are known, revisit this, perhaps adding an internal `Scaler` trait to cover all use cases.

PR #1086, opened 2024-08-15
[skrifa] autohint: public GlyphStyles type

dfrg:
Exposes the glyph style map as a public type and wrapped in an `Arc` to make it shareable among multiple autohinting instances.

### DIFF
--- a/skrifa/src/outline/autohint/instance.rs
+++ b/skrifa/src/outline/autohint/instance.rs
@@ -13,7 +13,7 @@ use raw::TableProvider;
 pub struct GlyphStyles(Arc<GlyphStyleMap>);
 
 impl GlyphStyles {
-    /// Precomputes the full set of glyph styles for the given font.
+    /// Precomputes the full set of glyph styles for the given outlines.
     pub fn new(outlines: &OutlineGlyphCollection) -> Self {
         if let Some(outlines) = outlines.common() {
             let glyph_count = outlines
--- a/skrifa/src/outline/autohint/instance.rs
+++ b/skrifa/src/outline/autohint/instance.rs
@@ -1,0 +1,32 @@
+//! Autohinting state for a font instance.
+
+use super::{super::OutlineGlyphCollection, style::GlyphStyleMap};
+use crate::MetadataProvider;
+use alloc::sync::Arc;
+use raw::TableProvider;
+
+/// Set of derived glyph styles that are used for automatic hinting.
+///
+/// These are invariant per font so can be precomputed and reused for multiple
+/// instances when requesting automatic hinting.
+#[derive(Clone, Debug)]
+pub struct GlyphStyles(Arc<GlyphStyleMap>);
+
+impl GlyphStyles {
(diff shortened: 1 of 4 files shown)


skrifa/src/outline/mod.rs
```diff
@@ -582,6 +584,14 @@ impl<'a> OutlineGlyphCollection<'a> {
             Some((gid, glyph))
         })
     }
+
+    pub(crate) fn common(&self) -> Option<&OutlinesCommon<'a>> {
```
drott, line 588: Not new in this change, but perhaps a naming related question: What "common" is, is not quite clear to me. Is that documented somewhere, or could we pick a more descriptive name?

skrifa/src/outline/autohint/instance.rs
```diff
@@ -0,0 +1,32 @@
...
+impl GlyphStyles {
+    /// Precomputes the full set of glyph styles for the given outlines.
+    pub fn new(outlines: &OutlineGlyphCollection) -> Self {
+        if let Some(outlines) = outlines.common() {
+            let glyph_count = outlines
+                .font
+                .maxp()
+                .map(|maxp| maxp.num_glyphs() as u32)
+                .unwrap_or_default();
+            Self(Arc::new(GlyphStyleMap::new(
```
drott, line 24: Do I understand correctly, that an `Arc` type is used here mostly for being able to be able to get the GlyphStyleMap "past the borrow checker" and being able to use it in multiple Skrifa calls in parallel threads? - 

Regarding write-safety: How is that ensured when IIRC info for additional scripts need to be added to the GlyphStyleMap?

What I am trying to say: Could you help me understand why this is an Arc type?

drott (commented):
No concerns in terms of API shape, a minor naming suggestion and question regarding the Arc type.

skrifa/src/outline/mod.rs
dfrg: Please bear with me while under construction :) This is a shim to resolve unexpected data dependencies for autohinting and is tracked for cleanup in #1085

skrifa/src/outline/autohint/instance.rs
dfrg: Yes, exactly. We want to be able to share the styles among multiple instances (potentially in multiple threads), so `Arc` is The Way to do this in safe Rust. The alternative would be to clone this for each new instance which would require an allocation and copy. 

The style map is computed in full on construction (FreeType does this as well) so it’s never mutated.

The mutable, lazy data structure we talked about is the one that holds unscaled style metrics (alignment zones and widths). That was added in #1077 and uses `RwLock` to handle updates.

drott: Thanks, got it. TIL about `RwLock`, how handy.

skrifa/src/outline/mod.rs
drott: Ok with me to track this in #1085.

drott (approved):
LGTM

PR #1087, opened 2024-08-15
[skrifa] new hinting options type

### LINKED ISSUE #1080: ClearType symmetric rendering
The TrueType interpreter has a [GETINFO](https://learn.microsoft.com/en-us/typography/opentype/spec/tt_instructions#get-information) instruction that allows the font to query information about the state and capabilities of the rasterizer. Of interest here is bit 18 which is described as follows:

```
Bit 18 is set to 1 if ClearType symmetric rendering is being used (for MS Rasterizer v.2.0 and later).
Depending on the horizontal LCD stripe orientation, this can impact the rendering of horizontal
features. It is set to 0 otherwise.
```

FreeType [always sets this to true](https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/truetype/ttinterp.c#L6772) when hinting for an anti-aliased target.

At some point in the Windows 10 era, the Calibri font (at least) was updated to query this bit and execute different instructions when enabled. As the note mentions, this does impact the rendering of horizontal features. Note the partially extended height of the top stem (rendered at 16px and zoomed 16x):

raster_ft

Skrifa currently produces the exact same output, but does better when that flag is disabled:

raster_skrifa

You can disable this in skrifa right now by enabling the `preserve_linear_metrics` setting but that can also affect horizontal rendering, so I propose that we add a new `symmetric_rendering` flag to control this directly, set to `true` by default to match FreeType.

It's conceivable that a supersampling rasterizer will handle this better (I assume ClearType does this when it sets this flag), but analytical area implementations do not.

cc @drott

dfrg:
Replaces the current `HintingMode` with a new `HintingOptions` type to support autohinting. Specifically, it allows for selecting the hinting engine (interpreter, auto or auto fallback). Also reintroduces a "light" mode to match `FT_RENDER_MODE_LIGHT` since this actually represents a different mode in the autohinter (but didn't for TrueType or CFF).

Adds the `symmetric_rendering` flag as well (closes #1080).

Based on #1086 since it depends on the new public `GlyphStyles` type.

@drott this is the same as the reference code in the preview branch


skrifa/src/outline/hint.rs
```diff
@@ -4,11 +4,179 @@ use super::{
...
+    /// The TrueType or PostScript interpreter.
+    Interpreter,
+    /// The automatic hinter that performs just-in-time adjustment of
+    /// outlines.
+    ///
+    /// Glyph styles can be precomputed per font and may be provided here
+    /// as an optimization to avoid recomputing them for each instance.
+    Auto(Option<GlyphStyles>),
+    /// Selects the engine based on the same rules that FreeType uses in
+    /// the default configuration.
```
drott, line 47: AFAICS FreeType uses load flags to control hinting behavior. What do you mean by "default configuration" here, do you mean no extra load flags? I would suggest to clarify that comment.

skrifa/src/outline/hint.rs
```diff
@@ -4,11 +4,179 @@ use super::{
...
+    ///
+    /// Specifically, PostScript fonts will always use the interpreter and
```
drott, line 49:
```suggestion
    /// Specifically, PostScript fonts will always use the interpreter and
```
Do you mean CFF fonts, or PostScript fonts here? And by interpreter, do you mean the CFF hinting engine? Or do PS fonts have TrueType style hinting instructions?

skrifa/src/outline/mod.rs
```diff
@@ -585,6 +587,21 @@ impl<'a> OutlineGlyphCollection<'a> {
...
 
+    /// Returns true if the interpreter engine should be used for hinting this
+    /// set of outlines.
+    ///
+    /// When this returns false, you likely want to use the automatic hinter
+    /// instead.
+    ///
+    /// This matches the logic used in FreeType when neither of the
+    /// `FT_LOAD_FORCE_AUTOHINT` or `FT_LOAD_NO_AUTOHINT` flags are specified.
+    pub fn prefer_interpreter(&self) -> bool {
```
drott, line 598: Perhaps explain in this comment from where this is used and how it relates to the `Engine` configuration option?

skrifa/src/outline/autohint/metrics.rs
```diff
@@ -218,40 +218,31 @@ pub(crate) struct Scale {
 }
 
 impl Scale {
-    /// Create initial scaling parameters from font size, units per em
-    /// and hinting mode.
-    pub fn new(size: f32, units_per_em: i32, mode: HintingMode) -> Self {
+    /// Create initial scaling parameters from metrics and hinting target.
+    pub fn new(size: f32, units_per_em: i32, target: Target, is_italic: bool) -> Self {
```
drott, line 222: is_italic is an odd argument here and does not feel in the right place. Can Skrifa determine on its own if the font is italic? What is supposed to be passed in here and what algorithm do you suggest for the client to determine this boolean - and if there is one, could that algorithm be run in Skrifa itself?

I think if there is really a client side decision needed on whether hinting should be activated for an italic font, then the client would need to adjust the hinting configuration / options for it?

drott (commented):
Thank you for continuing to upstream autohinting, great to see this. 

No concerns for `HintingOptions` and the configuration API. This looks as discussed earlier and seems good to be used in Skia (with a reservation that I have not tried it yet). 

But the `is_italic` argument to `Scale` feels odd to me and seems a surprising configuration side effect to hinting later. I don't know how to set this argument correctly from the client side. (Besides the general concern with boolean API args that are hard to read at the call site.)

skrifa/src/outline/autohint/metrics.rs
dfrg: Definitely not a client side decision… we’ll handle this internally when the autohinting instance state lands. See https://github.com/googlefonts/fontations/blob/cfcd5975d4e3050184ddd80b08cb48fbfdda19fc/skrifa/src/outline/autohint/instance.rs#L77 in the preview branch.

skrifa/src/outline/hint.rs
dfrg: Will add a comment similar to the one on `prefer_interpreter` that mentions the load flags.

skrifa/src/outline/hint.rs
dfrg: I’ve tried to be coherent in the docs wrt references so I use glyf/CFF/CFF2 when talking about tables and TrueType/PostScript when talking about the overall technologies.

This comment does refer to the CFF engine which is technically a PostScript interpreter :) If this is confusing, I’m open to suggestions for improvement.

skrifa/src/outline/mod.rs
dfrg: Will do.

skrifa/src/outline/hint.rs
drott: Thanks, then I understand what you mean. 

Perhaps something like this, to clarify that the two interpreters are different:
> Specifically, PostScript (CFF/CFF2) fonts will always use the PostScript interpreter and PostScript hinting engine. TrueType fonts will use the interpreter for TrueType instructions if one of the `fpgm` or `prep` tables is non-empty, falling back to the automatic hinter otherwise.

skrifa/src/outline/autohint/metrics.rs
drott: Good to hear this is not a client side setting. 

My concern was that we would expose a public `Scale` API with the odd boolean argument. But looking at this again, `Scale` is not a public Skrifa interface, is it? - If so, that addresses my concern. If possible, I would still prefer to turn the Boolean argument into an enum to make callsites clearer to read.

skrifa/src/outline/hint.rs
dfrg: That revision looks good to me, thanks!

skrifa/src/outline/autohint/metrics.rs
dfrg: Yes, this is internal code but fair point about the bare boolean. When I add the instance code, I’ll separate the hinting flags computation from scale setup and avoid the bool altogether.

skrifa/src/outline/hint.rs
dfrg: Done

skrifa/src/outline/hint.rs
dfrg: Also done

skrifa/src/outline/mod.rs
dfrg: And this one too

skrifa/src/outline/autohint/metrics.rs
dfrg: Reworked `Scale::new()` to take a `Style` parameter rather than a boolean

skrifa/src/outline/hint.rs
```diff
@@ -4,11 +4,181 @@ use super::{
...
+    /// Glyph styles can be precomputed per font and may be provided here
+    /// as an optimization to avoid recomputing them for each instance.
+    Auto(Option<GlyphStyles>),
+    /// Selects the engine based on the same rules that FreeType uses when
+    /// neither of the `FT_LOAD_NO_AUTOHINT` or `FT_LOAD_FORCE_AUTOHINT`
+    /// load flags are specified.
+    ///
+    /// Specifically, PostScript (CFF/CFF2) fonts will always use the hinting
+    /// engine in the PostScript interpreter and TrueType fonts will use the
+    /// interpreter for TrueType isntructions if one of the `fpgm` or `prep`
```
drott, line 52: Nit, typo:

```suggestion
    /// interpreter for TrueType instructions if one of the `fpgm` or `prep`
```

skrifa/src/outline/hint.rs
drott: Ack, thanks for rewording that. IMO very helpful to have these mappings.

skrifa/src/outline/autohint/metrics.rs
drott: Thanks for doing that, LGTM.

drott (approved):
LGTM, thank you!

skrifa/src/outline/hint.rs
```diff
@@ -28,6 +198,7 @@ use crate::alloc::{boxed::Box, vec::Vec};
 /// FreeType 2.7.
 ///
 /// The default value of this type is equivalent to `FT_LOAD_TARGET_NORMAL`.
+#[doc(hidden)]
```
DJMcNab, line 201: Should this be `#[deprecated]`?

dfrg: Yes, but we require transitional states for rolling updates in Chrome and I specifically avoided the deprecated attribute to avoid raising warnings at compile time.

DJMcNab: Thanks for the explanation. I wonder if whatever uses this in chrome should use [`allow(deprecated)`](https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#deprecated) instead. I think that should be entirely equivalent for that use case.

I suppose that could cause problems if a non-skrifa library were updated?

dfrg: Yeah, it seems like that could affect other things in undesirable ways. My plan was to mark these deprecated for the next cycle and then remove them which adds a bit of latency but should be fine.

PR #1088, opened 2024-08-15
Add 'if_cond(...)' to codegen.

garretrieger:
This allows for custom conditions to be tested against. Will be needed for IFT format 2 code gen. To begin with add any_flag condition which is true if any of the input flags are set.

### DIFF
--- a/font-codegen/README.md
+++ b/font-codegen/README.md
@@ -173,6 +173,9 @@ The following annotations are supported on top-level objects:
 - `#[if_flag($field, Flags::SOME_FLAG)]`: indicates that a given field is only
   present if a particular flag is set on the named field. The field is expected
   to be a bitset with a `contains` method.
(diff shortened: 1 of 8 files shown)


font-codegen/README.md
```diff
@@ -173,6 +173,9 @@ The following annotations are supported on top-level objects:
 - `#[if_flag($field, Flags::SOME_FLAG)]`: indicates that a given field is only
   present if a particular flag is set on the named field. The field is expected
   to be a bitset with a `contains` method.
+- `#[if_cond($field, Flags::SOME_FLAG_A, Flags::SOME_FLAG_B, ...)]`: indicates that a
```
cmyr, line 176: great, glad you found this

cmyr (approved):
This all looks good, but one thing I would suggest doing is adding a test case; you can do this by editing `resources/codegen_inputs/test_conditions.rs` to add a field that uses this annotation; and then you can add a test in `read-fonts/src/codegen_test.rs` to show that we are handling parsing correctly.

PR #1092, opened 2024-08-16
[kilppa] support subsetting OS/2 table

qxliu76:


klippa/src/lib.rs
```diff
@@ -461,6 +464,10 @@ fn subset_table<'a>(
         Maxp::TAG => {
             subset_maxp(font, plan, builder)?;
         }
+
+        Os2::TAG => {
+            subset_os2(font, plan, builder)?;
+        }
```
rsheeter, line 470: Here and above ther eseem to be a lot of extra {}, why not `Os2::TAG => subset_os2(font, plan, builder)?,`

klippa/src/os2.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+}
+
+// Returns the bit to be set in os/2 ulUnicodeOS2Range for a given codepoint.
+fn get_unicode_range_bit(cp: u32) -> Option<u8> {
+    match OS2_UNICODE_RANGES.binary_search_by(|&(a, b, _)| OS2Range::new(a, b).cmp(cp)) {
+        Ok(i) => Some(OS2_UNICODE_RANGES[i].2),
+        Err(_) => None,
+    }
```
cmyr, line 82: you can use `Result::ok` to convert a result to an option, like
```suggestion
    OS2_UNICODE_RANGES.binary_search_by(|&(a, b, _)| OS2Range::new(a, b).cmp(cp))
        .ok()
        .map(|i| OS2_UNICODE_RANGES[i].2)
```

cmyr (approved):
looks good!

klippa/src/lib.rs
rsheeter: Most of your branches have your functions return type. It looks like you could drop the final `Ok(())` and write this function as just the match, e.g.

```rust
fn subset_table<'a>(
    tag: Tag,
    font: &FontRef<'a>,
    plan: &Plan,
    builder: &mut FontBuilder<'a>,
) -> Result<(), SubsetError> {
    match tag {
        Glyf::TAG => subset_glyf_loca(plan, font, builder),
        // head is handled by glyf table if exists
        Head::TAG => font.glyf().map(|_| ()).or_else(|_| subset_head(font, plan, builder)),            
        //Skip, handled by Hmtx
        Hhea::TAG => Ok(()),
        Hmtx::TAG => subset_hmtx_hhea(font, plan, builder),
        //Skip, handled by glyf
        Loca::TAG => Ok(()),
        Maxp::TAG => subset_maxp(font, plan, builder),
        Os2::TAG => subset_os2(font, plan, builder),
        _ => if let Some(data) = font.data_for_tag(tag) {
                builder.add_raw(tag, data);
                Ok(())
            } else {
                Err(SubsetError::SubsetTableError(tag))
            }
    }
}
```

qxliu76: Thanks!

klippa/src/os2.rs
qxliu76: thanks!

klippa/src/os2.rs
```diff
@@ -0,0 +1,125 @@
...
+use write_fonts::{
+    read::{
+        collections::IntSet,
+        tables::os2::{Os2, OS2_UNICODE_RANGES},
+        FontRef, TableProvider, TopLevelTable,
+    },
+    FontBuilder,
+};
+
+pub fn subset_os2(
```
rsheeter, line 14: Would pub(crate) better reflect the intent?

rsheeter, line 14: Is there equivalent harfbuzz and/or fonttools code we could reference? If we're not doing that for our subset fns perhaps we could start? - doing it for freetype has been quite helpful.

klippa/src/os2.rs
```diff
@@ -0,0 +1,125 @@
...
+        .unwrap()
+        .copy_from_slice(&us_last_char_index.to_be_bytes());
+
+    if plan
+        .subset_flags
+        .contains(SubsetFlags::SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES)
+    {
+        builder.add_raw(Os2::TAG, out);
+        return Ok(());
+    }
```
rsheeter, line 38: If anyone adds code below this they expect to execute the return may surprise them, plus you duplicate adding to the builder. Could you do it the other way around, as in if plan doesn't contain no prune update ranges, no early return?

qxliu76:
> Question, where did all the test data come from? Are we copying it from harfbuzz?

yes, they're copied from harfbuzz, but I tailored them to drop those unsupported tables

garretrieger (approved):

PR #1093, opened 2024-08-16
Add #[if_cond(always_true())] to codegen.

garretrieger:
Used to allow the inclusion of a non-conditional field following a conditional field.

### DIFF
--- a/font-codegen/README.md
+++ b/font-codegen/README.md
@@ -173,9 +173,12 @@ The following annotations are supported on top-level objects:
 - `#[if_flag($field, Flags::SOME_FLAG)]`: indicates that a given field is only
   present if a particular flag is set on the named field. The field is expected
   to be a bitset with a `contains` method.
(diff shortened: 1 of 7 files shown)


garretrieger:
> I think we need to just rip off the bandaid and remove the 'field after conditional field' check.
> 
> I'm going to take the opportunity here to just explain how this came to be, for posterity:
> 
> * in the first pass of codegen, the only 'conditional field' construct that we supported was fields that were added with a new table version; these were always at the end of the data.
> * in this world, having a version-dependent field that proceeded an always-present field would have always been nonsensical, since it would break versioning.
> * when we added additional conditions, we reused some of the logic we'd used for versioned fields.
> * for those new conditions, this check doesn't make sense.
> 
> honestly I don't think this is that complicated, and I'll look into removing it.

Sounds good to me, that's definitely better then this.

PR #1094, opened 2024-08-16
Begin Implementation of IFT Format 2 Parsing.

garretrieger:
Define the structure and implement parsing of codepoint only entries.

Specification is here: https://w3c.github.io/IFT/Overview.html#patch-map-format-2

### DIFF
--- a/resources/codegen_inputs/ift.rs
+++ b/resources/codegen_inputs/ift.rs
@@ -209,6 +209,8 @@ record DesignSpaceSegment {
   end: Fixed,
 }
 
+// Storage for id strings, indexed by EntryData::entryIdStringLength
+// See: https://w3c.github.io/IFT/Overview.html#mapping-entry-entryidstringlength
 table IdStringData {
   #[count(..)]
   id_data: [u8],
--- a/resources/codegen_inputs/ift.rs
+++ b/resources/codegen_inputs/ift.rs
@@ -171,7 +171,9 @@ table EntryData {
   #[if_flag($format, EntryFormatFlags::PATCH_ENCODING)]
   patch_encoding: u8,
 
-  #[if_cond(any_flag($format, EntryFormatFlags::CODEPOINTS_BIT_1, EntryFormatFlags::CODEPOINTS_BIT_2))]
+  // hack to allow non-conditional field to follow conditional field
+  // this is non conditional as we always need to get a pointer to any remaining data after the entry.
+  #[if_cond(always_true())]
   #[count(..)]
   codepoint_data: [u8],
 }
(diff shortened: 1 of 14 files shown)


skrifa/src/patchmap.rs
```diff
@@ -278,8 +419,8 @@ impl PatchEncoding {
 /// the numeric index is supported.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct PatchUri {
-    template: String,
-    index: u32,
+    template: String, // TODO: Make this a reference?
```
cmyr, line 422: this is definitely doable; whether it's worth doing really depends on how hot the codepath is where this gets used. In any case it would require threading the lifetime of the font data through all the intermediate calls and types.

skrifa/src/patchmap.rs
```diff
@@ -293,6 +434,44 @@ impl PatchUri {
...
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+struct Entry {
+    // Key
+    codepoints: IntSet<u32>,
+    feature_tags: BTreeSet<Tag>,
+    design_space: Vec<DesignSpaceSegment>,
+
+    // Value
+    uri: PatchUri,
+    compatibility_id: [u32; 4], // TODO: Make this a reference?
```
cmyr, line 449: this feels lower value, a reference is only one machine word less than copying this around.

skrifa/src/patchmap.rs
```diff
@@ -293,6 +434,44 @@ impl PatchUri {
...
+}
+
+impl Entry {
+    fn new(template: &str, compat_id: &[u32; 4], default_encoding: &PatchEncoding) -> Entry {
+        Entry {
+            codepoints: IntSet::<u32>::empty(),
+            feature_tags: BTreeSet::<Tag>::new(),
```
cmyr, line 456: just fyi you don't need to specify types here, the compiler will figure it out. (If this is a matter of personal style, that's fine)

cmyr (approved):
This all looks sane, although it's a very big patch and easy to miss stuff. It's unfortunate that the github UI is still quite poor at handling chained/related patchsets...

skrifa/src/patchmap.rs
garretrieger: Sounds good, I figured I'd hold off on this until I have the other side (the client which uses this code) and then see if it makes sense from there to enable this. There could be hundreds to thousands of entries in a IFT font so not duplicating this template on each is probably fairly valuable.

skrifa/src/patchmap.rs
garretrieger: Good point.

resources/codegen_inputs/ift.rs
```diff
@@ -105,13 +105,111 @@ record EntryMapRecord {
...
+  RESERVED = 0b10000000,
+}
+
+record DesignSpaceSegment {
+  axis_tag: Tag,
+  start: Fixed,
+  end: Fixed,
+}
+
+table IdStringData {
```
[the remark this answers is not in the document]
garretrieger: This structure isn't currently implemented yet, but added some explanation of how it will work + a link to the spec.

PR #1096, opened 2024-08-19
Allow traversal of fields which are both nullable and conditional

### LINKED ISSUE #1095: Crash serializing GDEF
In diffenator3, we use traversal to serialize the tables for diffing. When using `get_field` on a `gdef`, this code path is executed:

```
            3usize => Some(Field::new(
                "lig_caret_list_offset",
                FieldType::offset(self.lig_caret_list_offset(), self.lig_caret_list()),
            )),
            4usize => Some(Field::new(
                "mark_attach_class_def_offset",
                FieldType::offset(
                    self.mark_attach_class_def_offset(),
                    self.mark_attach_class_def(),
                ),
            )),
            5usize if version.compatible((1u16, 2u16)) => Some(Field::new(
                "mark_glyph_sets_def_offset",
                FieldType::offset(
                    self.mark_glyph_sets_def_offset().unwrap(),
                    self.mark_glyph_sets_def().unwrap(),
                ),
            )),
```

Everything looks good until we get to the `unwrap`s in branch 5. (I feel like library code should never `unwrap`, since it takes the app down with it.) If there is no `markGlyphSetsDefOffset`, the code will crash. Simple reproducer:

```
println!("{:?}", font.gdef());
```

simoncozens:
Fixes #1095

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -367,7 +367,14 @@ fn traversal_arm_for_field(
 ) -> TokenStream {
     let name_str = &fld.name.to_string();
     let name = &fld.name;
-    let maybe_unwrap = fld.attrs.conditional.is_some().then(|| quote!(.unwrap()));
+    let is_single_nullable_offset = fld.is_nullable() && !fld.is_array();
+    let maybe_unwrap = fld
+        .attrs
+        .conditional
+        .as_ref()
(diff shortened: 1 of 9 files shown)


cmyr:
this isn't quite right, since we do want to unwrap the raw offset, just not the resolving getter fn. I've added an alternative at #1110

PR #1099, opened 2024-08-20
Make fuzzing able to use TTCs

rsheeter:
Contributes to #420

JMM


fuzz/fuzz_targets/helpers.rs
```diff
@@ -51,3 +57,19 @@ fn create_location(font: &FontRef, axis_positions: &[f32]) -> Location {
...
+        let _ = cr.len();
+        let _ = cr.is_empty();
+    }
+    match FileRef::new(data) {
+        Ok(fr) => Some(fr.fonts()),
+        Err(_) => None,
+    }
+    .into_iter()
+    .flatten()
+    .filter_map(|i| i.ok())
```
drott, line 74: Is the `Iterator` of fonts a bit much and creating too many similar iterations, slowing down the fuzzer? How about letting the fuzzer randomness choose the ttc index? That would allow the fuzzing to cover the ttc over time, but focus on the interesting paths, instead of repeating iterations on the same safe/non-interesting data.

font-test-data/src/lib.rs
```diff
@@ -375,3 +375,7 @@ pub mod ift {
         0x00, 0x00,               // last_entry_index
     ];
 }
+
+pub mod ttc {
+    pub static TTC: &[u8] = include_bytes!("../test_data/ttc/TTC.ttc");
```
drott, line 380: Similar to practices in Chromium or our other open source projects, I would suggest to create a `test_data/third_party/` or `test_data/third_party/harfbuzz` directory, and please a readme with provenance and license there, so we have a clear indication where the font is coming from.

drott (commented):
Yay for more fuzzing coverage, some comments on third-party test data and iteration count below.

font-test-data/src/lib.rs
rsheeter: I agree with that but as this applies to all our testdata I have defered it to https://github.com/googlefonts/fontations/issues/1127

fuzz/fuzz_targets/helpers.rs
rsheeter: Good point, made it so.

drott (approved):
LGTM

PR #1103, opened 2024-08-20
[write-fonts] post table: de-duplicate psNames

qxliu76:


garretrieger (approved):

write-fonts/src/tables/post.rs
```diff
@@ -14,21 +14,30 @@ pub struct PString(String);
 impl Post {
     /// Construct a new version 2.0 table from a glyph order.
     pub fn new_v2<'a>(order: impl IntoIterator<Item = &'a str>) -> Self {
-        let known_glyphs = read_fonts::tables::post::DEFAULT_GLYPH_NAMES
+        let standard_glyphs = read_fonts::tables::post::DEFAULT_GLYPH_NAMES
             .iter()
             .enumerate()
             .map(|(i, name)| (*name, i as u16))
             .collect::<HashMap<_, _>>();
+        const NUM_STANDARD: usize = 258;
```
cmyr, line 22: I would write this as, 
```suggestion
        const NUM_STANDARD: usize = read_fonts::tables::post::DEFAULT_GLYPH_NAMES.len();
```

because it communicates more clearly what is going on.

cmyr (approved):
looks like this should work!

For curiousity, do you know under what conditions these names would contain duplicates? It isn't obvious to me when that might happen...

write-fonts/src/tables/post.rs
qxliu76: Thanks, I'll make this change later in another PR.

qxliu76:
> looks like this should work!
> 
> For curiousity, do you know under what conditions these names would contain duplicates? It isn't obvious to me when that might happen...

I'm not aware why this would happen. But there're duplicate psNames in a few font files from "google fonts", and I found this during testing harfbuzz subsetter:)

PR #1104, opened 2024-08-21
Implement IFT format 2 design space parsing.

garretrieger:
Add tests for feature and design space intersection.

### DIFF
--- a/read-fonts/generated/generated_ift.rs
+++ b/read-fonts/generated/generated_ift.rs
@@ -924,7 +924,7 @@ pub struct EntryDataMarker {
 }
 
 impl EntryDataMarker {
-    fn format_byte_range(&self) -> Range<usize> {
+    fn format_flags_byte_range(&self) -> Range<usize> {
         let start = 0;
         start..start + EntryFormatFlags::RAW_BYTE_LEN
     }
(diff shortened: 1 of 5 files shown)


read-fonts/generated/generated_ift.rs
```diff
@@ -961,62 +961,29 @@ impl EntryDataMarker {
...
-                                                                    })
-                                                            })
-                                                    })
-                                            })
-                                    })
-                            })
-                    })
-            });
+        let start = self . patch_encoding_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . entry_id_delta_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . copy_indices_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . copy_count_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . design_space_segments_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . design_space_count_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . feature_tags_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . feature_count_byte_range () . map (| range | range . end) . unwrap_or_else (|| self . format_flags_byte_range () . end)))))))) ;
```
cmyr, line 964: so you've drifted too far to the right and rustfmt has left you. Can we figure out a different way to write this?

resources/codegen_inputs/ift.rs
```diff
@@ -141,35 +141,35 @@ table MappingEntries {
 }
 
 table EntryData {
-  format: EntryFormatFlags,
+  format_flags: EntryFormatFlags,
```
cmyr, line 144: was this running into some problem? It's possible we have code that treats 'format' as special or otherwise generates a 'format' token that would shadow a similarly named field..

skrifa/src/patchmap.rs
```diff
@@ -3,18 +3,19 @@
 //! The IFT and IFTX tables encode mappings from subset definitions to URL's which host patches
 //! that can be applied to the font to add support for the corresponding subset definition.
 
+use core::ops::RangeInclusive;
```
cmyr, line 6: this should be `std` (which is aliased to `core` if real 'std' is unavailable) but rust-analyzer will sometimes use `core` instead when auto-importing in a crate that has no-std support...

skrifa/src/patchmap.rs
```diff
@@ -3,18 +3,19 @@
...
 use crate::Tag;
-use raw::tables::ift::DesignSpaceSegment;
 use raw::tables::ift::EntryFormatFlags;
 use raw::types::Uint24;
 use raw::{FontData, FontRead, FontRef};
 use read_fonts::{
     tables::ift::{EntryData, EntryMapRecord, Ift, PatchMapFormat1, PatchMapFormat2},
     ReadError, TableProvider,
 };
+use std::collections::HashMap;
```
cmyr, line 18: total nit but if this could be up with the other `std` imports that would satisfy my OCD

cmyr (approved):
looks sane, some ignore-able notes inline..

read-fonts/generated/generated_ift.rs
garretrieger: Since this is generated code, does it matter if the formatting isn't right?

resources/codegen_inputs/ift.rs
garretrieger: Just changing this to match the naming we use in the spec: https://w3c.github.io/IFT/Overview.html#mapping-entry-formatflags

read-fonts/generated/generated_ift.rs
rsheeter: it IS nice to be able to read it at times

garretrieger: This looks to be a result of how conditional fields are handled, the codegen has created a giant if else chain to figure out which field preceeds this one. So if we want to fix it would probably require reworking how conditional fields are handled in codegen.

PR #1108, opened 2024-08-22
[klippa] subset post table

qxliu76:


garretrieger (approved):

klippa/src/post.rs
```diff
@@ -0,0 +1,128 @@
...
+// https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-post-table.hh#L96
+pub(crate) fn subset_post(
+    font: &FontRef,
+    plan: &Plan,
+    builder: &mut FontBuilder,
+) -> Result<(), SubsetError> {
+    let post = font.post().or(Err(SubsetTableError(Post::TAG)))?;
+    let mut out = Vec::with_capacity(post.offset_data().len());
+    // copy header
+    out.extend_from_slice(post.offset_data().as_bytes().get(0..32).unwrap());
```
cmyr, line 25: looking at this pattern happening everywhere, I wanted to point you to the `BeBuffer` type in read-fonts; it's intended for generating test cases, but perhaps it could be a useful demonstration of a more ergonomic way of expressing this pattern? https://github.com/googlefonts/fontations/blob/main/read-fonts/src/tests/test_helpers.rs

klippa/src/post.rs
```diff
@@ -0,0 +1,128 @@
...
+        .subset_flags
+        .contains(SubsetFlags::SUBSET_FLAGS_GLYPH_NAMES);
+    //version 3 does not have any glyph names
+    if !glyph_names {
+        let major_version = 0x3_u16.to_be_bytes();
+        out.get_mut(0..2).unwrap().copy_from_slice(&major_version);
+        out.get_mut(2..4).unwrap().fill(0);
+    }
+
+    if glyph_names && post.version().to_major_minor() == (2, 0) {
```
cmyr, line 37: Alternatively,
```suggestion
    if glyph_names && post.version() == MajorMinor::VERSION_2_0 {
```

klippa/src/post.rs
```diff
@@ -0,0 +1,128 @@
...
+    //init all glyphNameIndex as 0
+    let idx_start = out.len();
+    let new_len = out.len() + plan.num_output_glyphs * 2;
+    out.resize(new_len, 0);
+
+    let max_old_gid = plan.glyphset.last().unwrap().to_u32() as usize;
+    // for standard glyphs: name indices < 258
+    let glyph_index_iter = post
+        .glyph_name_index()
+        .unwrap()
```
cmyr, line 58: this will crash on a malformed post table, is that okay?

cmyr (approved):
looks good, a few notes inline but no changes necessary

klippa/src/post.rs
qxliu76: Thanks!

klippa/src/post.rs
qxliu76: good catch, fixed it

PR #1109, opened 2024-08-23
Change sparse bit set decode to take a max value instead of max number of elements.

garretrieger:
Values beyond the max are ignored instead of causing the decoding to fail. This matches changes in the IFT specification handling of sparse bit set decoding (See: https://github.com/w3c/IFT/pull/203).


cmyr (approved):
👍

PR #1112, opened 2024-08-26
Extend BeBuffer to allow locations to be tagged.

garretrieger:
This allows constructing test data which tests can then modify specific parts of the data via the tag names. Updates one of the IFT test data items to use BeBuffer.

### DIFF
--- a/read-fonts/src/tests/test_helpers.rs
+++ b/read-fonts/src/tests/test_helpers.rs
@@ -1,10 +1,14 @@
 //! small utilities used in tests
 
 use crate::{FontData, Scalar};
+use std::collections::HashMap;
 
 /// A convenience type for generating a buffer of big-endian bytes.
 #[derive(Debug, Clone, Default)]
-pub struct BeBuffer(Vec<u8>);
+pub struct BeBuffer {
+    data: Vec<u8>,
+    tagged_locations: HashMap<String, usize>,
+}
 
 impl BeBuffer {
     pub fn new() -> Self {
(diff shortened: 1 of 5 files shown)


read-fonts/src/tests/test_helpers.rs
```diff
@@ -1,10 +1,14 @@
...
 
 use crate::{FontData, Scalar};
+use std::collections::HashMap;
 
 /// A convenience type for generating a buffer of big-endian bytes.
 #[derive(Debug, Clone, Default)]
-pub struct BeBuffer(Vec<u8>);
+pub struct BeBuffer {
+    data: Vec<u8>,
+    tagged_locations: HashMap<String, usize>,
```
rsheeter, line 10: I don't entirely love the string-based tag mechanism. I suppose removing it would be a bit of a pain, you'd have to capture an offset and store it so you'd end up returning a (BeBuffer, struct full of offsets) here.

I wonder how best one might avoid that nicely.

rsheeter (approved):
LGTM

read-fonts/src/tests/test_helpers.rs
garretrieger: Since this code is for tests only, my primary concern is making it succinct as possible to utilize in tests. I suspect that if we want to do something without strings (eg. using enums are something else as markers) will require using macros in order to maintain the same level of simplicity as string tags.

garretrieger: I'll merge this for now, but open to reworking it if we can come up with a better approach.

rsheeter: I like the idea of using an enum but I'm not sure it's hugely different different than using strings as the key and sticking them in constants. I don't think you'd need a macro, you'd just make BeBuffer have a generic for the key type.

rsheeter: Ah I see. I wouldn't be enthused about a macro for that. It seems to me that typing the enum is almost exactly the same amount of work as typing a const for the str. I think we didn't do consts for the strs in this PR, arguably we should have.

PR #1115, opened 2024-08-28
[skrifa] autohint outline to path

dfrg:
Adds support for converting the autohinter outline data to a path.

Additionally adds the const qualifier to some `PointFlags` methods and fixes an oversight in CFF where we weren't emitting a close path command for unhinted outlines (this was detected here because the autohinting conversion always emits close path).

### DIFF
--- a/skrifa/src/outline/autohint/outline.rs
+++ b/skrifa/src/outline/autohint/outline.rs
@@ -2,14 +2,19 @@
 
 use super::{
     super::{
+        path,
+        pen::PathStyle,
         unscaled::{UnscaledOutlineSink, UnscaledPoint},
-        DrawError, LocationRef, OutlineGlyph,
+        DrawError, LocationRef, OutlineGlyph, OutlinePen,
     },
     metrics::Scale,
 };
 use crate::collections::SmallVec;
 use core::ops::Range;
-use raw::types::F2Dot14;
+use raw::{
+    tables::glyf::PointFlags,
+    types::{F26Dot6, F2Dot14},
+};
 
 /// Hinting directions.
 ///
@@ -196,6 +201,35 @@ impl Outline {
         self.units_per_em = 0;
         self.points.clear();
         self.contours.clear();
+    }
+
+    pub fn to_path(
+        &self,
+        style: PathStyle,
+        pen: &mut impl OutlinePen,
+    ) -> Result<(), path::ToPathError> {
+        for contour in &self.contours {
+            let Some(points) = self.points.get(contour.range()) else {
+                continue;
+            };
+            #[inline(always)]
+            fn to_contour_point(point: &Point) -> path::ContourPoint<F26Dot6> {
+                const FLAGS_MAP: [PointFlags; 3] = [
+                    PointFlags::on_curve(),
+                    PointFlags::off_curve_quad(),
+                    PointFlags::off_curve_cubic(),
+                ];
+                path::ContourPoint {
+                    x: F26Dot6::from_bits(point.x),
+                    y: F26Dot6::from_bits(point.y),
+                    flags: FLAGS_MAP[(point.flags & Point::CONTROL) as usize],
+                }
+            }
+            if let Some(last_point) = points.last().map(to_contour_point) {
+                path::contour_to_path(points.iter().map(to_contour_point), last_point, style, pen)?;
+            }
+        }
+        Ok(())
     }
 }
 
@@ -561,10 +595,10 @@ impl UnscaledOutlineSink for Outline {
 
 #[cfg(test)]
 mod tests {
-    use super::Outline;
+    use super::super::super::{pen::SvgPen, DrawSettings};
     use super::*;
-    use crate::MetadataProvider;
-    use raw::{types::GlyphId, FontRef};
+    use crate::{prelude::Size, MetadataProvider};
+    use raw::{types::GlyphId, FontRef, TableProvider};
 
     #[test]
     fn direction_from_vectors() {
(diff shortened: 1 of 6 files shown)


skrifa/src/outline/autohint/outline.rs
```diff
@@ -197,6 +202,35 @@ impl Outline {
...
+        &self,
+        style: PathStyle,
+        pen: &mut impl OutlinePen,
+    ) -> Result<(), path::ToPathError> {
+        for contour in &self.contours {
+            let Some(points) = self.points.get(contour.range()) else {
+                continue;
+            };
+            #[inline(always)]
+            fn to_contour_point(point: &Point) -> path::ContourPoint<F26Dot6> {
```
rsheeter, line 216: If you hoist this out to a free fn it would be *much* easier for me to read, inline it obfuscates the control flow.

skrifa/src/outline/autohint/outline.rs
```diff
@@ -197,6 +202,35 @@ impl Outline {
...
+                const FLAGS_MAP: [PointFlags; 3] = [
+                    PointFlags::on_curve(),
+                    PointFlags::off_curve_quad(),
+                    PointFlags::off_curve_cubic(),
+                ];
+                path::ContourPoint {
+                    x: F26Dot6::from_bits(point.x),
+                    y: F26Dot6::from_bits(point.y),
+                    flags: FLAGS_MAP[(point.flags & Point::CONTROL) as usize],
```
rsheeter, line 225: I initially expected `flags: PointFlags::from_bits(point.flags)` and was confused. autohint::outline::Point has `flags: u8` which looks an awful lot like it should fit there. 

At the absolute minimum I suggest comments here and in the autohint Point flags field to clarify they arne't the same. 

Unless there is a good reason not to I suggest we make autohint Point flags a newtype like glyf::PointFlags. Presumably it could then have a conversion method to produce glyf PointFlags and then this could would flow very idiomatically.

skrifa/src/outline/autohint/outline.rs
```diff
@@ -697,4 +731,56 @@ mod tests {
...
+        compare_outlines("MOSTLY_OFF_CURVE", font_test_data::MOSTLY_OFF_CURVE);
+        compare_outlines("STARTING_OFF_CURVE", font_test_data::STARTING_OFF_CURVE);
+        compare_outlines("CUBIC_GLYF", font_test_data::CUBIC_GLYF);
+        compare_outlines("VAZIRMATN_VAR", font_test_data::VAZIRMATN_VAR);
+        compare_outlines("CANTARELL_VF_TRIMMED", font_test_data::CANTARELL_VF_TRIMMED);
+        compare_outlines(
+            "NOTO_SERIF_DISPLAY_TRIMMED",
+            font_test_data::NOTO_SERIF_DISPLAY_TRIMMED,
+        );
+    }
```
rsheeter, line 749: This is many distinct tests hiding in one. It's a bad pattern that I've seen cause inconvenience many times because it denies us the ability to see which tests fail by just bailing on first failure.

Make individual `#[test]` entrypoints instead.

skrifa/src/outline/autohint/outline.rs
```diff
@@ -697,4 +731,56 @@ mod tests {
...
+                    point.x = point.fx << 6;
+                    point.y = point.fy << 6;
+                }
+                let mut autohint_svg = SvgPen::default();
+                outline.to_path(path_style, &mut autohint_svg);
+                let autohint_svg = autohint_svg.to_string();
+                assert_eq!(
+                    base_svg, autohint_svg,
+                    "autohint outline to path comparison failed for glyph {gid} of font {font_name} with style {path_style:?}"
+                );
```
rsheeter, line 782: Similar to above, you are going to try many things in this test but bail as soon as one fails. Maybe accumulate failures and then assert there aren't any at the end? - have to make sur ethe message has enough info to be useful ofc.

rsheeter (approved):
Couple of nits but generally LGTM

skrifa/src/outline/autohint/outline.rs
dfrg: Made this a method on `Point`

skrifa/src/outline/autohint/outline.rs
dfrg: Changed to just use the existing `glyf::PointFlags` entirely. Added a couple new "markers" and some new methods to support this.

skrifa/src/outline/autohint/outline.rs
dfrg: Done

skrifa/src/outline/cff/mod.rs
```diff
@@ -676,6 +681,38 @@ mod tests {
...
 
+    #[test]
+    fn unhinted_ends_with_close() {
+        let font = FontRef::new(font_test_data::CANTARELL_VF_TRIMMED).unwrap();
+        let glyph = font.outline_glyphs().get(GlyphId::new(1)).unwrap();
+        let mut svg = SvgPen::default();
+        glyph.draw(Size::unscaled(), &mut svg).unwrap();
+        assert_eq!(
+            svg.to_string(),
+            "M83.0,0.0 L163.0,0.0 L163.0,482.0 L83.0,482.0 Z M124.0,595.0 C160.0,595.0 181.0,616.0 181.0,652.0 C181.0,688.0 160.0,709.0 124.0,709.0 C88.0,709.0 67.0,688.0 67.0,652.0 C67.0,616.0 88.0,595.0 124.0,595.0 Z"
```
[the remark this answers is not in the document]
dfrg: Think I meant to do this but forgot. Now just checks `ends_with('Z')` which is all we care about.

PR #1116, opened 2024-08-28
[skrifa] autohint: latin edge metrics

dfrg:
Updates latin outline hinting to return a structure describing positions of the left and right-most edges which will be used for advance width adjustment. Updates the outer hinting function test to check these.

Includes a bit of additional code shuffling in the edge computation to better match FT.

### DIFF
--- a/skrifa/src/outline/autohint/hint.rs
+++ b/skrifa/src/outline/autohint/hint.rs
@@ -314,9 +314,9 @@ mod tests {
     };
 
     #[test]
-    fn hinted_coords() {
+    fn hinted_coords_and_metrics() {
         let font = FontRef::new(font_test_data::NOTOSERIFHEBREW_AUTOHINT_METRICS).unwrap();
-        let outline = hint_latin_outline(
+        let (outline, metrics) = hint_latin_outline(
             &font,
             16.0,
             Default::default(),
@@ -364,6 +364,14 @@ mod tests {
             .map(|point| (point.x, point.y))
             .collect::<Vec<_>>();
         assert_eq!(coords, expected_coords);
+        let expected_metrics = latin::HintedMetrics {
+            x_scale: 67109,
+            left_opos: 15,
+            left_pos: 0,
+            right_opos: 210,
+            right_pos: 192,
+        };
+        assert_eq!(metrics.unwrap(), expected_metrics);
     }
 
     // Specific test case for <https://issues.skia.org/issues/344529168> which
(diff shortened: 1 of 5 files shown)


skrifa/src/outline/autohint/latin/mod.rs
```diff
@@ -22,18 +22,41 @@ fn derived_constant(units_per_em: i32, value: i32) -> i32 {
...
+    pub left_opos: i32,
+    pub left_pos: i32,
+    pub right_opos: i32,
+    pub right_pos: i32,
+}
+
+/// Applies the complete hinting process to a latin outline.
+///
+/// See <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/aflatin.c#L3554>
+pub(crate) fn hint_outline(
```
rsheeter, line 39: I don't find the interpretation of no HintedMetrics (None) very clear, suggest an additional comment to say what that means.

skrifa/src/outline/autohint/hint.rs
rsheeter, line 1: It would be really helpful to beef up the comments at the top. For example:

* What's a weak or no-weak point?
* What does directly attached to edges mean, is it the same as "on curve"?
* What the heck is a blue zone?
* What types of zones are there? - one comment references neutral ones. Are there also warlike ones?!

rsheeter (approved):
I think there is a great deal of room to better explain what on earth hinting is actually doing but that seems out of scope here :)

skrifa/src/outline/autohint/hint.rs
```diff
@@ -364,6 +364,14 @@ mod tests {
...
         assert_eq!(coords, expected_coords);
+        let expected_metrics = latin::HintedMetrics {
+            x_scale: 67109,
+            left_opos: 15,
+            left_pos: 0,
+            right_opos: 210,
+            right_pos: 192,
+        };
+        assert_eq!(metrics.unwrap(), expected_metrics);
     }
```
rsheeter, line 375: We should have a test where the expected metrics is None?

skrifa/src/outline/autohint/latin/mod.rs
dfrg: This returns `None` when we have fewer than two edges because there isn't enough data to adjust metrics. Think of empty (spacing) glyphs or degenerate glyphs where all x coords are (roughly) the same. Noted as such.

skrifa/src/outline/autohint/hint.rs
dfrg: Added comments elaborating on the first two. Blue zones aren't relevant to this file so those will need further explanation at a later time. If you're looking for a short summary here, blue zones are basically just font-wide alignment regions to ensure that hinting adjustments are consistent across all glyphs. Neutral zones are those that are not bound to a particular metric (top/bottom/x-height) and only apply to flat segments (they're ignored for overshoots).

skrifa/src/outline/autohint/hint.rs
dfrg: Added another test with space glyph which triggers this.

PR #1117, opened 2024-08-28
[klippa] nameid_closure

qxliu76:
No tests added for some tables(layout, CPAL) because I did't find a testcase.


klippa/src/lib.rs
```diff
@@ -394,6 +449,30 @@ fn get_font_num_glyphs(font: &FontRef) -> usize {
     ret.max(maxp.num_glyphs() as usize)
 }
 
+fn remap_indices<T: Domain + std::cmp::Eq + std::hash::Hash>(
+    indices: IntSet<T>,
+) -> FnvHashMap<T, T> {
+    indices
+        .iter()
+        .enumerate()
+        .map(|x| (x.1, T::from_u32(InDomain::new(x.0 as u32))))
```
garretrieger, line 458: InDomain construction is meant to be only done internally in IntSet. I think you can add the From<u32> trait to T to allow you to construct one from a u32 in this function.

read-fonts/src/collections/int_set/mod.rs
```diff
@@ -98,6 +98,10 @@ impl InDomain {
     pub fn value(&self) -> u32 {
         self.0
     }
+
+    pub fn new(value: u32) -> Self {
```
garretrieger, line 102: Can you remove this, InDomain is not meant to be constructable outside of IntSet (see my above comment for how to work around).

font-types/src/name_id.rs
```diff
@@ -153,6 +153,10 @@ impl NameId {
     /// If present in a variable font, it may be used as the family prefix in the PostScript Name Generation
     /// for Variation Fonts algorithm.
     pub const VARIATIONS_POSTSCRIPT_NAME_PREFIX: Self = Self(25);
+
+    /// Invalid `name_id`.
+    /// Value to represent a nonexistent name ID.
+    pub const INVALID: Self = Self(0xFFFF);
```
dfrg, line 159: I think we should prefer `Option` rather than in-band signaling whenever possible. I have a suggestion for this below.

klippa/src/fvar.rs
```diff
@@ -0,0 +1,54 @@
...
+            .instances()
+            .iter()
+            .filter_map(|x| x.ok())
+        {
+            plan.name_ids.insert(instance_record.subfamily_name_id);
+            if let Some(name_id) = instance_record.post_script_name_id {
+                if name_id != NameId::INVALID {
+                    plan.name_ids.insert(name_id);
+                }
+            }
```
dfrg, line 25: Since the spec does allow these to be sparse, I think we should handle it in `InstanceRecord::read_with_args()` rather than here, returning `None` when the PostScript name id is `0xFFFF`.

Specifically, here: https://github.com/googlefonts/fontations/blob/5e0b6ae79bbfb98afaa35173a5752e8ec6d2a71c/read-fonts/src/tables/instance_record.rs#L52

I'd change that code to something like:
```rust
let id: NameId = cursor.read()?;
// From <https://learn.microsoft.com/en-us/typography/opentype/spec/fvar#instancerecord>:
// "If the value is 0xFFFF, then the value is ignored, and no
// PostScript name equivalent is provided for the instance."
(id.to_u16() != 0xFFFF).then_some(id)
```

And then we can avoid the invalid check (and remove the INVALID const in NameId).

klippa/src/cpal.rs
```diff
@@ -0,0 +1,27 @@
+//! impl subset() for CPAL table
+
+use crate::{NameidClosure, Plan};
+use write_fonts::read::{tables::cpal::Cpal, types::NameId};
+
+impl<'a> NameidClosure for Cpal<'a> {
```
dfrg, line 6: Small nit: can we rename this trait to `NameIdClosure`?

I'm parsing "Nameid" as some sort of Greek hero which is spoiling our Old Norse theme for me :)

klippa/src/fvar.rs
qxliu76: oh yes this makes more sense. Thanks!

klippa/src/cpal.rs
qxliu76: Done.

dfrg (approved):
Looks good!

PR #1118, opened 2024-08-29
Add Hash impl for Location

LaurenzV:
I hope that's okay to add, I want to be able to store location's in a hashmap so I can map combinations of fonts + location to specific PDF fonts in my crate.

### DIFF
--- a/skrifa/src/collections.rs
+++ b/skrifa/src/collections.rs
@@ -6,7 +6,7 @@ use alloc::vec::Vec;
 ///
 /// Note that unlike the real `SmallVec`, this only works with types that
 /// are `Copy + Default` to simplify our implementation.
-#[derive(Clone)]
+#[derive(Clone, Hash)]
 pub(crate) struct SmallVec<T, const N: usize>(Storage<T, N>);
 
 impl<T, const N: usize> SmallVec<T, N>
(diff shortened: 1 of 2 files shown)


skrifa/src/collections.rs
```diff
@@ -224,7 +224,7 @@ where
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Hash)]
```
dfrg, line 227: For consistency, we’ll want to manually impl `Hash` here based on the slice data the same way we do for `PartialEq`, otherwise we may end up with situations where equal values produce different hashes due to the intermediate structure.

dfrg (commented):
Happy to add this functionality once we ensure consistency between `PartialEq` and `Hash`.

skrifa/src/collections.rs
LaurenzV: You are talking about the `Hash` derive for `SmallVec`, not `Storage`, right? As far as I can tell, `Storage` doesn't implement `PartialEq`. Let me know if I misunderstood, I've updated it now.

dfrg: Oops, yes, I meant the derive on `SmallVec`. Your change looks good now.

Looks like the now unused `Hash` derive is still on `Storage`. Let's remove that and then I'll merge this.

dfrg (approved):
Looks good, thanks!

PR #1119, opened 2024-08-29
Add be_buffer! macro.

garretrieger:
Simplifies the construction of test data via BeBuffer.


rsheeter (approved):
I'm not entirely sold on this macro being worth the complication, I guess if it feels like it's helping you write tests it's OK.

PR #1120, opened 2024-08-30
avar2 support

simoncozens:
Parsing so far; possible writing (untested); but no segment map application yet. This has required messing about with the codegen to allow variable length subtables in the middle of a table - see commit message of first commit for more details.


font-test-data/test_data/ttx/avar2checker.ttx
rsheeter, line 1: Two small suggestions:

1. You don't seem to be using most of the contents of the file, could it be minimized? For example:
   * do you need many glyphs?
   * Do they need to have outlines? 
   * If they do - perhaps you will use them in a future test? - could they be minimizl outlines, e.g. a single glyph with triangle or box?
1. It would be helpful to add comments annotating the interesting bits, e.g. what does the avar2 *do*
   * Doesn't have to be verbose, ex https://github.com/googlefonts/fontations/blob/main/font-test-data/test_data/ttx/mostly_off_curve.ttx#L189

read-fonts/src/array.rs
```diff
@@ -147,6 +147,17 @@ impl<'a, T: FontRead<'a> + VarSize> VarLenArray<'a, T> {
             Some(next)
         })
     }
+
+    pub fn counted_size(&self, elements: usize) -> Result<usize, ReadError> {
```
rsheeter, line 151: I think I would have to check the code to know what a counted size is. Suggest at least adding a comment, potentially renaming, for example element_size or size_of_elements would be more obvious.

I immediately wonder if one would ever want the size of something other than 0..elements, e.g. should we take a range rather than just #elements. Admittedly I don't have a specific use in mind so feel free to ignore.

rsheeter:
Took a quick skim and only saw minor nits but I think we want @cmyr to approve given the codegen changes.

font-test-data/test_data/ttx/avar2checker.ttx
simoncozens: The file only has two glyphs and the outlines are already pretty minimal. It's Laurence's avar2 test font, so I think it's preferable just to keep it in sync with Laurence's version and drop in any updates rather than customise it too much ourselves.

rsheeter: Fair enough. Could you add an entry to https://github.com/googlefonts/fontc/blob/main/resources/testdata/README.md that specifies the repo/path to obtain the file? - will help with https://github.com/googlefonts/fontations/issues/1127

cmyr (changes requested):
Sorry for the delay, this fell through the cracks while I was off.

*I'm going to just focus on the codegen changes for now, since that's really the meat of this PR; once we have the codegen right then everything else is pretty straight-forward.*

avar2 introduces a pattern we have not seen in a previous table, which is a sequence of variable-length items which is not the last item in a table.

One of the major considerations for codegen is to try, whenever possible, to reduce the number of special cases. This makes it easier for contributors, but it also is important for reducing as much as possible the complexity of the codegen code, since it can be particularly difficult to reason about.

Given that goal, I think there is a slightly more aggressive change we can make to how `VarLenArray` works that will be an overall improvement, and also support your case.

### design proposal 

I think you're on the right track here, but after looking at how `VarLenArray` is used elsewhere, I think we should avoid having two different paths for 'counted' and 'uncounted', and just require that `VarLenArray` is always provided with a count.

- for avar/gvar/fvar this will be fairly trivial; we'll need to define a new count helper that extracts the count from the `TupleVariationCount` field.
- for `post` we will need a slightly more involved count helper that looks at the `glyphNameIndex` array and determines how many strings we're storing by looking at the largest referenced index.

Then, instead of adding a method to the specific type we should make use of the existing [`VarSize` trait](https://github.com/googlefonts/fontations/blob/6d01f7425b52f06d23d8e24f433f09d2e3779ac5/read-fonts/src/read.rs#L85), which must already be implemented for any type used with `VarLenArray`. It looks like we can handle this pretty easily with a single inherit method on the trait, that just calls the existing `read_len_at` method, which would look something like,


```rust
pub trait VarSize {
    /// The type of the first (length) field of the item.
    ///
    /// When reading this type, we will read this value first, and use it to
    /// determine the total length.
    type Size: Scalar + Into<u32>;

    #[doc(hidden)]
    fn read_len_at(data: FontData, pos: usize) -> Option<usize> {
        let asu32 = data.read_at::<Self::Size>(pos).ok()?.into();
        (asu32 as usize).checked_add(Self::Size::RAW_BYTE_LEN)
    }


    /// Determine the total length required to store `count` items of `Self` in
    /// `data` starting from `start`.
    #[doc(hidden)]
    fn total_len_for_count(data: FontData, start: usize, count: usize) -> Result<usize, ReadError> {
        (0..count).try_fold(start, |current_pos, _i| {
            Self::read_len_at(data, current_pos)
                .and_then(|i_len| current_pos.checked_add(i_len))
                .ok_or(ReadError::OutOfBounds)
        })
    }
}
```

(this impl is untested, but gives the general idea)

Alongside this change, it would be good if we also had specific codegen tests for this logic. This would involve updating `resources/codegen_inputs/test_offsets_arrays.rs` to add a table that uses `VarLenArray`, and then writing some tests in `read-fonts/src/codegen_test.rs` that construct the test table under various conditions and ensure that everything is sane.

My strong preference is to do these codegen changes on their own, as a separate PR; that can be reviewed and merged in isolation, and then we can have a much simpler avar2 PR on top of it.


If you're interested in writing this patch you are very welcome to, otherwise I can find time to do it next week.

read-fonts/src/lib.rs
```diff
@@ -60,6 +60,8 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
+extern crate alloc;
```
cmyr, line 63: This shouldn't be necessary... if you have any code that requires allocation it should be gated on `std`.

cmyr (approved):
Okay this all looks sane, but I have a few little questions & thoughts inline, but big picture I think this is basically mergeable.

font-test-data/test_data/extracted/avar2checker-glyphs.txt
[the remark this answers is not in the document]
cmyr: @dfrg how does this work? Is there a mechanism for adding a font here that doesn't get the text file generated?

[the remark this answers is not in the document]
dfrg: Rebuild automatically does this but feel free to remove this generated file if you’d like. I plan on ripping this all out and replacing it with fauntlet in CI soon since this doesn’t handle hinting at all.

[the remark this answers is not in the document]
dfrg: In fact, if you want to change rebuild.py to not call extract_glyphs, that’s fine too

read-fonts/src/lib.rs
dfrg: Yes, avar2 fundamentally requires allocation. Though I think we can cover 100% of real world fonts by attempting allocation on the stack first. An array of size 64 is probably of reasonable size. If you’d prefer not to add that here, I’ll do it as a follow up.

read-fonts/src/tables/fvar.rs
```diff
@@ -71,6 +74,35 @@ impl<'a> Fvar<'a> {
...
+            if let Some(Ok(varstore)) = var_store.as_ref() {
+                if let Ok(delta) =
+                    varstore.compute_float_delta(var_index.unwrap(), normalized_coords)
+                {
+                    new_coords[i] = F2Dot14::from_f32((*v).apply_float_delta(delta))
+                        .clamp(F2Dot14::MIN, F2Dot14::MAX);
+                }
+            }
+        }
+        normalized_coords.copy_from_slice(&new_coords);
```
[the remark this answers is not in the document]
dfrg: If you’d prefer not to have the unconditional dependency on alloc, we can change `user_to_normalized` to not take the avar parameter and then `Avar` would get a new method that takes both input and output slices?

read-fonts/src/lib.rs
cmyr: My issue is more with our pattern for using `std` vs `alloc`; so far we've just been either `std` or not, and bringing in alloc sort of confuses that?

If we can just do `let mut buf = [F2Dot14::ZERO; 64];` and work in there that seems like a reasonable way to punt on `alloc` for now.

font-test-data/test_data/extracted/avar2checker-glyphs.txt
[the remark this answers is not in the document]
cmyr: okay @simoncozens let's just manually delete the file from this PR.

read-fonts/src/tables/fvar.rs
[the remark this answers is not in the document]
cmyr: the inline array approach seems like a reasonable one for now? But I defer to you if you think one approach is better than another.

read-fonts/src/lib.rs
simoncozens: Yeah, I can do that. The only reason I didn't is because my head was in "just implement the spec" mode, `axisCount` is a uint16, and that's a lot of bytes to put on the stack. Maybe we can do both - in a `no-std` environment, use a stack-allocated array and blow up if there are more than 64 axis: in `std`, DTRT with allocation.

cmyr: yea let's just do the 64 for the time being to get this merged, and then I we'll open an issue to figure out what the best long-term approach is.

dfrg: This sgtm as well. I assume most people are going to be using skrifa anyway so we can muck about with the read-fonts side to find something suitable.

read-fonts/src/tables/fvar.rs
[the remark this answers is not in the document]
simoncozens: I've gone with an inline array. One function to do the mapping which works everywhere is a much nicer interface.

cmyr (approved):
This looks good to me, thanks! @simoncozens ping me to confirm there's nothing else to do here and I'll merge.

PR #1121, opened 2024-08-30
Convert all IFT test data to use BeBuffer

garretrieger:
Also move it to it's own module.


font-test-data/src/ift.rs
```diff
@@ -0,0 +1,412 @@
+use read_fonts::types::Tag;
```
rsheeter, line 1: Suggest a brief intro comment, e.g. `//! Test data for incremental transfer [link]

PR #1124, opened 2024-08-30
[IFT] Add support for handling Format 2 entries with string IDs.

garretrieger:
When a string data block is present then entry id deltas are instead interpretted as string lengths in order to assign string ids to entries (instead of the usual numeric ones).

See: https://w3c.github.io/IFT/Overview.html#mapping-entry-entryidstringlength

### DIFF
--- a/read-fonts/src/tables/ift.rs
+++ b/read-fonts/src/tables/ift.rs
@@ -46,6 +46,8 @@ impl ReadArgs for IdDeltaOrLength {
 
 impl ComputeSize for IdDeltaOrLength {
     fn compute_size(entry_id_string_data_offset: &Offset32) -> Result<usize, ReadError> {
+        // This field is either a u16 or an int24 depending on whether or not string data
+        // is present. See: <https://w3c.github.io/IFT/Overview.html#mapping-entry-entryiddelta>
         Ok(if entry_id_string_data_offset.is_null() {
             3
         } else {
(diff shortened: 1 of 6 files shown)


read-fonts/src/tables/ift.rs
```diff
@@ -37,6 +37,43 @@ impl U8Or16 {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct IdDeltaOrLength(i32);
+
+impl ReadArgs for IdDeltaOrLength {
+    type Args = Offset32;
+}
+
+impl ComputeSize for IdDeltaOrLength {
+    fn compute_size(entry_id_string_data_offset: &Offset32) -> Result<usize, ReadError> {
+        Ok(if entry_id_string_data_offset.is_null() {
```
rsheeter, line 49: nit: it would be nice to cite spec here, the 3/2 is a wee bit magical

read-fonts/src/tables/ift.rs
```diff
@@ -37,6 +37,43 @@ impl U8Or16 {
...
+            data.read_at::<Int24>(0).map(|v| Self(i32::from(v)))
+        } else {
+            data.read_at::<u16>(0).map(|v| Self(v as i32))
+        }
+    }
+}
+
+impl IdDeltaOrLength {
+    #[inline]
+    pub fn get(self) -> i32 {
```
rsheeter, line 72: since this consumes self by convention I think it should be `into_inner`?

rsheeter (approved):

PR #1125, opened 2024-09-01
[skrifa] autohint: prep for CJK

dfrg:
Just adds a CJK test font, a test for CJK widths (exact same computation as Latin), and shuffles the current Latin blues code around to make way for the CJK stuff.

No functional changes.

Refactoring the Latin blues code from method to free function generated a ton of noise in the diff so this was pulled out to make the CJK blues work actually reviewable without the noise.


font-test-data/test_data/ttx/notoseriftc_autohint_metrics.ttx
rsheeter, line 1: I'm belatedly realizing we've done a really bad job documenting where our ttx files came from. Suggest adding a bullet to ... maybe font-test-data/test_data/README.md ... saying how to produce this file, e.g. get the whole binary from repo, subset using _command_ and ttx

skrifa/src/outline/autohint/latin/widths.rs
```diff
@@ -134,6 +134,32 @@ mod tests {
         );
     }
 
+    #[test]
+    fn cjk_computed_widths() {
+        check_widths(
+            font_test_data::NOTOSERIFTC_AUTOHINT_METRICS,
+            super::ScriptClass::HANI,
+            [
```
rsheeter, line 142: suggest comment explaining where the expected value scome from. FT I assume, but how do you actually create the data? Or is it just what we observed our code to produce?

rsheeter (approved):

font-test-data/test_data/ttx/notoseriftc_autohint_metrics.ttx
rsheeter: Context https://github.com/googlefonts/fontations/issues/1127, it'll make our lives easier to start capturing the info ^

dfrg: Added a section for this to the test_data readme with details about the font source, license, usage and command used to subset

skrifa/src/outline/autohint/latin/widths.rs
dfrg: Noted that these values are scraped from a FT debugging session.

In this case, I just copied the test skeleton and manually transcribed the values from the debugger. I figured the odds of me making a transcription error that matches a bug are vanishingly low :) For tests with more data (segments, edges, etc), I have some ugly printf code in my local FT copy that prints the expected values to the console and I copy that into the tests. This is really handy in cases where FT uses pointers but we use indices.

I tend to only do "regression style" tests (copying in values that we produce) when we already have known good tests to start with.

PR #1126, opened 2024-09-01
[skrifa] autohint: the CJK blues

dfrg:
The actual CJK blue zone computation.

Based on #1125


skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+        }
+        let mut n_flats = 0;
+        let mut n_fills = 0;
+        let mut is_fill = true;
+        for ch in *blue_chars {
+            if *ch == '|' {
+                is_fill = false;
+                continue;
+            }
+            // TODO: do some shaping
```
rsheeter, line 476: Unless the shaping is coming soon I'd prefer an issue and a reference to that issue

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+            }
+            let Some(glyph) = glyphs.get(gid) else {
+                continue;
+            };
+            outline_buf.clear();
+            if glyph.draw_unscaled(coords, None, &mut outline_buf).is_err() {
+                continue;
+            }
+            let outline = outline_buf.as_ref();
+            // Reject glyphs that don't produce any rendering
```
rsheeter, line 491: s/don't/can't/ because a glyph with > 2 points still might not render anything?

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+        let is_top = blue_flags & blue_flags::TOP != 0;
+        let blues = &mut blues[!is_horizontal as usize];
+        if blues.len() >= MAX_BLUES {
+            continue;
+        }
+        let mut n_flats = 0;
+        let mut n_fills = 0;
+        let mut is_fill = true;
+        for ch in *blue_chars {
+            if *ch == '|' {
```
rsheeter, line 472: This magic perhaps merits a comment?

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
     blues
 }
 
+/// Compute unscaled blue values for the CJK script set.
+///
+/// See <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/afcjk.c#L277>
+fn compute_cjk_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass) -> [UnscaledBlues; 2] {
+    let mut blues = [UnscaledBlues::new(), UnscaledBlues::new()];
```
rsheeter, line 442: latin only got one, suggest a comment explaining why we need two

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+    let mut outline_buf = UnscaledOutlineBuf::<MAX_INLINE_POINTS>::new();
+    let mut flats = [0; BLUE_STRING_MAX_LEN];
+    let mut fills = [0; BLUE_STRING_MAX_LEN];
```
rsheeter, line 445: The blues functions are long and somewhat duplicative. Perhaps we can exploit. For example, the use of UnscaledOutlineBuf<256>, [i32; 51], [i32; 51] is common across both blues fns so I imagine we could do something like:

```rust
#[inline(always)]
fn buffers() -> (UnscaledOutlineBuf<MAX_INLINE_POINTS>, [i32; BLUE_STRING_MAX_LEN], [i32; BLUE_STRING_MAX_LEN]) {
    (UnscaledOutlineBuf::<MAX_INLINE_POINTS>::new(), [0; BLUE_STRING_MAX_LEN], [0; BLUE_STRING_MAX_LEN])
}
...
// compute_default_blues
let (mut outline_buf, mut flats, mut rounds) = buffers();
// compute_cjk_blues doesn't quite use the same names ... perhaps a comment explaining this should exist somewhere?
let (mut outline_buf, mut flats, mut fills) = buffers();


````

File: skrifa/src/outline/autohint/latin/blues.rs [hunk 6]
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+    let glyphs = font.outline_glyphs();
+    let charmap = font.charmap();
+    let units_per_em = font
+        .head()
+        .map(|head| head.units_per_em())
+        .unwrap_or_default() as i32;
```
rsheeter, line 451: as with buffers this seems like it could be hoisted out. `let (glyphs, charmap, units_per_em) = things_all_blues_need(font);` (we must get our Dr Seuss refs in; https://seuss.fandom.com/wiki/Thneed)

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+            } else {
+                outline.find_last_contour(|point| {
+                    if best_pos.is_none() || Some(point.y) < best_pos {
+                        best_pos = Some(point.y);
+                        true
+                    } else {
+                        false
+                    }
+                });
+            }
```
rsheeter, line 536: Isn't `find_last_contour` just `reduce` with unnecessary bookkeeping in this case?

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -436,6 +435,153 @@ fn compute_default_blues(font: &FontRef, coords: &[F2Dot14], style: &StyleClass)
...
+            let Some(best_pos) = best_pos else {
```
rsheeter, line 537: You checked if the outline has points above. find_last_contour walks all the points. Unless I'm missing something this can't be None.

skrifa/src/outline/autohint/latin/blues.rs
rsheeter: Actually per below I think you have to have points so you could write this with fold and best_pos would never be None?

skrifa/src/outline/autohint/latin/blues.rs
```diff
@@ -527,4 +673,29 @@ mod tests {
...
+            UnscaledBlue {
+                position: -78,
+                overshoot: -66,
+                ascender: 0,
+                descender: 0,
+                flags: 0,
+            },
+        ];
+        assert_eq!(values, &expected);
+    }
```
rsheeter, line 700: Seems a little light, I suppose we plan to lean on fuzzing and fauntlet for more substantial testing?

skrifa/src/outline/autohint/latin/blues.rs
rsheeter: (resolving because this goes away if you reduce)

rsheeter (approved):
LGTM with minor suggestions about deduplication

skrifa/src/outline/autohint/latin/blues.rs
dfrg: #1128

skrifa/src/outline/autohint/latin/blues.rs
dfrg: Indeed. I waffled on this but kept the common code so that it mirrored the default blue computation because I felt the similarity would make the code easier to understand. I'm fine with using a reduction though.. made it so based on your code, ty :)

The only part I didn't include is the character to glyph mapping. I deliberately avoided fanciness here because shaping will require changes anyway which will unify this.

skrifa/src/outline/autohint/latin/blues.rs
dfrg: Yes, the plan is to use broad scale testing to detect bugs which may be used to inform additional unit tests.

PR #1131, opened 2024-09-04
gasp is a top-level table

simoncozens:
Add a tag and accessor for it.

Tests are failing for me locally, but they are on main as well.


rsheeter:
I think some tests require a venv and do not clearly report as much, guessing thats why some tests fail.

rsheeter (approved):

Issue #1129, opened 2024-09-04
[skrifa] CJK autohinting: scaled widths are always 0?


dfrg:
The FreeType CJK autohinter never compute scaled/fitted width values and debugging suggests they are always 0 which seems suspicious.

ref: https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/afcjk.c#L661

We'll just match this behavior for now but should investigate later to see what's going on.

rsheeter:
Sounds worth FreeType issues?

PR #1132, opened 2024-09-04
[skrifa] autohint: CJK segments

dfrg:
Adds the segment computation for CJK.

There's some dead code here that exists in FT but is never executed due to bugs. See #1129

Based on #1130 which should land first


rsheeter (approved):
I don't feel confident at all about what it does but it does seem sort of similar to the FT version :)

dfrg:
At this point, I'll take sort of similar :)

PR #1134, opened 2024-09-05
[skrifa] autohint: CJK edges

dfrg:
This was similar enough to latin that changes were folded into the current code with some branches.


skrifa/src/outline/autohint/axis.rs
```diff
@@ -184,6 +184,10 @@ impl Segment {
         edges.get(self.edge_ix.map(|ix| ix as usize)?)
     }
 
+    pub fn edge_next<'a>(&self, segments: &'a [Segment]) -> Option<&'a Segment> {
```
rsheeter, line 187: is there a reason this isn't `next_edge`? Does it mean something different?

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -21,10 +21,11 @@ pub(crate) fn compute_edges(
     metrics: &ScaledAxisMetrics,
     top_to_bottom_hinting: bool,
     y_scale: i32,
+    group: ScriptGroup,
 ) {
     axis.edges.clear();
     let scale = metrics.scale;
-    let top_to_bottom_hinting = if axis.dim == Axis::HORIZONTAL {
+    let top_to_bottom_hinting = if axis.dim == Axis::HORIZONTAL || group != ScriptGroup::Default {
```
rsheeter, line 28: horizontal implies not top-to-bottom makes sense. The group limitation somewhat less so. I suggest a comment.

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -39,77 +40,130 @@ pub(crate) fn compute_edges(
...
-                continue 'segments2;
+    }
+    if group == ScriptGroup::Default {
+        // Loop again to find single point segments without a direction and
+        // associate them with an existing edge if possible
+        'segments: for segment_ix in 0..axis.segments.len() {
+            let segment = &axis.segments[segment_ix];
+            if segment.dir != Direction::None {
+                continue;
+            }
```
rsheeter, line 147: Why not `for (ix, segment) in axis.segments.iter().enumerate().filter(no direction)`? - particularly the `.iter().enumerate()`

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -39,77 +40,130 @@ pub(crate) fn compute_edges(
...
+            // Find a matching edge
```
rsheeter, line 148: Suggest `// Find the first matching edge`

skrifa/src/outline/autohint/latin/edges.rs
rsheeter: Would that create unpleasently overlapping borrows? - if so suggest a comment noting as much

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -39,77 +40,130 @@ pub(crate) fn compute_edges(
...
+            for edge_ix in 0..axis.edges.len() {
+                let edge = &axis.edges[edge_ix];
+                let dist = (segment.pos as i32 - edge.fpos as i32).abs();
+                if dist < edge_distance_threshold {
+                    // We found an edge, link everything up
+                    axis.append_segment_to_edge(segment_ix, edge_ix);
+                    // Move to next segment
+                    continue 'segments;
+                }
```
rsheeter, line 157: At a glance this also looks possible to do via iterators? - I didn't try so perhaps not; if not perhaps add a comment explaining why.

`if let Some(edge_ix) = axis.edges.iter().enumerate().filter_map(if dist < threshold retain index).find()` type of deal?

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -223,16 +277,25 @@ pub(crate) fn compute_blue_edges(
     scale: &Scale,
     unscaled_blues: &[UnscaledBlue],
     blues: &[ScaledBlue],
+    group: ScriptGroup,
 ) {
-    if axis.dim != Axis::VERTICAL {
+    // For the default script group, we only handle vertical blues
```
rsheeter, line 282: because ...

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -244,27 +307,42 @@ pub(crate) fn compute_blue_edges(
...
-                    scale.y_scale,
-                );
+                let (ref_pos, matching_blue) = if group == ScriptGroup::Default {
+                    (unscaled_blue.position, blue.position)
+                } else {
+                    // For CJK, we take the blue with the smallest delta
+                    // from the edge
+                    // See <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/afcjk.c#L1356>
+                    if (edge.fpos as i32 - unscaled_blue.position).abs()
+                        > (edge.fpos as i32 - unscaled_blue.overshoot).abs()
```
rsheeter, line 317: I suggest adding some methods to things like `ScaledBlue` and `Edge`. For example, instead of `let is_top = blue.flags & (blue_flags::TOP | blue_flags::LATIN_SUB_TOP) != 0;` we could have `blue.is_top()`, instead of `(edge.fpos as i32 - unscaled_blue.position).abs()` we could have `edge.distance_to(unscaled_blue)` (...or would it need to be specifically unscaled_distance_to?)

This could be deferred to an issue if we don't want to address here. I feel like autohinting would be quite a bit easier to read if widely deployed.

skrifa/src/outline/autohint/latin/edges.rs
```diff
@@ -223,16 +277,25 @@ pub(crate) fn compute_blue_edges(
...
+    let axis_scale = if axis.dim == Axis::HORIZONTAL {
+        scale.x_scale
+    } else {
+        scale.y_scale
+    };
+    // Initial threshold
+    let initial_best_dest = fixed_mul(scale.units_per_em / 40, axis_scale).min(64 / 2);
     for edge in &mut axis.edges {
         let mut best_blue = None;
         let mut best_is_neutral = false;
```
rsheeter, line 295: Optional. I don't have time to try it but finding the best values for an edge smells a lot like reduce. Is it possible it would make sense to write this in terms of reduce? - if it does but we don't want to do it now feel free to defer to an issue.

rsheeter (approved):

skrifa/src/outline/autohint/axis.rs
dfrg: This returns the next segment in the parent edge. This name was copied from FT but I agree that it's ambiguous. Renamed to `next_in_edge` for clarity.

skrifa/src/outline/autohint/latin/edges.rs
dfrg: I'm probably being overly defensive here but the CJK code always passes constant 0 where this value is expected. I added a comment with a link to FT code

skrifa/src/outline/autohint/latin/edges.rs
dfrg: Yes, we make mutable calls into `axis` below which trips up the borrow checker. Added a comment

skrifa/src/outline/autohint/latin/edges.rs
dfrg: Changed to "Try to find an edge that coincides with this segment within the threshold" which better explains what's happening here

skrifa/src/outline/autohint/latin/edges.rs
dfrg: Good catch. Done

skrifa/src/outline/autohint/latin/edges.rs
dfrg: ... FreeType does this. I added a link to the relevant code.

In reality, FT doesn't use any alignment zones in the horizontal direction but the infra exists for it. I believe this is an attempt at better hinting for vertical text but the code is permanently disabled with an `#undef`. After this all lands, I think it's worth considering simplifying our code and removing this altogether. Filed #1140

skrifa/src/outline/autohint/latin/edges.rs
dfrg: There's a lot that can be done to clean up the whole autohint module but I prioritized (mostly) matching the FT code in the first pass. I agree that we can do much better but I would like to defer until after we land. Filed #1141

skrifa/src/outline/autohint/latin/edges.rs
dfrg: Added to #1141. This seems possible but the interior code would need to be factored out into a function to make the change worthwhile.

PR #1136, opened 2024-09-05
[klippa] Use Subset trait

qxliu76:


klippa/src/lib.rs
```diff
@@ -508,6 +501,16 @@ pub trait NameIdClosure {
     fn collect_name_ids(&self, plan: &mut Plan);
 }
 
+pub trait Subset {
```
rsheeter, line 504: Clarify who should implement, e.g. top level tables

klippa/src/lib.rs
```diff
@@ -508,6 +501,16 @@ pub trait NameIdClosure {
...
+    /// Subset this table, if any error occurs, the table will be dropped
```
rsheeter, line 505: Dropped? Or just not added to builder in the first place?

Optional alternate phrasing:

`/// if successful a subset version of this table will have been added to builder`

rsheeter (approved):

PR #1138, opened 2024-09-06
[skrifa] autohint: CJK edge hinting

dfrg:
Completes the CJK hinting algorithm

Based on #1134


skrifa/src/outline/autohint/latin/hint.rs
```diff
@@ -45,62 +53,64 @@ pub(crate) fn hint_edges(
...
+    for edge_ix in 0..axis.edges.len() {
+        let edges = axis.edges.as_mut_slice();
+        let edge = &edges[edge_ix];
+        if edge.flags & Edge::DONE != 0 {
+            continue;
+        }
+        let edge2_ix = edge.link_ix.map(|x| x as usize);
+        let edge2 = edge2_ix.map(|ix| &edges[ix]);
+        // If we have two neutral zones, skip one of them.
+        if let (true, Some(edge2)) = (edge.blue_edge.is_some(), edge2) {
```
cmyr, line 73: can't believe we still don't have if-let chains!? but apparently not (https://github.com/rust-lang/rust/issues/53667)

(if we did, this could be,

```rust
if let Some(edge2) = edge2 && edge.blue_edge.is_some() { .. }
```
)

skrifa/src/outline/autohint/latin/hint.rs
```diff
@@ -129,102 +142,133 @@ fn align_stem_edges(
...
+                );
+                // Some "voodoo" to specially round edges for small stem widths
+                let (u_off, d_off) = if cur_len <= 64 {
+                    // width <= 1px
+                    (32, 32)
+                } else {
+                    // 1px < width < 1.5px
+                    (38, 26)
+                };
+                if cur_len < 96 {
```
cmyr, line 233: 🤨

cmyr (approved):
There is much to be dismayed about, but all of it is faithful to the original and I think that's definitely the way to go.

skrifa/src/outline/autohint/hint.rs
```diff
@@ -388,12 +409,156 @@ mod tests {
...
+    fn hinted_coords_and_metrics_cjk() {
+        let font = FontRef::new(font_test_data::NOTOSERIFTC_AUTOHINT_METRICS).unwrap();
+        let (outline, metrics) = hint_outline(
+            &font,
+            16.0,
+            Default::default(),
+            GlyphId::new(9),
+            &style::STYLE_CLASSES[style::StyleClass::HANI],
+        );
+        let expected_coords = [
```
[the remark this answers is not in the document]
dfrg: Noted that these are generated with some lovely printf debugging in FreeType

skrifa/src/outline/autohint/latin/hint.rs
dfrg: This seems to always be just over the horizon... at least we can avoid the double `if` with the tuple pattern matching so I don't want to complain too much

skrifa/src/outline/autohint/latin/hint.rs
dfrg: same :/

dfrg:
> There is much to be dismayed about, but all of it is faithful to the original and I think that's definitely the way to go.

Agreed. I filed #1141 to clean all of this up after we ensure matching behavior.

PR #1142, opened 2024-09-07
Generate docs for most features on docs.rs

waywardmonkeys:
This generates docs for all features in `font-types`, `skrifa`, and `write-fonts`. In `read-fonts`, only for `libm`, `serde`, and `std`.

### DIFF
--- a/font-types/Cargo.toml
+++ b/font-types/Cargo.toml
@@ -9,6 +9,11 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[package.metadata.docs.rs]
+# To build locally:
+# RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
+all-features = true
+
 [features]
 std = []
 bytemuck = ["dep:bytemuck"]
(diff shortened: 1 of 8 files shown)


waywardmonkeys:
There's an interesting detail here where I didn't enable `experimental_traversal` for `read-fonts` because it is marked as experimental and not part of the public API for semver. However, `skrifa` does enable it by default, but I figure that's for its internal usage and so we don't want to have docs for it on docs.rs.

alerque:
Can't the docs annotate any API surface that is marked as experimental so that it can both be document and not cause any semver issues if something changes?

waywardmonkeys:
My understanding is that it isn't meant for broader use, so my logic is that there's no point in documenting it on docs.rs at this point in time.

cmyr:
yes, let's leave traverse out of the docs :)

cmyr (approved):
This looks good, thanks!

cmyr:
rustfmt is displeased, otherwise happy to merge :)

waywardmonkeys:
@cmyr Fixed the rustfmt issue.

PR #1144, opened 2024-09-09
[skrifa] expose autohinting in public API

dfrg:
Adds all the plumbing to enable autohinting from a `HintingInstance`.

### DIFF
--- a/skrifa/src/outline/autohint/instance.rs
+++ b/skrifa/src/outline/autohint/instance.rs
@@ -1,9 +1,21 @@
 //! Autohinting state for a font instance.
 
-use super::{super::OutlineGlyphCollection, style::GlyphStyleMap};
-use crate::MetadataProvider;
+use crate::{attribute::Style, prelude::Size, MetadataProvider};
+
+use super::{
+    super::{
+        pen::PathStyle, AdjustedMetrics, DrawError, OutlineGlyph, OutlineGlyphCollection,
+        OutlinePen, SmoothMode, Target,
+    },
+    metrics::{fixed_mul, pix_round, Scale, UnscaledStyleMetricsSet},
+    outline::Outline,
+    style::GlyphStyleMap,
+};
 use alloc::sync::Arc;
-use raw::TableProvider;
+use raw::{
+    types::{F26Dot6, F2Dot14},
+    FontRef, TableProvider,
+};
 
 /// Set of derived glyph styles that are used for automatic hinting.
 ///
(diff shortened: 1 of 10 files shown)


skrifa/src/outline/autohint/instance.rs
```diff
@@ -30,3 +42,134 @@ impl GlyphStyles {
...
+}
+
+impl Instance {
+    pub fn new(
+        font: &FontRef,
+        outlines: &OutlineGlyphCollection,
+        coords: &[F2Dot14],
+        target: Target,
+        styles: Option<GlyphStyles>,
+        lazy_metrics: bool,
```
drott, line 62: Doesn't have to be in this CL, but perhaps this could be changed from a raw bool to something more expressive? But it looks like this isn't public API, so I have no strong opinion on this.

skrifa/src/outline/autohint/instance.rs
```diff
@@ -30,3 +42,134 @@ impl GlyphStyles {
...
+    }
+
+    pub fn draw(
+        &self,
+        size: Size,
+        coords: &[F2Dot14],
+        glyph: &OutlineGlyph,
+        path_style: PathStyle,
+        pen: &mut impl OutlinePen,
+    ) -> Result<AdjustedMetrics, DrawError> {
```
drott, line 94: I thought `HintingOptions`, `Target::Smooth` etc. were already exposed, so the subject of the PR means that the hinted metrics etc. would start to be returned? Otherwise I might be missing what is new public API in this PR. I was trying to look at it from the perspective of whether it matches the autohinting preview PR, which I assume it does.

drott (approved):
LGTM, thank you!

skrifa/src/outline/autohint/instance.rs
dfrg: Yeah, this option isn’t actually exposed yet so you get lazy metrics when compiling with `std` and precomputed metrics otherwise (since without `std` we don’t have access to the synchronization primitives). I’m not really sure yet how to present this option to the user.

drott: Ack, that's fine with me.

skrifa/src/outline/autohint/instance.rs
dfrg: A better title for this PR might be “Actually run the autohinter when requested” since the public API already exposed the option but the underlying code ignored it :) There shouldn’t be any changes wrt use from the preview branch.

drott: Ack, makes sense now.

PR #1146, opened 2024-09-09
[codegen] Support explicit #[count] arg for VarLenArray

cmyr:
Previously this only worked at the end of a table, with the #[count(..)] attribute.

- unblocks #1120

cc @simoncozens This is slightly different than what I'd described; specifically it doesn't _require_ an explicit count, and `#[count(..)]` will still work. I think this is better since computing the length from an explicit count requires an additional redundant iteration, so it seems silly to require that unnecessarily. Please check that this works your avar2 needs?


simoncozens:
I've rebased the avar2 branch on top of this, and t he tests pass. Thank you very much!

simoncozens (approved):

PR #1152, opened 2024-09-12
Improve head.macStyle docs

simoncozens:
It's true that the OpenType spec doesn't contain a description for this field, but "See somewhere else" was an odd thing to see pop up in my IDE.


rsheeter (approved):
Ha, good catch

PR #1154, opened 2024-09-17
[read-fonts] layout: selecting scripts for shaping

dfrg:
Adds HB equivalent functionality for selecting an OpenType script for a given Unicode script code.


read-fonts/src/tables/layout/script.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// Returned by the [`ScriptList::select`] method.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub struct SelectedScript {
+    /// The actual OpenType tag of the chosen script.
+    pub tag: Tag,
+    /// Index of the script in the [`ScriptList`].
+    pub index: u16,
+    /// True if a script was chosen that wasn't in the requested list.
+    pub is_fallback: bool,
```
rsheeter, line 16: Speculation: I wonder if a Rusty SelectedScript should be an enum instead?

read-fonts/src/tables/layout/script.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+}
+
+/// A prioritized list of OpenType script tags mapped from a Unicode script
+/// tag.
+///
+/// Generated by [`script_tags_from_unicode`].
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
+pub struct ScriptTags {
+    tags: [Tag; 3],
```
rsheeter, line 76: this merits a comment explaining what it is

read-fonts/src/tables/layout/script.rs
```diff
@@ -0,0 +1,269 @@
...
+            // Else, just change the first char to lowercase
+            bytes[0] = bytes[0].to_ascii_lowercase();
+            &bytes
+        }
+    };
+    Tag::new(tag_bytes)
+}
+
+// See <https://github.com/harfbuzz/harfbuzz/blob/63d09dbefcf7ad9f794ca96445d37b6d8c3c9124/src/hb-ot-tag.cc#L84>
+fn new_tag_from_unicode(unicode_script: Tag) -> Option<Tag> {
```
cmyr, line 142: I've definitely impl'd this somewhere before... looks like it's in `fontbe`? Let's open an issue to deduplicate.

read-fonts/src/tables/layout/script.rs
```diff
@@ -0,0 +1,269 @@
...
+/// tag.
+///
+/// Generated by [`script_tags_from_unicode`].
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
+pub struct ScriptTags {
+    tags: [Tag; 3],
+    len: usize,
+}
+
+impl Deref for ScriptTags {
```
rsheeter, line 80: Optional, would an explicit call like `as_tags()` be overly horrid? - not a huge fan of Deref.

rsheeter (approved):
LGTM subject to the typo rules being updated to fix the warnings

read-fonts/src/tables/layout/script.rs
cmyr: `as_slice()` would be the idiomatic conversion fn here, but since this type is essentially an array deref to slice is totally reasonable (having both would be the most idiomatic approach)

cmyr (approved):
looks good; noted inline that some of this is duplicating code that exists elsewhere, so let's open an issue to resolve that at some point.

read-fonts/src/tables/layout/script.rs
cmyr: it's in a different repo, at the very least we need to do a release of read-fonts before we can change anything there.

rsheeter: Oh sorry, I skimmed pastit being in fontc not here. Then +1 to issue or fast-follow PR to update fontc.

read-fonts/src/tables/layout/script.rs
dfrg: That's an equivalent representation but an enum with all variants carrying the same payload where the variant itself is secondary information is often awkward to use. In other words, we'll always want the tag/index but might occasionally care if we're using a fallback so it doesn't make sense to primarily categorize on that aspect.

read-fonts/src/tables/layout/script.rs
dfrg: Agreed with Colin on both points. Added `as_slice()` for those that prefer explicit calls.

read-fonts/src/tables/layout/script.rs
dfrg: @cmyr I exposed a `UNICODE_TO_NEW_OPENTYPE_SCRIPT_TAGS` const that is basically the same as `NEW_SCRIPT_TAGS` in fontbe except the key is `&[u8; 4]` instead of `&str`. Does that work for you?

cmyr: i'm not going to have time to dig into this before I take off, so let's merge as-is and we can figure that out when I get back.

PR #1156, opened 2024-09-18
[klippa] subset gvar table

qxliu76:


klippa/src/gvar.rs
```diff
@@ -0,0 +1,175 @@
...
+                    && !plan
+                        .subset_flags
+                        .contains(SubsetFlags::SUBSET_FLAGS_NOTDEF_OUTLINE)
+                {
+                    return None;
+                }
+                match self.data_for_gid(x.0) {
+                    Err(_) => None,
+                    Ok(font_data) => Some(font_data.len() as u32),
+                }
```
cmyr, line 43: fyi you can write this as,

```rust
self.data_for_gid(x.0).ok().map(|data| data.len() as u32)
```

cmyr (approved):
looks reasonable, one little observation inline but nothing blocking!

klippa/src/gvar.rs
qxliu76: Thanks!

klippa/src/gvar.rs
```diff
@@ -0,0 +1,172 @@
...
+            1_u16
+        } else {
+            0_u16
+        };
+        // flags
+        out.extend_from_slice(&long_offset.to_be_bytes());
+
+        // calculate sharedTuplesOffset
+        // shared tuples array follow the GlyphVariationData offsets array at the end of the 'gvar' header.
+        let glyph_var_data_offset_size = if long_offset > 0 { 4 } else { 2 };
```
garretrieger, line 56: In the if long_offset == 1 { .. } else { .. } blocks here and below the code in both branches looks to be nearly identical, differing only by the use of a 2 or 4 width, a divide by 2 for the offset values, and the use of u32/u16 for glyph offsets.

It should be possible to refactor this code to avoid the duplicated code blocks, here's what I'm thinking:
- move the code from here down into a helper function.
- Make it generic and take either u16 or u32 as a type parameter and then write the code in terms of the generic type. This should allow you to avoid switching on long_offset throughout.

garretrieger:
Looks like you're now getting the point of needing to manually assemble table binaries including offsets. That's fine for this PR but before you get further into the more complex tables like GSUB/GPOS we probably want to create something like hb_serialize_context_t from harfbuzz to assist in manually assembling tables like this.

klippa/src/gvar.rs
qxliu76: I've tried my best to generalize them. The checks on long_offset cannot be avoided, put it outside of each for loop is better from a performance perspective

PR #1158, opened 2024-09-20
[skrifa] autohint shaper abstraction

dfrg:
Adds the infrastructure to support shaping without making any functional changes.

A little bit of churn but the intention is to allow future PRs to be more focused.

### DIFF
--- a/skrifa/src/outline/autohint/shape.rs
+++ b/skrifa/src/outline/autohint/shape.rs
@@ -1,0 +1,87 @@
+//! Shaping support for autohinting.
+
+use raw::{tables::gsub::Gsub, TableProvider};
+
+use crate::{charmap::Charmap, collections::SmallVec, FontRef, GlyphId, MetadataProvider};
+
+/// Determines the fidelity with which we apply shaping in the
+/// autohinter.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub(crate) enum ShaperMode {
+    /// Characters are mapped to nominal glyph identifiers and layout tables
+    /// are not used for style coverage.
+    ///
+    /// This matches FreeType when HarfBuzz support is not enabled.
+    Nominal,
+    /// Simple substitutions are applied according to script rules and layout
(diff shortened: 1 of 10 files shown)


drott (approved):
LGTM - do we later still want Nominal? Is this transitional?

dfrg:
Nominal allows us to keep testing against a known FreeType configuration which will at least ensure that the internal algorithms are a match. It's unlikely that the "best effort" mode will ever truly match FT+HB so I think we should keep nominal around until some future point that we have access to a full shaper.

skrifa/src/outline/autohint/shape.rs
```diff
@@ -0,0 +1,87 @@
...
+pub(crate) struct ShapedGlyph {
+    pub id: GlyphId,
+    pub y_offset: i32,
+}
+
+/// Arbitrarily chosen to cover our max input size plus some extra to account
+/// for expansion from multiple substitution tables.
+const SHAPED_CLUSTER_INLINE_SIZE: usize = 16;
+
+pub(crate) type ShapedCluster = SmallVec<ShapedGlyph, SHAPED_CLUSTER_INLINE_SIZE>;
```
rsheeter, line 32: cluster is not so widely known outside of a very small circle, suggest adding a reference that explains what clusters are

skrifa/src/outline/autohint/shape.rs
```diff
@@ -0,0 +1,87 @@
+//! Shaping support for autohinting.
+
+use raw::{tables::gsub::Gsub, TableProvider};
+
+use crate::{charmap::Charmap, collections::SmallVec, FontRef, GlyphId, MetadataProvider};
+
+/// Determines the fidelity with which we apply shaping in the
+/// autohinter.
```
rsheeter, line 8: I think I'd expect an enum that determines fidelity to give clearer guidance about the tradeoffs and/or which is better, can we beef up the comment a touch?

rsheeter (approved):

PR #1161, opened 2024-09-30
[skrifa] autohint: capture styles from GSUB

dfrg:
This adds "best effort" support for using `GSUB` to assign styles to glyphs that are not present in `cmap`.

Per discussion, the new shaping behavior is feature gated (and enabled by default) to support testing against FreeType (-HarfBuzz) with fauntlet which is a known good configuration.

This should get us FreeType equivalent autohinting for basic ligatures. Further work is needed to compute alignment zones for various features (`c2sc`, `titl`, etc) and for complex scripts that might require some substitutions.


skrifa/src/outline/autohint/style.rs
```diff
@@ -58,6 +61,31 @@ impl GlyphStyle {
             self.0 = (self.0 & !Self::STYLE_INDEX_MASK) | other.0;
         }
     }
+
+    pub(super) fn set_gsub_output(&mut self) {
```
drott, line 65: Naming nits: These names do not fit together so well IMO, `clear_shaping` vs. `set_gsub_output`. What does "gsub output" really mean here? Does it mean the glyph was found as corresponding to some group/class through gsub? Then maybe "set_from_gsub_substitution" + "clear_from_gsub"? But not sure if I got the semantics right.

skrifa/src/outline/autohint/style.rs
```diff
@@ -198,6 +237,16 @@ impl GlyphStyleMap {
                 }
             })
     }
+
+    fn use_style(&mut self, style_ix: usize) {
+        let mapped = &mut self.metrics_map[style_ix];
+        if *mapped == 0xFF {
```
drott, line 243: This is a bit cryptic with magic values...

skrifa/src/outline/autohint/style.rs
```diff
@@ -434,13 +482,46 @@ mod tests {
             (96, Some(("Adlam", true))),
             (97, Some(("Adlam", false))),
         ];
+        check_styles(font_test_data::AUTOHINT_CMAP, ShaperMode::Nominal, expected);
+    }
+
+    #[test]
+    fn shaped_glyph_styles() {
+        // generated by printf debugging in FreeType
+        // (gid, Option<(script_name, is_non_base)>)
```
drott, line 491: What does `is_none_base` mean? That it's not from a basic char to glyph mapping?

drott (approved):
I think the integration with the existing code looks smooth thanks to the prep PRs for that. LGTM with some comments on readability perhaps.

skrifa/src/outline/autohint/style.rs
dfrg: It means that the glyph is specifically the _output_ of a GSUB substitution. I kept the names distinct because we may want to handle GPOS as well but I agree that the disconnect is jarring with having the additional GPOS code in there to tie it together. I'll rename to `set_from_gsub_output` and `clear_from_gsub`.

skrifa/src/outline/autohint/style.rs
dfrg: Oops, I meant to name this constant when I made the change.

skrifa/src/outline/autohint/style.rs
drott: sgtm.

skrifa/src/outline/autohint/style.rs
drott: Got it thanks, not base as in non base character... ok, perhaps at a comment?

dfrg: Done... hopefully will clear up confusion for future readers :)

PR #1164, opened 2024-10-01
[read-fonts] add hdmx table

dfrg:
Also includes a small codegen change that uses `cursor.read()` instead of `cursor.read_be()` for `u8` fields in records.


cmyr (approved):
Looks good!

PR #1165, opened 2024-10-01
[skrifa] use hdmx table if available

dfrg:
In `ScaledOutline::adjusted_advance_width()` prefer widths from `hdmx` over phantom points to better match FreeType.


drott (approved):
LGTM, this would benefit from a test, but in the interest of getting this into a release, I am okay with doing this in a follow-up. This will have some coverage in Chromium's web tests.

dfrg:
I'll add tests for adjusted advance width in fauntlet after I address the problem with empty glyphs. I can also add a unit test when I find a font with an hdmx table that has known provenance.

PR #1171, opened 2024-10-03
[skrifa] path conversion perf improvements

dfrg:
Reworks the TrueType->path conversion code to capture states explicitly to reduce branches along with some inlining and other code motion.

This gives a roughly 10% improvement based on measured fonts.

Example for Noto Sans Regular (all glyphs 100x):
| FreeType | skrifa (current) | skrifa (this PR) | skrifa (this PR + LTO) |
|-----------|-------------|--------------|--------------------|
| 308ms     | 337ms      |  295ms        | 213ms                  |

The significant difference when enabling LTO suggests that we might be missing some inlining opportunities across the `read-fonts`/`font-types` boundaries. Note that current skrifa + LTO is roughly the same as this PR without it.

### DIFF
--- a/fauntlet/src/font/skrifa.rs
+++ b/fauntlet/src/font/skrifa.rs
@@ -60,20 +60,21 @@ impl<'a> SkrifaInstance<'a> {
             .advance_width(glyph_id)
     }
 
+    /// Returns the scaler adjusted advance width when available.
     pub fn outline(
         &mut self,
         glyph_id: GlyphId,
         pen: &mut impl OutlinePen,
-    ) -> Result<(), DrawError> {
+    ) -> Result<Option<f32>, DrawError> {
         let outline = self
             .outlines
             .get(glyph_id)
             .ok_or(DrawError::GlyphNotFound(glyph_id))?;
-        if let Some(hinter) = self.hinter.as_ref() {
-            outline.draw(DrawSettings::hinted(hinter, false), pen)?;
+        let metrics = if let Some(hinter) = self.hinter.as_ref() {
+            outline.draw(DrawSettings::hinted(hinter, false), pen)?
         } else {
-            outline.draw((self.size, self.coords.as_slice()), pen)?;
-        }
-        Ok(())
+            outline.draw((self.size, self.coords.as_slice()), pen)?
+        };
+        Ok(metrics.advance_width)
     }
 }
--- a/skrifa/src/outline/autohint/style.rs
+++ b/skrifa/src/outline/autohint/style.rs
@@ -197,6 +197,17 @@ impl GlyphStyleMap {
         if need_hani {
             map.use_style(StyleClass::HANI);
         }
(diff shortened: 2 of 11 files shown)


dfrg:
cc @drott

drott (approved):
Faster and IMO better to read, great improvement, LGTM!

drott:
I think the target branch is set wrong on this PR, shouldn't it be main?

dfrg:
> I think the target branch is set wrong on this PR, shouldn't it be main?

I chained this locally because I wanted to capture the advance with changes in my performance measurements. Should be based on main now that the other PR is merged.

fauntlet/src/font/skrifa.rs
```diff
@@ -60,20 +60,21 @@ impl<'a> SkrifaInstance<'a> {
...
-        if let Some(hinter) = self.hinter.as_ref() {
-            outline.draw(DrawSettings::hinted(hinter, false), pen)?;
+        let metrics = if let Some(hinter) = self.hinter.as_ref() {
+            outline.draw(DrawSettings::hinted(hinter, false), pen)?
         } else {
-            outline.draw((self.size, self.coords.as_slice()), pen)?;
-        }
-        Ok(())
+            outline.draw((self.size, self.coords.as_slice()), pen)?
+        };
```
rsheeter, line 77: Optional, I feel like this would be easier to read if

1. The if statement produced DrawSettings
1. The unhinted was explicit
1. The result of draw was used to form the result

```rust
        let draw_settings = if let Some(hinter) = self.hinter.as_ref() {
            DrawSettings::hinted(hinter, false)
        } else {
            DrawSettings::unhinted(self.size, self.coords.as_slice())
        };
        outline.draw(draw_settings, pen).map(|metrics| metrics.advance_width)
```

skrifa/src/outline/autohint/style.rs
```diff
@@ -197,6 +197,17 @@ impl GlyphStyleMap {
...
+        // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/57617782464411201ce7bbc93b086c1b4d7d84a5/src/autofit/afglobal.c#L251>
+        for digit_char in '0'..='9' {
+            if let Some(style) = shaper
+                .charmap()
+                .map(digit_char)
+                .and_then(|gid| map.styles.get_mut(gid.to_u32() as usize))
+            {
+                style.0 |= GlyphStyle::DIGIT;
+            }
+        }
```
rsheeter, line 210: style assignment seems very amenable to unit testing

fauntlet/src/font/skrifa.rs
dfrg: Definitely looks better. #1177

skrifa/src/outline/autohint/style.rs
dfrg: Indeed.. #1178

PR #1176, opened 2024-10-03
Add shared-brotli crate which provides a library for decoding shared …

garretrieger:
…brotli patches.

Under the hood it's a small wrapper around brotlic-sys.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
     "fauntlet",
     "klippa",
     "fuzz",
+    "shared-brotli",
 ]
 
 [workspace.package]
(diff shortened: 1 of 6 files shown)


cmyr (approved):
This looks broadly reasonable. If you anticipate needing to add more functionality in the future it might make sense to take a slightly more modular approach (like have a `Decoder` struct that wraps the decoder pointer) but if we really just need this one bit of functionality than this approach seems fine.

Since we'll have to publish this crate, I would be inclined to give it a more annoying & descriptive name, and we'll need to add a readme etc before publishing.

garretrieger:
> This looks broadly reasonable. If you anticipate needing to add more functionality in the future it might make sense to take a slightly more modular approach (like have a `Decoder` struct that wraps the decoder pointer) but if we really just need this one bit of functionality than this approach seems fine.

Unless things change in the spec this should be all that's needed. The other patch type will also need Brotli decompression but it should be able to re-use this method as well.

> Since we'll have to publish this crate, I would be inclined to give it a more annoying & descriptive name, and we'll need to add a readme etc before publishing.

Sounds good.

cmyr:
> > Since we'll have to publish this crate, I would be inclined to give it a more annoying & descriptive name, and we'll need to add a readme etc before publishing.
> 
> Sounds good.

To clarify my 'annoying name' point; I mean this is very unlikely to be useful to anyone but us, and so I'd name it something like `brotli-shared-data-decode` and make clear in the readme that it has a very limited scope that is not expected to expand. Just my two cents though!

PR #1181, opened 2024-10-08
[int-set] Add iter_excluded_ranges()

garretrieger:
This iterates over all ranges of values not contained in the set.


fuzz/fuzz_targets/int_set_op_processor.rs
```diff
@@ -721,6 +721,64 @@ where
...
+        let mut btree_ranges: Vec<RangeInclusive<T>> = vec![];
+        let mut cur_range: Option<RangeInclusive<T>> = None;
+
+        let mut inverted = BTreeSet::<T>::new();
+        for v in T::ordered_values() {
+            let v = T::create(v).unwrap();
+            if !input.btree_set.contains(&v) {
+                inverted.insert(v);
+            }
+        }
```
cmyr, line 751: just  golf but I'd write this whole thing as,

```rust
let inverted: BTreeSet<_> = T::order_values()
    .map(|v| T::create(v).unwrap())
    .filter(|val| !input.btree_set.contains(&v))
    .collect();

cmyr (approved):
one little style observation that you're welcome to ignore.

fuzz/fuzz_targets/int_set_op_processor.rs
garretrieger: looks good, thanks

PR #1185, opened 2024-10-10
[skrifa] cff: allow empty private dict and fix scaling

### LINKED ISSUE #1184: Path extraction fails for CFF contour for various glyphs of Material Icons
With https://fonts.gstatic.com/s/googlematerialicons/v142/Gw6kwdfw6UnXLJCcmafZyFRXb3BL9rvi0QZG3Q.woff2 Material Icons CFF font, Skia path extraction fails, and cbrokaw@ reports, when reading the CFF path an error `PostScript(MissingPrivateDict)` occurs.

Example glyph ids: 
```
~/dev/harfbuzz/build/util/hb-shape --no-glyph-names  Gw6kwdfw6UnXLJCcmafZyFRXb3BL9rvi0QZG3Q.ttf "mood present_to_all"
[1829=0+512|1=4+512|2261=5+512]
```

dfrg:
We were erroneously rejecting a Private DICT with length 0. Also, the value `Fixed::ONE` was used as a sentinel for unscaled drawing but this value can be produced when upem is 512 and requested ppem is 8, resulting in incorrect scaling. Replaced the scale factor with an option to correct this.

Fixes #1184


font-test-data/test_data/ttf/material_icons_subset.ttf
drott, line 1: Did the subsetting preserve the empty private dict?

font-test-data/test_data/ttx/material_icons_subset.ttx
```diff
@@ -0,0 +1,256 @@
...
+      <UnderlinePosition value="-100"/>
+      <UnderlineThickness value="50"/>
+      <PaintType value="0"/>
+      <CharstringType value="2"/>
+      <FontMatrix value="0.001953125 0 0 0.001953125 0 0"/>
+      <FontBBox value="-1 0 513 512"/>
+      <StrokeWidth value="0"/>
+      <!-- charset is dumped separately as the 'GlyphOrder' element -->
+      <Encoding name="StandardEncoding"/>
+      <Private>
```
drott, line 171: Does this compile to an empty one?

drott (approved):
LGTM and big thanks for the quick fix!

font-test-data/test_data/ttx/material_icons_subset.ttx
dfrg: Yes, this confused me as well but those are all default values so the compiler just omits the dict entirely. There's a test to check for this.

skrifa/src/outline/cff/mod.rs
```diff
@@ -707,6 +709,36 @@ mod tests {
         assert!(svg.to_string().ends_with('Z'));
     }
 
+    /// Ensure we don't reject an empty Private DICT
+    #[test]
+    fn empty_private_dict() {
+        let font = FontRef::new(font_test_data::MATERIAL_ICONS_SUBSET).unwrap();
+        let common = OutlinesCommon::new(&font).unwrap();
+        let outlines = super::Outlines::new(&common).unwrap();
```
drott, line 717: So before, outlines came back as Err()? Do we want to access any contours here as part of the test?

skrifa/src/outline/cff/mod.rs
```diff
@@ -707,6 +709,36 @@ mod tests {
...
+        assert!(outlines.top_dict.private_dict_range.is_empty());
```
drott, line 718: This assert answers my earlier question.

font-test-data/test_data/ttx/material_icons_subset.ttx
drott: Ack, thanks!

skrifa/src/outline/cff/mod.rs
dfrg: Constructing Outlines was fine, but making a Subfont would return an error so we couldn't generate an outline. This is checked, although indirectly, in the following test.

dfrg: Although the call to `Outlines::private_dict_range()` would fail in this case, which is checked here.

PR #1187, opened 2024-10-10
[IFT] Add code gen definitions for GlyphKeyed patches

garretrieger:
Second patch format for IFT, see: https://w3c.github.io/IFT/Overview.html#glyph-keyed

### DIFF
--- a/read-fonts/src/tables/ift.rs
+++ b/read-fonts/src/tables/ift.rs
@@ -77,6 +77,41 @@ impl U8Or16 {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct U16Or24(u32);
+
+impl ReadArgs for U16Or24 {
+    type Args = GlyphKeyedFlags;
+}
+
+impl ComputeSize for U16Or24 {
+    fn compute_size(flags: &GlyphKeyedFlags) -> Result<usize, ReadError> {
+        // See: https://w3c.github.io/IFT/Overview.html#glyph-keyed-patch-flags
+        Ok(if flags.contains(GlyphKeyedFlags::WIDE_GLYPH_IDS) {
+            3
+        } else {
+            2
+        })
+    }
+}
+
+impl FontReadWithArgs<'_> for U16Or24 {
+    fn read_with_args(data: FontData<'_>, flags: &Self::Args) -> Result<Self, ReadError> {
+        if flags.contains(GlyphKeyedFlags::WIDE_GLYPH_IDS) {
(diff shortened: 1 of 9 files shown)


read-fonts/src/tables/ift.rs
```diff
@@ -193,6 +228,92 @@ impl<'a> Iterator for GidToEntryIter<'a> {
...
+impl<'a> GlyphPatches<'a> {
+    /// Returns an iterator over the per glyph data for the table with the given index.
+    pub fn glyph_data_for_table(
+        &'a self,
+        table_index: u32,
+    ) -> impl Iterator<Item = Result<(GlyphId, &'a [u8]), ReadError>> {
+        let glyph_count = self.glyph_count() as usize;
+        let start_index = table_index as usize * glyph_count;
+        let start_it = self.glyph_data_offsets().iter().skip(start_index);
+        let end_it = self.glyph_data_offsets().iter().skip(start_index + 1);
```
cmyr, line 240: I wonder if `skip` is smart enough in this instance to avoid actually doing a linear scan, and just compiles down to `self.glyph_data_offsets()[start_index..].iter()`?

resources/codegen_inputs/ift.rs
```diff
@@ -250,4 +251,47 @@ table TablePatch {
...
+/// [GlyphPatches](https://w3c.github.io/IFT/Overview.html#glyphpatches)
+#[read_args(flags: GlyphKeyedFlags)]
+table GlyphPatches {
+  glyph_count: u32,
+  table_count: u8,
+
+  #[count($glyph_count)]
+  #[read_with($flags)]
+  #[traverse_with(skip)]
+  #[compile(skip)] // TODO remove this once write fonts side is implemented.
```
cmyr, line 284: are we trying to compile this table? If not maybe it makes more sense to skip the whole table, instead of generating write-fonts code that is going to be incorrect. You can add `#[skip_font_write]` to skip just this table, although it might also break other tables that reference it..

cmyr (approved):
one concern about a broken `FontWrite` impl, but otherwise looks good, nice to see you seemingly already quite comfortable with the dustier corners of codegen.

resources/codegen_inputs/ift.rs
garretrieger: In this first pass I'm not going to implement compilation, but I will be adding eventually. So far I've just been using #[compile(skip)] on anything that causes any issues with the plan being to go back and fix it all up once I get to implementing compilation.

PR #1188, opened 2024-10-11
[IFT] Add beginning implementation for Glyph Keyed patches

garretrieger:
https://w3c.github.io/IFT/Overview.html#glyph-keyed

Currently only supports glyf/loca modification and is missing functionality to mark applied patches in the IFT tables.

### DIFF
--- a/font-test-data/src/ift.rs
+++ b/font-test-data/src/ift.rs
@@ -560,7 +560,7 @@ pub fn test_font_for_patching() -> Vec<u8> {
     font_builder.add_table(&head).unwrap();
 
     // ## glyf ##
-    // glyphs are padded to even number of bytes since loca format wil be short.
+    // glyphs are padded to even number of bytes since loca format will be short.
     let glyf = BeBuffer::new()
         .push_with_tag(1u8, "gid_0")
         .extend([2, 3, 4, 5u8, 0u8])
--- a/font-test-data/src/ift.rs
+++ b/font-test-data/src/ift.rs
@@ -666,6 +666,48 @@ pub fn glyf_u16_glyph_patches() -> BeBuffer {
 }
 
 // Format specification: https://w3c.github.io/IFT/Overview.html#glyphpatches
+pub fn glyf_u16_glyph_patches_2() -> BeBuffer {
+    let mut buffer = be_buffer! {
+      3u32,       // glyph count
+      {1u8: "table_count"},        // table count
+
+      // glyph ids * 3
+      7u16,
+      12u16,
+      14u16,
+
+      (Tag::new(b"glyf")),   // tables * 1
+
+      // glyph data offsets * 6
+      {0u32: "gid_7_offset"},
+      {0u32: "gid_12_offset"},
+      {0u32: "gid_14_offset"},
+      {0u32: "end_offset"},
+
+      // data blocks
+      {b'q': "gid_7_data"},
+      [b'r'],
+
+      {b's': "gid_12_data"},
+      [b't', b'u'],
+
+      {b'v': "gid_14_data"}
+    };
+
+    let offset = buffer.offset_for("gid_7_data") as u32;
+    buffer.write_at("gid_7_offset", offset);
+
+    let offset = buffer.offset_for("gid_12_data") as u32;
+    buffer.write_at("gid_12_offset", offset);
+
+    let offset = buffer.offset_for("gid_14_data") as u32;
+    buffer.write_at("gid_14_offset", offset);
+    buffer.write_at("end_offset", offset + 1);
+
+    buffer
+}
+
+// Format specification: https://w3c.github.io/IFT/Overview.html#glyphpatches
 pub fn glyf_u24_glyph_patches() -> BeBuffer {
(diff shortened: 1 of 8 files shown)


font-test-data/src/ift.rs
```diff
@@ -5,6 +5,10 @@
 
 use read_fonts::types::Tag;
 use read_fonts::{be_buffer, be_buffer_add, test_helpers::BeBuffer, types::Int24, types::Uint24};
+use write_fonts::tables::head::Head;
+use write_fonts::tables::loca::Loca;
+use write_fonts::tables::maxp::Maxp;
+use write_fonts::FontBuilder;
```
cmyr, line 11: these scattered imports trigger me so bad 😅

incremental-font-transfer/src/font_patch.rs
```diff
@@ -72,8 +78,26 @@ pub enum PatchingError {
...
     FontParsingFailed(ReadError),
     IncompatiblePatch,
+    IncompatiblePatches,
     NonIncrementalFont,
     InvalidPatch(&'static str),
+    InternalError,
+}
+
+impl PatchingError {
+    pub(crate) fn from(decoding_error: DecodeError) -> Self {
```
cmyr, line 88: this looks like it should be an impl of `From<DecodeError> for PatchingError`.

incremental-font-transfer/src/font_patch.rs
```diff
@@ -127,14 +156,41 @@ impl IncrementalFontPatchBase for FontRef<'_> {
             return Err(PatchingError::NonIncrementalFont);
         }
 
-        patch.apply_to(self)
+        let first = patches.first().ok_or(PatchingError::InvalidPatch(
```
cmyr, line 159: I'd personally probably just make this a variant, `PatchingError::EmptyPatch` or something.

incremental-font-transfer/src/font_patch.rs
```diff
@@ -12,20 +12,25 @@
...
+
 use write_fonts::FontBuilder;
 
 /// An incremental font patch which can be used to extend a font.
 ///
 /// See: <https://w3c.github.io/IFT/Overview.html#font-patch-formats>
+#[derive(Clone)]
 pub enum IncrementalFontPatch<'a> {
     TableKeyed(TableKeyedPatch<'a>),
+    GlyphKeyed(GlyphKeyedPatch<'a>),
```
cmyr, line 33: looking at the code below, instead of using `&[IncrementalFontPatch]` (with the condition that there can be only one table-keyed patch but multiple glyph-keyed patches) do you think it would make more sense to try and represent that in the type system here, somehow?

What I mean is ending up with something like,

```rust
pub enum IncrementalFontPatch<'a, 'b> {
     TableKeyed(TableKeyedPatch<'a>),
     GlyphKeyed(&'b [GlyphKeyedPatch<'a>]),
}
```

which is annoying for the extra lifetime but might otherwise make life easier? But maybe not, it's not totally clear to me at what stage you would have the slice.

incremental-font-transfer/src/font_patch.rs
```diff
@@ -127,14 +156,41 @@ impl IncrementalFontPatchBase for FontRef<'_> {
...
+                        IncrementalFontPatch::GlyphKeyed(patch_data) => Some(patch_data.clone()),
+                        _ => None,
+                    })
+                    .collect();
+                // glyph keyed patches are no invalidation so any number of them may be applied simultaneously
+                if glyph_keyed_patches.len() != patches.len() {
+                    return Err(PatchingError::IncompatiblePatches);
+                }
+                apply_glyph_keyed_patch(&glyph_keyed_patches, self)
+            }
```
cmyr, line 185: see above: looking at this big whack of validation code I wonder if we're missing a chance to represent this in the type system.

cmyr (approved):
some notes inline to be addressed at your discretion, overall looks good!

incremental-font-transfer/src/font_patch.rs
garretrieger: Interesting idea, but IncrementalFontPatch's are derived from a single PatchUri + binary data via a .into_patch() method on the URI. So we'd need to modify the API for deriving the patches from uri's and I think it would end up kind of clunky, and ultimately I think this just shifts the compatibility check up one level. Either way the caller of the API is ultimately responsible for not feeding in an incompatible set of patches.

garretrieger: I think maybe a better approach is to have two distinct apply_patch methods that take either a single TableKeyed or multiple GlyphKeyed and leave it up to the caller to extract the appropriate types from the generic IncrementalFontPatch list (like I mentioned it's already something a caller will need to do before calling this API).

garretrieger: I'll give that a try and see what it ends up looking like.

garretrieger: Looking into this more I think we probably want more broad changes to the entire set of IFT APIs, so I'm going to leave this as is for now but revisit it later with a more intentional design for the full API surface.

PR #1191, opened 2024-10-18
[write] Add common util for computing searchRange

cmyr:
This is shared in a few places, so it makes sense to have a single implementation. 

Ran into this in the context of the `kern` table, which I'll PR shortly; this code is shared in a few places.

(CI is failing for other reasons, so let's hold off on this until that gets cleared up)

### DIFF
--- a/write-fonts/src/util.rs
+++ b/write-fonts/src/util.rs
@@ -100,3 +100,59 @@ impl Default for FloatComparator {
 pub fn isclose(a: f64, b: f64) -> bool {
     FloatComparator::default().isclose(a, b)
 }
+
+/// Search range values used in various tables
+#[derive(Clone, Copy, Debug)]
+pub struct SearchRange {
+    pub search_range: u16,
+    pub entry_selector: u16,
+    pub range_shift: u16,
+}
+
+impl SearchRange {
+    //https://github.com/fonttools/fonttools/blob/729b3d2960ef/Lib/fontTools/ttLib/ttFont.py#L1147
+    /// calculate searchRange, entrySelector, and rangeShift
+    ///
+    /// these values are used in various places, such as [the base table directory]
+    /// and [cmap format 4].
+    ///
+    /// [the base table directory]: https://learn.microsoft.com/en-us/typography/opentype/spec/otff#table-directory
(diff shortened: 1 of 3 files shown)


write-fonts/src/util.rs
```diff
@@ -100,3 +100,29 @@ impl Default for FloatComparator {
...
+    pub search_range: u16,
+    pub entry_selector: u16,
+    pub range_shift: u16,
+}
+
+impl SearchRange {
+    //https://github.com/fonttools/fonttools/blob/729b3d2960ef/Lib/fontTools/ttLib/ttFont.py#L1147
+    /// calculate searchRange, entrySelector, and rangeShift
+    ///
+    /// these values are used in various tables.
```
rsheeter, line 116: I suggest enumerating them non-exclusively, e.g. 

````md
These values are used in at least the following tables:

* [OpenType font table directory](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#table-directory)
* [cmap subtable format 4](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#format-4-segment-mapping-to-delta-values)
...others...
```

write-fonts/src/util.rs
```diff
@@ -100,3 +100,29 @@ impl Default for FloatComparator {
...
+        let search_range = (2.0_f64.powi(entry_selector as i32) * item_size as f64) as usize;
+        // The result doesn't really make sense with 0 tables but ... let's at least not fail
+        let range_shift = (n_items * item_size).saturating_sub(search_range);
+        SearchRange {
+            search_range: search_range.try_into().unwrap(),
+            entry_selector: entry_selector.try_into().unwrap(),
+            range_shift: range_shift.try_into().unwrap(),
+        }
+    }
+}
```
rsheeter, line 128: Add a test, perhaps based on the sample at https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#format-4-segment-mapping-to-delta-values?

rsheeter: ...and based on you comment, one for 0?

rsheeter (approved):

PR #1192, opened 2024-10-19
Add support for the 'kern' table

### LINKED ISSUE #1183: Legacy kern table
fontations doesn't currently read or write the [`kern`](https://learn.microsoft.com/en-us/typography/opentype/spec/kern#format-0) table. However, support for this table is needed for kerning in Microsoft Powerpoint. It is also the subject of a fontbakery check.

cmyr:
This supports only the kern0 subtable type, and we should probably add a nice builder to make this a bit easier.

this is based off of #1191, which should go in first.

- fixes #1183


resources/codegen_inputs/kern.rs
```diff
@@ -0,0 +1,94 @@
...
+    num_tables: u16,
+    #[count($num_tables)]
+    #[traverse_with(skip)]
+    subtables: VarLenArray<Kern0>,
+}
+
+///// The different kern subtable formats.
+//format u16 KernSubtable {
+    //Format0(Kern0),
+    //// Nope.
```
rsheeter, line 20: lol ... explain why? Even just note we're starting simple?

rsheeter (approved):

resources/codegen_inputs/kern.rs
rsheeter: Just noticed https://github.com/googlefonts/fontations/issues/1183#issuecomment-2421870991 has some good reasons, "Windows only supports format 0, and FontTools does not know of it either. FontTools also mentions a "new" Apple-specific format  kern table (not subtable) with version 1.0 and completely different structures." ... should we mention 1 as well?

cmyr:
superseded by #1519

PR #1201, opened 2024-10-23
[skrifa] replace todo! with None for color bbox

dfrg:
Fixes abort caused by `todo!()` and caught by fuzzing.

ref
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=71277 and https://oss-fuzz.com/testcase?key=5794538109599744

JMM if happy

### DIFF
--- a/skrifa/src/color/mod.rs
+++ b/skrifa/src/color/mod.rs
@@ -294,27 +294,34 @@ impl<'a> ColorGlyph<'a> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


rsheeter (approved):
LGTM, only nit is that None for COLRv0 seems like an obvious unit test :)

dfrg:
Perhaps I thought it was too obvious :)

I'll go ahead and add before merging.

PR #1206, opened 2024-10-28
[Skrifa] Fix panic in degenerate linear gradient drawing

### LINKED ISSUE #1205: Panic in COLRv1 with degenerate linear gradient
Spun off from
https://issues.oss-fuzz.com/375768991

If a degenerate linear color gradient is attempted to be painted with an empty color line, this leads to a panic when accessing the first color line element.

drott:
Avoid a panic when a degenerate linear gradient occurs and an empty color line is passed.

Fixes #1205.


dfrg (approved):
LGTM but we should file an issue for reducing the test case. We had a problem recently where fuzzer generated fonts happened to depend on implementation defined behavior in `std` and our tests broke when that behavior changed.

drott:
Filed as #1207

PR #1208, opened 2024-10-29
[read-fonts] helper methods for layout tables

dfrg:
Adds some convenience methods to various layout types to enable simpler access to scripts, language systems and features.

There's a new `TaggedElement<T>` type that is basically `(Tag, T)` which is used to represent an array element where we have a sequence of records with a tag and an offset.

Designed mostly to match the needs of HarfRuzz but also useful for an upcoming autohinting patch and may be helpful for other projects.


read-fonts/src/tables/layout/script.rs
```diff
@@ -67,6 +80,51 @@ impl<'a> ScriptList<'a> {
...
+        let records = list.feature_records();
+        for feature_ix in self.feature_indices() {
+            let feature_ix = feature_ix.get();
+            if let Some(rec) = records.get(feature_ix as usize) {
+                if rec.feature_tag() == tag {
+                    return Some(feature_ix);
+                }
+            }
+        }
+        None
```
cmyr, line 124: could also just be,

```rust
self.feature_indices().map(|ix| ix.get())
    .find(|feature_ix| records.get(feature_ix as usize).map(|rec| rec.feature_tag()) == Some(tag))
```

but I'm just having fun

read-fonts/src/table_ref.rs
```diff
@@ -31,3 +33,28 @@ impl<'a, T> TableRef<'a, T> {
 impl<U, T: Format<U>> Format<U> for TableRef<'_, T> {
     const FORMAT: U = T::FORMAT;
 }
+
+/// Combination of a tag and a child table.
+///
+/// This is used in cases where a data structure has an array of records
```
cmyr, line 39: instead of being overly general I'd just say it's used in the script & feature lists? Do we anticipate any other cases?

cmyr (approved):
looks like a nice little ergonomics improvement.

read-fonts/src/tables/layout/script.rs
dfrg: snazzy... taken with a sprinkle of `.iter()` and `&`

read-fonts/src/table_ref.rs
dfrg: There are three cases here. I did think this pattern was more general but I'm not coming up with any additional cases. 

Going to move the type to the `layout` module and I'll narrow the docs. We can change in the future if other cases come up.

PR #1209, opened 2024-10-30
define OtRound for/to float types

anthrotype:
this will come in handy for rounding master values before passing them on to VariationModel::deltas while keeping them as floats (without the need to temporarily cast them to integer types), https://github.com/googlefonts/fontc/issues/1043

JMM

### DIFF
--- a/write-fonts/src/round.rs
+++ b/write-fonts/src/round.rs
@@ -42,6 +42,20 @@ impl OtRound<u16> for f32 {
     }
 }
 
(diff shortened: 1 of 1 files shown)


cmyr (approved):
huh just noticed that this also probably has the wrong signature, should be,


```rust
trait OtRound {
    type Output;
    fn ot_round(self) -> Self::Output;
}
```

but that's not a problem with this PR, so lgtm

PR #1210, opened 2024-10-30
[klippa] custom serializer step 1: add struct definition

qxliu76:
Also added a few APIs like push/embed/copy_bytes

### DIFF
--- a/klippa/src/serialize.rs
+++ b/klippa/src/serialize.rs
@@ -1,0 +1,354 @@
+//! serializer
+//! ported from Harfbuzz Serializer: <https://github.com/harfbuzz/harfbuzz/blob/5e32b5ca8fe430132b87c0eee6a1c056d37c35eb/src/hb-serialize.hh>
+
+use fnv::FnvHashMap;
+use write_fonts::types::Scalar;
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[allow(dead_code)]
+pub(crate) struct SerializeErrorFlags(u16);
+
+#[allow(dead_code)]
+impl SerializeErrorFlags {
+    pub const SERIALIZE_ERROR_NONE: Self = Self(0x0000);
+    pub const SERIALIZE_ERROR_OTHER: Self = Self(0x0001);
+    pub const SERIALIZE_ERROR_OFFSET_OVERFLOW: Self = Self(0x0002);
+    pub const SERIALIZE_ERROR_OUT_OF_ROOM: Self = Self(0x0004);
+    pub const SERIALIZE_ERROR_INT_OVERFLOW: Self = Self(0x0008);
+    pub const SERIALIZE_ERROR_ARRAY_OVERFLOW: Self = Self(0x0010);
+}
+
+impl Default for SerializeErrorFlags {
+    fn default() -> Self {
+        Self::SERIALIZE_ERROR_NONE
+    }
+}
+
+impl std::ops::BitOrAssign for SerializeErrorFlags {
+    /// Adds the set of flags.
+    #[inline]
+    fn bitor_assign(&mut self, other: Self) {
+        self.0 |= other.0;
+    }
+}
+
+impl std::ops::Not for SerializeErrorFlags {
+    type Output = bool;
+    #[inline]
+    fn not(self) -> bool {
+        self == SerializeErrorFlags::SERIALIZE_ERROR_NONE
+    }
+}
+
+#[allow(dead_code)]
+#[derive(Default)]
+// Offset relative to the current object head (default)/tail or
+// Absolute: from the start of the serialize buffer
+enum OffsetWhence {
+    #[default]
+    Head,
+    Tail,
+    Absolute,
+}
+
+type PoolIdx = usize;
+// Reference Harfbuzz implementation:
+// <https://github.com/harfbuzz/harfbuzz/blob/5e32b5ca8fe430132b87c0eee6a1c056d37c35eb/src/hb-serialize.hh#L69>
+#[derive(Default)]
+#[allow(dead_code)]
+struct Object {
+    // head/tail: indices of the output buffer for this object
+    head: usize,
+    tail: usize,
+    // real links are associated with actual offsets
+    real_links: Vec<Link>,
+    // virtual links not associated with actual offsets,
+    // they exist merely to enforce an ordering constraint.
+    virtual_links: Vec<Link>,
+
+    // pool index of the object that will be worked on next
+    next_obj: Option<PoolIdx>,
+}
+
+// LinkWidth:2->offset16, 3->offset24 and 4->offset32
+#[allow(dead_code)]
+#[derive(Default)]
+enum LinkWidth {
+    #[default]
+    Two,
+    Three,
+    Four,
+}
+
+type ObjIdx = u32;
+
+#[allow(dead_code)]
+#[derive(Default)]
+struct Link {
+    width: LinkWidth,
+    is_signed: bool,
+    whence: OffsetWhence,
+    bias: u32,
+    position: u32,
+    objidx: Option<ObjIdx>,
+}
+
+// reference harfbuzz implementation:
+//<https://github.com/harfbuzz/harfbuzz/blob/5e32b5ca8fe430132b87c0eee6a1c056d37c35eb/src/hb-serialize.hh#L57>
+#[derive(Default)]
+#[allow(dead_code)]
+pub(crate) struct Serializer {
+    start: usize,
+    end: usize,
+    head: usize,
+    tail: usize,
+    // TODO: zerocopy, debug_depth
+    errors: SerializeErrorFlags,
+
+    data: Vec<u8>,
+    object_pool: ObjectPool,
+    // index for current Object in the object_pool
+    current: Option<PoolIdx>,
+
+    packed: Vec<Option<ObjIdx>>,
+    packed_map: FnvHashMap<usize, ObjIdx>,
+}
+
+#[allow(dead_code)]
+impl Serializer {
+    pub(crate) fn new(size: u32) -> Self {
+        let buf_size = size as usize;
+        let mut this = Serializer {
+            data: vec![0; buf_size],
+            end: buf_size,
+            tail: buf_size,
+            packed: Vec::new(),
+            ..Default::default()
+        };
+
+        this.packed.push(None);
+        this
+    }
+
+    // Embed a single Scalar type
+    pub(crate) fn embed(&mut self, obj: impl Scalar) -> Result<usize, SerializeErrorFlags> {
+        let raw = obj.to_raw();
+        let bytes = raw.as_ref();
+        let size = bytes.len();
+
+        let ret = self.allocate_size(size)?;
+        self.data[ret..ret + size].copy_from_slice(bytes);
+        Ok(ret)
+    }
+
+    // Allocate size
+    pub(crate) fn allocate_size(&mut self, size: usize) -> Result<usize, SerializeErrorFlags> {
+        if self.in_error() {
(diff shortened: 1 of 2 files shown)


klippa/src/serialize.rs
```diff
@@ -0,0 +1,378 @@
...
+
+        obj.head = self.head;
+        obj.tail = self.tail;
+        obj.next_obj = self.current;
+        self.current = obj_idx;
+
+        Ok(())
+    }
+
+    pub(crate) fn copy_bytes(&self) -> Vec<u8> {
```
garretrieger, line 204: Since I don't believe we ever need multiple copies of the serialize result, you could make this more efficient by having it pass ownership of the internal vector (consuming the serialize context) instead of creating a copy. You can use https://doc.rust-lang.org/std/primitive.slice.html#method.copy_within to shift the tail bytes into the correct place then shrink the internal buffer as needed.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,378 @@
...
+        self.errors |= error_type;
+        self.errors
+    }
+
+    pub(crate) fn push(&mut self) -> Result<(), SerializeErrorFlags> {
+        if self.in_error() {
+            return Err(self.errors);
+        }
+
+        let obj_idx = self.object_pool.alloc();
```
garretrieger, line 191: The term object index is being used in two places with different meanings:
- Here meaning the index in the object pool
- Second, in Link's to represent the index in the packed array.

To avoid confusion I suggest giving these two things distinct types:

PackedIndex and PoolIndex
or
ObjIndex and PoolIndex if you want to stay closer to hb terminology.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,378 @@
...
+        !self.errors
+    }
+
+    #[inline]
+    pub(crate) fn in_error(&self) -> bool {
+        !!self.errors
+    }
+
+    #[inline]
+    pub(crate) fn err(&mut self, error_type: SerializeErrorFlags) -> SerializeErrorFlags {
```
garretrieger, line 181: Can you rename this to set_err to make it clear it's a mutating the serializer.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        let ret = self.next;
+        self.next = self.chunks[self.next].next;
+
+        ret
+    }
+
+    pub fn release(&mut self, obj: usize) {
+        if obj == NULL_IDX || obj >= self.chunks.len() {
```
garretrieger, line 282: If you use self.chunks.get_mut(obj) then you can handle this case using the returned Option. That avoids having to do the bounds check twice (here and later in self.chunks[obj])

klippa/src/serialize.rs
```diff
@@ -0,0 +1,378 @@
...
+            Err(SerializeErrorFlags::SERIALIZE_ERROR_OUT_OF_ROOM)
+        );
+        assert_eq!(
+            s.push(),
+            Err(SerializeErrorFlags::SERIALIZE_ERROR_OUT_OF_ROOM)
+        );
+    }
+
+    #[test]
+    fn test_object_pool() {
```
garretrieger, line 348: Can you add a test which triggers a new chunk being added during alloc().

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+use fnv::FnvHashMap;
+use write_fonts::types::Scalar;
+
+#[derive(Clone, Copy, Debug)]
+#[allow(dead_code)]
+pub(crate) struct SerializeErrorFlags(u16);
+
+#[allow(dead_code)]
+impl SerializeErrorFlags {
+    pub const SERIALIZE_ERROR_NONE: Self = Self(0x0000);
```
cmyr, line 11: Unless there's something fancy here I think this should just be a rust enum?

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn default() -> Self {
+        Self::SERIALIZE_ERROR_NONE
+    }
+}
+
+impl PartialEq for SerializeErrorFlags {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
```
cmyr, line 29: both of these could be derived..

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+    #[default]
+    Head,
+    Tail,
+    Absolute,
+}
+
+// the maximal len of vec in the object_pool is not expected to exceed u32::MAX
+const NULL_POOL_IDX: usize = u32::MAX as usize;
+
+type PoolIdx = usize;
```
cmyr, line 69: if we want to really leverage the type system we could consider making this an actual newtype, e.g.

```rust
struct PoolIdx(usize);
```

this would be more verbose but would also make it impossible to misuse.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+#[allow(dead_code)]
+struct Object {
+    head: usize,
+    tail: usize,
+    real_links: Vec<Link>,
+    virtual_links: Vec<Link>,
+    next_obj: PoolIdx,
```
cmyr, line 76: Could we just make this `Option<PoolIdx>`, and forget about the null sentinal value? A nice bonus of this approach is that we can now derive default.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+enum LinkWidth {
+    #[default]
+    Two,
+    Three,
+    Four,
+}
+
+type ObjIdx = u32;
+#[allow(dead_code)]
+const INVALID_OBJ_IDX: ObjIdx = ObjIdx::MAX;
```
cmyr, line 102: How is this used? Can we represent invalid states with option?

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+    Tail,
+    Absolute,
+}
+
+// the maximal len of vec in the object_pool is not expected to exceed u32::MAX
+const NULL_POOL_IDX: usize = u32::MAX as usize;
+
+type PoolIdx = usize;
+#[allow(dead_code)]
+struct Object {
```
cmyr, line 71: I think it's worth documenting some of these fields, it isn't clear to a casual reader what this is representing.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+    objidx: ObjIdx,
+}
+
+#[derive(Default)]
+#[allow(dead_code)]
+pub(crate) struct Serializer {
+    pub(crate) start: usize,
+    pub(crate) end: usize,
+    pub(crate) head: usize,
+    pub(crate) tail: usize,
```
cmyr, line 121: I'd document what these describe. I'm also slightly cautious that these are `pub(crate)`, since I assume we would be violating some invariants if somebody changed them out from under the serializer?

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+    // TODO: zerocopy, debug_depth
+    pub(crate) errors: SerializeErrorFlags,
```
cmyr, line 123: it would be more idiomatic to return `Result` from places where errors could occur, and let those results cascade.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+            packed: Vec::new(),
+            ..Default::default()
+        };
+
+        this.packed.push(INVALID_OBJ_IDX);
+        this
+    }
+
+    // Embed a single Scalar type
+    pub(crate) fn embed(&mut self, obj: impl Scalar) -> Result<usize, SerializeErrorFlags> {
```
cmyr, line 152: this logic is very similar to what happens in the `write.rs` module of `write-fonts`, maybe take a look at that and see if there's some inspiration?

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+        let raw = obj.to_raw();
+        let bytes = raw.as_ref();
+        let size = bytes.len();
```
cmyr, line 155: If you have acess to the type of `obj` (by using a named generic parameter, instead of `impl Trait`) than you can get this length with `T::RAW_BYTE_LEN`.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+            return Err(self.set_err(SerializeErrorFlags::SERIALIZE_ERROR_OUT_OF_ROOM));
+        }
+
+        //TODO: add support for clear?
+        let ret = self.head;
+        self.head += size;
+        Ok(ret)
+    }
+
+    #[inline]
```
cmyr, line 181: this should only really be necessary if this is called across crate boundaries; within a crate the compiler is intelligent about inlining (and if you see it failing after profiling you can use `#[inline(always)]`, `#[inline]` is just a hint)

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
...
+        obj.next_obj = self.current;
+        self.current = pool_idx;
+
+        Ok(())
+    }
+
+    pub(crate) fn copy_bytes(mut self) -> Vec<u8> {
+        if !self.successful() {
+            return Vec::new();
+        }
```
cmyr, line 218: I think this should be returning an error, and then the caller can choose to substitute an empty vec as needed?

klippa/src/serialize.rs
cmyr: Okay so after reading the rest of this:

I wonder if it makes sense to have some kind of buffer type that lives outside of the serializer, and which manages just the buffer? this could then be managed by the serializer?

It feels this is a reasonably isolated task, and would reduce overall complexity somewhat?

klippa/src/serialize.rs
```diff
@@ -0,0 +1,394 @@
+//! serializer
```
cmyr, line 1: assuming this is based on harfbuzz, it would be useful to link to the particular source that we're referencing, both here and for particular items below.

cmyr (commented):
This looks like it's ported from CPP? I have various notes inline that are mostly about things that don't feel very idiomatic to me. I appreciate these changes will take us further away from the reference implementation, so you'll have to decide where that's worth it...

klippa/src/serialize.rs
qxliu76: We'll need |= operator on this flag, so I'm keep it as a unsigned number

klippa/src/serialize.rs
qxliu76: We allow some errors(e.g offset overflow error) to exist and don't bail out, because they don't compromise the serializer state and could be resolved by the repacker later.

klippa/src/serialize.rs
```diff
@@ -0,0 +1,354 @@
...
+    tail: usize,
+    // TODO: zerocopy, debug_depth
+    errors: SerializeErrorFlags,
+
+    data: Vec<u8>,
+    object_pool: ObjectPool,
+    // index for current Object in the object_pool
+    current: Option<PoolIdx>,
+
+    packed: Vec<Option<ObjIdx>>,
```
garretrieger, line 113: The only item in this vector that will be None is the first one right? If so maybe we should just do the implementation in a way that doesn't require to have a single None object at the start? Then this can just be a Vec<ObjIdx>

garretrieger (approved):

klippa/src/serialize.rs
qxliu76: Agreed. I'll address this in later PRs

garretrieger: One other thing when you address this in the later PR: I believe this should be pool index instead of object index.

PR #1211, opened 2024-10-30
[skrifa] autohinting: apply GSUB to styles with features

dfrg:
Some autohinter styles have associated features like `c2sc` (caps to small caps). When generating metrics for these styles, we want to compute widths and alignment zones on the result of applying the relevant GSUB feature to our "blue" and standard width strings.

This patch adds support for simple single substitution GSUB lookups and applies them when requested.


drott (approved):
LGTM

PR #1212, opened 2024-10-30
[IFT] Add tests for error cases in Glyph Keyed patching.

garretrieger:
Adds a variety of tests that cover error cases when parsing and applying glyph keyed patches.

### DIFF
--- a/incremental-font-transfer/src/glyph_keyed.rs
+++ b/incremental-font-transfer/src/glyph_keyed.rs
@@ -32,7 +32,7 @@ pub(crate) fn apply_glyph_keyed_patches(
 
     for patch in patches {
         if patch.format() != Tag::new(b"ifgk") {
-            return Err(PatchingError::InvalidPatch("Patch file tag is not 'iftk'"));
+            return Err(PatchingError::InvalidPatch("Patch file tag is not 'ifgk'"));
         }
 
         decompression_buffer.push(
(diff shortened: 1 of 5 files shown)


incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -24,15 +24,15 @@ use std::ops::RangeInclusive;
...
+pub(crate) fn apply_glyph_keyed_patches(
     patches: &[GlyphKeyedPatch<'_>],
     font: &FontRef,
 ) -> Result<Vec<u8>, PatchingError> {
     let mut decompression_buffer: Vec<Vec<u8>> = Vec::with_capacity(patches.len());
 
     for patch in patches {
         if patch.format() != Tag::new(b"ifgk") {
-            return Err(PatchingError::InvalidPatch("Patch file tag is not 'iftk'"));
+            return Err(PatchingError::InvalidPatch("Patch file tag is not 'ifgk'"));
```
cmyr, line 35: briefly read this as `idfk` which should probably be a reserved table name

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -98,20 +104,31 @@ pub(crate) fn apply_glyph_keyed_patch(
...
+fn table_tag_list(glyph_patches: &[GlyphPatches]) -> Result<BTreeSet<Tag>, PatchingError> {
+    for patches in glyph_patches {
+        if patches
+            .tables()
+            .iter()
+            .zip(patches.tables().iter().skip(1))
+            .any(|(prev_tag, next_tag)| next_tag <= prev_tag)
+        {
+            return Err(PatchingError::InvalidPatch(
+                "Duplicate or unsorted table tag.",
```
cmyr, line 116: Not really the fault of this PR but I want to register my dislike for using strings as error types; using distinct variants for each failure case forces you to actually enumerate the failure cases, and also makes it easier to handle particular failure cases with a match statement.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -219,6 +236,12 @@ fn synthesize_glyf_and_loca<OffsetType: LocaOffset + TryFrom<usize>>(
     new_glyf: &mut [u8],
     new_loca: &mut [u8],
 ) -> Result<(), PatchingError> {
+    if !loca.all_offsets_are_ascending() {
+        return Err(PatchingError::FontParsingFailed(ReadError::MalformedData(
+            "loca contains unordered offsets.",
```
cmyr, line 241: this is a good counter example; we want to use a string here because the particular details of how data are malformed depend on context, and it's useful to let the user add that context where needed.

cmyr (approved):
looks good!

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: So in the case of the InvalidPatch error type the callers handling of the error isn't going to change depending on how specifically the patch was invalid, the string here is to just help with debugging to provide some additional info about why we're rejecting it.

Issue #1182, opened 2024-10-09
Font constants crate?


simoncozens:
One thing I have noticed in the Python font ecosystem is endless duplication of core constants; various packages have their own lists of:

* Valid language/script/feature tags
* Conversions between usWeightClass/usWidthClass values and fvar values
* Which scripts go through the Indic or USE shapers
* Well-known names for various weight instances
* Unicode range / codepage bit definitions
* Registered axis tags

and so on. Would there be any interest in hosting a crate in this repo with a common set of font-related constants?

rsheeter:
Makes sense to me. I have questions about whether it should be one crate, several crates, if some of it belongs in font-types, if some of it belongs in harfruzz (what shaper), etc.

cmyr:
We've discussed this sort of thing in the past, and I think it's a good idea. I would make it an additional crate in this repo, and called something like... `font-common` or similar?

We want to avoid adding much to `font-types` because it has some unsafe and requires additional review, so I think a new `#[forbid(unsafe_code)]` crate would be the best approach.

cmyr:
@rsheeter does this sound reasonable?

rsheeter:
Broadly yes. Needs a norse name of course. Is it possible to enumerate what we currently believe would go into this crate and what we think might consume it? It would also be nice to capture where current python and/or Rust definitions live. For example, fontc has a port of the python codepage range code @ https://github.com/googlefonts/fontc/blob/ab35d44b16267fa97c12c110bdc1b5dc53a01166/fontbe/src/os2.rs#L341. Presumably that should use the new library once it exists.

simoncozens:
> Is it possible to enumerate what we currently believe would go into this crate 

I made a list above. :-)

> what we think might consume it? 

The immediate motivation for this issue is fontspector (fontbakery port), but I imagine fontc and fea-rs also have collections of magical data. Python constants are scattered around `opentypespec-py`, `fontbakery.constants`, `ufo2ft.constants`, `fontTools.agl`, fontTools.ttLib.standardGlyphOrder`, `gftools.constants`, `gftools.util.styles`... 

> https://cleasby-vigfusson-dictionary.vercel.app/word/kunnasta perhaps?

Sounds gud, ya. I wanted fróðleikr just to see what cargo would do with utf8 crate names.

rsheeter:
> I made a list above. :-)

If that's the list all good, I wasn't sure if that was just quick examples or all the examples we currently have.

[...]

PR #1214, opened 2024-10-31
Add font constants crate

simoncozens:
This adds the `kunnasta` font constants library as described in https://github.com/googlefonts/fontations/issues/1182

Almost certainly more constants to come.


waywardmonkeys (commented):
This should get jobs added to `.github/workflows/rust.yml` for `test-stable`, `check-no-std`, and `check-wasm`

That'll mean either adding a `std` feature or making it always be `no_std`.

PR #1220, opened 2024-11-04
[font-types] wrapping add/sub for fixed point types

dfrg:
Per f2f, let's just stop chasing these panics.

JMM


rsheeter (approved):

drott:
Can we please clean up the existing explicit calls where we added them for `Fixed`?

dfrg:
Will do. We should also remove the wrapping methods since they’re unnecessary now.

PR #1222, opened 2024-11-05
Bump read-fonts, font-type, skrifa patch version

drott:


drott:
Changes in:
https://gist.github.com/drott/8e176849eb5fcd7321711467564ca360

Looks like patch version is sufficient, and I would need the overflow fixes released for Chromium.

anthrotype:
oops - looking at the cargo release changes it looks like i had forgotten to push the tags for read-fonts-v0.23.1 and write-fonts-v0.30.0 last week. sorry
I have created them manually at the commit they were supposed to be created immediately after merging my last version bumping PR. If you re-run cargo release changes, you should now see the actual changes since the last release, excluding the stuff that was alread released.

anthrotype:
actually, it's not that I had forgotten. I probably did ctrl-C on the release.sh bash script. I think the script prompts you to do that if you don't want to wait until the crates.io confirms it has gone through. And only at the end of the script (which I had skipped) it creates the git tags.

dfrg (approved):
LGTM

drott:
Thanks for the quick reviews - now I released those three crates.

PR #1228, opened 2024-11-12
More efficient cmap compilation

cmyr:
This produces significantly smaller cmap tables by combining segments where possible, and taking advantage of of the `offset_range_id` array to store glyphids directly where this is beneficial.

This fixes the issue where our cmap tables were often significantly larger than the ones produced by the python toolchain.

I believe there may be one further optimization possible here, which i am curious to explore but am holding back because I think the size savings are negligible: but I believe it is theoretically possible to reuse glyph ids in the `glyphIDArray`, if two segments have glyphs identical spacing patterns, due to the fact that the idDelta is still applied even if the glyphidarray is used.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -39,7 +39,7 @@ impl CmapSubtable {
             // no chars in BMP
             return None;
         }
-        let n_segments = segments.len();
+        let n_segments = segments.len() + 1;
         for (i, segment) in segments.into_iter().enumerate() {
             let start = mappings[segment.start_ix].0;
             let end = mappings[segment.end_ix].0;
@@ -294,12 +294,15 @@ impl<'a> Format4Ranges<'a> {
 
     /// a convenience method called from our iter in the various cases where
     /// we emit a segment.
+    ///
+    /// a 'seg_len' of 0 means start == end, e.g. a segment of one glyph.
     fn make_segment(&mut self, seg_len: usize) -> Format4Segment {
+        // if start == end, we should always use a delta.
+        let use_delta = self.currently_contiguous || seg_len == 0;
         let result = Format4Segment {
             start_ix: self.seg_start,
             end_ix: self.seg_start + seg_len,
-            id_delta: self
-                .currently_contiguous
+            id_delta: use_delta
                 .then_some(
                     self.mappings
                         .get(self.seg_start)
@@ -359,7 +362,7 @@ impl<'a> Iterator for Format4Ranges<'a> {
 
         // if we're done looping then create the last segment:
         let last_idx = self.mappings.len() - 1;
-        return Some(self.make_segment(last_idx - self.seg_start));
+        Some(self.make_segment(last_idx - self.seg_start))
     }
 }
 
@@ -719,34 +722,18 @@ mod tests {
         assert_eq!(result, Err(CmapConflict { ch, gid1, gid2 }))
     }
 
-    // input is a sequence of ranges representing contiguous gids,
-    // output is how they're grouped into segments
-    fn compute_ranges(
-        iter: impl IntoIterator<Item = RangeInclusive<char>>,
-    ) -> Vec<RangeInclusive<char>> {
-        let mut next_gid = 0u16;
-        let mappings = iter
-            .into_iter()
-            .flat_map(|range| {
-                // start of new range means we aren't contiguous:
-                let start_gid = next_gid;
-                next_gid += 1 + range.clone().count() as u16;
-                range
-                    .enumerate()
-                    .map(move |(i, c)| (c, GlyphId::new((start_gid + i as u16) as _)))
-            })
-            .collect::<Vec<_>>();
-
-        super::compute_format_4_segments(&mappings)
-            .into_iter()
-            .map(|seg| mappings[seg.start_ix].0..=mappings[seg.end_ix].0)
-            .collect()
-    }
-
-    #[derive(Default)]
     struct MappingBuilder {
         mappings: Vec<(char, GlyphId)>,
         next_gid: u16,
+    }
+
+    impl Default for MappingBuilder {
+        fn default() -> Self {
+            Self {
+                mappings: Default::default(),
+                next_gid: 1,
+            }
+        }
     }
 
     impl MappingBuilder {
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+        return Default::default();
+    };
+
+    let mut result = vec![first];
+
+    // now we want to collect the segments, combining smaller segments where
+    // that leads to a size savings.
+    //
+    // This differs from the python, which starts from larger segments and then
+    // subdivides them, but the overall idea is the same.
```
rsheeter, line 383: Why do we choose to differ from python?

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+        };
+        self.seg_start += seg_len + 1;
+        self.currently_contiguous = false;
+        result
+    }
+}
+
+// this iterator creates segments greedily, that is it will create a new segment
+// at every opportunity. These are then recombined by the caller to generate
+// the most efficient overall sequence of segments.
```
rsheeter, line 321: I'm not quite sure what this is trying to tell me, why this is important. Perhaps we could mention that?

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+    id_delta: Option<i32>,
+}
+
+impl Format4Segment {
+    fn len(&self) -> usize {
+        self.end_ix - self.start_ix + 1
+    }
+
+    // cost in bytes of this segment
+    fn cost(&self) -> usize {
```
rsheeter, line 260: add a spec ref?

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+        let glyph_id_cost = self
+            .id_delta
+            .is_none()
+            .then(|| self.len() * u16::RAW_BYTE_LEN)
+            .unwrap_or(0);
+        BASE_COST + glyph_id_cost
+    }
+
+    fn combine(&self, other: Format4Segment) -> Format4Segment {
+        assert_eq!(other.start_ix, self.end_ix + 1);
```
rsheeter, line 271: Not loving the assert, I suggest modelling the failure explicitly, e.g. try_combine or similar.

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+        let Some(((mut prev_cp, mut prev_gid), rest)) =
+            self.mappings[self.seg_start..].split_first()
+        else {
+            // if this is the last element, make a final segment
+            return Some(self.make_segment(0));
+        };
+
+        for (i, (cp, gid)) in rest.iter().enumerate() {
+            // first: all segments must be a contiguous range of codepoints
+            if *cp as u32 - prev_cp as u32 > 1 {
```
rsheeter, line 338: The check is throwing me, can' t we just check `if cp != prev_cp + 1`, e.g. more directly say "if cp doesn't continue the run"

write-fonts/src/tables/cmap.rs
rsheeter: Actually that sort of describes the code in general. I think it's saying it tries to find runs, prioritizing, from most to least favorable:

1. A run of at least 2 where codepoint *and* gid are contiguous
1. A run of at least 2 where codepoints are contiguous (but gids aren't)
1. A run of 1

I think the comment might want to elaborate.

rsheeter: https://github.com/googlefonts/fontations/pull/1229 attempts to simplify, eliminating the currently_contiguous field and making the iterator just grab the next run, prioritized as ^, each cycle.

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+            continue;
+        }
+
+        // next: if chars are contiguous and neither has a delta, we always combine
+        // this will mostly happen if we've combined the previous, previously
+        // delta-having segment
+        if is_contiguous_with_prev && prev.id_delta.is_none() && segment.id_delta.is_none() {
+            *prev = prev.combine(segment);
+            continue;
+        }
```
rsheeter, line 407: Why isn't the combine if bytes saved check sufficient?

write-fonts/src/tables/cmap.rs
```diff
@@ -237,6 +236,187 @@ impl Cmap {
...
+            let combined = prev.combine(segment);
+            if combined.cost() < prev.cost() + segment.cost() {
+                *prev = combined;
+                continue;
+            }
+        }
+        result.push(segment);
+    }
+    result
+}
```
rsheeter, line 419: I'm slightly surprised the loop doesn't mirror the comment more directly, e.g. see if we can combine, do so if it saves bytes? - naive pseudocode:

```rust
while let Some(next) = iter.next() {
  // we initialized result with the first result from iter
  if !result.last_mut().unwrap().try_combine(next) {
    result.push(next);
  }
}
```

Checking next and prev and checking id delta instead of letting the cost check handle it both throw me.

rsheeter: https://github.com/googlefonts/fontations/pull/1230 tries ^ and seems to pass the tests.

write-fonts/src/tables/cmap.rs
rsheeter: https://github.com/googlefonts/fontations/pull/1230 explores this idea

rsheeter (changes requested):
Generally looks good but I found some of the code hard to follow. I sent suggestions for simplifications. I'm not tied ot the specific simplifications I proposed but I would like to look again once it's been altered to be a bit easier to follow so marking as Request changes.

write-fonts/src/tables/cmap.rs
cmyr: I'm trying to explain a slightly arbitrary design choice; that this iterator returns segments but that they need further optimization. In my initial design the iterator produced the segments directly, but you need to do some lookahead to determine if you should split a segment or not.

write-fonts/src/tables/cmap.rs
cmyr: this is just documenting an invariant, we only combine contiguous segments.  Can add a comment...

write-fonts/src/tables/cmap.rs
cmyr: good catch, I didn't write the `cost()` check until I'd gotten to that lat branch that has the tricky logic, but you're right that it should cover this case as well.

write-fonts/src/tables/cmap.rs
cmyr: yea I think this is a nice idea, will play around a bit.

write-fonts/src/tables/cmap.rs
cmyr: there's no specific part of the spec that describes a segment, but I'll try to make this clearer..

write-fonts/src/tables/cmap.rs
```diff
@@ -238,6 +237,222 @@ impl Cmap {
...
+            let can_merge_left = prev.can_merge(&current);
+
+            if !can_merge_left {
+                result.push(current);
+                continue;
+            }
+
+            let combined = prev.combine(current);
+            let should_merge = match next.as_ref() {
+                // if we can merge on both sides, consider total cost
```
rsheeter, line 438: Why do you need to check both sides at once? That is, why doesn't checking if current could merge with the last entry in result suffice?

write-fonts/src/tables/cmap.rs
```diff
@@ -238,6 +237,222 @@ impl Cmap {
...
+        self.seg_start += seg_len + 1;
+        self.gids_in_order = false;
+        result
+    }
+
+    /// Find the next possible segment.
+    ///
+    /// A segment _must_ be a contiguous range of chars, but we where such a range
+    /// contains subranges that are also contiguous ranges of glyph ids, we will
+    /// split those subranges into separate segments.
```
rsheeter, line 345: This seems more complicated than just finding the next run more directly? - we've got various special cases that go away if we do that:

1. `if prev gids were ordered but this one isn't, end prev segment`
1. tracking `self.gids_in_order`
1. `if gids were not previously ordered but are now`
1. `done looping then create the last segment`

https://github.com/googlefonts/fontations/pull/1229/files avoids all this. Is there a reason this is better that I'm missing?

write-fonts/src/tables/cmap.rs
rsheeter: Per IM, let's add a comment explaining the arithmatic of merging that leads to potential for merging 3 to be a win where 2 wasn't, and no need to check 4+

rsheeter (approved):
LGTM subject to remaining comments being addressed

Issue #1221, opened 2024-11-05
Non-standard extension should be behind flag


drott:
Compare https://github.com/googlefonts/fontations/pull/927

I would prefer if we kept not yet fully standardized extensions behind a flag to avoid breakage and being able to follow Chromium's release process.

dfrg:
We can either feature gate this or just mask out the cubic bit on read with a comment that it should be re-enabled when the spec finalizes. I lean toward the latter unless we have a case where we really need to parse cubic glyf outlines. 

@rsheeter any preference?

drott:
>  I lean toward the latter [...]

Yes, let's move forward with that, could you do a PR? Thanks in advance.

PR #1232, opened 2024-11-12
[read-fonts] feature gate cubic glyf support

dfrg:
We don't want to ship experimental functionality in Chrome. Resolved per #1221 to mask out the bit and re-enable later. Also ignores two tests for this.

cc @drott


rsheeter:
Why not put it behind a feature and make sure we run tests for that feature?

rsheeter:
For context, I anticipate we'll want to add more OFF.Next functionality in the not too distant future and we won't want it all commented out/ignored

dfrg:
Per f2f, resolved that we’ll add a `spec_next` feature (off by default) to capture this and any future proposed features.

rsheeter (approved):

PR #1233, opened 2024-11-12
[skrifa] cff: prevent overflow in subrs offset

dfrg:
Detected add with overflow with computing the private subroutines offset.

ref https://issues.oss-fuzz.com/issues/377965575

Factored out the private dict parsing to enable testing this directly. First commit contains failing test.


cmyr (approved):
👍

PR #1234, opened 2024-11-12
[klippa] serializer: Added packed_map hash table and pop_pack()

qxliu76:


klippa/src/serialize.rs
```diff
@@ -165,6 +178,12 @@ impl Serializer {
         !!self.errors
     }
 
+    pub(crate) fn only_overflow(&self) -> bool {
+        self.errors == SerializeErrorFlags::SERIALIZE_ERROR_OFFSET_OVERFLOW
+            || self.errors == SerializeErrorFlags::SERIALIZE_ERROR_INT_OVERFLOW
+            || self.errors == SerializeErrorFlags::SERIALIZE_ERROR_ARRAY_OVERFLOW
+    }
```
cmyr, line 185: iirc this check is returning `false` if the bitset contains some combination of these flags?

klippa/src/serialize.rs
```diff
@@ -188,6 +207,106 @@ impl Serializer {
...
+            let hash_old = hash_one_pool_idx(to.0, &self.data, &self.object_pool);
+            self.packed_map
+                .del(to.0, hash_old, &self.data, &self.object_pool);
+        }
+
+        {
+            let to_obj = self.object_pool.get_obj_mut(to.0).unwrap();
+            to_obj.virtual_links.append(&mut from_vec);
+        }
+
```
cmyr, line 304: I don't think these braces are doing anything? (this still compiles fine without them)

klippa/src/serialize.rs
```diff
@@ -188,6 +207,106 @@ impl Serializer {
...
+            self.head = obj.head;
+            self.tail -= len;
+
+            obj_head = obj.head;
+            obj.head = self.tail;
+            obj.tail = self.tail + len;
+
+            //TODO: consider zerocopy here
+            self.data.copy_within(obj_head..obj_head + len, self.tail);
+        }
```
cmyr, line 248: These braces don't seem necessary? (maybe they were in an earlier iteration?)

cmyr (approved):
Just a quick pass looking at rust stuff; maybe @garretrieger can take a look as well, since he is more familiar with the harfbuzz code?

klippa/src/serialize.rs
qxliu76: Oh yeah it seems these braces are not needed with mem::take. Thank you!

klippa/src/serialize.rs
```diff
@@ -70,20 +76,27 @@ struct Object {
     next_obj: Option<PoolIdx>,
 }
 
+impl Object {
+    fn fini(&mut self) {
```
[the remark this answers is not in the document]
garretrieger: or maybe reset()?

klippa/src/serialize.rs
```diff
@@ -188,6 +207,106 @@ impl Serializer {
...
+            obj.tail = self.tail + len;
+
+            //TODO: consider zerocopy here
+            self.data.copy_within(obj_head..obj_head + len, self.tail);
+        }
+
+        let mut hash = 0_u64;
+        // introduce flags to avoid mutable borrow and immutable borrow issues
+        let mut obj_duplicate = false;
+        let mut dup_obj_idxes = (0, 0);
```
garretrieger, line 253: obj_duplicate and dup_obj_idxes could be combined into a single Option<...> field

klippa/src/serialize.rs
```diff
@@ -188,6 +207,106 @@ impl Serializer {
...
+                obj_duplicate = true;
+                dup_obj_idxes = *entry;
+            };
+        }
+
+        if obj_duplicate {
+            self.merge_virtual_links(pool_idx, dup_obj_idxes);
+            let obj = self.object_pool.get_obj_mut(pool_idx).unwrap();
+            obj.fini();
+            self.object_pool.release(pool_idx);
```
garretrieger, line 269: could we have release automatically call .fini()?

klippa/src/serialize.rs
```diff
@@ -351,4 +551,171 @@ mod test {
...
+        let n = UfWord::new(10);
+        assert_eq!(s.embed(n), Ok(3));
+        assert_eq!(s.head, 5);
+        assert_eq!(s.pop_pack(true), Some(2));
+        assert_eq!(s.packed_map.hash_table.len(), 2);
+        //check to make sure head is rewinded
+        assert_eq!(s.head, 3);
+        //check that tail is updated
+        assert_eq!(s.tail, 95);
+    }
```
garretrieger, line 720: Some additional tests to add:
- pop_pack(false) when a duplicate object exists
- de-duplication with virtual links (ie. checking merge virtual links works, and that virtual links don't affect equality for de-duping).
- de-duplication on an object that has real links to test they are being included in hash and equality computations.

klippa/src/serialize.rs
qxliu76: Good question! It's ported from Harfbuzz. I think these flags are exclusive, maybe @garretrieger can help to confirm?

klippa/src/serialize.rs
qxliu76: right, no braces are needed here after I moved code around. thank you!

klippa/src/serialize.rs
qxliu76: good idea! We should do the same thing in harfbuzz as well.

klippa/src/serialize.rs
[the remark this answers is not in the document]
qxliu76: Done.

klippa/src/serialize.rs
qxliu76: Thanks! Adding de-duplication test with virtual links uncovered a typo in the code: I used real_links in the merge_virtual_links() API.

klippa/src/serialize.rs
garretrieger: Not sure why specifically it's that way in hb but unless we have a strong reason otherwise we should just match hb.

garretrieger (approved):

PR #1237, opened 2024-11-13
Add initial implementation of `meta`

Hoolean:
Hello!

This PR adds adds read and write for the `meta` table, including a codegen definition, some manual parts to handle the offset slices, and plumbing to surface the table within `read-fonts` and `write-fonts`.

This is my first contribution, so where possible I've borrowed liberally from the implementation of other tables, especially `name` (for its handling of offset prescribed-length data).

This is an early implementation, and has not had the manual and automated testing it will require yet. Additionally, there is an implementation of the `FontWrite` trait for `&[u8]` that may be redundant.

As such, I'm making the PR as a draft for now, until I can gather feedback and dig a bit deeper :)

<sub>Note: this is a personal contribution independent of my employer, and so I've submitted from a fork under my personal profile and email to make this distinction</sub>


Hoolean:
Thanks for reviewing Colin, and an extra thanks for furthering the PR!

I am very happy every time Rust enables us to do more strict type-based validation on top of that which fonttools provided, and so am more than happy to close this in favour of #1244 - it is exactly how I hoped this initial branch could be used, and it is great to see and learn from the improvements you have made :)

PR #1238, opened 2024-11-13
[read] Expose table marker, make range fns public

cmyr:
This is useful for applications like subsetting.

The individual markers are still #[doc(hidden)] because they really aren't intended to be public API, so this flood of new methods should also not show up in the docs.


qxliu76:
LGTM

PR #1239, opened 2024-11-13
[IFT] major rework to the IFT API around patch intersection and application.

garretrieger:
Adds a new patch_group module which generates a candidate set of URIs to be applied next based on the list of intersecting patches. The candidate URI group is structured to enforce compatibility constraints between patches. PatchGroup's can then be used to trigger the next iteration of patch application.

Example usage:

```
loop {
  let g = PatchGroup::select_next_patches(font, &subset_definition).unwrap();
  if !g.has_uris() {
    break;
  }

  // g.uris() lists patches to be downloaded, 
  // patch_data is populated with downloaded data.
  let patch_data = ...;

  font = g.apply_next_patches(&mut patch_data).unwrap();
}
```

### DIFF
--- a/incremental-font-transfer/src/glyph_keyed.rs
+++ b/incremental-font-transfer/src/glyph_keyed.rs
@@ -1,25 +1,28 @@
+//! Implementation of Glyph Keyed patch application.
+//!
+//! Glyph Keyed patches are a type of incremental font patch which stores opaque data blobs
+//! keyed by glyph id. Patch application places the data blobs into the appropriate place
+//! in the base font based on the associated glyph id.
+//!
+//! Glyph Keyed patches are specified here:
+//! <https://w3c.github.io/IFT/Overview.html#glyph-keyed>
 use crate::font_patch::PatchingError;
-/// Implementation of Glyph Keyed patch application.
-///
-/// Glyph Keyed patches are a type of incremental font patch which stores opaque data blobs
-/// keyed by glyph id. Patch application places the data blobs into the appropriate place
-/// in the base font based on the associated glyph id.
-///
-/// Glyph Keyed patches are specified here:
-/// <https://w3c.github.io/IFT/Overview.html#glyph-keyed>
 use crate::table_keyed::copy_unprocessed_tables;
 
 use font_types::Scalar;
-use read_fonts::collections::IntSet;
-use read_fonts::tables::ift::{GlyphKeyedPatch, GlyphPatches};
-use read_fonts::tables::loca::Loca;
-use read_fonts::types::Tag;
-use read_fonts::ReadError;
-use read_fonts::{FontData, FontRef, TableProvider};
+use read_fonts::{
+    collections::IntSet,
+    tables::{
+        ift::{GlyphKeyedPatch, GlyphPatches},
+        loca::Loca,
+    },
+    types::Tag,
+    FontData, FontRef, ReadError, TableProvider,
+};
+
 use shared_brotli_patch_decoder::shared_brotli_decode;
 use skrifa::GlyphId;
-use std::collections::BTreeSet;
-use std::collections::HashMap;
+use std::collections::{BTreeSet, HashMap};
 use std::ops::RangeInclusive;
 
 use write_fonts::FontBuilder;
--- a/incremental-font-transfer/src/glyph_keyed.rs
+++ b/incremental-font-transfer/src/glyph_keyed.rs
@@ -414,7 +414,7 @@ fn patch_glyf_and_loca<'a>(
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use std::{collections::BTreeSet, io::Write};
 
     use brotlic::CompressorWriter;
@@ -432,7 +432,7 @@ mod tests {
 
     use crate::{font_patch::PatchingError, glyph_keyed::apply_glyph_keyed_patches};
 
-    fn assemble_patch(mut header: BeBuffer, payload: BeBuffer) -> BeBuffer {
+    pub(crate) fn assemble_glyph_keyed_patch(mut header: BeBuffer, payload: BeBuffer) -> BeBuffer {
         let payload_data: &[u8] = &payload;
         let mut compressor = CompressorWriter::new(Vec::new());
         compressor.write_all(payload_data).unwrap();
@@ -466,7 +466,8 @@ mod tests {
 
     #[test]
     fn noop_glyph_keyed() {
-        let patch = assemble_patch(glyph_keyed_patch_header(), noop_glyf_glyph_patches());
+        let patch =
+            assemble_glyph_keyed_patch(glyph_keyed_patch_header(), noop_glyf_glyph_patches());
         let patch: &[u8] = &patch;
         let patch = GlyphKeyedPatch::read(FontData::new(patch)).unwrap();
 
@@ -481,7 +482,8 @@ mod tests {
 
     #[test]
     fn basic_glyph_keyed() {
-        let patch = assemble_patch(glyph_keyed_patch_header(), glyf_u16_glyph_patches());
+        let patch =
+            assemble_glyph_keyed_patch(glyph_keyed_patch_header(), glyf_u16_glyph_patches());
         let patch: &[u8] = &patch;
         let patch = GlyphKeyedPatch::read(FontData::new(patch)).unwrap();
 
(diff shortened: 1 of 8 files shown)


incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -1,3 +1,4 @@
+use crate::font_patch::PatchingError;
 /// Implementation of Glyph Keyed patch application.
```
cmyr, line 2: not sure how this happened, but if these are documenting the overall module they should start with `//!` not `///`.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -1,3 +1,4 @@
+use crate::font_patch::PatchingError;
```
cmyr, line 1: probably an auto-import misfire?

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+use std::collections::{BTreeMap, HashMap};
+
+use crate::{
+    font_patch::{IncrementalFontPatchBase, PatchingError},
+    patchmap::{intersecting_patches, IftTableTag, PatchEncoding, PatchUri, SubsetDefinition},
+};
+
+/// A group of patches derived from a single IFT font which can be applied simultaneously
+/// to that font. Patches are initially missing data which must be fetched and supplied to
+/// the group before it can be applied to the font.
```
cmyr, line 11: doc comments sort of work like git commits; you want a short first line, then whitespace, then you can elaborate.

The first line is used as a summary when the docs are rendered.

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+pub struct PatchGroup<'a> {
+    font: FontRef<'a>,
+    patches: Option<CompatibleGroup>,
+}
+
+impl<'a> PatchGroup<'a> {
+    /// Intersect the available and unapplied patches in ift_font against subset_definition and return a group
+    /// of patches which would be applied next.
+    pub fn select_next_patches<'b>(
```
cmyr, line 20: is this `'b` necessary? afaict the lifetimes are the same...

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+            ift_font.iftx().ok().map(|t| t.compatibility_id()),
+        )?;
+
+        Ok(PatchGroup {
+            font: ift_font,
+            patches: Some(compat_group),
+        })
+    }
+
+    /// Returns the list of URIs in this group.
```
cmyr, line 44:
```suggestion
    /// Returns an iterator over URIs in this group.
```

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+    fn invalidating_patch_iter(&self) -> impl Iterator<Item = &PatchInfo> {
+        let full = match &self.patches {
+            Some(CompatibleGroup::Full(info)) => Some(&info.0),
+            _ => None,
+        };
+
+        let partial_1 = match &self.patches {
+            Some(CompatibleGroup::Mixed {
+                ift: ScopedGroup::PartialInvalidation(v),
+                iftx: _,
```
cmyr, line 74: you don't need to name the field you don't care about here; you can just write,

```rust
let MyStruct { x, y, .. } = thing;
```

that is you can use two trailing dots to indicate "and whatever else is in here".

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+            }) => Some(&v.0),
+            _ => None,
+        };
+
+        full.into_iter().chain(partial_1).chain(partial_2)
+    }
+
+    fn non_invalidating_patch_iter(&self) -> impl Iterator<Item = &PatchInfo> {
+        let ift = match &self.patches {
+            Some(CompatibleGroup::Mixed { ift, iftx: _ }) => Some(ift),
```
cmyr, line 92: ditto here, it could just be,
```suggestion
            Some(CompatibleGroup::Mixed { ift, .. }) => Some(ift),
```

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+            }
+        }
+
+        // Step 2 - now make patch selections in priority order: first full invalidation, second partial, lastly none.
+        if let Some(patch) = full_invalidation.into_iter().next() {
+            // TODO(garretrieger): use a heuristic to select the best patch
+            return Ok(CompatibleGroup::Full(patch));
+        }
+
+        let mut ift_selected_uri: Option<String> = None;
```
cmyr, line 174: instead of declaring this mut and the assigning to it, would it make sense to just return a tuple from the map?

as in,

```rust
let (red, blue) = iter.next().map(|y| (y.0, y.1)).unzip();
```

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+impl PatchInfo {
+    pub(crate) fn tag(&self) -> &IftTableTag {
+        &self.source_table
+    }
+}
+
+impl From<PatchUri> for PatchInfo {
+    fn from(value: PatchUri) -> Self {
+        PatchInfo {
+            uri: value.uri_string(),
```
cmyr, line 313: You could consider using something other than `String` to store URI, if you'r eat all concerned about the cloning costs? A simple option would be `Rc<str>`, which is a shared reference counted `str`. If you need thread safety, you could use `Arc<str>` instead.

incremental-font-transfer/src/patch_group.rs
```diff
@@ -0,0 +1,1253 @@
...
+            ScopedGroup::NoInvalidation(map) => NoInvalidationPatchesIter {
+                it: Some(map.values()),
+            },
+        }
+    }
+}
+
+struct NoInvalidationPatchesIter<'a, T>
+where
+    T: Iterator<Item = &'a NoInvalidationPatch>,
```
cmyr, line 367: you can probably remove this bound, and just leave it on the impl block. (below, as well)

incremental-font-transfer/src/patchmap.rs
```diff
@@ -528,30 +574,83 @@ pub struct PatchUri {
...
     id: PatchId,
     encoding: PatchEncoding,
-    expected_compatibility_id: CompatibilityId,
+    source_table: IftTableTag,
     // TODO(garretrieger): add a resolve method which when supplied the bytes associated with the URL
     //                     produces an object suitable for passing to the font_patch API.
 }
 
 impl PatchUri {
+    const BASE32HEX_NO_PADDING: data_encoding::Encoding = new_encoding! {
```
cmyr, line 583: huh, neat.

incremental-font-transfer/src/table_keyed.rs
```diff
@@ -0,0 +1,441 @@
...
+    let mut processed_tables = BTreeSet::<Tag>::new();
+    // TODO(garretrieger): enforce a max combined size of all decoded tables? say something in the spec about this?
+    for i in 0..patch.patches_count() {
+        let i = i as usize;
+        let next = i + 1;
+
+        let table_patch = patch
+            .patches()
+            .get(i)
+            .map_err(PatchingError::PatchParsingFailed)?;
```
cmyr, line 32: would it make sense to just be `iter().enumerate()`ing the patch?

incremental-font-transfer/src/table_keyed.rs
```diff
@@ -0,0 +1,441 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use font_test_data::ift::{noop_table_keyed_patch, table_keyed_patch};
+    use read_fonts::FontData;
+    use read_fonts::FontRead;
+    use read_fonts::FontRef;
+    use read_fonts::ReadError;
```
cmyr, line 131: rust-analyzer has a 'organize imports' or 'cleanup imports' action...

incremental-font-transfer/src/table_keyed.rs
```diff
@@ -0,0 +1,441 @@
...
+    use write_fonts::FontBuilder;
+
+    const IFT_TABLE: &[u8] = "IFT PATCH MAP".as_bytes();
```
cmyr, line 134: you can write a byte literal be prepending a string literal with 'b', e.g. `b"IFT PATCH MAP"`.

cmyr (approved):
I have no familiarity with the IFT spec, but reading through this nothing strikes me as egregious. I've left some little rust style nits inline.

incremental-font-transfer/src/patch_group.rs
garretrieger: It appears so, I'm unable to get it to compile without.

incremental-font-transfer/src/patch_group.rs
garretrieger: Compiler says no :(

PR #1240, opened 2024-11-13
[write] Minor improvment to cmap 4 segmentation

cmyr:
There was one case where fonttools was saving several bytes on us, which felt bad. We are now equal or better, for every font I've examined.

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -434,16 +434,20 @@ impl<'a> Format4SegmentComputer<'a> {
             }
 
             let combined = prev.combine(current);
-            let should_merge = match next.as_ref() {
-                // if we can merge on both sides, consider total cost
-                Some(next) if current.can_merge(next) => {
+            // if we should merge, based just on the previous segment cost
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/cmap.rs
```diff
@@ -434,16 +434,18 @@ impl<'a> Format4SegmentComputer<'a> {
...
-            if should_merge {
+            // if we should merge, based just on the previous segment cost
+            let should_merge = combined.cost() < prev.cost() + current.cost();
+            // but if we can also merge with the next segment we want to check
+            // that too
+            let should_merge_considering_next = next
+                .as_ref()
+                // don't bother computing if we have already decided to merge
+                .filter(|next| !should_merge && current.can_merge(next))
+                .map(|next| combined.combine(*next).cost() < prev.cost() + current.cost())
```
cmyr, line 445: this looks wrong 😒

rsheeter (approved):
makes sense to add to cmap-size

cmyr:
I'll close this and add it as a commit to #1228

PR #1244, opened 2024-11-14
Custom types for the meta table

cmyr:
This is based on #1237; I think it makes sense to use custom types for the actual data in this table, but I wouldn't have been able to explain what to do without trying it first, so here we are.

(also this needs tests)


write-fonts/src/tables/meta.rs
```diff
@@ -0,0 +1,137 @@
...
+                }
+                len
+            }
+            Metadata::Other(vec) => {
+                vec.write_into(writer);
+                vec.len()
+            }
+        };
+
+        let len: u32 = len.try_into().unwrap();
```
rsheeter, line 85: I don't love the unwrap, is this common in write-fonts? Should file an issue to make write_into fallible and remove that?

rsheeter (approved):

rsheeter (approved):
LGTM subject to tests to prove it works

write-fonts/src/tables/meta.rs
cmyr: By design `FontWrite` is infallible; we do all the error checking in an earlier validation pass.

In generated code, we do add a check that an array's length can be represented by the type that stores the length, but I didn't bother hand-writing that here, given that any single font table's length must be representable in 32 bits or else its offset cannot be stored in the root table directory.

rsheeter: Maybe we should comment as much and do an infallible production of len?

resources/codegen_inputs/meta.rs
```diff
@@ -0,0 +1,38 @@
...
+    version: u32,
+    /// Flags — currently unused; set to 0.
+    #[compile(0)]
+    flags: u32,
+    /// Not used; set to 0.
+    #[skip_getter]
+    #[compile(0)]
+    reserved: u32,
+    /// The number of data maps in the table.
+    #[compile(array_len($data_maps))]
```
Hoolean, line 17: There is one part here that I wanted to gather opinion on for the initial implementation :mag_right:

[`array_len` mandates that its calculated value fits within `u16`](https://github.com/googlefonts/fontations/blob/a41651ce0ee061bbda491f011894fd183d053ff0/write-fonts/src/lib.rs#L185-L188); it is unlikely that any production fonts are compiling `meta` tables with more than ~60,000 records, but is it worth fontations using a split or generic `array_len` function to fully support these wider types? (from memory, this case appears in a few other codegen inputs too)

cmyr: good catch! I had been thinking that array_len returned a usize, so if this is returning a `u16` it's definitely not right, let me see if there's an easy fix.

write-fonts/src/tables/meta.rs
cmyr: So you've brought up something that is ultimately a larger philosophical question that we should discuss, which is what we want to do when the user really wants to write nonsensical table data.

Some possible options:

- let them
- let them, but with hand guards: for instance always truncating collections to their maximum allowed length
- crash

bear in mind that we do already return an error if they fail validation, so this decision only really concerns the case where,

- someone is manually skipping validation (which would require very intentional work, validation happens automatically if you use `dump_table` so you'd need to basically reimplement that function yourself)
- we have failed to add some validation check (this is definitely going to be the case _somewhere_)

so my personal preference right now would be to crash, because it tells us in this latter case very clearly that we need to add a validation check for something, but maybe this is not appropriate for all cases?

In this specific case, I'll add the validation check manually and leave the unwrap, since that is consistent with what we do elsewhere. The only way you could hit the unwrap is if you manually write your own serialization code that skips validation, and then construct a completely unreasonable `meta` table that you pass to your custom serialization code, so it's not going to be hit by accident...

resources/codegen_inputs/meta.rs
cmyr: I've opened #1251 to fix the issue with array_len, which as you note predates this PR, so I'll go ahead and merge this. :)

PR #1250, opened 2024-11-19
[IFT] add patch application call to patch group fuzzer.

garretrieger:
To make this effective the brotli decoder has been configured to be disabled during fuzzing. This is needed because the brotli decoder code is c++ based and the fuzzer isn't able to instrument and track coverage through it, which results in the fuzzer being extremely unlikely to generate valid brotli encoded data.


cmyr (approved):
Looks good!

PR #1251, opened 2024-11-19
[codegen] Better handling of computed lengths

cmyr:
This fixes an issue where we were incorrectly truncating all computed lengths to u16, even if the resulting value is storedwould be stored as a u32.

We now keep computed values as usize and then generate a conversion for the appropriate type only at the callsite, instead of blindly converting to a u16 earlier on.

(this issue was pointed out in https://github.com/googlefonts/fontations/pull/1244#discussion_r1847439465; thanks @Hoolean!)


dfrg (approved):
LGTM. It’s amusing that we haven’t seemed to trigger this with crater.

cmyr:
even when permitted I still don't think it's very common to have an array with > 65536 items...

PR #1252, opened 2024-11-19
[klippa] serializer: add_link(), resolve_links() and a few more APIs

qxliu76:


klippa/src/serialize.rs
```diff
@@ -283,6 +314,126 @@ impl Serializer {
...
+    }
+
+    pub(crate) fn add_link(
+        &mut self,
+        offset_byte_range: Range<usize>,
+        obj_idx: ObjIdx,
+        whence: OffsetWhence,
+        bias: u32,
+        is_signed: bool,
+    ) -> bool {
```
cmyr, line 393: I would document the meaning of the return type.

klippa/src/serialize.rs
```diff
@@ -303,6 +454,86 @@ impl Serializer {
...
+        }
+
+        for (offset_start_pos, offset_width, offset) in offset_links.iter() {
+            self.assign_offset(*offset_start_pos, *offset_width, *offset);
+        }
+    }
+
+    //TODO: take care of signed offset
+    fn assign_offset(&mut self, offset_start_pos: usize, link_width: LinkWidth, offset: usize) {
+        let (offset_width, max_offset) = match link_width {
```
cmyr, line 508: maybe move this to a method on `LinkWidth`?

klippa/src/serialize.rs
```diff
@@ -303,6 +454,86 @@ impl Serializer {
...
+        else {
+            self.set_err(SerializeErrorFlags::SERIALIZE_ERROR_OTHER);
+            return;
+        };
+
+        let be_bytes = (offset as u32).to_be_bytes();
+        match link_width {
+            LinkWidth::Two => offset_data_bytes.copy_from_slice(&[be_bytes[2], be_bytes[3]]),
+            LinkWidth::Three => {
+                offset_data_bytes.copy_from_slice(&[be_bytes[1], be_bytes[2], be_bytes[3]])
```
cmyr, line 531: could maybe just be,
```suggestion
            LinkWidth::Two => offset_data_bytes.copy_from_slice(&be_bytes[2..=3]),
            LinkWidth::Three => {
                offset_data_bytes.copy_from_slice(&be_bytes[1..=3])
```

cmyr (approved):
looks reasonable to me!

klippa/src/serialize.rs
qxliu76: I followed what HB does, returning void. :)

klippa/src/serialize.rs
qxliu76: it updates some values in the serializer.data so I'd like to keep it as an serializer method

klippa/src/serialize.rs
```diff
@@ -283,6 +314,125 @@ impl Serializer {
...
+        if self.in_error() || self.current.is_none() {
+            return;
+        }
+
+        let pool_idx = self.current.unwrap();
+        let Some(current) = self.object_pool.get_obj_mut(pool_idx) else {
+            return;
+        };
+
+        let Some(link_width) = LinkWidth::new_checked(offset_byte_range.len()) else {
```
garretrieger, line 403: In the error case here we shouldn't silently ignore the invalid width.  If we have a caller accidentally pass in a range of an unsupported width it will seem like everything worked which will possibly lead to an unexpected bug in the produced serialization. Recommend either:
1. set the serializer into an error state.
2. add Result<(), ...> as the return type to signal the failure back to the caller.
3. or add an assert to cause a panic.

I prefer option 1 or 2.

garretrieger (approved):

PR #1255, opened 2024-11-20
[types] API additions to NameId

cmyr:
- define a LAST_RESERVED_NAME_ID const
- mark a bunch of things as #[inline]
- add is_reserved and saturating_add methods
- tweak the debug representation

### DIFF
--- a/font-types/src/name_id.rs
+++ b/font-types/src/name_id.rs
@@ -153,28 +153,63 @@ impl NameId {
     /// If present in a variable font, it may be used as the family prefix in the PostScript Name Generation
     /// for Variation Fonts algorithm.
     pub const VARIATIONS_POSTSCRIPT_NAME_PREFIX: Self = Self(25);
+
+    /// The last value that is explicitly reserved for standard names.
+    ///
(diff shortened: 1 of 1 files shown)


font-types/src/name_id.rs
```diff
@@ -153,28 +153,53 @@ impl NameId {
...
+        self.0 <= Self::LAST_RESERVED_NAME_ID.0
+    }
+
+    /// Returns a new `NameId` by adding `rhs` to the current value.
+    ///
+    /// This is a useful convenience method when constructing a name table.
+    #[inline]
+    pub const fn saturating_add(self, rhs: u16) -> Self {
+        Self(self.0.saturating_add(rhs))
+    }
```
rsheeter, line 198: Let's just accept it's fallible (however unlikely) and use checked?

rsheeter (approved):
LGTM except saturating as noted in comments

font-types/src/name_id.rs
cmyr: neat in double-checking this apparently the largest allowed NameID is 32767, so we want to check for that explicitly in any case.

PR #1256, opened 2024-11-20
[write] Use Vec instead of BTreeSet for name records

cmyr:
It is an invariant that entries in the name_records array are sorted, and to enforce this at the type level the initial design used a BTreeSet for these records.

This didn't fully help, though, because it was still possible to have a duplicate entry in the array (the same
platform/encoding/language/nameId, mapped to different strings) since the string itself is included in the Eq method of `NameRecord`.

This patch tries to be less cute: it is now the responsibility of the caller to make sure that their data is right, and we will verify this at validation time.

Ultimately this probably deserves a builder (there's one in fea-rs that should maybe get moved over here?)

cc @simoncozens; I recall you not enjoying the ergonomics of the `BTreeSet` approach, so here you go 💁

### DIFF
--- a/resources/codegen_inputs/name.rs
+++ b/resources/codegen_inputs/name.rs
@@ -17,7 +17,7 @@ table Name {
     #[count($count)]
     #[offset_data_method(string_data)]
     #[offset_adjustment(self.compute_storage_offset() as u32)]
-    #[compile_type(BTreeSet<NameRecord>)]
+    #[validate(check_sorted_and_unique_name_records)]
     name_record: [NameRecord],
     /// Number of language-tag records.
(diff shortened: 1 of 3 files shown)


rsheeter:
> since the string itself is included in the Eq method of NameRecord

I appreciate this being a problem but ... wouldn't using a *map* model this better? - (plat/enc/lang/nameid) => value

rsheeter (approved):

cmyr:
This is in generated code in `write-fonts`, where the types we're using are very close to what is in the spec. If we want to do something with a map, that would go in a `NameBuilder` or similar.

PR #1264, opened 2024-11-25
[IFT] track the index of the bit which allows a glyph keyed patch to be marked as applied.

garretrieger:


incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -100,7 +102,32 @@ pub(crate) fn apply_glyph_keyed_patches(
         }
     }
 
-    // TODO(garretrieger): mark the patch applied in the appropriate IFT table.
+    // Mark patches applied in IFT and IFTX as needed, copy the modified tables into the font builder.
+    let mut new_itf_data = font
+        .table_data(Tag::new(b"IFT "))
```
cmyr, line 107: I'd probably declare these as `const` somewhere? you use them a few times.

cmyr (approved):
looks good!

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: Agreed, I'll do that in a separate PR.

PR #1265, opened 2024-11-27
[IFT] Implement spec compliant selection order for invalidating patches

garretrieger:
When multiple candidate invalidating patches match a target subset def the client must pick a specific one to proceed with. The specification provides specific criteria for making this selection: https://w3c.github.io/IFT/Overview.html#invalidating-patch-selection

This adds an implementation of a spec compliant selection criteria. We track the size of the intersections of the subset definition sets and then use those to make the selection.

### DIFF
--- a/incremental-font-transfer/src/patch_group.rs
+++ b/incremental-font-transfer/src/patch_group.rs
@@ -1310,4 +1310,6 @@ mod tests {
             assert_eq!(ReadError::ValidationError, err);
         }
     }
+
+    // TODO(garretrieger): tests of selection order criteria for invalidating patches.
 }
--- a/incremental-font-transfer/src/patch_group.rs
+++ b/incremental-font-transfer/src/patch_group.rs
@@ -40,12 +40,16 @@ impl<'a> PatchGroup<'a> {
             });
         }
 
-        // TODO(garretrieger): following the spec disallow cases where both tables have the same compat id.
-        let compat_group = Self::select_next_patches_from_candidates(
-            candidates,
-            ift_font.ift().ok().map(|t| t.compatibility_id()),
-            ift_font.iftx().ok().map(|t| t.compatibility_id()),
-        )?;
+        let ift_compat_id = ift_font.ift().ok().map(|t| t.compatibility_id());
+        let iftx_compat_id = ift_font.iftx().ok().map(|t| t.compatibility_id());
+        if ift_compat_id == iftx_compat_id {
+            // The spec disallows two tables with same compat ids.
+            // See: https://w3c.github.io/IFT/Overview.html#extend-font-subset
+            return Err(ReadError::ValidationError);
+        }
+
+        let compat_group =
+            Self::select_next_patches_from_candidates(candidates, ift_compat_id, iftx_compat_id)?;
 
         Ok(PatchGroup {
             font: ift_font,
@@ -125,9 +129,6 @@ impl<'a> PatchGroup<'a> {
         ift_compat_id: Option<CompatibilityId>,
         iftx_compat_id: Option<CompatibilityId>,
     ) -> Result<CompatibleGroup, ReadError> {
-        // TODO(garretrieger): disallow the case where IFT and IFTX have the same compat id. Make it an error in the
-        //                     specification.
-
         // Some notes about this implementation:
         // - From candidates we need to form the largest possible group of patches which follow the selection criteria
         //   from: https://w3c.github.io/IFT/Overview.html#extend-font-subset and won't invalidate each other.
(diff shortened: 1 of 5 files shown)


incremental-font-transfer/src/patch_group.rs
```diff
@@ -142,72 +146,52 @@ impl<'a> PatchGroup<'a> {
...
+            &mut full_invalidation,
+            &mut partial_invalidation_ift,
+            &mut partial_invalidation_iftx,
+            &mut no_invalidation_ift,
+            &mut no_invalidation_iftx,
+        );
 
         // Step 2 - now make patch selections in priority order: first full invalidation, second partial, lastly none.
-        if let Some(patch) = full_invalidation.into_iter().next() {
+        if let Some(patch) = Self::select_invalidating_candidate(full_invalidation.into_iter()) {
```
cmyr, line 169: if the generic param here as `T: IntoIterator` instead of `T: Iterator` you wouldn't need the `into_iter` calls.

incremental-font-transfer/src/patch_group.rs
```diff
@@ -243,6 +227,63 @@ impl<'a> PatchGroup<'a> {
...
+    #[allow(clippy::too_many_arguments)]
+    fn group_patches(
+        candidates: Vec<PatchUri>,
+        ift_compat_id: Option<CompatibilityId>,
+        iftx_compat_id: Option<CompatibilityId>,
+        full_invalidation: &mut Vec<CandidatePatch>,
+        partial_invalidation_ift: &mut Vec<CandidatePatch>,
+        partial_invalidation_iftx: &mut Vec<CandidatePatch>,
+        no_invalidation_ift: &mut BTreeMap<String, NoInvalidationPatch>,
+        no_invalidation_iftx: &mut BTreeMap<String, NoInvalidationPatch>,
```
cmyr, line 239: are these only used to aggregate results? If so we can just create them internally and return them, and I would make a little helper struct with a name like `PatchGrouping` that had these five items as members, e.g.

```rust
struct PatchGrouping {
    full_invalidation: Vec<CandidatePatch>,
    partial_invalidation_ift: Vec<CandidatePatch>,
    partial_invalidation_iftx: Vec<CandidatePatch>,
    no_invalidation_ift: BTreeMap<String, NoInvalidationPatch>,
    no_invalidation_iftx: BTreeMap<String, NoInvalidationPatch>,
}

fn group_patches(
    candidates: Vec<PatchUri>,
    ift_compat_id: Option<CompatibilityId>,
    iftx_compat_id: Option<CompatibilityId>,
) -> PatchGrouping { .. }
```

And then if we get this far we could consider making `group_patches` into a constructor on `PatchGrouping`.

incremental-font-transfer/src/patchmap.rs
```diff
@@ -104,93 +105,132 @@ fn add_intersecting_format1_patches(
...
 
-    // Step 1: Collect the glyph map entries.
-    let mut entries = IntSet::<u16>::empty();
-    intersect_format1_glyph_map(font, map, codepoints, &mut entries)?;
-
-    // Step 2: Collect feature mappings.
-    intersect_format1_feature_map(map, features, &mut entries)?;
+    // Step 1: Collect the glyph and feature map entries.
+    let charmap = Charmap::new(font);
+    let mut entries = BTreeMap::<u16, SubsetDefinition>::new();
```
cmyr, line 110: not sure if this is a personal preference towards explicitness, but most of these type annotations you use on collections can be omitted.

cmyr: also, why is this mutable and passed in? Unless it's necessary for some other reason I would always prefer to have the function return a result, instead of passing in something mutable; in general I would encourage you to avoid `mut` if there's another reasonable alternative, I find it makes code much easier to understand (as in, while reading the code you don't need to keep track of whether or not something might possibly be getting mutated later, and the implications of this)

incremental-font-transfer/src/patchmap.rs
```diff
@@ -523,6 +602,21 @@ pub enum PatchEncoding {
 }
 
 impl PatchEncoding {
+    fn is_invalidating(&self) -> bool {
+        match self {
+            PatchEncoding::GlyphKeyed => false,
+            PatchEncoding::TableKeyed { .. } => true,
+        }
```
cmyr, line 609: can write this as, `matches!(self, PatchEncoding::TableKeyed { .. })`.

incremental-font-transfer/src/patchmap.rs
```diff
@@ -676,19 +818,31 @@ impl PatchUri {
...
             source_table,
             application_flag_bit_index,
             encoding,
+            intersection_info,
+        }
+    }
+}
+
+impl IntersectionInfo {
+    fn from(value: SubsetDefinition, order: usize) -> Self {
```
cmyr, line 835: I personally wouldn't use `from` as a fn name (outside of the `From` trait) and I would just call this `new`, if that isn't taken? (or something like `for_subset_definition`, if it is)

cmyr (approved):
looks reasonable!

PR #1268, opened 2024-11-28
[IFT] track design space intersection size in IntersectionInfo.

garretrieger:


incremental-font-transfer/src/patchmap.rs
```diff
@@ -11,11 +11,13 @@ use std::io::Cursor;
...
 
+use font_types::Fixed;
 use font_types::Tag;
 
 use data_encoding::BASE64URL;
 use data_encoding_macro::new_encoding;
 
+use read_fonts::collections::RangeSet;
 use read_fonts::FontRead;
 use read_fonts::{
```
cmyr, line 22: once again the auto importing hurts me

cmyr (approved):
looks good.

I'm curious if you're enforcing the invariant that start <= end, for your ranges? And/or have some test coverage for this case?

garretrieger:
> looks good.
> 
> I'm curious if you're enforcing the invariant that start <= end, for your ranges? And/or have some test coverage for this case?

Yeah, here: https://github.com/googlefonts/fontations/blob/main/incremental-font-transfer/src/patchmap.rs#L437

Also internally RangeSet ignores attempts to add ranges with start > end so it's not possible to construct one that has bad ranges.

garretrieger:
> > looks good.
> > I'm curious if you're enforcing the invariant that start <= end, for your ranges? And/or have some test coverage for this case?
> 
> Yeah, here: https://github.com/googlefonts/fontations/blob/main/incremental-font-transfer/src/patchmap.rs#L437
> 
> Also internally RangeSet ignores attempts to add ranges with start > end so it's not possible to construct one that has bad ranges.

and tests here: https://github.com/googlefonts/fontations/blob/main/incremental-font-transfer/src/patchmap.rs#L1998

PR #1269, opened 2024-11-29
MATH table support

simoncozens:
This is an attempt to implement support for the MATH table but I'm coming unstuck with the codegen.

@cmyr, is it possible that the codegen assumes that every `record` is part of an array? It's doing bad things with `MathValueRecord`, which seems like a simple enough structure except that it is directly embedded into various tables (`MathConstants` and `GlyphAssembly`). These are giving errors. Where the `MathValueRecord` *is* part of an array (`MathKern`, for example), then all is well.


cmyr:
codegen doesn't exactly expect that records are always part of arrays, but the records that aren't are special cases... I'll take a look.

simoncozens:
I'm still waiting for Colin to come back to me after he's had a look...

dfrg:
We should support embedded records now since #2077

PR #1283, opened 2024-12-10
[skrifa] empty glyphs have phantom points too

dfrg:
Fixes a small oversight: we explicitly didn't account for phantom point memory for empty glyphs. Just removes the explicit check and adds a test to ensure it works.

JMM


cmyr (approved):
I'm still regularly confused by the presence of phantom points in debug output...

dfrg:
> I'm still regularly confused by the presence of phantom points in debug output...

I still find them surprising sometimes

PR #1284, opened 2024-12-10
[skrifa] tthint: fix overflow in flip region instruction

dfrg:
first commit has failing test

ref https://issues.oss-fuzz.com/issues/377736138


cmyr (approved):
👍

PR #1285, opened 2024-12-10
[read-fonts] var: fix overflow in packed point numbers

dfrg:
first commit contains failing test

ref https://issues.oss-fuzz.com/issues/378159154


cmyr (approved):
👍

PR #1286, opened 2024-12-11
[skrifa] tthint: overflow in move_index stack op

dfrg:
ref https://oss-fuzz.com/testcase-detail/5765856825507840 which wasn't filed as an issue yet due to similar failing test case


cmyr (approved):
wow being careful is exhausting huh

PR #1292, opened 2024-12-17
[klippa] subset cmap table

qxliu76:


klippa/src/cmap.rs
```diff
@@ -0,0 +1,1105 @@
...
+                .map(|(cp, gid)| (*cp, *gid))
+                .collect();
+            format12_objidx =
+                serialize_encoding_record(record, &subtable, s, &cp_to_new_gid_list, plan)?;
+        } else if format == 14 {
+            serialize_encoding_record(record, &subtable, s, &plan.unicode_to_new_gid_list, plan)?;
+        }
+    }
+
+    let num_retained_records = (s.length() - 4) / EncodingRecord::RAW_BYTE_LEN;
```
cmyr, line 229: I assume `4` is the header length? Might be worth a `const`, I think you use it elsewhere too and a const with a helpful name is more robust documentation than a comment.

klippa/src/cmap.rs
```diff
@@ -0,0 +1,1105 @@
...
+}
+
+fn is_gid_consecutive(
+    end_char_code: u32,
+    start_char_code: u32,
+    gid: GlyphId,
+    cp: u32,
+    new_gid: GlyphId,
+) -> bool {
+    cp - 1 == end_char_code && new_gid.to_u32() == gid.to_u32() + (cp - start_char_code)
```
cmyr, line 289: do we know this will never be called with `cp == 0`?

klippa/src/cmap.rs
```diff
@@ -0,0 +1,1105 @@
...
+    }
+}
+
+impl UnicodeRangeWriter for RangeCounter {
+    fn write_one_range(&mut self, _start: u16, _end: u16, _delta: i16, _s: &mut Serializer) {
+        self.seg_count += 1;
+    }
+}
+
+fn to_ranges(
```
cmyr, line 415: if this is ported from something in harfbuzz can we include a link? (and if not, it deserves some documentation, it's a non-trivial fn)

klippa/src/lib.rs
```diff
@@ -445,8 +504,11 @@ fn remove_invalid_gids(gids: &mut IntSet<GlyphId>, num_glyphs: usize) {
 }
 
 fn get_font_num_glyphs(font: &FontRef) -> usize {
-    let loca = font.loca(None).expect("Error reading loca table");
-    let ret = loca.len();
+    let ret = if let Ok(loca) = font.loca(None) {
+        loca.len()
+    } else {
+        0
+    };
```
cmyr, line 511: or just `let ret = font.loca(None).map(|loca| loca.len().unwrap_or_default();`

cmyr (approved):
looks reasonable!

klippa/src/cmap.rs
qxliu76: yes I think so

klippa/src/cmap.rs
```diff
@@ -0,0 +1,1105 @@
...
+        for (_, record) in retained_encoding_records.iter() {
+            if record
+                .subtable(self.offset_data())
+                .is_ok_and(|t| t.format() == 12)
+            {
+                has_format12 = true;
+            }
+
+            if record.platform_id() == PlatformId::Unicode && record.encoding_id() == 3 {
+                has_unicode_bmp = true;
```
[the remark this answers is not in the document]
qxliu76: thanks!

garretrieger (approved):

PR #1293, opened 2024-12-19
[read-fonts] hdmx: use explicitly defined record size

dfrg:
We were not accounting for the 32-bit alignment requirement.

Also updated the test to account for this.


qxliu76:
LGTM, Thanks!

PR #1294, opened 2024-12-20
[read-fonts] revert midpoint change to match FT

dfrg:
A recent PR to fix an overflow changed the midpoint calculation from `(a + b) / 2` to the "equivalent" `a + (b - a) / 2` to attempt to avoid overflow with large values but this ends up producing results that don't exactly match FreeType in some cases.

This just changes the expression to `a.wrapping_add(b) / 2` to restore compatibility while still avoiding the panic on overflow.

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -803,13 +803,14 @@ impl PointCoord for i32 {
 }
 
 // Midpoint function that avoids overflow on large values.
+#[inline(always)]
 fn midpoint_i32(a: i32, b: i32) -> i32 {
     // Original overflowing code was: (a + b) / 2
     // Choose wrapping arithmetic here because we shouldn't ever
     // hit this outside of fuzzing or broken fonts _and_ this is
     // called from the outline to path conversion code which is
     // very performance sensitive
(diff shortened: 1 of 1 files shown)


drott (approved):
Change LGTM, but please modify commit message:

> to the "equivalent" a + (b - a) / 2 to

Please explicitly point out why the formula change was not equivalent, instead of leaving that to the reader interpreting the quotes.

dfrg:
Will add:

"The results differ when `a > b` and the sum is not even. For example, when a = 7 and b = 4, the first expression is `(7 + 4) / 2` which yields 5 (and matches FreeType) while the second is `7 + (4 - 7) / 2` which yields 6."

drott:
> A recent PR to fix an overflow ...

This was e7161e71bd20b1fcb12f92a4d8d44a797ca31cd2 in 0.26.2 - which I was starting to roll in Chromium but eventually did not land yet. https://chromium-review.googlesource.com/c/chromium/src/+/6088164 - it would be good to have this fix in a release, together with the fix for #1295 - then we need to update the roll into Chromium. I hope I get to it before the branch point (even though I am OOO).

PR #1296, opened 2024-12-20
[skrifa] autohint: cycle detection for GSUB lookups

### LINKED ISSUE #1295: Suspected endless loop in autohinting Shaper coverage computation with Poppins Semibold
In Chrome `content_shell` and https://rtsh.es/test/poppins.html the content tab starts to hang. 
Running this and content_shell with gdb
$ gdb --args out/gnrelease/content_shell --single-process https://rtsh.es/test/poppins.html
Then stopping and running
`thread apply all backtrace`

reveals the stack trace at which this is hanging:

```
[#65 repeats]
#65 0x00007ffff5913703 in skrifa::outline::autohint::shape::GsubHandler::process_lookup::h05a71b4249d913f6 () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#66 0x00007ffff5910d39 in skrifa::outline::autohint::shape::Shaper::compute_coverage::h8b6af0d2971cd068 () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#67 0x00007ffff58fe6fd in skrifa::outline::autohint::style::GlyphStyleMap::new::he7232822c79b7d06 () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#68 0x00007ffff590021b in skrifa::outline::autohint::instance::GlyphStyles::new::h8993ddbc473295b4 () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#69 0x00007ffff5900407 in skrifa::outline::autohint::instance::Instance::new::h1fd5544fce0fc93c () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#70 0x00007ffff587e861 in skrifa::outline::hint::HintingInstance::reconfigure::hff00a995aaa3c43c () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#71 0x00007ffff587efe2 in skrifa::outline::hint::HintingInstance::new::h8760a010858fb68a () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#72 0x00007ffff5866eaf in skia_cbridge_urust_uside::make_hinting_instance::h8f1b0d2f10919329 () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#73 0x00007ffff57b1c94 in fontations_ffi::make_hinting_instance(fontations_ffi::BridgeOutlineCollection const&, float, fontations_ffi::BridgeNormalizedCoords const&, bool, bool, bool, bool) () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
#74 0x00007ffff53492d7 in SkFontationsScalerContext::SkFontationsScalerContext(sk_sp<SkTypeface_Fontations>, SkScalerContextEffects const&, SkDescriptor const*, sk_sp<SkTypeface>) () at /usr/local/google/home/drott/dev/blink/src/out/gnrelease/libskia.so
```

dfrg:
Adds a stack to detect cycles in GSUB lookups while computing shaper coverage for autohinting.

Fixes #1295

### DIFF
--- a/skrifa/src/outline/autohint/shape.rs
+++ b/skrifa/src/outline/autohint/shape.rs
@@ -700,7 +700,7 @@ mod tests {
         let shaper = Shaper::new(&font, ShaperMode::BestEffort);
         let lookup_list = shaper.gsub.as_ref().unwrap().lookup_list().unwrap();
         let style = &style::STYLE_CLASSES[style::StyleClass::LATN];
-        let mut gsub_handler = GsubHandler::new(&shaper.charmap, &lookup_list, &style, &mut []);
+        let mut gsub_handler = GsubHandler::new(&shaper.charmap, &lookup_list, style, &mut []);
         for ix in 0..MAX_NESTING_DEPTH {
             // all okay up to max depth...
             assert!(gsub_handler.enter_lookup(ix as u16));
@@ -715,7 +715,7 @@ mod tests {
         let shaper = Shaper::new(&font, ShaperMode::BestEffort);
         let lookup_list = shaper.gsub.as_ref().unwrap().lookup_list().unwrap();
         let style = &style::STYLE_CLASSES[style::StyleClass::LATN];
-        let mut gsub_handler = GsubHandler::new(&shaper.charmap, &lookup_list, &style, &mut []);
+        let mut gsub_handler = GsubHandler::new(&shaper.charmap, &lookup_list, style, &mut []);
         for ix in 0..10 {
             // add 0 through 9 to the stack, all succeeding...
             assert!(gsub_handler.enter_lookup(ix as u16));
--- a/skrifa/src/outline/autohint/shape.rs
+++ b/skrifa/src/outline/autohint/shape.rs
@@ -18,7 +18,7 @@ use raw::{
 
 // To prevent infinite recursion in contextual lookups. Matches HB
 // <https://github.com/harfbuzz/harfbuzz/blob/c7ef6a2ed58ae8ec108ee0962bef46f42c73a60c/src/hb-limits.hh#L53>
-const MAX_NESTING_DEPTH: u32 = 64;
+const MAX_NESTING_DEPTH: usize = 64;
 
 /// Determines the fidelity with which we apply shaping in the
 /// autohinter.
@@ -223,7 +223,7 @@ impl<'a> Shaper<'a> {
                     };
                     // And now process associated lookups
                     for index in feature.lookup_list_indices().iter() {
-                        gsub_handler.process_lookup(index.get(), 0);
+                        gsub_handler.process_lookup(index.get());
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


skrifa/src/outline/autohint/shape.rs
```diff
@@ -592,6 +589,29 @@ impl<'a> GsubHandler<'a> {
...
+        }
+        self.lookup_stack[self.lookup_depth] = lookup_index;
+        self.lookup_depth += 1;
+        true
+    }
+
+    /// Called when we finish processing a single lookup.
+    fn exit_lookup(&mut self) {
+        self.lookup_depth = self.lookup_depth.saturating_sub(1);
+    }
```
rsheeter, line 613: I don't love that this makes it relatively easy to add code that enters without calling exit. This would be somewhat harder if it was directly in the process lookup fn? - depicted below. Or by a guard that reduces depth on drop but I suppose that path takes you towards multiple mut refs.

```rust
    fn process_lookup(&mut self, lookup_index: u16) {
        // Guard: don't cycle and don't go exceed depth limit
        // Note: we use a linear search here under the assumption that
        // most fonts have shallow contextual lookup chains
        if self.lookup_depth != 0
            && (self.lookup_depth == MAX_NESTING_DEPTH
                || self.lookup_stack[..self.lookup_depth].contains(&lookup_index))
        {
            return;
        }
        self.lookup_stack[self.lookup_depth] = lookup_index;
        self.lookup_depth += 1;

        // Actually process the lookup
        self.process_lookup_inner(lookup_index);

        // Out we go again
        self.lookup_depth -= 1;
    }
```

skrifa/src/outline/autohint/shape.rs
```diff
@@ -411,13 +414,19 @@ impl<'a> GsubHandler<'a> {
...
-        if nesting_depth > MAX_NESTING_DEPTH {
-            return;
+    fn process_lookup(&mut self, lookup_index: u16) {
+        if self.enter_lookup(lookup_index) {
+            self.process_lookup_inner(lookup_index);
+            self.exit_lookup();
         }
+    }
+
+    fn process_lookup_inner(&mut self, lookup_index: u16) {
```
rsheeter, line 429: might as well mark inline always?

rsheeter (approved):

skrifa/src/outline/autohint/shape.rs
dfrg: I agree with the sentiment but I kept the stack ops separate to support isolated unit testing of that functionality. If we embed them then we need to build a lookup list with the desired structure which seems less clear and more fragile to me.

skrifa/src/outline/autohint/shape.rs
dfrg: sgtm

dfrg:
For reference, contextual lookup depth in all of Google Fonts and Noto (10604 font files):

| depth | # top level lookups |
|--------|----------------------|
| 2 | 76338 |
| 3 |  933 |
| 4 | 14 |
| 5 | 14 |
| 6 | 42 |

skrifa/src/outline/autohint/shape.rs
rsheeter: I would have thought we need to build lookups with good/bad structure (one that cycles, one that has max depth, one that has max depth + 1, etc) to unit test regardless so... that's kind of fine?

Issue #1280, opened 2024-12-09
Fuzzer aborts/panics in IFT implementation


drott:
See
https://g-issues.oss-fuzz.com/issues/382721852
and 
https://g-issues.oss-fuzz.com/issues/382927070

@garretrieger, could you please fix those? Thanks.

drott:
> Unfortunately it looks like the crate doesn't implement proper error handling. I'm leaning towards writing our own uri template expansion implementation instead.

SGTM, also in the interest of reducing dependencies.

garretrieger:
Fuzzer issues have been marked as fixed.

PR #1299, opened 2025-01-02
[IFT] switch to uri_template_system crate for URI template expansion.

garretrieger:
The uri-template crate panics when encountering errors so switch to the uri-template-system crate which has error handling built into the API. Fixes fuzzer crashes from #1280.

### DIFF
--- a/incremental-font-transfer/src/patch_group.rs
+++ b/incremental-font-transfer/src/patch_group.rs
@@ -10,7 +10,7 @@ use crate::{
     font_patch::{IncrementalFontPatchBase, PatchingError},
     patchmap::{
         intersecting_patches, IftTableTag, IntersectionInfo, PatchFormat, PatchUri,
-        SubsetDefinition,
+        SubsetDefinition, UriTemplateError,
     },
 };
 
@@ -153,7 +153,8 @@ impl PatchGroup<'_> {
             partial_invalidation_iftx,
             mut no_invalidation_ift,
             mut no_invalidation_iftx,
-        } = GroupingByInvalidation::group_patches(candidates, ift_compat_id, iftx_compat_id);
+        } = GroupingByInvalidation::group_patches(candidates, ift_compat_id, iftx_compat_id)
+            .map_err(|_| ReadError::MalformedData("Malformed URI templates."))?;
 
         // Step 2 - now make patch selections in priority order: first full invalidation, second partial, lastly none.
         if let Some(patch) = Self::select_invalidating_candidate(full_invalidation) {
@@ -306,7 +307,7 @@ impl GroupingByInvalidation {
         candidates: Vec<PatchUri>,
         ift_compat_id: Option<CompatibilityId>,
         iftx_compat_id: Option<CompatibilityId>,
-    ) -> GroupingByInvalidation {
+    ) -> Result<GroupingByInvalidation, UriTemplateError> {
         let mut result: GroupingByInvalidation = Default::default();
 
         for uri in candidates.into_iter() {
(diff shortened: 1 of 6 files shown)


incremental-font-transfer/src/patch_group.rs
```diff
@@ -153,7 +153,8 @@ impl PatchGroup<'_> {
             partial_invalidation_iftx,
             mut no_invalidation_ift,
             mut no_invalidation_iftx,
-        } = GroupingByInvalidation::group_patches(candidates, ift_compat_id, iftx_compat_id);
+        } = GroupingByInvalidation::group_patches(candidates, ift_compat_id, iftx_compat_id)
+            .map_err(|_| ReadError::MalformedData("Malformed URI templates."))?;
```
cmyr, line 157: we have an inner error here that should provide more detail, why not preserve that in a `BadUri` error variant?

incremental-font-transfer/src/patchmap.rs
```diff
@@ -734,12 +734,26 @@ impl Ord for IntersectionInfo {
     }
 }
 
+/// Indicates a malformed URI template was encountered.
+///
+/// More info: <https://datatracker.ietf.org/doc/html/rfc6570#section-3>
+#[derive(Debug, PartialEq, Eq)]
+pub struct UriTemplateError;
```
cmyr, line 741: what error does the new crate that we're using return? do we need our own error type?

incremental-font-transfer/src/patchmap.rs
```diff
@@ -734,12 +734,26 @@ impl Ord for IntersectionInfo {
...
+
+impl std::error::Error for UriTemplateError {}
+
 impl PatchUri {
     const BASE32HEX_NO_PADDING: data_encoding::Encoding = new_encoding! {
         symbols: "0123456789ABCDEFGHIJKLMNOPQRSTUV",
     };
 
-    pub fn uri_string(&self) -> String {
+    pub fn uri_string(&self) -> Result<String, UriTemplateError> {
```
cmyr, line 756: would it make sense to do our error checking at construction time?

cmyr (commented):
overall makes sense but I have some questions about the API...

incremental-font-transfer/src/patchmap.rs
garretrieger: I want to avoid exposing details of the internal crate used in our api.

incremental-font-transfer/src/patchmap.rs
garretrieger: We can't determine errors until the template is parsed and for performance reasons we want to avoid parsing until we actually need a concrete URL (since only a small number of the patch URI's will be instantiated in typical cases).

If we switch to sharing the template between multiple URI's then we could pre parse, but that will require larger changes to how things are setup.

incremental-font-transfer/src/patch_group.rs
garretrieger: I'm not sure I want to add a ReadError type that's specific to IFT since ReadError is more general than IFT. MalformedData does capture the essence of the error and in this case the inner type doesn't have any additional information that the error string doesn't already capture.

drott:
In relation to your earlier comment on the issue? Do we benefit a lot from the dependency? What is the main functionality we use and could we do without the dependency, what's the tradeoff?

cmyr (approved):
I'm satisfied, although there's an outstanding Q from dominik.

garretrieger:
> In relation to your earlier comment on the issue? Do we benefit a lot from the dependency? What is the main functionality we use and could we do without the dependency, what's the tradeoff?

This dep is an implementation of https://datatracker.ietf.org/doc/html/rfc6570 which is needed for the IFT mechanism. Writing our own would be a sizable task so I decided to stick with an existing crate in order to get this fuzzer issue sorted out in a timely fashion. I haven't ruled out replacing it with our own at some future point if we're looking to trim off dependencies.

drott:
> This dep is an implementation of https://datatracker.ietf.org/doc/html/rfc6570 which is needed for the IFT mechanism. Writing our own would be a sizable task so I decided to stick with an existing crate in order to get this fuzzer issue sorted out in a timely fashion.

Thanks for the clarifications. Which parts, as in: replacement rule/format, of the RFC are we using and expecting to use in the future? Is the IFT spec allowing multiple types of replacements, or only few of them? 

Ok with me to land this for now, but I would like to understand if we can strip a dependency.

garretrieger:
> > This dep is an implementation of https://datatracker.ietf.org/doc/html/rfc6570 which is needed for the IFT mechanism. Writing our own would be a sizable task so I decided to stick with an existing crate in order to get this fuzzer issue sorted out in a timely fashion.
> 
> Thanks for the clarifications. Which parts, as in: replacement rule/format, of the RFC are we using and expecting to use in the future? Is the IFT spec allowing multiple types of replacements, or only few of them?
> 
> Ok with me to land this for now, but I would like to understand if we can strip a dependency.

We are using nearly the entirety of that spec. Specifically in the IFT client we do expansion of templates utilizing the syntax from the RFC and as the IFT spec is currently written all three operator levels are supported. The only part we're not using is expansion of [composite values](https://datatracker.ietf.org/doc/html/rfc6570#section-2.4.2). For reference [here's](https://w3c.github.io/IFT/Overview.html#uri-templates) the relevant section from IFT.

drott:
> We are using nearly the entirety of that spec. Specifically in the IFT client we do expansion of templates utilizing the syntax from the RFC and as the IFT spec is currently written all four operator levels are supported. The only part we're not using is expansion of [composite values](https://datatracker.ietf.org/doc/html/rfc6570#section-2.4.2). For reference [here's](https://w3c.github.io/IFT/Overview.html#uri-templates) the relevant section from IFT.

Makes sense, thanks.

PR #1300, opened 2025-01-06
[read-fonts] avoid overflow when computing glyph data range

dfrg:
First commit contains failing test.

ref https://g-issues.oss-fuzz.com/issues/385918147


cmyr (approved):
another year, another overflow 🥲

PR #1302, opened 2025-01-07
[IFT] Add 'ift_graph' cli utility

garretrieger:
This utility will be used in testing of encoder implementations. It generates a graph representation of an IFT font and associated patches.


incremental-font-transfer/Cargo.toml
```diff
@@ -43,3 +43,8 @@ path = "src/lib.rs"
 name = "ift_extend"
 path = "src/ift_extend.rs"
 required-features = ["cli"]
+
+[[bin]]
+name = "ift_graph"
+path = "src/ift_graph.rs"
```
cmyr, line 49: missed this the first time, but by convention bins live in src/bin, in which case you don't need to specify the path here: https://doc.rust-lang.org/1.82.0/cargo/reference/cargo-targets.html#binaries

incremental-font-transfer/src/ift_graph.rs
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    Ok(result)
+}
+
+fn get_node_name(font: &FontRef<'_>) -> Result<String, ReadError> {
+    let chars: BTreeSet<char> = font
+        .charmap()
+        .mappings()
+        .map(|(cp, _)| char::from_u32(cp).unwrap())
```
cmyr, line 158: do we care if the font has invalid data here?

incremental-font-transfer/src/ift_graph.rs
```diff
@@ -0,0 +1,261 @@
...
+        let axis_strings: Vec<_> = font
+            .axes()
+            .iter()
+            .map(|axis| {
+                axis.tag().to_string()
+                    + "["
+                    + &axis.min_value().to_string()
+                    + ".."
+                    + &axis.max_value().to_string()
+                    + "]"
```
cmyr, line 182: I... didn't even know there was an `Add` impl for String. That said, I think it would be more idiomatic (and presumably more efficient) to just use the `format!` macro here.

incremental-font-transfer/src/ift_graph.rs
```diff
@@ -0,0 +1,261 @@
...
+
+fn to_next_font(
+    base_path: &Path,
+    font: &FontRef<'_>,
+    patch_uri: PatchUri,
+) -> Result<Vec<u8>, UriTemplateError> {
+    let path = base_path.join(patch_uri.uri_string()?);
+    let patch_bytes = std::fs::read(path.clone())
+        .unwrap_or_else(|e| panic!("Unable to read patch file ({}): {:?}", path.display(), e));
+
```
cmyr, line 201: if you want to panic, I think it would be cleaner to do,

```rust
let Ok(patch_bytes) = std::fs::read(&path) else {
    panic!("oh no")
};
```

(also no need to clone path, `read` can take a reference?)

incremental-font-transfer/src/ift_graph.rs
```diff
@@ -0,0 +1,261 @@
...
+    let node_name = get_node_name(&font).unwrap();
+    graph.entry(node_name.clone()).or_default();
+
+    for patch in patches {
+        match patch.encoding() {
+            PatchFormat::TableKeyed { .. } => {}
+            // This graph only considers invalidating (that is table keyed patches), so skip all other types.
+            _ => continue,
+        };
```
cmyr, line 225: I'd probably write this as `if !matches!(patch.encoding(), PatchFormat::TableKeyed { .. }) { .. }` but this is totally fine 🤷

cmyr (approved):
looks good!

incremental-font-transfer/src/ift_graph.rs
garretrieger: No, this utility is really only meant for use in testing. Just panicking and failing in the face of a bad font is fine.

incremental-font-transfer/src/ift_graph.rs
garretrieger: I need to get a copy of the error object so can't use .. else {}.

Issue #1303, opened 2025-01-07
Incorrect contour extraction and rendering of DFKai-SB glyphs


drott:
https://rtsh.es/test/dfkai_pdf.html hosts a `PDF.js` view of a PDF that displays glyphs from Microsofts's *DFKai-SB* font 
https://learn.microsoft.com/en-us/typography/font-list/dfkai-sb

Two of the glyphs appear compressed with incomplete contour shapes. 

<img width="730" alt="Image" src="https://github.com/user-attachments/assets/a36358c5-6682-4763-8473-c2f9fe231b56" />

vs expected:

<img width="546" alt="Image" src="https://github.com/user-attachments/assets/aaf638e8-1ff4-4f75-b2cf-a3acfb7a6a01" />

I'll try to provide a further reduction.

The font is available as an attachment of the upstream bug report: https://issues.chromium.org/387045808

drott:
> The font is available as an attachment of the upstream bug report: https://issues.chromium.org/387045808

Sorry, the font is not, only the PDF that embeds it. I am working on extracting it.

drott:
According to https://learn.microsoft.com/en-us/typography/fonts/windows_11_font_list#traditional-chinese and https://learn.microsoft.com/en-us/typography/font-list/dfkai-sb#products-that-supply-this-font 5.01 seems to be the latest version of this font.

dfrg:
So it turns out that this is one of the "tricky" fonts that requires the bytecode interpreter to generate a correct outline. If I run these through skrifa with the target set to `Mono`, I get the intended results.

Since our API doesn't handle hinting internally, this isn't something we can just fix. What we can do is copy FreeType's tricky font detection and expose it as a method (maybe `OutlineGlyphCollection::required_hinting_options() -> Option<HintingOptions>`). If that returns `Some` then the user should apply hinting with those options. This method won't be fast since it requires checking the name table and potentially computing checksums so for Skia, this should be done when creating a typeface.

drott:
Fixed in https://issues.skia.org/issues/388445690

PR #1305, opened 2025-01-07
[skrifa] detect "tricky" fonts by family name

dfrg:
Implements the family name part of FreeType's "tricky" font detection. This is sufficient to address the problematic font in #1303.

Adds an `OutlineGlyphCollection::require_interpreter()` method that can be used to determine if hinting (with the interpreter engine) is required to correctly render glyphs from the font.

FreeType has some additional checks on length/checksums of various tables. I'll do that in a follow up.

### DIFF
--- a/skrifa/src/outline/mod.rs
+++ b/skrifa/src/outline/mod.rs
@@ -84,6 +84,7 @@ mod glyf;
 mod hint;
 mod metrics;
 mod path;
+mod tricky;
 mod unscaled;
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


skrifa/src/outline/tricky.rs
```diff
@@ -0,0 +1,140 @@
+//! Match FreeType's "tricky" font detection.
```
drott, line 1: Suggest to rephrase to "Name detection for fonts that require hinting to be run for correct contours (FreeType calls these "tricky" fonts."

We could also call it something like hinting-reliant or so, if we need a single word.

skrifa/src/outline/mod.rs
```diff
@@ -608,6 +609,27 @@ impl<'a> OutlineGlyphCollection<'a> {
...
+    ///
+    /// When this returns `true`, you should construct a [`HintingInstance`]
+    /// with [`HintingOptions::engine`] set to [`Engine::Interpreter`] and
+    /// [`HintingOptions::target`] set to [`Target::Mono`].
+    ///
+    /// # Performance
+    /// This digs through the name table and potentially computes checksums
+    /// so it may be slow. You should cache the result of this function if
+    /// possible.
+    pub fn require_interpreter(&self) -> bool {
```
drott, line 627: This name works well. Optional: If we go with `hinting-reliant` elsewhere, we could call it that here, too, i.e. `is_hinting_reliant()`

skrifa/src/outline/tricky.rs
```diff
@@ -0,0 +1,140 @@
...
+//!
+//! Tricky fonts are those that have busted outlines and require the bytecode
+//! interpreter to produce something that makes sense.
+
+use crate::{string::StringId, FontRef, MetadataProvider};
+
+pub(super) fn is_tricky(font: &FontRef) -> bool {
```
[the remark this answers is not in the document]
dfrg: Makes sense. I renamed this top level function to `require_interpreter` to match the public API.

skrifa/src/outline/tricky.rs
dfrg: I like this comment so I used it verbatim.

skrifa/src/outline/tricky.rs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            is_tricky_name(core::str::from_utf8(&buf[..len]).unwrap_or_default())
+        })
+        .unwrap_or_default()
+}
+
+/// Does this family name belong to a "tricky" font?
+///
+/// <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/truetype/ttobjs.c#L174>
+fn is_tricky_name(name: &str) -> bool {
```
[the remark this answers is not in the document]
dfrg: Changed.

skrifa/src/outline/mod.rs
drott: Thanks for making these changes.

PR #1308, opened 2025-01-08
Add gasp to write-fonts

rsheeter:
fontc needs it for sources that use https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/#opentype-gasp-table-fields, like the Wix families.


write-fonts/src/tables/gasp.rs
```diff
@@ -0,0 +1,43 @@
+//! The gasp table
+
+include!("../../generated/generated_gasp.rs");
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn gasping() {
```
cmyr, line 10: +1

cmyr (approved):
more tables like this, please!

PR #1316, opened 2025-01-10
[IFT] Update ift_extend to support design space

garretrieger:
Adds --design-space flag to ift_extend to specify design space as part of the target subset definition.

Also fix a bug in glyph keyed patch application on long loca tables.

### DIFF
--- a/font-test-data/src/ift.rs
+++ b/font-test-data/src/ift.rs
@@ -4,6 +4,7 @@
 //! <https://w3c.github.io/IFT/Overview.html>
 
 use std::collections::HashMap;
+use std::iter;
 
 use read_fonts::types::Tag;
 use read_fonts::{be_buffer, be_buffer_add, test_helpers::BeBuffer, types::Int24, types::Uint24};
@@ -591,6 +592,7 @@ pub fn noop_table_keyed_patch() -> BeBuffer {
 }
 
 pub fn test_font_for_patching_with_loca_mod<F>(
+    short_loca: bool,
     loca_mod: F,
     additional_tables: HashMap<Tag, &[u8]>,
 ) -> Vec<u8>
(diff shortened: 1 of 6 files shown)


font-test-data/src/ift.rs
```diff
@@ -4,6 +4,7 @@
 //! <https://w3c.github.io/IFT/Overview.html>
 
 use std::collections::HashMap;
+use std::iter::repeat;
```
cmyr, line 7: very much a nit, but a rust idiom I've gotten kind of into is not to import free fns directly but to import the containing module, and then call the fn through that, e.g. do `iter::repeat(0)` instead of just `repeat(0)`; I find it makes for easier reading?

anyway not a blocker obviously, just a note..

incremental-font-transfer/src/bin/ift_extend.rs
```diff
@@ -84,3 +107,102 @@ fn main() {
...
+    f64::from_str(value)
+        .map_err(|_| ParsingError::DesignSpaceParsingFailed {
+            flag_value: flag_value.to_string(),
+            message: "Bad axis position value".to_string(),
+        })
+        .map(Fixed::from_f64)
+}
+
+fn parse_design_space(args: Vec<String>) -> Result<DesignSpace, ParsingError> {
+    let re = Regex::new(r"^([a-zA-Z][a-zA-Z0-9 ]{3})@([0-9.]+)(:[0-9.]+)?$").unwrap();
```
cmyr, line 140: I wouldn't personally use a regex for this, the parsing seems straightforward if you just split at `@`? regex is fairly heavyweight.

cmyr (approved):
A few observations, overall seems reasonable. I'd have preferred separate patches for the bugfix and the new feature code, unless they're related in some way that isn't clear? But not worth splitting apart now.

incremental-font-transfer/src/bin/ift_extend.rs
garretrieger: Agreed it's typically heavy weight; however, in this case performance isn't much of a concern as this is a command line utility for testing/debugging.

PR #1317, opened 2025-01-13
[IFT] Add support for feature tags to ift_extend.

garretrieger:


incremental-font-transfer/src/bin/ift_extend.rs
```diff
@@ -127,6 +138,30 @@ fn parse_unicodes(args: Vec<String>, codepoints: &mut IntSet<u32>) -> Result<(),
...
+
+        if tag_string.len() != 4 {
+            return Err(ParsingError::FeatureTagParsingFailed(tag_string));
+        }
+
+        let Ok(tag) = Tag::new_checked(&tag_string.as_bytes()[0..4]) else {
+            return Err(ParsingError::FeatureTagParsingFailed(tag_string));
+        };
+        tags.insert(tag);
+    }
```
cmyr, line 160: this is fine but also a place where rust pattern matching might help you clean it up a little bit? Specifically I could imagine something like,

```rust
match tag_string.as_str() {
    "" => continue,
   "*" => return Ok(FeatureSet::All),
   tag => Tag::new_checked(tag.as_bytes()).map_err(|_| ParsingError::FeatureTagParsingFailed(tag_string)),
}
```

also I'm not bothering with the length check since `Tag::new_checked` will error if it's wrong, although it would accept _less_ than 4 bytes, adding padding as needed. If you want to be strict about length you could have an extra branch `tag if tag.len() != 4 => Err(ParsingError::FeatureTagParsingFailed(tag_string)),`.

incremental-font-transfer/src/bin/ift_extend.rs
```diff
@@ -201,6 +237,9 @@ impl std::fmt::Display for ParsingError {
             ParsingError::UnicodeCodepointParsingFailed(value) => {
                 write!(f, "Invalid unicode code point value: {}", value,)
             }
+            ParsingError::FeatureTagParsingFailed(value) => {
+                write!(f, "Invalid feature tag value: {}", value,)
```
cmyr, line 241: Or just,

```suggestion
                write!(f, "Invalid feature tag value: {value}")
```

cmyr (approved):
looks good, some style notes inline but nothing blocking.

PR #1318, opened 2025-01-15
[IFT] Add support for applying glyph keyed patches to gvar tables.

garretrieger:
Glyph keyed format: https://w3c.github.io/IFT/Overview.html#glyph-keyed

Due to the nature of gvar, complex serialization with offset calculation was needed so I made the klippa Serializer public and used it here as it is designed specifically to handle this type of situation.

Also added a method to read_fonts::Gvar to provide more direct access to underlying bytes.

### DIFF
--- a/incremental-font-transfer/src/glyph_keyed.rs
+++ b/incremental-font-transfer/src/glyph_keyed.rs
@@ -586,14 +586,16 @@ impl GlyphDataOffsetArray for Gvar<'_> {
         // typical gvar layout (see: https://learn.microsoft.com/en-us/typography/opentype/spec/gvar):
         //   Part 1 - Header
         //   Part 2 - glyphVariationDataOffsets[glyphCount + 1]
-        //   Part 3 - Shared Tuples + Glyph Variation Data Header
-        //   Part 4 - per glyph variation data
+        //   Part 3 - Shared Tuples
+        //   Part 4 - Array of per glyph variation data
         //
         // When constructing a new gvar from the newly synthesized data we'll be replacing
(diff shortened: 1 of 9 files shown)


incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -72,30 +77,43 @@ pub(crate) fn apply_glyph_keyed_patches(
...
             // glyf patch application also generates a loca table.
             processed_tables.insert(table_tag);
-            processed_tables.insert(Tag::new(b"loca"));
-        } else if table_tag == Tag::new(b"CFF ")
-            || table_tag == Tag::new(b"CFF2")
-            || table_tag == Tag::new(b"gvar")
-        {
-            // TODO(garretrieger): add CFF, CFF2, and gvar support as well.
+            processed_tables.insert(Loca::TAG);
+        } else if table_tag == Tag::new(b"gvar") {
```
cmyr, line 99: maybe Gvar::TAG, for consistency?

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -213,74 +231,74 @@ fn retained_glyphs_in_font(
...
-                "loca entries are not in ascending order",
+                "offset entries are not in ascending order",
             )))? as u64;
     }
 
     Ok(total_size)
 }
 
-trait LocaOffset {
+trait WritableOffset<const DIV: usize> {
```
cmyr, line 262: I think the DIV param is worth a comment, it feels slightly mysterious (where is it used, what are the expected values?)

cmyr (approved):
looks good!

PR #1324, opened 2025-01-22
impl Error for DrawError

RickyDaMa:
Helps with error composition & handling (e.g. using `anyhow`)

### DIFF
--- a/skrifa/src/outline/error.rs
+++ b/skrifa/src/outline/error.rs
@@ -84,3 +84,6 @@ impl fmt::Display for DrawError {
         }
     }
 }
+
+#[cfg(feature = "std")]
+impl std::error::Error for DrawError {}


skrifa/src/outline/error.rs
```diff
@@ -84,3 +85,5 @@ impl fmt::Display for DrawError {
         }
     }
 }
+
+impl Error for DrawError {}
```
cmyr, line 89: the only possible concern here is that I think `Error` is only available via `std`, so I would remove the top-level import of the trait and do,

```suggestion
#[cfg(feature = "std")]
impl std::error::Error for DrawError {}
```

@dfrg does this sound right to you?

cmyr (commented):
looks good with one note about std, I want chad to double-check my thinking but otherwise happy to get this in.

skrifa/src/outline/error.rs
dfrg: I believe `Error` in `core` is relatively recent so we may want to stick with the optional `std` type until we confirm that our primary users are compiling with a suitable Rust version.

rsheeter: MSRV is a bit squirrely, it's the floor of our major consumers and desired consumers so I don't actually know the authoritative answer.  That would lead me to be +1 on this suggestion:

```rust
#[cfg(feature = "std")]
impl std::error::Error for DrawError {}
```

cmyr (approved):
thanks!

PR #1325, opened 2025-01-22
[klippa] subset BASE table

qxliu76:


klippa/src/inc_bimap.rs
```diff
@@ -0,0 +1,48 @@
...
+use fnv::FnvHashMap;
+
+#[derive(Default)]
+pub(crate) struct IncBiMap {
+    forw_map: FnvHashMap<u32, u32>,
+    back_map: Vec<u32>,
+}
+
+impl IncBiMap {
+    pub(crate) fn new(capacity: usize) -> Self {
```
cmyr, line 14: for symmetry with other basic data structures I would expect `new` to not take any arguments, and for there to be a `with_capacity` constructor if you wanted to pass a capacity.

klippa/src/inc_bimap.rs
```diff
@@ -0,0 +1,48 @@
...
+            forw_map: FnvHashMap::with_capacity_and_hasher(capacity, Default::default()),
+            back_map: Vec::with_capacity(capacity),
+        }
+    }
+
+    pub(crate) fn len(&self) -> usize {
+        self.forw_map.len()
+    }
+
+    pub(crate) fn add(&mut self, lhs: u32) -> u32 {
```
cmyr, line 25: I think this method deserves a doc comment, since the semantics do not match what one would generally expect given the method name.

klippa/src/inc_bimap.rs
```diff
@@ -0,0 +1,48 @@
...
+                rhs
+            }
+        }
+    }
+
+    pub(crate) fn get(&self, lhs: u32) -> Option<&u32> {
+        self.forw_map.get(&lhs)
+    }
+
+    pub(crate) fn backward(&self, rhs: u32) -> Option<&u32> {
```
cmyr, line 41: maybe `get_backward` or `get_key` or something?

klippa/src/inc_bimap.rs
```diff
@@ -0,0 +1,48 @@
...
+        self.back_map.get(rhs as usize)
+    }
+
+    pub(crate) fn keys(&self) -> Iter<'_, u32> {
+        self.back_map.iter()
+    }
+}
```
cmyr, line 48: Do we expect this implementation to get any more complicated in the future? Would it make sense to have at least a single illustrative test?

klippa/src/lib.rs
```diff
@@ -10,19 +11,25 @@ mod gvar;
...
 mod os2;
 mod parsing_util;
 mod post;
 pub mod serialize;
 mod stat;
+mod variations;
+use inc_bimap::IncBiMap;
 pub use parsing_util::{
-    parse_drop_tables, parse_name_ids, parse_name_languages, parse_unicodes, populate_gids,
+    parse_name_ids, parse_name_languages, parse_tag_list, parse_unicodes, populate_gids,
```
cmyr, line 28: just noticing now, but are these methods genuinely part of the public API?

klippa/src/lib.rs
```diff
@@ -62,6 +70,98 @@ const MAX_NESTING_LEVEL: u8 = 64;
 // See <https://github.com/googlefonts/fontations/issues/997>
 const MAX_GID: GlyphId = GlyphId::new(0xFFFFFF);
 
+// ref: <https://github.com/harfbuzz/harfbuzz/blob/021b44388667903d7bc9c92c924ad079f13b90ce/src/hb-subset-input.cc#L82>
+pub static DEFAULT_LAYOUT_FEATURES: [Tag; 72] = [
```
cmyr, line 74: I would normally just make this a static slice instead of an array but that's just me?

(I mean like:

```rust
pub static DEFAULT_LAYOUT_FEATURES: &[Tag] = &[ ..
```
)

klippa/src/lib.rs
```diff
@@ -600,6 +800,50 @@ pub trait Subset {
...
+pub(crate) trait SubsetTableWithArgs: SubsetArgs {
+    fn subset_with_args(
+        &self,
+        plan: &Plan,
+        s: &mut Serializer,
+        args: &Self::ArgsForSubset,
+    ) -> Result<(), SerializeErrorFlags>;
+}
+
+pub(crate) trait SubsetTwoArgs {
```
cmyr, line 823: is this necessary? would it be possible to just have the args be a 2-tuple in some cases? (this is what we do in codegen for `FontReadWithArgs`)

klippa/src/lib.rs
```diff
@@ -600,6 +800,50 @@ pub trait Subset {
...
+        &self,
+        plan: &Plan,
+        s: &mut Serializer,
+        arg1: &Self::Arg1ForSubset,
+        arg2: &Self::Arg2ForSubset,
+    ) -> Result<(), SerializeErrorFlags>;
+}
+
+// A trait for subtables that require parent table's fontdata
+pub(crate) trait SubsetTableWithFontData {
```
cmyr, line 838: In general I'd see if I couldn't fold all of this into one trait, if we end up passing in the data to some tables that don't use it that seems fine?

klippa/src/lib.rs
```diff
@@ -600,6 +800,50 @@ pub trait Subset {
...
+    /// Subset this table and write a subset version of this table into serializer
+    fn subset(&self, plan: &Plan, s: &mut Serializer) -> Result<(), SerializeErrorFlags>;
+}
+
+// A trait for subtables that require additional data in order to be subsetted
+pub(crate) trait SubsetArgs {
+    type ArgsForSubset: Default;
+}
+
+pub(crate) trait SubsetTableWithArgs: SubsetArgs {
```
cmyr, line 814: is the trait dependency necessary? Do you sometimes impl `SubsetArgs` but not `SubsetTableWithArgs`?

klippa/src/variations.rs
```diff
@@ -0,0 +1,532 @@
...
+    if item >= var_data.item_count() as usize || region >= var_data.region_index_count() as usize {
+        return 0;
+    }
+
+    let p = item * row_size;
+    let word_delta_count = var_data.word_delta_count();
+    let word_count = (word_delta_count & 0x7FFF) as usize;
+    let is_long = word_delta_count & 0x8000 != 0;
+
+    #[allow(clippy::collapsible_else_if)]
```
cmyr, line 322: In general if we're ignoring a lint I think it's worth a comment explaining why we think it's justified.

In this case I assume this is ported directly from the harfbuzz, and so you want to keep the translation as literal as possible?

If so I'd just say this, and include a link to the original.

cmyr (commented):
I have a few questions inline about whether we need so many traits, but overall this seems reasonable :)

klippa/src/lib.rs
qxliu76: oh this is because I need to use these methods in integration_test, do you know how can I export them to integration_test without making them public? I tried `pub(crate) use` but that doesn't work.

cmyr: yea integration test only works on public API.

cmyr (approved):
Looks good, it's kind of annoying to have to pass the empty args in but I think it's an improvement on balance. :)

qxliu76:
> Looks good, it's kind of annoying to have to pass the empty args in but I think it's an improvement on balance. :)

thanks very much for your valuable review comments!

garretrieger (approved):

PR #1326, opened 2025-01-23
[klippa] subset VORG table

qxliu76:


klippa/src/vorg.rs
```diff
@@ -0,0 +1,108 @@
...
+impl Subset for Vorg<'_> {
+    fn subset(
+        &self,
+        plan: &Plan,
+        _font: &FontRef,
+        s: &mut Serializer,
+        _builder: &mut FontBuilder,
+    ) -> Result<(), SubsetError> {
+        s.embed(self.version())
+            .map_err(|_| SubsetError::SubsetTableError(Vorg::TAG))?;
```
dfrg, line 21: The `map_err` duplication is a bit noisy. If we’re just converting all errors to the same thing, it might be cleaner to move the logic to a private function in this module that returns `Result<(), SerializeErrorFlags>` and call that from the public API, doing the error conversion once.

dfrg (approved):
LGTM with a small suggestion for removing some duplication.

garretrieger (approved):

Issue #1263, opened 2024-11-25
Moving variable font logic into `read-fonts`


LaurenzV:
I was trying to look into whether I can use `read-fonts` to build an instancer for TTF fonts, for usage in PDFs. Building a full-blown instancer is obviously a huge task, but luckily, for PDF, I only need to instance the `hmtx` and `glyf` tables, which I'm hoping will be much more doable. 

I still need to do some proper reading to assess what I need to do to achieve this, however, at a first glance, when looking into this I noticed that most of the VF handling for glyph outlines seems to be inside `skrifa` in `deltas.rs`. Would it make sense to move some of this into `read-fonts`, so that other crates can re-use the code, or is there a particular reason why it's not in `read-fonts`? Another option would be to expose it in skrifa, but it seems to me like this would go against its philosophy, since `skrifa` is supposed to be very high-level and it also would be unfortunate to have to depend on skrifa just for that.

dfrg:
I don’t see any problem with moving this to read-fonts, perhaps as methods on the gvar table. The API requires passing in a lot of buffers (to avoid internal allocation) so that might need to be cleaned up and better documented for public use.

rsheeter:
> Building a full-blown instancer is obviously a huge task

FYI, we do plan to tackle this, likely after `klippa` (memory safe `hb-subset`) wraps up.

PR #1329, opened 2025-01-24
Declare 1.0 for libraries actively used by Chrome

rsheeter:
We are LIVE from the sold out Chrome stable user base. Declare 1.0 for relevant libraries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,11 +40,11 @@ core_maths = "0.1"
 # default-features enabled by `skrifa`. Other crates using `read-fonts`
 # that want default features will have to enable them directly.
 font-test-data = { path = "font-test-data" }
-font-types = { version = "0.8.2", path = "font-types" }
-read-fonts = { version = "0.25.3", path = "read-fonts", default-features = false }
+font-types = { version = "1.0.0", path = "font-types" }
+read-fonts = { version = "1.0.0", path = "read-fonts", default-features = false }
 # Disable default-features so that fauntlet can use skrifa without autohint
 # shaping support
-skrifa = { version = "0.26.5", path = "skrifa", default-features = false, features = ["std"] }
+skrifa = { version = "1.0.0", path = "skrifa", default-features = false, features = ["std"] }
 write-fonts = { version = "0.33.1", path = "write-fonts" }
 shared-brotli-patch-decoder = { version = "0.1.0", path = "shared-brotli-patch-decoder" }
 incremental-font-transfer = { version = "0.1.0", path = "incremental-font-transfer" }
--- a/font-test-data/Cargo.toml
+++ b/font-test-data/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "font-test-data"
-version = "0.2.1"
+version = "1.0.0"
 description = "Test data for the fontations crates"
 readme = "README.md"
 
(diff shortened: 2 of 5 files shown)


cmyr (commented):
I do think this makes sense but I have some minor hesitancy about doing it super casually; at the very least I'd like to go through open issues and see if there's anything obvious that we'll want to break API for.

other current blockers:

- we know we're breaking the api in the ivs module of font_test_data, so it's worth at least doing that first
- ...nothing I can think of offhand 🤷

rsheeter:
> doing it super casually

I would argue we aren't doing it casually. We're acknowledging that we're shipping to a very large userbase and now need to be a bit more careful wrt compatibility. This makes the versioning reflect reality on the ground.

> breaking the api in the ivs module of font_test_data, so it's worth at least doing that first

Agreed, imho that's the only thing we need to block on

nicoburns:
Based on fontations' history, I'm assuming there will still be regular releases that are breaking from an API perspective (if not from a behaviour perspective)? So we should expect 2.0, 3.0 release soon and this just indicated maturity? Or is this indicating a change in policy with regard to API breaks?

cmyr:
> Based on fontations' history, I'm assuming there will still be regular releases that are breaking from an API perspective (if not from a behaviour perspective)? So we should expect 2.0, 3.0 release soon and this just indicated maturity? Or is this indicating a change in policy with regard to API breaks?


write-fonts will still stay on minor versions and should expect regular breaking changes, but we're going to try and avoid too much churn in `read-fonts` and `skrifa`; hopefully the majority of the work we need to do in these crates is just adding new API, not touching the old API.

That said, I expect we likely will need at least _some_ breaking changes in read-fonts and skrifa in the future, in which case we would go to 2.0, 3.0 etc.

nicoburns:
> we're going to try and avoid too much churn in read-fonts and skrifa; hopefully the majority of the work we need to do in these crates is just adding new API, not touching the old API.
> That said, I expect we likely will need at least some breaking changes in read-fonts and skrifa in the future, in which case we would go to 2.0, 3.0 etc.

This sounds good to me as an open source user of these crates (I'm not currently using `write-fonts`)! I wouldn't want y'all to hamstring yourself by forbidding breaking releases entirely. On the other hand, I had noticed that fontations crates were making breaking particularly frequently which has caused a lot of releases in downstream crates I use just to update those versions (especially `font-types` and `read-fonts` that often end up in the public API and so can't be updated in patch releases).

cmyr:
> On the other hand, I had noticed that fontations crates were making breaking particularly frequently which has caused a lot of releases in downstream crates I use just to update those versions (especially `font-types` and `read-fonts` that often end up in the public API and so can't be updated in patch releases).

Fair point, it is 100% true that we are not currently thinking very much about consumers other than chrome and `fontc`.  That said, I do hope that our churn will start to slow down soon; in particular `font-types` should be almost done, and `read-fonts` should be getting close (although there will be a few more breaking releases before we finally get to 1.0, such as #1263)

nicoburns:
Related to this issue, I believe the most recent two version minor version bumps for `font-types` were unecessary, and that `font-types` hasn't otherwise had any breaking changes since `80cfce87dd3cd334118828f0a8b76218dcabce22` in November 2024.

PR #1330, opened 2025-01-24
Rename Hhea::number_of_long_metrics => ...h_metrics

### LINKED ISSUE #1213: hhea/vhea numberOfHMetrics has become number_of_long_metrics
*Generally* the names of fields in `read-fonts` follows the names in the OpenType specification; so it's confusing when they don't.

The spec has

uint16 | numberOfHMetrics | Number of hMetric entries in 'hmtx' table
-- | -- | --

but `hhea.rs` has:

```
    /// Number of LongMetric entries in 'hmtx'/'vmtx' table
    number_of_long_metrics: u16,
```

`read-fonts`'s is probably a *better* name, but it causes a mental "wait, are these the same thing or not?" moment. A doc comment would be helpful.

cmyr:
This matches the spec; the old name was a relic of when we used a single common type to represent both vhea and hhea.

Also fixes the docs of both tables.

- closes #1213

### DIFF
--- a/fuzz/fuzz_targets/fuzz_basic_metadata.rs
+++ b/fuzz/fuzz_targets/fuzz_basic_metadata.rs
@@ -27,7 +27,7 @@ fn do_metadata_things(data: &[u8]) -> Result<(), Box<dyn Error>> {
     }
 
     if let Ok(hhea) = font.hhea() {
-        let _ = hhea.number_of_long_metrics();
+        let _ = hhea.number_of_h_metrics();
         if let Ok(hmtx) = font.hmtx() {
             let _ = hmtx.h_metrics().iter().count();
             let _ = hmtx.left_side_bearings().iter().count();
--- a/read-fonts/generated/generated_hhea.rs
+++ b/read-fonts/generated/generated_hhea.rs
@@ -91,7 +91,7 @@ impl HheaMarker {
         start..start + i16::RAW_BYTE_LEN
     }
 
-    pub fn number_of_long_metrics_byte_range(&self) -> Range<usize> {
+    pub fn number_of_h_metrics_byte_range(&self) -> Range<usize> {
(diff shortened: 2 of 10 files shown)


rsheeter:
Note that this will break Chrome due to https://crsrc.org/c/third_party/skia/src/ports/fontations/src/ffi.rs;drc=5f2e30e0c8bd00b2ff314a6c751ceac95fedb913;l=1211. IIRC that may mean we need to mark deprecated in one release then drop in a subsequent release?

rsheeter (approved):
LGTM subject to confirmation around whether we need to temporarily have both names to enable client migration. If we do I'm not completely sure the change is worthwhile?

cmyr:
I think long term it's worth maintaining the overall principle that the names we use for fields are based on the names those fields have in the spec.

In any case going to 1.0 is somewhere that we can break things, so we could just break it now and then chromium and other consumers will need to fix it when they pull in the new version; the deprecation notice is mostly just a nice way of pointing the user to the new method name.

drott:
Because the Fontations code is in Skia, and that rolls into Chromium too, I know ran into the situation that when rolling Skrifa/Fontations in Chromium, I can't compile as Rust is configured to fail on deprecation warnings there. On the other hand, if I don't have the latest Fontations in Chromium, I can't change the function use to the new name. I think I can break the loop by putting in `#[allow(deprecated)]` somewhere. In the future, I think it would be ideal to publish one release without a deprecation, then publish another with the deprecation if possible. That still requires extra intermediate roll commits in Chromium, but it would not require insertion of `#[allow(deprecated)]` workaround.

cmyr:
@drott ah okay I didn't know you had `deny(deprecated)]` on; the only real purpose for having the deprecation was to guide you to the new method.

PR #1331, opened 2025-01-24
Shuffle things to avoid circular dep

rsheeter:
`font-test-data/Cargo.toml` no longer depends on read/write fonts, meaning there isn't a cycle formed by a regular dependency from font-test-data => read/write-fonts and the dev-dependency from read/write-fonts to font-test-data. Despite being theoretically OK this was confusing things during import into other enviornments.

Meant to unblock #1329 , if this looks ok please mark that approved as well.

### DIFF
--- a/font-test-data/Cargo.toml
+++ b/font-test-data/Cargo.toml
@@ -8,8 +8,7 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
-
+# do NOT create circular deps, even with dev-dependencies. It messes with google3 import.
 [dependencies]
-read-fonts = { workspace = true }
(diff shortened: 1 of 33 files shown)


cmyr (approved):
slightly confused about moving both `BeBuffer` into `font-test-data` and also moving code that uses it _out_ of `font-test-data`, I assume there was some rationale there? Otherwise lgtm...

rsheeter:
BeBuffer is useful for creation of binary testdata so it is handy to have in font-test-data (and used). It's just the use of read/write types that's problemmatic so that part moves and then gets BeBuffer from font-test-data.

PR #1332, opened 2025-01-24
Gvar delta getters return Result<Option<_>,_>>

### LINKED ISSUE #1277: `glyph_variation_data` returns Err for glyphs with no variation
`font.gvar()?.glyph_variation_data(GlyphId::new(0))` for a font where `.notdef` does not vary returns "Error: An offset was out of bounds".

This is surprising, because nothing actually went *wrong*, and I didn't even provide any offsets, so this looks to the user like a bug in the library. If the return type was `Result<Option<...>>` or `Option<Result<...>>`,instead of `Result`, then the lack of variation could be returned more clearly.

cmyr:
Instead of just Result<_,_>. This better represents the underlying semantics, where there is a potentially important distinction between deltas that do not exist at all and deltas that are expected to exist but cannot be read.

- closes #1277

### DIFF
--- a/read-fonts/src/tables/gvar.rs
+++ b/read-fonts/src/tables/gvar.rs
@@ -57,9 +57,14 @@ impl<'a> GlyphVariationDataHeader<'a> {
 }
 
 impl<'a> Gvar<'a> {
-    pub fn data_for_gid(&self, gid: GlyphId) -> Result<FontData<'a>, ReadError> {
-        let range = self.data_range_for_gid(gid)?;
-        self.data.slice(range).ok_or(ReadError::OutOfBounds)
+    /// Return the raw data for this gid.
+    ///
+    /// If there is no variation data for the glyph, returns `Ok(None)`.
+    pub fn data_for_gid(&self, gid: GlyphId) -> Result<Option<FontData<'a>>, ReadError> {
+        match self.data_range_for_gid(gid)? {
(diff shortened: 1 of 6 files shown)


read-fonts/src/tables/gvar.rs
```diff
@@ -57,9 +57,14 @@ impl<'a> GlyphVariationDataHeader<'a> {
...
 
 impl<'a> Gvar<'a> {
-    pub fn data_for_gid(&self, gid: GlyphId) -> Result<FontData<'a>, ReadError> {
-        let range = self.data_range_for_gid(gid)?;
-        self.data.slice(range).ok_or(ReadError::OutOfBounds)
+    /// Return the raw data for this gid.
+    ///
+    /// If there is no variation data for the glyph, returns `Ok(None)`.
+    pub fn data_for_gid(&self, gid: GlyphId) -> Result<Option<FontData<'a>>, ReadError> {
+        match self.data_range_for_gid(gid)? {
```
dfrg, line 64: I believe `data_range_for_gid` should remain unchanged and this should return `Ok(None)` when the range is empty.

dfrg: And we should still return an error when the range is out of bounds.

cmyr: good call, I'd misunderstood the logic here.

PR #1334, opened 2025-01-27
[skrifa] reject composite outlines with more than 64k points

### LINKED ISSUE #1328: Fuzzer identifies contour point shift overflow in components
See https://issues.oss-fuzz.com/391753684

Fuzzer test case in the above issue manages to overflow contour shift in in:
https://github.com/googlefonts/fontations/blob/865c1132c88998b2e43a93a43d294cad5af23197/skrifa/src/outline/glyf/mod.rs#L735

dfrg:
Fuzzing caught an overflow in contour end point shifts which occurs when a composite glyph contains more than 64k points. Since we compute the total point count up front, this simply rejects those glyphs when we try to generate an outline, returning an error instead.

For good measure, also ensures that contour end point indices are properly ordered when loading a glyph so that we don't end up with intermediate values that might overflow since point count only checks the final end point index.

ref https://issues.oss-fuzz.com/391753684

Fixes #1328

### DIFF
--- a/skrifa/src/outline/glyf/mod.rs
+++ b/skrifa/src/outline/glyf/mod.rs
@@ -12,7 +12,7 @@ use core_maths::CoreFloat;
 use deltas::AvailableVarMetrics;
 pub use hint::{HintError, HintInstance, HintOutline};
 pub use outline::{Outline, ScaledOutline};
-use raw::FontRef;
+use raw::{FontRef, ReadError};
 
 use super::{DrawError, GlyphHMetrics, Hinting};
 use crate::GLYF_COMPOSITE_RECURSION_LIMIT;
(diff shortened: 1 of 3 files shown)


cmyr (approved):
64k points should be enough for anybody

skrifa/src/outline/glyf/mod.rs
```diff
@@ -1424,4 +1443,58 @@ mod tests {
         assert!(advance_hvar != F26Dot6::ZERO);
         assert_eq!(advance_hvar, advance_gvar);
     }
+
+    // fuzzer overflow for composite glyph with too many total points
+    // <https://g-issues.oss-fuzz.com/issues/391753684
```
drott, line 1448: Tiny nit:
```suggestion
    // <https://issues.oss-fuzz.com/issues/391753684
```

skrifa/src/outline/glyf/mod.rs
```diff
@@ -1424,4 +1443,58 @@ mod tests {
...
+    #[test]
+    fn composite_with_too_many_points() {
```
drott, line 1450: Great minimal test case!

drott (approved):

PR #1337, opened 2025-01-29
Do not inline single bytes if they are followed by a zero

Hoolean:
This gives a small file size saving with a large test font, although I am unsure if it is always optimal; most importantly, it makes fontations [match exactly the behaviour of fontTools](https://github.com/fonttools/fonttools/blob/eeaa499981c587/Lib/fontTools/ttLib/tables/TupleVariation.py#L467), as linked in the adjacent comment of the modified source.


cmyr (approved):
Agree that this definitely fixes a mismatch with fonttools. Staring at this logic though I do now wonder if we are handling zeros sub-optimally? But that is orthogonal.

Good catch!

PR #1338, opened 2025-02-03
[skrifa] avoid timeout in DELTA instructions

### LINKED ISSUE #1290: Timeout in fuzz_skrifa_outline
See https://issues.oss-fuzz.com/issues/42538387

dfrg:
Adds a new `ValueStack::pop_count_checked()` method that checks for negative stack values when we're popping a value intended for a count. Changing `ValueStack::pop_usize()` directly could affect our compatibility with FreeType when dealing with other instructions so that one is left unchanged.

Additionally limits the counts for DELTAP and DELTAC to point count and cvt size respectively since we don't expect those instructions to modify the same value more than once.

ref https://issues.oss-fuzz.com/issues/42538387

Fixes #1290

### DIFF
--- a/skrifa/src/outline/glyf/hint/value_stack.rs
+++ b/skrifa/src/outline/glyf/hint/value_stack.rs
@@ -16,7 +16,7 @@ use HintErrorKind::{ValueStackOverflow, ValueStackUnderflow};
 pub struct ValueStack<'a> {
     values: &'a mut [i32],
     len: usize,
-    is_pedantic: bool,
+    pub(super) is_pedantic: bool,
 }
 
 impl<'a> ValueStack<'a> {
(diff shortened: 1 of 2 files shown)


skrifa/src/outline/glyf/hint/value_stack.rs
```diff
@@ -112,6 +112,24 @@ impl<'a> ValueStack<'a> {
...
+            if self.is_pedantic {
+                Err(HintErrorKind::InvalidStackValue(value))
+            } else {
+                Ok(0)
+            }
+        } else {
+            // Positive i32 always fits into usize
+            Ok(value as usize)
+        }
+    }
```
cmyr, line 131: i'd probably have written,

```rust
if value < 0 && self.is_pedantic {
    // err path
} else {
    Ok(value.max(0) as usize)
}
```

but of course your version is totally fine

cmyr (approved):
having an additional method feels like a clean solution.

skrifa/src/outline/glyf/hint/value_stack.rs
dfrg: I like your version better, thanks!

PR #1339, opened 2025-02-03
[types] Add Tag::NULL associated constant

cmyr:
We'd already decided to use this value (four spaces) as the value of Tag::default(), and it turns out that it is also allowed in OS/2, so exposing it as a constant seems useful.

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -23,6 +23,15 @@ use std::{
 pub struct Tag([u8; 4]);
 
 impl Tag {
+    /// A 'null' tag consists of four spaces.
+    ///
+    /// This tag is not a valid identifier, but can be used to represent the
+    /// absence of a tag. See the comments on the [OS/2 `achVendID`][achVendID]
+    /// field.
+    ///
+    /// [achVendID]: https://learn.microsoft.com/en-us/typography/opentype/spec/os2#achvendid
+    pub const NULL: Self = Tag::new(&[0x20; 4]);
+
     /// Construct a `Tag` from raw bytes.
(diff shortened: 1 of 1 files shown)


font-types/src/tag.rs
```diff
@@ -23,6 +23,12 @@ use std::{
 pub struct Tag([u8; 4]);
 
 impl Tag {
+    /// A 'null' tag consists of four spaces.
+    ///
+    /// This tag is not a valid identifier, but can be used to represent the
+    /// absence of a tag.
+    pub const NULL: Self = Tag::new(&[0x20; 4]);
```
rsheeter, line 30: I think there is spec ref for this that could be added?

behdad:
At the risk of bikeshedding, I suggest renaming this to ::Spaces. At least in HB we use NONE for all-zero tag, to specify lack of a tag...

anthrotype:
I merged too quickly sorry. I kind of agree with behdad, calling this 'null' implies it uses null byte (0x0) which this is not.

Since the bike is already out of the shed, I suggest `Tag::BLANK` to match the description in the OS/2 table

> This field can also be left _blank_ (set to null, or a tag comprised of four space characters)

Though technically there isn't any semantical difference between all-zero and all-space tag, both mean the absence of a tag.

cmyr:
okay so maybe I misread that comment?

I heard "set to null, _equivalent_ to four blank spaces", but maybe it means "set to all zeros, or to four blank spaces"?

cmyr:
in any case this constant isn't really doing anything, so happy to just revert...

PR #1340, opened 2025-02-03
[read-fonts] expose IndexSubtableList for bitmaps

dfrg:
To better represent the actual structure of the bitmap location tables in the spec.

Adds `BitmapSize::index_subtable_list()` which can be used to enumerate the index subtables for a given size.

@qxliu76 please let me know if this is sufficient for subsetting

### DIFF
--- a/read-fonts/generated/generated_bitmap.rs
+++ b/read-fonts/generated/generated_bitmap.rs
@@ -10,11 +10,11 @@ use crate::codegen_prelude::*;
 #[repr(C)]
 #[repr(packed)]
 pub struct BitmapSize {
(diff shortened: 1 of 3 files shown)


qxliu76:
Nice! This is exactly what I need. Thank you so much!

dfrg:
> Nice! This is exactly what I need. Thank you so much!

Great, thanks for confirming!

dfrg:
> Are you aware of any other places in read-fonts where we might have similar gaps in codegen?

Not off the top of my head but I'm sure there are a few. I'll be on the lookout.

PR #1344, opened 2025-02-05
Update copy indices implementation to match spec changes in https://github.com/w3c/IFT/pull/252

garretrieger:
Copy indices becomes child indices. Intersection is now based on the intersection of the child entries instead of forming a union of the child entry subset definitions.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -1087,7 +1087,7 @@ impl Field {
             let condition = condition.condition_tokens_for_read();
             if self.read_at_parse_time {
                 quote! {
-                    let #name = #condition.then(|| cursor.read::<#typ>()).transpose()?.unwrap_or(Default::default());
+                    let #name = #condition.then(|| cursor.read::<#typ>()).transpose()?.unwrap_or_default();
                 }
             } else {
                 quote! {
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -157,7 +157,7 @@ impl Fields {
             .find(|fld| &fld.name == name)
             .map(|fld| match &fld.typ {
                 FieldType::Scalar { typ } => typ,
-                _ => panic!("not a scalar field: {:?}", &fld.typ),
+                _ => panic!("not a scalar field"),
             })
             .or_else(|| {
                 self.read_args.as_ref().and_then(|args| {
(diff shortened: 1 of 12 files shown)


font-codegen/src/fields.rs
```diff
@@ -1087,7 +1087,7 @@ impl Field {
             let condition = condition.condition_tokens_for_read();
             if self.read_at_parse_time {
                 quote! {
-                    let #name = #condition.then(|| cursor.read::<#typ>()).transpose()?.unwrap_or(0);
+                    let #name = #condition.then(|| cursor.read::<#typ>()).transpose()?.unwrap_or(Default::default());
```
cmyr, line 1090: nit: prefer `unwrap_or_default()`:

```suggestion
                    let #name = #condition.then(|| cursor.read::<#typ>()).transpose()?.unwrap_or_default();
```

font-codegen/src/parsing.rs
```diff
@@ -318,6 +318,8 @@ pub(crate) enum CountTransform {
     MaxValueBitmapLen,
     /// requires exactly two args, defined as $arg1 - $arg2 + 2
     SubAddTwo,
+    /// requires exactly one arg. Get the count from the $arg1.`try_into::<usize>`().
+    Custom,
```
cmyr, line 322: can we just call this `TryInto`? It's clearer what is going on.

font-codegen/src/parsing.rs
```diff
@@ -1662,6 +1666,9 @@ impl Count {
                 (CountTransform::SubAddTwo, [a, b]) => {
                     quote!(transforms::subtract_add_two(#a, #b))
                 }
+                (CountTransform::Custom, [a]) => {
+                    quote!(transforms::custom_count(#a))
```
cmyr, line 1670: I'd probably just generate the code directly here, like:

```suggestion
                    quote!(usize::try_from(#a).unwrap_or_default())
```

read-fonts/src/tables/ift.rs
```diff
@@ -7,6 +7,55 @@ use std::str;
 pub const IFT_TAG: types::Tag = Tag::new(b"IFT ");
 pub const IFTX_TAG: types::Tag = Tag::new(b"IFTX");
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct MatchModeAndCount(u8);
```
cmyr, line 11: Docs please :)

cmyr (approved):
looks good!

PR #1345, opened 2025-02-06
Only write explicit tuple min/maxes if there is an intermediate tent

Hoolean:
Moved from original implementation in googlefonts/fontc#1225.

Motivation for moving:

> I've hit a road block adding tests: fontations' `GlyphDeltas` is public but its fields are not, and so `intermediate_region` cannot be inspected after construction.
> 
> We could make this field public, but I am tempted to argue that we make the wider change of moving the abstraction boundary lower into fontations, and treat this as serialisation logic.
> 
> e.g. give `GlyphDeltas` a `Vec<(F2Dot14, F2Dot14, F2Dot14)>` (or custom struct equivalent) and determine whether explicit intermediate coordinates are needed to be written at serialisation time.
> 
> I think this optimisation is higher-level than delta set optimisation but lower-level than IUP optimisation -- it is lossless, ttx does not expose it, and there has never been motivation to toggle it -- and so it may be more ergonomic there. We could avoid unzipping until serialisation time, ditch the assert that the tuple lengths are the same, and gain a bit more immutability for example.

---

This PR roughly sketches how the optimisation could look in a new home here instead. In general I think it has reduced the complexity of the code in the original PR, at the cost of some additional `Vec`s that could be targeted later if we end up in the hot path, and some boilerplate when we already know we do not have intermediate regions in the test cases.

If we are happy with this direction then we can polish the types and tidy the PR, and I will look to furnish it with unit tests too.

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -358,7 +358,7 @@ impl Intermediate {
     }
 
     /// Derive intermediate coordinates from a peak position as OpenType would
-    /// do. For non-zero peak positions, this is equivalent to definining a
+    /// do. For non-zero peak positions, this is equivalent to defining a
     /// non-intermediate region. For zero peak positions, this is equivalent to
     /// specifying that the axis does not contribute.
     /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#tuple-variation-store-header>
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -327,16 +327,25 @@ impl GlyphDelta {
 
 pub struct AxisCoordinates {
     peak: F2Dot14,
-    intermediates: Option<(F2Dot14, F2Dot14)>,
+    intermediate: Option<Intermediate>,
 }
 
-impl AxisCoordinates {
-    /// Get the minimum and maximum only if they differ from their defaults as
-    /// calculated from the peak.
-    fn required_intermediates(&self) -> Option<(F2Dot14, F2Dot14)> {
-        self.intermediates.filter(|(min, max)| {
-            (min, max) != (&self.peak.min(F2Dot14::ZERO), &self.peak.max(F2Dot14::ZERO))
-        })
+#[derive(PartialEq, Clone, Copy)]
+pub struct Intermediate {
+    minimum: F2Dot14,
+    maximum: F2Dot14,
+}
+
+impl Intermediate {
+    pub fn explicit(minimum: F2Dot14, maximum: F2Dot14) -> Self {
+        Self { minimum, maximum }
+    }
+
+    pub fn implied_by_peak(peak: F2Dot14) -> Self {
+        Self {
+            minimum: peak.min(F2Dot14::ZERO),
+            maximum: peak.max(F2Dot14::ZERO),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/gvar.rs
```diff
@@ -325,22 +325,41 @@ impl GlyphDelta {
...
+    pub fn new(axis_coordinates: Vec<AxisCoordinates>, deltas: Vec<GlyphDelta>) -> Self {
+        let peak_tuple = Tuple::new(axis_coordinates.iter().map(|coords| coords.peak).collect());
+
+        // File size optimisation: only write explicit minimums and maximums if
+        // at least one of the tents requires them.
+        // https://github.com/fonttools/fonttools/blob/b467579c/Lib/fontTools/ttLib/tables/TupleVariation.py#L184-L193
+        let required_intermediates = axis_coordinates
+            .iter()
+            .map(|coords| coords.required_intermediates())
+            .collect::<Option<Vec<_>>>();
```
cmyr, line 357: we should be able to get rid of this allocation and do this all in one iter chain? I think? 🤔

write-fonts/src/tables/gvar.rs
```diff
@@ -325,22 +325,41 @@ impl GlyphDelta {
     }
 }
 
+pub struct AxisCoordinates {
```
cmyr, line 328: if this is public api we'll want docs, and we'll also need some way to construct it.

cmyr (commented):
I think this is a good direction, would be happy to get this in after a bit of polish.

write-fonts/src/tables/gvar.rs
Hoolean: This did indeed turn out to be elegant and possible, with this neat new addition: https://users.rust-lang.org/t/unzip-with-error-handling/110250/2

...but I then realised that when reimplementing in fontations I had accidentally switched the semantics from `any(...)` to `all(...)`, and so unfortunately it all had to be scrapped again ☔

With the intention corrected, there doesn't seem to be a better semantic match than a mutable fold, which @RickyDaMa and I have now implemented, although it does at least stay readable and keep allocations to a minimum.

In general, representing a combination of `collect(...)` and `any(...)` in an iter chain has turned out to be a bit of an abstraction tar-pit; each time you follow an improvement, you get bumped back into the same narrow path :(

Hoolean:
We should be good to undraft here - ready for review!

write-fonts/src/tables/gvar.rs
```diff
@@ -325,22 +325,82 @@ impl GlyphDelta {
...
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#tuple-variation-store-header>
+    /// <https://learn.microsoft.com/en-us/typography/opentype/spec/otvaroverview#variation-data>
+    pub fn implied_by_peak(peak: F2Dot14) -> Self {
+        Self {
+            minimum: peak.min(F2Dot14::ZERO),
+            maximum: peak.max(F2Dot14::ZERO),
+        }
+    }
+}
+
```
cmyr, line 373: Okay so I'm trying to think through this, and understand if this is ultimately the API that makes the most sense.

Thinking out loud:

- I like the term 'tent', which although it was not immediately intuitive to me has the advantage that it is used elsewhere in the font ecosystem, and is fairly easy to understand.
- having an explicit `Intermediate` type feels maybe too verbose. If I look at `AxisCoordinates::new`, it even takes `Option<Intermediate>`, so if I don't need intermediates am I supposed to pass `None`, or am I supposed to call `implied_by_peak`?

My instinct, then, would be to keep using the name `Tent`, instead of `AxisCoordinate`, and so I'd have something like,

.... okay so I started sketching out what I would've expected to see, and then it just ended up easier for me to push a commit. Check out https://github.com/googlefonts/fontations/commit/e61f2cfaa62f98a09a195d61faffce8d06db3d7e and see if it makes sense?

Hoolean: Looks smashing, thanks Colin - I've applied it immediately!

Hoolean: I'll update the fontc PR; we may be nearing ready to merge 🚀

cmyr (approved):
I can't really have any complaints about this now then, can I? @Hoolean I'm happy to merge unless you are aware of any blockers.

Hoolean:
I'll do a ttx_diff run-through of a large font through the whole shebang (googlefonts/fontc#1253) just to check we're still capturing everything, but after that I'm confident we're ready to go :)

Hoolean:
Identical 💪 Let's merge!

write-fonts/src/tables/gvar.rs
```diff
@@ -325,22 +325,57 @@ impl GlyphDelta {
...
+///
+/// The values here end up serialized in the peak/start/end tuples in the
+/// [`TupleVariationHeader`].
+///
+/// The name 'Tent' is taken from HarfBuzz.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Tent {
+    peak: F2Dot14,
+    min: F2Dot14,
```
anthrotype, line 338: this is probably just cosmetic, as this isn't really a tuple but a struct with named fields, but I'm used to seeing a "tent" as a { min, peak, max } in this particular order, where peak is between min and max, by definition. I wonder if we should reoder.
Also about min/max.. I believe the spec prefers to call them start and end respectively, though the meaning is clear anyway.

anthrotype (approved):

write-fonts/src/tables/gvar.rs
cmyr: we're using min/max elsewhere, so I'm happy to just keep it consistent. ordering is an internal detail so we can change that later, I think it makes sense to get this merged. :)

PR #1346, opened 2025-02-07
[ivs_builder] match fonttools VarStore optimizer's stable ordering

anthrotype:
the fonttools VarStore.optimize() algorithm performs a sorting step on the input encodings (i.e. groups of delta sets characterized by the same row shape) to be processed before the heap-based priority queue is populated. This is to make sure the optimizer's output is stable no matter the order in which delta sets are processed.
Our Rust port currently does not perform any such sorting on the inputs; these get stored in an IndexMap and thus their order is not random, but depends on the order in which delta sets are added to the builder (usually e.g. for HVAR this means the glyph order); we only sort the output matching fonttools sort key.

But since this is a greedy algorithm the order in which things are processed matters. 
Moreover, while the priority queue is sorted by decreasing gain in both implementations, there are situations when multiple pairs of encodings have the same gain, thus the order in which they were originally pushed onto the heap determines the order in which they get popped.

There is also an additional difference, if you like this is an implementation detail. Python heapq module used by fonttools is a "min heap" (where the smallest item gets popped first) whereas Rust's BinaryHeap is a "max heap". In fonttools the encoding pairs are pushed with a negative gain, but positive i and j indexes, i.e. `heappush(heap, (-gain, i, j))`, so items are correctly sorted by decreasing gain like in our Rust's max heap; but in case of identical gains, items with smaller indexes are popped first from python's min heap, while the opposite happens in rust's max heap.

So the lack of sorting of the inputs on the Rust end, and the different heap implementations, contribute to produce slightly different ItemVariationData subtables. Even though the difference in size of the VarStores compiled by fontc vs fontmake is negligible (in the order of bytes, mere luck), we can hold the temptation to normalize this away, as we can easily match fonttools here - and we should.

I'll first push a failing test case, then I will follow with the actual fix.

@cmyr run the crater locally and said that with this patch, we gain +20 identical fonts, including Google Sans Flex.

### DIFF
--- a/write-fonts/Cargo.toml
+++ b/write-fonts/Cargo.toml
@@ -38,6 +38,7 @@ read-fonts = { workspace = true, features = [
 ] }
 rstest = "0.18.0"
 bincode = "1.0"
+rand = "0.8.5"
 
 pretty_assertions.workspace = true
 env_logger.workspace = true
--- a/write-fonts/src/tables/variations/ivs_builder.rs
+++ b/write-fonts/src/tables/variations/ivs_builder.rs
@@ -1521,8 +1521,8 @@ mod tests {
 
         let var_data = var_data_array.get(0).unwrap().unwrap();
         assert_eq!(var_data.region_indexes(), &[0, 1]);
-        assert_eq!(var_data.item_count(), 8); // not 7!
-        // [0, -1] should be in here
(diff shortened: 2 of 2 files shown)


cmyr (approved):
excellent find!

PR #1351, opened 2025-02-11
[klippa] subset CBLC/CBDT tables

qxliu76:
Also updated SubsetTable trait.
Harfbuzz implementation is suboptimal, rust code is using a different approach, following fonttools' method:
https://github.com/fonttools/fonttools/blob/7854669acd63be43e1ad41d0486297d8d6da325d/Lib/fontTools/subset/__init__.py#L1799


klippa/src/lib.rs
```diff
@@ -343,6 +346,13 @@ impl Plan {
         this.populate_gids_to_retain(font);
         this.create_old_gid_to_new_gid_map();
 
+        //update the unicode to new gid list
+        let num = this.unicode_to_new_gid_list.len();
```
cmyr, line 350: is this related to the cblc/cbdt changes?

klippa/src/lib.rs
```diff
@@ -807,13 +818,14 @@ pub trait Subset {
 // A helper trait providing a 'subset' method for various subtables that have no associated tag
 pub(crate) trait SubsetTable<'a> {
     type ArgsForSubset: 'a;
+    type SubsetOutput: 'a;
     /// Subset this table and write a subset version of this table into serializer
     fn subset(
         &self,
         plan: &Plan,
```
cmyr, line 825: I'd just call this `Output`, we know it's part of the `SubsetTable` trait. (Looking back I would have also suggested `Args` instead of `ArgsForSubset` but maybe that name was about avoiding clashing with `FontRead`?

klippa/src/cblc.rs
```diff
@@ -0,0 +1,762 @@
+//! impl subset() for CBLC
+use crate::serialize::{OffsetWhence, SerializeErrorFlags, Serializer};
+use crate::{Plan, Subset, SubsetError, SubsetTable};
+use skrifa::raw::collections::IntSet;
+use write_fonts::types::FixedSize;
+use write_fonts::{
```
cmyr, line 6: total nit but this can be cleaned up a bit..

klippa/src/cblc.rs
```diff
@@ -0,0 +1,762 @@
...
+    fn subset(
+        &self,
+        plan: &Plan,
+        s: &mut Serializer,
+        args: Self::ArgsForSubset,
+    ) -> Result<(), SerializeErrorFlags> {
+        let cblc = args.0;
+        let cbdt = args.1;
+        let src_bytes = args.2;
+        let cbdt_out = args.3;
```
cmyr, line 95: or,

```rust
let (cblc, cbdt, src_bytes, cbdt_out) = args;
```

klippa/src/cblc.rs
```diff
@@ -0,0 +1,762 @@
...
+            }
+        }
+        Ok(())
+    }
+}
+
+impl<'a> SubsetTable<'a> for IndexSubtableList<'a> {
+    type ArgsForSubset = (&'a Cbdt<'a>, &'a mut Vec<u8>);
+    // min_gid(new), max_gid(new), indexSubtableListSize, numberOfIndexSubtables
+    type SubsetOutput = (GlyphId, GlyphId, usize, usize);
```
cmyr, line 150: you could consider defining a simple struct for passing in these args, so that you have named fields? in this case in particular it seems easy to get the argument order wrong.

cmyr (approved):
A few notes inline but nothing major is jumping out at me...

klippa/src/lib.rs
qxliu76: no, but it's an important fix to avoid OTS failure

garretrieger (approved):

PR #1352, opened 2025-02-11
Apply limits to cmap12 iterator

dfrg:
Meant to address fuzzing timeouts that all seem to point back to iterating cmap format 12 subtables. We previously limited characters to the max assigned codepoint and this extends that to limit glyphs to `maxp.numGlyphs`.

Reverts `Cmap12::iter()` to yield the raw mappings and adds a new `Cmap12::iter_with_limits()` that allows custom limits for both characters and glyphs. `Cmap12IterLimits::default_for_font(&FontRef)` produces the expected limits.

We _could_ have modified the cmap table to add these limits directly but reasons for not doing so are two-fold:
1. This prevents the read-fonts code from accurately representing the font data
2. Would require a lot of manual definitions and impls for many cmap types

Skrifa changed to use the limited version internally.

Avoids timeout for fuzzing bug: https://issues.oss-fuzz.com/issues/394638728

Note: these are breaking changes for skrifa due to change from `impl TableProvider` to `FontRef` for various charmap constructors.

### DIFF
--- a/read-fonts/src/tables/cmap.rs
+++ b/read-fonts/src/tables/cmap.rs
@@ -4,7 +4,8 @@ include!("../../generated/generated_cmap.rs");
 
 #[cfg(feature = "std")]
 use crate::collections::IntSet;
-use std::ops::{Range, RangeInclusive};
+use crate::{FontRef, TableProvider};
+use std::ops::Range;
 
 /// Result of mapping a codepoint with a variation selector.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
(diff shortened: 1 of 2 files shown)


cmyr (approved):
looks good, happy to see us avoid another immediate breakage in read-fonts 😏

dfrg:
> looks good, happy to see us avoid another immediate breakage in read-fonts 😏

doing my best!

drott:
> Note: these are breaking changes for skrifa due to change from `impl TableProvider` to `FontRef` for various charmap constructors.

As you may have seen, I released a new skrifa and read-fonts and did only bump patch version. I hope that's suitable, as I understood this change to be only internally breaking?

PR #1353, opened 2025-02-11
[skrifa] tthint: adjust DELTAP/DELTAC limits

dfrg:
We limited these to point count/cvt size respectively but there are fonts (Arial is one) where these limits are not sufficient. Notably, you can have one exception per point/entry _and_ ppem so we can't use these counts for hard limits.

Instead limit both to `(stack_size / 2)` since each iteration requires two elements from the stack.


cmyr (approved):
godspeed 🫡

PR #1354, opened 2025-02-13
[chore] Bump patch version on read/write

cmyr:


anthrotype:
given that there were some backward incompatible changes in write-fonts, we should have bumped the minor version not just the patch

E.g. see this seemingly innocuous PR https://github.com/googlefonts/fontc/pull/1260 fails like this:

https://github.com/googlefonts/fontc/actions/runs/13306572423/job/37158822593?pr=1260

Issue #1139, opened 2024-09-06
write_fonts should update head.checkSumAdjustment


anthrotype:
it's always set to 0x0, see https://github.com/googlefonts/fontc/issues/945

@rsheeter noted

> https://github.com/googlefonts/fontations/blob/6d01f7425b52f06d23d8e24f433f09d2e3779ac5/write-fonts/src/font_builder.rs#L102 is computing table directory checksums and strikes me as the appropriate place to do it, update head if there is a head.

Issue #1357, opened 2025-02-13
write-fonts should respect the suggested top-level table ordering


cmyr:
This doesn't seem very high priority, but there is a suggested ordering so we might as well respect it:

 https://learn.microsoft.com/en-us/typography/opentype/spec/recom#optimized-table-ordering

anthrotype:
Fixed by https://github.com/googlefonts/fontations/pull/1360

PR #1360, opened 2025-02-18
[font_builder] compute head.checksum_adjustment and sort table in recommended order

anthrotype:
Will fix https://github.com/googlefonts/fontations/issues/1139 and https://github.com/googlefonts/fontations/issues/1357

by basically matching fonttools `sortedTagList` and SFTNWriter's checksum adjustment computation.

~I still need to add unit test and fix up some failing ift tests.~ Added tests and fixed failing ones.

This is required for https://github.com/googlefonts/fontc/issues/945 and https://github.com/googlefonts/fontc/issues/1268


anthrotype:
@qxliu76 I had to recompile all the binary test files used in klippa integration tests (as these compare the bytes of expected and output file). The only diff is the head.checksum_adjustment and table order.

This file contains the full ttx diff if you would like to inspect:
[kippa-test-data.diff.txt](https://github.com/user-attachments/files/18865790/kippa-test-data.diff.txt)

In case you don't know this trick already, you can enable readable diffs for binary TTF/OTF font files in git using ttx by doing this:

```
# configure Git to use ttx for diffing fonts
git config --global diff.ttx.textconv "ttx -q -o -"
# set up global .gitattributes for font files
echo "*.ttf diff=ttx" >> ~/.gitattributes
echo "*.otf diff=ttx" >> ~/.gitattributes
# tell Git to use the global .gitattributes
git config --global core.attributesFile ~/.gitattributes
```

You can then do the usual `git diff` or `git show` etc. to see textual diffs in binary ttf/otf files using ttx.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -752,6 +752,15 @@ pub(crate) mod tests {
 
         for (tag_a, tag_b) in it_a.zip(it_b) {
             assert_eq!(tag_a, tag_b);
+            if tag_a == Tag::new(b"head") {
+                // zero out checksum_adjustment field
+                let mut a = a.table_data(tag_a).unwrap().as_bytes().to_vec();
+                let mut b = b.table_data(tag_b).unwrap().as_bytes().to_vec();
+                a[8..12].copy_from_slice(&[0, 0, 0, 0]);
```
cmyr, line 759: just trivia, but I think you could write this as,
```suggestion
                a[8..12].fill(0);
```

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -752,6 +752,15 @@ pub(crate) mod tests {
...
+                b[8..12].copy_from_slice(&[0, 0, 0, 0]);
+                assert_eq!(a, b, "{}", tag_a);
+                continue;
+            }
```
cmyr, line 763: after staring at this for a second I think I would probably express this whole thing as,

```rust
        for (tag_a, tag_b) in it_a.zip(it_b) {
            assert_eq!(tag_a, tag_b);
            let data_a = a.table_data(tag_a).unwrap();
            let data_b = b.table_data(tag_b).unwrap();
            if tag_a == Tag::new(b"head") {
                // ignore the checksum, which will necessarily differ
                assert_eq!(data_a.as_bytes()[..8], data_b.as_bytes()[..8]);
                assert_eq!(data_a.as_bytes()[12..], data_b.as_bytes()[12..]);
            } else {
                assert_eq!(data_a.as_bytes(), data_b.as_bytes(), "{}", tag_a);
            }
```

but 🤷

write-fonts/src/font_builder.rs
```diff
@@ -93,35 +127,99 @@ impl<'a> FontBuilder<'a> {
...
+        //   Group 0: tags that are in the recommended order, sorted accordingly.
+        //   Group 1: tags not in the recommended order, sorted alphabetically.
+        //   Group 2: 'DSIG' is always sorted last, matching fontTools' behavior.
+        let mut ordered_tags: Vec<Tag> = self.tables.keys().copied().collect();
+        let dsig = Tag::new(b"DSIG");
+        ordered_tags.sort_unstable_by_key(|rtag| {
+            let tag = *rtag;
+            if tag == dsig {
+                (2, 0, tag)
+            } else if let Some(idx) = recommended_order.iter().position(|t| t == rtag) {
```
cmyr, line 156: my "accidentally quadratic" sense is tingling, although we know that the values of M & N are bounded and small so it's fine...

write-fonts/src/font_builder.rs
```diff
@@ -93,35 +127,99 @@ impl<'a> FontBuilder<'a> {
...
 
         let mut writer = TableWriter::default();
         directory.write_into(&mut writer);
         let mut data = writer.into_data().bytes;
-        for table in self.tables.values() {
-            data.extend_from_slice(table);
+        checksums.push(read_fonts::tables::compute_checksum(&data));
+
+        let checksum = checksums.into_iter().fold(0u32, u32::wrapping_add);
+        let checksum_adjustment = 0xB1B0_AFBAu32.wrapping_sub(checksum); // BiboAfba!
```
cmyr, line 211: this deserves a more expository comment :)

cmyr (approved):
looks good! some little nits and notes inline but nothing major :)

incremental-font-transfer/src/glyph_keyed.rs
anthrotype: :+1 that's much simpler, thanks

write-fonts/src/font_builder.rs
anthrotype: fair enough :) I wanted to keep the magic aura 🤣

anthrotype:
the clippy failures are unrelated (probably a new toolchain was released) so I don't think they should block this

Issue #1261, opened 2024-11-22
[skrifa] Please add a glyph name function


simoncozens:
I find myself carrying this bit of code everywhere I go:

```rust
    fn glyph_name_for_id_impl(&self, gid: impl Into<GlyphId>, synthesize: bool) -> Option<String> {
        let gid: GlyphId = gid.into();
        if self._glyphnames.borrow().is_empty() {
            if let Ok(post) = self.font().post() {
                match post.version() {
                    Version16Dot16::VERSION_1_0 => {
                        let names = DEFAULT_GLYPH_NAMES.into_iter().map(|x| Some(x.to_string()));
                        self._glyphnames.borrow_mut().extend(names);
                    }
                    Version16Dot16::VERSION_2_0 => {
                        let strings: Vec<Option<read_fonts::tables::post::PString>> =
                            post.string_data()?.iter().map(|x| x.ok()).collect();
                        if let Some(index) = post.glyph_name_index() {
                            let names = (0..self.glyph_count).map(|gid| {
                                let idx = index.get(gid)?.get() as usize;
                                if idx < 258 {
                                    Some(DEFAULT_GLYPH_NAMES[idx].to_string())
                                } else {
                                    let entry = strings.get(idx - 258)?;
                                    entry.map(|x| x.to_string())
                                }
                            });
                            self._glyphnames.borrow_mut().extend(names);
                        }
                    }
                    _ => {}
                }
            }
        }
        if let Some(Some(n)) = self._glyphnames.borrow().get(gid.to_u32() as usize) {
            Some(n.to_string())
        } else if synthesize {
            Some(format!("gid{:}", gid))
        } else {
            None
        }
    }
```

(Note that this gets all the glyph names at once and then uses a _glyphnames cache because random access on the post table is unpleasantly slow #1106; I guess in a library function you may have to bite the bullet and do random access, and end users can cache it if they want to.)

dfrg:
Agreed that this is a hole in the API and we’ll also need it for HarfRuzz soon. I’ve mostly avoided doing it so far because we don’t currently have support for parsing glyph names from the cff table. I’ll add it to my list.

behdad:
Would be nice to match HB on this.

behdad:
> Would be nice to match HB on this.

HB has two APIs. One low-level, that can return false if the font has no glyph names. The other will fall back to `gidDDD` where DDD is the glyph index in decimal.

PR #1361, opened 2025-02-19
[read-fonts] CFF charsets

dfrg:
Lets us parse CFF charsets which enables support for reading glyph names from CFF fonts.

Makes some progress on #1261 and will soon be needed for harfruzz tests which depend on glyph names.


cmyr (approved):
looks good!

PR #1362, opened 2025-02-20
[klippa] subset HVAR/VVAR tables

qxliu76:


klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+pub(crate) struct IndexMapSubsetPlan {
+    map_count: u16,
+    max_inners: Vec<u16>,
+    outer_bit_count: u8,
+    inner_bit_count: u8,
+    output_map: FnvHashMap<GlyphId, u32>,
+}
+
+impl IndexMapSubsetPlan {
+    fn init(
```
cmyr, line 122: by convention Rust uses the name `new` for constructors 🤷

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+        index_map: &Option<DeltaSetIndexMap>,
```
cmyr, line 123: you rarely want `&Option<T>`; `Option<&T>` is more flexible, and it's easier to convert from `&Option<T>` with `.as_ref()`.

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+        let mut this = IndexMapSubsetPlan::default();
+        if bypass_empty && index_map.is_none() {
+            return Ok(this);
+        }
+
+        let entry_format = if index_map.is_none() {
+            EntryFormat::from_bits_truncate(1)
+        } else {
+            index_map.as_ref().unwrap().entry_format()
+        };
```
cmyr, line 138: slightly more idiomatic would be,

```suggestion
        let entry_format = index_map
            .as_ref()
            .map(DeltaSetIndexMap::entry_format)
            .unwrap_or(EntryFormat::from_bits_truncate(1));
```

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+        let mut this = HvarVvarSubsetPlan::default();
+        let vardata_count = var_store.item_variation_data_count() as usize;
+
+        let mut inner_sets = Vec::new();
+        inner_sets.resize(vardata_count, Default::default());
+
+        let mut outer_map = IncBiMap::default();
+        // don't bypass empty adv map
+        let index_map_subset_plan =
+            IndexMapSubsetPlan::init(&index_maps[0], plan, false, &mut outer_map, &mut inner_sets)?;
```
cmyr, line 281: is there a particular reason for passing these in? In general I would prefer to construct them in the called method, and return a tuple if needed...

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+            && plan
+                .subset_flags
+                .contains(SubsetFlags::SUBSET_FLAGS_RETAIN_GIDS);
+
+        if retain_adv_map {
+            let mut inner_map = IncBiMap::with_capacity(plan.new_to_old_gid_list.len());
+            for (_, old_gid) in plan.new_to_old_gid_list.iter() {
+                inner_map.add(old_gid.to_u32());
+            }
+            this.inner_maps.push(inner_map);
```
cmyr, line 311: why not just do,

```suggestion
            let inner_map =  plan.new_to_old_gid_list.iter().map(|(_, old_gid)| old_gid.to_u32()).collect();
            this.inner_maps.push(inner_map);
```

you might need to implement `FromIterator` for `IncBiMap` but that's a good exercise.

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+        } else {
+            let mut inner_map = IncBiMap::default();
+            for old_gid in adv_set.iter() {
+                inner_map.add(old_gid as u32);
+            }
```
cmyr, line 316: same story. In general if there's a way to express something that avoids the`mut` keyword, you should try to do that, since it makes the code easier to read (every `mut` declaration is something that the reader needs to keep track of, in case it is subsequently modified)

klippa/src/hvar.rs
```diff
@@ -0,0 +1,640 @@
...
+            this.inner_maps.push(map);
+        }
+
+        for (i, index_map) in index_maps.iter().enumerate() {
+            this.index_map_subset_plans[i].remap(plan, index_map, &outer_map, &this.inner_maps);
+        }
+        Ok(this)
+    }
+
+    pub(crate) fn inner_maps(&self) -> &Vec<IncBiMap> {
```
cmyr, line 340: unless there's a good reason (i.e. you need `&mut Vec<T>`) you should generally prefer to pass `&[T]` to `&Vec<T>`.

klippa/src/variations.rs
```diff
@@ -419,6 +422,102 @@ fn collect_region_refs(
...
+        self.inner_bit_count
+    }
+
+    pub(crate) fn get_output_map(&self) -> &'a FnvHashMap<GlyphId, u32> {
+        self.output_map
+    }
+
+    pub(crate) fn get_map_count(&self) -> u32 {
+        self.map_count
+    }
```
cmyr, line 461: By convention we generally leave the `get_` prefix off of getters (but use `set_` for setters)

cmyr (approved):
A few optional things inline, all related to style & idiom.

If you rebase on `main` the clippy issues should be resolved.

klippa/src/hvar.rs
qxliu76: oh it's because these are more global and would need to be updated in each constructor call, we cannot just construct them and return

garretrieger (approved):

PR #1371, opened 2025-03-03
[write] Fix table splitting dropping markFilteringSet

cmyr:
An oversight. This was causing an OTS failure in fontc's CI.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -1670,35 +1670,38 @@ mod tests {
         assert_eq!(graph.order.len(), EXPECTED_N_TABLES);
     }
 
+    fn make_big_pair_pos(glyph_range: Range<u16>) -> crate::tables::gpos::PositionLookup {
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -1670,36 +1670,37 @@ mod tests {
...
     }
 
+    fn make_big_pair_pos(glyph_range: Range<u16>) -> crate::tables::gpos::PositionLookup {
+        use crate::tables::{gpos, layout};
+        let coverage = glyph_range.clone().map(GlyphId16::new).collect();
+        let pair_sets = glyph_range
+            .map(|id| {
+                let value_rec = gpos::ValueRecord::new().with_x_advance(id as _);
+                gpos::PairSet::new(
+                    (id..id + 165)
```
rsheeter, line 1680: Can 165 gain a const or comment? - bit magical as written here

rsheeter (approved):

PR #1374, opened 2025-03-04
Set up shared-brotli-patch-decoder to be able to utilize a rust or c brotli decoder implementation.

garretrieger:
The pure rust decoder implementation is needed when compiling to WASM which can't easily work with the c version.

### DIFF
--- a/shared-brotli-patch-decoder/Cargo.toml
+++ b/shared-brotli-patch-decoder/Cargo.toml
@@ -11,8 +11,16 @@ repository.workspace = true
 # RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
 all-features = true
 
+[features]
+default = ["c-brotli"]
+c-brotli = ["dep:brotlic-sys"]
+rust-brotli = ["dep:brotli-decompressor"]
+
+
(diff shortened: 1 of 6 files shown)


shared-brotli-patch-decoder/Cargo.toml
```diff
@@ -11,8 +11,16 @@ repository.workspace = true
 # RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
 all-features = true
 
+[features]
+default = ["c-brotli"]
```
rsheeter, line 15: Optional, I wonder if rust-brotli isn't a more natural default for a rust program?

rsheeter (approved):

shared-brotli-patch-decoder/Cargo.toml
garretrieger: A couple of things I'd like to resolve first before making rust the default:
1. Need to test the relative performance of the two versions. If rust is comparable to c then it should be fine as a default.
2. Need to fix the inability of the rust version to detect an input that has unused trailing data, since that's required by the spec.

PR #1375, opened 2025-03-04
[write-fonts] Make COLR available

rsheeter:
Generate write-fonts types for colr. I suspect in time we'll want some sort of builder or helper but this is a start.

Note that I added `#[allow(clippy::derivable_impls)]` to generated records to enable the generated code to build.


font-codegen/src/record.rs
```diff
@@ -273,6 +273,7 @@ pub(crate) fn generate_compile_impl(
         let default_field_inits = fields.iter_compile_default_inits();
         quote! {
         impl #default_impl_params Default for #name <#generic_param> {
+            #[allow(clippy::derivable_impls)]
```
cmyr, line 276: You'll notice above that this path is only taken if we believe that we can't derive the impl, otherwise we just derive it, so perhaps you're hitting some case where `can_derive_default` is returning `false` incorrectly?

resources/codegen_inputs/colr.rs
```diff
@@ -6,6 +6,7 @@
 table Colr {
     /// Table version number - set to 0 or 1.
     #[version]
+    #[default(0)]
```
cmyr, line 9: I think the best approach for this is going to be `#[compile(self.compute_version())]`, and then add a method to `Color` that picks the correct version based on what fields are used.

cmyr (approved):
One change to make, which will obviate needing to make any changes in codegen, otherwise looks good!

font-codegen/src/record.rs
rsheeter: Yes. If you `#[default(the default value for the type)]`, e.g. `#[default(0)]` and all other fields are defaultable we don't identify that we could just take the derived default. We generate roughly:

```rust
fn default() -> Self {
Self {
  my_int_field_with_default_zero: 0,
  other_fields: Default::default()
}
```

 It's a bug in codegen but not one I want to block on. Your other change will avoid it but the bug will remain, will file an issue.

resources/codegen_inputs/colr.rs
rsheeter: That works, will so change.

font-codegen/src/record.rs
rsheeter: https://github.com/googlefonts/fontations/issues/1376

PR #1378, opened 2025-03-04
Wire up the c-brotli/rust-brotli features into incremental-font-transfer.

garretrieger:


incremental-font-transfer/Cargo.toml
```diff
@@ -15,16 +15,18 @@ repository.workspace = true
 all-features = true
 
 [features]
-default = ["read-fonts/std"]
+default = ["read-fonts/std", "c-brotli"]
 cli = ["clap", "regex"]
+rust-brotli = ["shared-brotli-patch-decoder/rust-brotli"]
+c-brotli = ["shared-brotli-patch-decoder/c-brotli"]
```
cmyr, line 21: welcome to one of the annoying corners of rust: features are supposed to be 'additive', that is you can add them in any combination and things work, see https://doc.rust-lang.org/cargo/reference/features.html?highlight=additive#feature-unification.


In any case there's actually a section in the docs now on "mutually incompatible features", so that would be worth a skim at least: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features

cmyr (approved):
lgtm with an asterisk, you should go read through the doc links mentioned below.

incremental-font-transfer/Cargo.toml
garretrieger: I set up shared-brotli-patch-decoder to work if both are enabled simultaneously (c takes priority) using the cfg-if mechanism as suggested there.

PR #1381, opened 2025-03-07
Fix invalid state when compiling certain IUP deltas (googlefonts/fontc#1116)

Hoolean:
Our transition to unsigned values when porting the Python original introduced a slight semantic difference that could cause an invalid state error: [in the original source](https://github.com/googlefonts/fonttools/blob/430e364db7baf5494d1bdfe077154edc4dbd6b71/Lib/fontTools/varLib/iup.py#L441), this condition is also true when `i` is non-negative and `start - n` is negative.

This PR introduces a checked comparison to match this behaviour.

(The only negative values possible on either side of the comparison are `-1`, so we will continue matching Python's behaviour for negative `i` too.)

---

After this commit, four fonts on fontc_crater (the Instrument Sans family) go from crashing with fontc to compiling, and yield an identical `gvar`. This is the same for the test case in the original issue, which has been included here as a unit test, and our proprietary library font that originated the issue too :)

### DIFF
--- a/write-fonts/src/tables/gvar/iup.rs
+++ b/write-fonts/src/tables/gvar/iup.rs
@@ -522,11 +522,14 @@ fn iup_contour_optimize(
             // Assemble solution
             let mut solution = HashSet::new();
             let mut i = Some(start);
-            while i > Some(start.saturating_sub(n)) {
+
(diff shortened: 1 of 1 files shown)


cmyr (approved):
good catch, I've been avoiding digging into the IUP code too much, so happy that one of us is brave.

PR #1384, opened 2025-03-07
[skrifa] impl TableProvider -> FontRef in ctors (breaking)

dfrg:
We needed a full `FontRef` for outlines anyway so this changes all other constructors to take `FontRef` as well for consistency.

JMM


cmyr (approved):
we love consistency 👍

PR #1387, opened 2025-03-07
[skrifa] glyph names API

### LINKED ISSUE #1261: [skrifa] Please add a glyph name function

dfrg:
Adds an API for accessing glyph names. Tries the `post` and `CFF` tables in that order. If neither are available or do not contain glyph names, then this generates synthesized names in the format `gidDDD` to match HarfBuzz.

Fixes #1261

### DIFF
--- a/skrifa/src/glyph_name.rs
+++ b/skrifa/src/glyph_name.rs
@@ -27,6 +27,14 @@ pub struct GlyphNames<'a> {
     inner: Inner<'a>,
 }
 
+#[derive(Clone)]
+enum Inner<'a> {
+    // Second field is num_glyphs
+    Post(Post<'a>, u32),
+    Cff(Cff<'a>, Charset<'a>),
+    Synthesized(u32),
+}
+
 impl<'a> GlyphNames<'a> {
     /// Creates a new object for accessing glyph names from the given font.
     pub fn new(font: &FontRef<'a>) -> Self {
@@ -164,6 +172,13 @@ impl GlyphName {
         name
     }
 
+    /// Appends the given bytes to `self` while keeping the maximum length
+    /// at 63 bytes.
+    ///
+    /// This exists primarily to support the [`core::fmt::Write`] impl
+    /// (which is used for generating synthesized glyph names) because
+    /// we have no guarantee of how many times `write_str` might be called
+    /// for a given format.
     fn append(&mut self, bytes: &[u8]) {
         // We simply truncate when length exceeds the max since glyph names
         // are expected to be <= 63 chars
@@ -221,14 +236,6 @@ impl core::fmt::Write for GlyphNameWrite<'_> {
         self.0.append(s.as_bytes());
         Ok(())
     }
-}
-
-#[derive(Clone)]
-enum Inner<'a> {
-    // Second field is num_glyphs
-    Post(Post<'a>, u32),
-    Cff(Cff<'a>, Charset<'a>),
-    Synthesized(u32),
 }
 
 #[derive(Clone)]
--- a/skrifa/src/glyph_name.rs
+++ b/skrifa/src/glyph_name.rs
@@ -20,7 +20,7 @@ const MAX_GLYPH_NAME_LEN: usize = 63;
 ///
 /// This sources glyph names from the `post` and `CFF` tables in that order.
 /// If glyph names are not available in either, then they are synthesized
-/// as `gidNNN` where `NNN` is the glyph identifier. Use the
+/// as `gidDDD` where `DDD` is the glyph identifier in decimal. Use the
 /// [`source`](Self::source) to determine which source was chosen.
 #[derive(Clone)]
 pub struct GlyphNames<'a> {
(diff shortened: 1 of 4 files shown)


dfrg:
@behdad @simoncozens Does this work for you?

It will automatically fall back to synthesis but `GlyphNames::source() == GlyphNameSource::Synthesized` when it does so the user is aware if it matters.

khaledhosny:
How does this handle `post` table with partial glyph names coverage?

dfrg:
It doesn’t do anything clever; just returns whatever is in `post`. Do you have an example font for this?

khaledhosny:
Here is a font with three glyphs, but `post` table has 2 names only: [test.zip](https://github.com/user-attachments/files/19146425/test.zip)

dfrg:
> Here is a font with three glyphs, but `post` table has 2 names only: [test.zip](https://github.com/user-attachments/files/19146425/test.zip)

Thanks. For this font, we produce `[(GlyphId(0), ".notdef"), (GlyphId(1), ".space")]` which is what I would expect since we don't synthesize names for partial coverage. I suppose we could? I'm not sure what the desired behavior is here.

behdad:
SGTM.

skrifa/src/glyph_name.rs
```diff
@@ -0,0 +1,394 @@
...
+        use core::fmt::Write;
+        let mut name = Self {
+            is_synthesized: true,
+            ..Self::default()
+        };
+        let _ = write!(GlyphNameWrite(&mut name), "gid{}", glyph_id.to_u32());
+        name
+    }
+
+    fn append(&mut self, bytes: &[u8]) {
```
cmyr, line 167: do we ever expect this to be called on a non-empty name? I'm sure there's some rationale here but it seems like a funny API choice. I suppose it's just to work with `write!`? In that case I think it deserves a little expository comment.

skrifa/src/glyph_name.rs
```diff
@@ -0,0 +1,394 @@
...
+}
+
+struct GlyphNameWrite<'a>(&'a mut GlyphName);
+
+impl core::fmt::Write for GlyphNameWrite<'_> {
+    fn write_str(&mut self, s: &str) -> core::fmt::Result {
+        self.0.append(s.as_bytes());
+        Ok(())
+    }
+}
```
cmyr, line 224: I'm wondering if this is necessary, since `&mut [u8]` impls write? but I guess it's hard to determine the written length in that case.

skrifa/src/glyph_name.rs
```diff
@@ -0,0 +1,394 @@
...
+
+#[derive(Clone)]
+enum Inner<'a> {
```
cmyr, line 227: totally unnecessary but I would personally declare this up right below the type that uses it, I think it would make it a bit clearer how the root type works.

cmyr (approved):
looks good!

skrifa/src/glyph_name.rs
dfrg: correct, the append functionality exists just for `write!` since we have no idea how many times `write_str` may be called for a given format. Added a comment explaining this.

skrifa/src/glyph_name.rs
dfrg: exactly, we need to track the length so this needs a tiny wrapper. It's the only builtin method I could think of that does int to string without allocation.

skrifa/src/glyph_name.rs
dfrg: Makes sense! Done.

behdad:
For `util/hb-shape /home/behdad/harfbuzz/test/shape/data/in-house/fonts/a34a7b00f22ffb5fd7eef6933b81c7e71bc2cdfb.ttf --unicodes U+180A,U+1868,U+180A,U+1868,U+180B,U+180A,U+1868,U+180C,U+180A,U+1868,U+180D,U+180A --font-funcs=fontations` I'm was getting empty glyph names. That is, I suppose, Some(""). So I had to filter those out in my code.

PR #1388, opened 2025-03-07
[skrifa] malloc free cycle detection

dfrg:
This is similar to Behdad's implementation: an extension of Floyd's tortoise and hare algorithm to DFS traversals but we use a fixed sized array rather than a linked list for obvious Rust reasons (the borrow checker doesn't like linked lists).

See https://github.com/harfbuzz/harfbuzz/pull/5074 for some background

### DIFF
--- a/skrifa/src/decycler.rs
+++ b/skrifa/src/decycler.rs
@@ -1,0 +1,178 @@
+//! Support for cycle detection in DFS graph traversals.
+
+use core::ops::{Deref, DerefMut};
+
+#[derive(Copy, Clone, Debug)]
+pub(crate) enum DecyclerError {
+    DepthLimitExceeded,
+    CycleDetected,
+}
+
+/// Cycle detector for DFS traversal of a graph.
+///
+/// The graph is expected to have unique node identifiers of type `T`
+/// and traversal depth is limited to the constant `D`.
+///
+/// This is based on `hb_decycler_t` in HarfBuzz (https://github.com/harfbuzz/harfbuzz/blob/a2ea5d28cb5387f4de2049802474b817be15ad5b/src/hb-decycler.hh)
+/// which is an extension of Floyd's tortoise and hare algorithm (https://en.wikipedia.org/wiki/Cycle_detection#Floyd's_tortoise_and_hare)
+/// to DFS traversals.
+///
+/// Unlike the implementation in HB which supports traversals of arbitrary
+/// depth, this is limited to a constant value. Instead of building a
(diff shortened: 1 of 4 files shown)


behdad:
Looks good to me. I thought about moving my depth-limit tracking to the decycler as well, but didn't in the end.

behdad:
Wait a second. The reason I didn't integrate the depth detection in the decycler in HB is this:
- Since paint offsets are positive, we don't need to do decycling work in each paint visit.
- Only paints that can "jump back" (PaintColrLayers and PaintColrGlyphs) will consume a decycler node; and we have different decycler for those two.
- The depth counter however applies to all paints.

That's why we have them separate. Though, maybe the cost of updating the decycler on each paint is not measurable for you.

dfrg:
Right, this actually tracks traversal depth twice in the current COLR implementation. The decycler needs to check the array bounds anyway so that one is essentially free (and might be useful for VARC?). The additional check handles the full graph including nodes that can’t cycle.

behdad:
Yeah we use our decycler in COLRv1, VARC, and glyf composites so far.

behdad:
Notably GSUB/GPOS should NOT have a decycler since it's legit to recurse to the same lookup...

skrifa/src/decycler.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Returns an error when a cycle is detected or the max depth of the
+    /// traversal is exceeded. Otherwise, increases the current depth and
+    /// returns a guard object that will decrease the depth when dropped.
+    ///
+    /// The guard object derefs to the decycler, so it can be passed to
+    /// a recursive traversal function to check for cycles in descendent
+    /// nodes in a graph.
+    pub fn enter(&mut self, node_id: T) -> Result<DecyclerGuard<T, D>, DecyclerError> {
+        if self.depth < D {
+            if self.depth == 0 || self.node_ids[self.depth / 2] != node_id {
```
cmyr, line 57: is the idea here that if a cycle exists, it will eventually be the case that the second condition fails?

(nm, read linked wikipedia page)

cmyr (approved):
looks nice and clean.

skrifa/src/decycler.rs
dfrg: Exactly, this algorithm _eventually_ detects cycles. This is essentially a trade off: the happy path is fast and doesn't allocate but we don't catch cycles immediately for malformed data. We do seem to always catch them long before we hit the recursion limit so there's still a definite benefit with very little cost.

PR #1390, opened 2025-03-10
Collection of glyf/gvar performance improvements

dfrg:
Stems from an investigation where @behdad pointed out that our performance on Roboto Flex was... less than satisfying.

This PR provides the following improvements:

1. Don't use `HVAR` for computing deltas for phantom points. We're already processing deltas from `gvar` so just make use of those instead. There are some caveats to this:
    * FreeType rounds these differently so we need a bit of extra code to account for this.
    * There appear to be some fonts where HVAR/gvar don't agree on phantom point deltas. In these cases, we won't produce advance widths that match FreeType. 
        * I assume these are not spec-conforming and the only cases I've seen so far are the "slim-variable-ttf" fonts in Noto.
2. More generally, don't read anything from `HVAR` when we have an empty location.
3. When reading a tuple variation, defer parsing of point numbers and counting of deltas. We don't need this information if the computed scalar is 0 for a given tuple and this is quite common in some fonts (hi Roboto Flex!)
4. When parsing a delta stream, read each (typed) run into a slice and apply the deltas directly to a buffer given by the user. This avoids a lot of branching and significantly reduces the number of bounds checked reads.
5. Similarly, read simple glyph flags as a slice rather than reading each individually with a `Cursor`. 
6. Extract the inner `ComputedArray<Tuple>` from `SharedTuples`. We were reparsing this every time we needed a peak tuple.
7. Small bits of refactoring, code motion and inlining (or preventing inlining) driven by measurement.

The overall result is a measurable boost across the board but the primary gains come from variable fonts. For comparison, reading all glyphs from Roboto Flex (100x) on main takes about 890ms. With this patch, it's about 320ms.


dfrg:
re advance width diffs: the max diff is always -1 at font unit scale. When examining the fonts, it appears that tuples with a -1 delta for the second phantom point are being dropped from gvar. As mentioned, this only seems to occur for fonts `slim-variable-ttf` directories in Noto.

behdad:
Amazing work. Here's new benchmark numbers:
```
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
BM_Font/draw_glyph/RobotoFlex.ttf/ot                  0.331 ms        0.330 ms         2118
BM_Font/draw_glyph/RobotoFlex.ttf/ft                  0.851 ms        0.847 ms          827
BM_Font/draw_glyph/RobotoFlex.ttf/fontations          0.539 ms        0.537 ms         1298
BM_Font/draw_glyph/RobotoFlex.ttf/var/ot               1.54 ms         1.53 ms          409
BM_Font/draw_glyph/RobotoFlex.ttf/var/ft               3.01 ms         3.00 ms          234
BM_Font/draw_glyph/RobotoFlex.ttf/var/fontations       2.18 ms         2.17 ms          316
```

drott:
Testing this in https://chromium-review.googlesource.com/c/chromium/src/+/6347556

read-fonts/src/tables/variations.rs
```diff
@@ -275,6 +300,7 @@ impl<'a> PackedPointNumbers<'a> {
     }
 
     /// the number of bytes to encode the packed point numbers
+    #[inline(never)]
```
cmyr, line 303: ?

read-fonts/src/tables/variations.rs
```diff
@@ -191,7 +196,26 @@ impl<'a> TupleVariationHeader<'a> {
...
+                },
+                Tuple {
+                    values: self.data.read_array(end_range).unwrap(),
+                },
+            )
+        })
+    }
+
     /// Compute the actual length of this table in bytes
+    #[inline(always)]
```
cmyr, line 218: huh, where is this even used?

read-fonts/src/tables/variations.rs
```diff
@@ -661,6 +688,7 @@ impl<'a, T> TupleVariationIter<'a, T>
 where
     T: TupleDelta,
 {
+    #[inline(never)]
```
cmyr, line 691: I'm curious about what inline(never) is buying us?

read-fonts/src/tables/variations.rs
```diff
@@ -850,11 +884,180 @@ where
...
+            (
+                self.shared_point_numbers.clone().unwrap_or_default(),
+                self.serialized_data,
+            )
+        }
     }
 }
 
+impl TupleVariation<'_, GlyphDelta> {
+    pub fn accumulate_dense_deltas<D: PointCoord>(
```
cmyr, line 911: If it's public API I'd love some docs 👼

read-fonts/src/tables/variations.rs
```diff
@@ -850,11 +884,180 @@ where
...
+                if let Some(delta) = deltas.get_mut(ix) {
+                    delta.y += D::from_fixed(Fixed::from_i32(new_delta) * scalar);
+                }
+            })?;
+        }
+        Ok(())
+    }
+}
+
+fn read_dense_deltas<T>(
```
cmyr, line 978: ditto, given that we're adding this fn specifically because to address performance concerns (and thus it was not the most obvious implementation approach) I think it's worth noting that, and explaining briefly why it's better.

cmyr (approved):
Looks good; I haven't actually run benchmarks so I am deferring to you on the performance aspect but overall these changes make sense in terms of avoiding redundant reads and otherwise reducing unnecessary work.

A few questions & notes inline, but happy to see this get in!

drott:
Can we get this landed and released? I would like to re-evaluate crbug.com/402451027 with this.

dfrg:
> Can we get this landed and released? I would like to re-evaluate crbug.com/402451027 with this.

Yes, I’ll get a release out this evening.

read-fonts/src/tables/variations.rs
dfrg: Curiously, I added these to generate finer grained profiles, but ended up with consistent and measurable performance gains due to the lack of inlining. I'm chalking this up to sometimes the optimizer blows the inlining budget before it gets deep enough into the call graph to reach functions where const prop actually makes a huge difference.

read-fonts/src/tables/variations.rs
dfrg: This is used when parsing the tuple variation headers, each of which is variable size based on the header flags. Pretty sure you wrote this code :)

read-fonts/src/tables/variations.rs
dfrg: see the explanation above

read-fonts/src/tables/variations.rs
dfrg: oops, I totally meant to write docs for these before requesting review. done now :)

PR #1391, opened 2025-03-10
In ift_graph util add option to include patch paths in the output graph.

garretrieger:


incremental-font-transfer/src/bin/ift_graph.rs
```diff
@@ -199,15 +203,24 @@ fn to_next_font(
         .expect("Patch application failed."))
 }
 
+#[derive(Clone, Default, Ord, PartialEq, PartialOrd, Eq)]
+struct NodeName(String);
```
cmyr, line 207: Are these frequently cloned? If so a type other than `String` might be preferable.

PR #1393, opened 2025-03-13
[klippa] subset COLR and CPAL

qxliu76:


klippa/src/colr.rs
```diff
@@ -0,0 +1,1708 @@
...
+            .map_err(|_| SubsetError::SubsetTableError(Colr::TAG))?;
+
+        //LayerList offset pos = 18
+        if let Some(layer_list) = self
+            .layer_list()
+            .transpose()
+            .map_err(|_| SubsetError::SubsetTableError(Colr::TAG))?
+        {
+            match Offset32::serialize_subset(&layer_list, s, plan, (), 18) {
+                Ok(()) | Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY) => (),
```
cmyr, line 83: Is the distinction between Ok(()) and `Err(ERROR_EMPTY)` ever meaningful? If it isn't, I think you should update `Offset32::serialize_subset` to check this condition and just return `Ok(())`.

klippa/src/colr.rs
```diff
@@ -0,0 +1,1708 @@
...
+                )
+                .map_err(|_| SubsetError::SubsetTableError(Colr::TAG))?;
+            }
+        }
+        Ok(())
+    }
+}
+
+// serialize header and V0 tables, format is decided by subset_to_v0 flag
+fn serialize_v0(
```
cmyr, line 129: if this has a direct parallel in harfbuzz, it would be useful to link to it here.

klippa/src/lib.rs
```diff
@@ -549,18 +558,79 @@ impl Plan {
...
+
+                // if DeltaSetIndexMap exists, we need to use deltaset index instead of var_idx
+                if var_index_map.is_some() {
+                    let (mut new_deltaset_idx_varidx_map, mut deltaset_idx_delta_map) =
+                        remap_delta_set_indices(
+                            &delta_set_indices,
+                            &deltaset_idx_var_idx_map,
+                            &self.colr_varidx_delta_map,
+                        );
+                    std::mem::swap(
```
cmyr, line 627: if we aren't going to use the old value, it would be more idiomatic to use `mem::replace` instead of `mem::swap`.

klippa/src/lib.rs
```diff
@@ -945,6 +1040,18 @@ fn subset_table<'a>(
             .map_err(|_| SubsetError::SubsetTableError(Cmap::TAG))?
             .subset(plan, font, s, builder),
 
+        Colr::TAG => font
+            .colr()
+            .map_err(|_| SubsetError::SubsetTableError(Colr::TAG))?
+            .subset(plan, font, s, builder),
+
+        //T ODO: if SVG is present and we support subsetting SVG table, pass through CPAL table
```
cmyr, line 1048:
```suggestion
        //TODO: if SVG is present and we support subsetting SVG table, pass through CPAL table
```

cmyr (approved):
The classic trick of "make the PR so big that nobody has the patience to be maximally nitpicky", eh?

Overall looks fine? I'm not too familiar with the overall logic of subsetting, and I'm avoiding making many comments about things that are not particularly idiomatic since the code feels like a pretty literal port from CPP and so is sort of necessarily not idiomatic..

PR #1395, opened 2025-03-17
Reduce typo count

waywardmonkeys:
This includes changes to use "Unicode" as a proper noun more often as well as to use the term "codepoint" rather than the occasional usage of "code points".


rsheeter:
Thank you :)

PR #1397, opened 2025-03-17
[klippa] subset GDEF table step 1: collect/map variation indices

qxliu76:


garretrieger (approved):

klippa/src/gdef.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn collect_variation_indices(&self, plan: &Plan, varidx_set: &mut IntSet<u32>) {
+        let Ok(coverage) = self.coverage() else {
+            return;
+        };
+
+        let lig_glyphs = self.lig_glyphs();
+        for (gid, lig_glyph) in coverage.iter().zip(lig_glyphs.iter()) {
+            if lig_glyph.is_err() {
+                return;
+            }
```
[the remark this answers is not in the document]
cmyr: (also, do you want to return if any item is an error, instead of continuing? probably doesn't matter either way, an error implies something wrong with the font)

klippa/src/gdef.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+
+impl CollectVaritionaIndices for LigGlyph<'_> {
+    fn collect_variation_indices(&self, plan: &Plan, varidx_set: &mut IntSet<u32>) {
+        for caret in self.caret_values().iter() {
+            if caret.is_err() {
+                return;
+            }
```
cmyr, line 44: ditto

klippa/src/gdef.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl CollectVaritionaIndices for CaretValueFormat3<'_> {
+    fn collect_variation_indices(&self, plan: &Plan, varidx_set: &mut IntSet<u32>) {
+        if let Ok(device) = self.device() {
+            device.collect_variation_indices(plan, varidx_set);
+        }
+    }
+}
```
cmyr, line 64: I might just inline this into the impl above, assuming that's the only place that gets here..

klippa/src/gdef.rs
```diff
@@ -0,0 +1,130 @@
...
+pub(crate) trait CollectUsedMarkSets {
+    fn collect_used_mark_sets(&self, plan: &Plan, used_mark_sets: &mut IntSet<u16>);
+}
+
+impl CollectUsedMarkSets for Gdef<'_> {
+    fn collect_used_mark_sets(&self, plan: &Plan, used_mark_sets: &mut IntSet<u16>) {
+        let Some(Ok(mark_glyph_sets)) = self.mark_glyph_sets_def() else {
+            return;
+        };
+        mark_glyph_sets.collect_used_mark_sets(plan, used_mark_sets);
```
cmyr, line 75: because the logic is so simple I'd just express this as `if let` I think?
```suggestion
        if let Some(Ok(mark_glyph_sets)) = self.mark_glyph_sets_def() else {
            mark_glyph_sets.collect_used_mark_sets(plan, used_mark_sets);
        }
```

klippa/src/gdef.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+}
+
+impl CollectUsedMarkSets for MarkGlyphSets<'_> {
+    fn collect_used_mark_sets(&self, plan: &Plan, used_mark_sets: &mut IntSet<u16>) {
+        for (i, coverage) in self.coverages().iter().enumerate() {
+            if coverage.is_err() {
```
cmyr, line 82: same as earlier

read-fonts/src/tables/layout.rs
```diff
@@ -182,6 +194,28 @@ impl CoverageFormat1<'_> {
...
+    pub fn intersects(&self, glyphs: &IntSet<GlyphId>) -> bool {
+        let glyph_count = self.glyph_count() as u32;
+        let num_bits = 32 - glyph_count.leading_zeros();
+        if glyph_count > (glyphs.len() as u32) * num_bits / 2 {
+            for g in glyphs.iter() {
+                if self.get(g).is_some() {
+                    return true;
+                }
+            }
+            false
```
cmyr, line 209:
```suggestion
            glyphs.iter().any(|g| self.get(g).is_some())
```

read-fonts/src/tables/layout.rs
```diff
@@ -182,6 +194,28 @@ impl CoverageFormat1<'_> {
...
+        } else {
+            for g in self.glyph_array() {
+                if glyphs.contains(GlyphId::from(g.get())) {
+                    return true;
+                }
+            }
+            false
```
cmyr, line 216:
```suggestion
            self.glyph_array().iter().any(|g| glyphs.contains(GlyphId::from(g.get())))
```

read-fonts/src/tables/layout.rs
```diff
@@ -204,12 +238,39 @@ impl CoverageFormat2<'_> {
...
             })
     }
+
+    /// Returns if this table contains at least one glyph in the 'glyphs' set.
+    #[cfg(feature = "std")]
+    pub fn intersects(&self, glyphs: &IntSet<GlyphId>) -> bool {
+        let range_count = self.range_count() as u32;
+        let num_bits = 32 - range_count.leading_zeros();
+        if range_count > (glyphs.len() as u32) * num_bits / 2 {
+            for g in glyphs.iter() {
```
cmyr, line 248: this can use `any()` too?

cmyr (approved):
lots of little nits but the logic looks good!

PR #1401, opened 2025-03-17
Fix buffer overflow in font builder.

garretrieger:
There is code for modifying 'head' for checksum computation which didn't ensure there was enough bytes in the head table.

Fixes fuzzer crash: https://oss-fuzz.com/testcase-detail/5496811662147584


rsheeter (approved):
My programming 101 instructor is mad about all the literal 8's and 12's ;)

garretrieger:
> My programming 101 instructor is mad about all the literal 8's and 12's ;)

Might as well fix those while I'm here.

anthrotype:
> There is code for modifying 'head' for checksum computation which didn't ensure there was enough bytes in the head table.

oops my fault. I naively thought that for a table to qualify as `head` it ought to have enough bytes.

PR #1403, opened 2025-03-18
[skrifa] Synthesize glyph name if it's empty string

behdad:
Not the most idiomatic, but does the job for me.

### DIFF
--- a/skrifa/src/glyph_name.rs
+++ b/skrifa/src/glyph_name.rs
@@ -93,6 +93,10 @@ impl<'a> GlyphNames<'a> {
                 .and_then(|sid| GlyphName::from_cff_sid(cff, sid)),
             _ => None,
         };
+        // If name is empty string, synthesize it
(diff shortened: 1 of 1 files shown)


skrifa/src/glyph_name.rs
```diff
@@ -93,6 +93,10 @@ impl<'a> GlyphNames<'a> {
                 .and_then(|sid| GlyphName::from_cff_sid(cff, sid)),
             _ => None,
         };
+        // If name is empty string, synthesize it
+        if name.as_ref().map_or(true, |s| s.is_empty()) {
```
dfrg, line 97:
```suggestion
        if name.as_ref().is_none_or(|s| s.is_empty()) {
```
Apparently not quite idiomatic enough for clippy :)

dfrg (approved):
Looks good once the suggested change is made to pass CI. 

We should make a similar change to the iterator code for consistency. Feel free to have a go at that in this PR or we can do it as a follow up.

PR #1412, opened 2025-03-20
Prototype implementation of the IFT specific subset of URI templates.

garretrieger:
In the IFT spec we're restricting the URI template to a small subset of the general URI template syntax (see: https://github.com/w3c/IFT/pull/263). This makes it reasonable to write our own template expansion functionality instead of relying on a general parser implementation.

This takes a pass at implementing one (to see how it looks in practice), but does not yet hook it up to the rest of the IFT client. Ultimately it should allow us to drop the dependency on "uri-template-system".

### DIFF
--- a/incremental-font-transfer/src/uri_templates.rs
+++ b/incremental-font-transfer/src/uri_templates.rs
@@ -8,7 +8,8 @@
 //!
 //! By implementing our own we avoid pulling in a much larger general purpose template expansion library
 //! and improve performance versus a more general implementation.
-use std::{ops::RangeInclusive, sync::OnceLock};
+use std::fmt::Write;
+use std::sync::OnceLock;
 
 enum ParseState {
     // Literal parsing
@@ -108,7 +109,7 @@ impl OutputBuffer {
     /// - Something invalid encountered.
     fn handle_literal(&mut self, byte_info: &ByteInfo) -> Result<ParseState, UriTemplateError> {
         match byte_info.literal_class {
-            LiteralClass::Invalid | LiteralClass::EndExpression => Err(UriTemplateError),
+            LiteralClass::Invalid => Err(UriTemplateError),
             LiteralClass::Percent => {
                 self.append(byte_info.value);
                 Ok(ParseState::LiteralPercentEncoded(Digit::One))
@@ -118,10 +119,9 @@ impl OutputBuffer {
                 self.append(byte_info.value);
                 Ok(ParseState::Literal)
             }
-            LiteralClass::LiteralPercentEncoded => {
-                self.append_percent_encoded(byte_info.value);
-                Ok(ParseState::Literal)
-            }
+            LiteralClass::LiteralPercentEncoded => self
+                .append_percent_encoded(byte_info.value)
+                .map(|_| ParseState::Literal),
         }
     }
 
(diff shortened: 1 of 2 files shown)


incremental-font-transfer/src/uri_templates.rs
```diff
@@ -0,0 +1,682 @@
...
+///
+/// Specification: <https://w3c.github.io/IFT/Overview.html#uri-templates>
+///
+/// IFT uri templates are a subset of the more general <https://datatracker.ietf.org/doc/html/rfc6570>
+/// uri templates. Notably, only level one substitution expressions are supported and there are a fixed
+/// set of variables used in the expansion (id, id64, d1, d2, d3, and d4).
+///
+/// All arguments are assumed to be utf8 encoded strings.
+pub(crate) fn expand_template(
+    template_string: &str,
```
cmyr, line 69: would it make more sense to pass this in as bytes?

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    // Appends a single byte to the output.
+    fn append(&mut self, byte: u8) {
+        self.0.push(byte.into());
+    }
+
+    // Appends the percent encoded representation (%XX) of a byte to the output.
+    fn append_percent_encoded(&mut self, byte: u8) {
+        self.0.push_str(&format!("%{:02X}", byte));
```
cmyr, line 254: `format!` allocates; you can skip that by writing directly into the buffer, like:

```rust
write!(&mut self.0, "%{:02X}", byte);
```

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -0,0 +1,682 @@
...
+            is_varchar: Self::is_varchar(value),
+            value,
+        }
+    }
+
+    /// Returns true if byte is a hexdig.
+    ///
+    /// As defined here: <https://datatracker.ietf.org/doc/html/rfc6570#section-1.5>
+    fn is_hexdig(byte: u8) -> bool {
+        DIGIT.contains(&byte) || HEX_ALPHA_UPPER.contains(&byte) || HEX_ALPHA_LOWER.contains(&byte)
```
cmyr, line 293: could you use [`u8::is_ascii_hexdigit`](https://doc.rust-lang.org/1.85.0/std/primitive.u8.html#method.is_ascii_hexdigit)?

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -0,0 +1,682 @@
...
+    }
+}
+
+const NUM_U8S: usize = 256;
+const ALPHA_UPPER: RangeInclusive<u8> = 0x41..=0x5A;
+const ALPHA_LOWER: RangeInclusive<u8> = 0x61..=0x7A;
+const DIGIT: RangeInclusive<u8> = 0x30..=0x39;
+const CTL_AND_SPACE: RangeInclusive<u8> = 0x00..=0x20;
+const HEX_ALPHA_UPPER: RangeInclusive<u8> = 0x41..=0x46;
+const HEX_ALPHA_LOWER: RangeInclusive<u8> = 0x61..=0x66;
```
cmyr, line 316: many of these are available as methods on `u8`.

- `u8::is_ascii_lowercase`
- `u8::is_ascii_uppercase`
- `u8::is_ascii_digit`
- `u8::is_ascii_hexdigit`

cmyr (approved):
👍

incremental-font-transfer/src/uri_templates.rs
garretrieger: It's important that we're operating on the bytes of a utf8 encoding, and it's my understanding that being a &str implies the contents are utf8 encoded.

PR #1415, opened 2025-03-20
[skrifa] synthesize glyph names for empty strings in iterator

### LINKED ISSUE #1410: [skrifa] Skip empty glyph names in iterator
The mapping code does that now. We should make a similar change to the iterator code for consistency.

_Originally posted by @dfrg in https://github.com/googlefonts/fontations/pull/1403#pullrequestreview-2698506838_

dfrg:
Follow up to #1403 which did this for `GlyphNames::get()`.

Fixes #1410

### DIFF
--- a/skrifa/src/glyph_name.rs
+++ b/skrifa/src/glyph_name.rs
@@ -308,7 +308,7 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


behdad:
LGTM. Thanks.

PR #1417, opened 2025-03-20
Drop uri-template-system and instead use our own template expander in IFT client.

garretrieger:
- Also syncs up to some recent spec changes with regards to handling undefined expansion variables.

### DIFF
--- a/incremental-font-transfer/src/uri_templates.rs
+++ b/incremental-font-transfer/src/uri_templates.rs
@@ -86,10 +86,11 @@ pub(crate) fn expand_template(
     let mut id64_string_encoded: OutputBuffer = Default::default();
     for byte in id64_string.as_bytes() {
         let byte_info = &byte_info_map[*byte as usize];
-        if byte_info.is_unreserved {
-            id64_string_encoded.append(*byte)
-        } else {
-            id64_string_encoded.append_percent_encoded(*byte).unwrap()
+        match byte_info {
+            ByteInfo::CopiedLiteralUnreserved | ByteInfo::CopiedLiteralHexDigit => {
+                id64_string_encoded.append(*byte)
+            }
+            _ => id64_string_encoded.append_percent_encoded(*byte).unwrap(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


incremental-font-transfer/src/uri_templates.rs
```diff
@@ -8,9 +8,13 @@
 //!
 //! By implementing our own we avoid pulling in a much larger general purpose template expansion library
 //! and improve performance versus a more general implementation.
+use data_encoding::BASE64URL;
+use data_encoding_macro::new_encoding;
```
cmyr, line 12: I wonder why your imports never end up getting combined? Maybe this is some setting in rust-analyzer..

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -65,13 +66,58 @@ impl std::error::Error for UriTemplateError {}
 /// uri templates. Notably, only level one substitution expressions are supported and there are a fixed
 /// set of variables used in the expansion (id, id64, d1, d2, d3, and d4).
 ///
-/// All arguments are assumed to be utf8 encoded strings.
+/// template_string is assumed to be a utf8 encoded string.
```
cmyr, line 69: this isn't really an assumption, it's an invariant (there's no way to get a non-utf8 `&str` without unsafe)

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -65,13 +66,58 @@ impl std::error::Error for UriTemplateError {}
...
+fn count_leading_zeroes(id: &[u8]) -> usize {
+    let mut leading_bytes = 0;
+    for b in id {
+        if *b != 0 {
+            break;
+        }
+        leading_bytes += 1;
+    }
+    // Always keep at least one digit.
+    leading_bytes.min(id.len() - 1)
```
cmyr, line 112: `leading_zeros = id.iter().take_while(|b| *b == 0).count();`?

cmyr (approved):
👍

incremental-font-transfer/src/uri_templates.rs
cmyr: oops 😅

Issue #1418, opened 2025-03-21
[Skrifa/COLRv1] Pass `composite_mode` to `pop_layer` instead of or as well as `push_layer`


behdad:
From my current investigation into malloc usage of various font drawing and painting libraries and technologies:
https://docs.google.com/document/d/1S47f3E--yqvFdG7lmmufxRoFi_wMzotC03v8UvS_p54/edit?tab=t.0#heading=h.m9bdd1pt4wv0

Quoting:

When processing a COLRv1 PaintComposite, the Skrifa API makes a push_layer(composite_mode) followed later by a pop_layer(). In HarfBuzz paint API, and I assume in most graphics libraries though, the composite_mode is needed during the layer popping. HarfBuzz API has push_group(), later followed up by pop_group(composite_mode). To gap this API mismatch, hb-fontations layer keeps a stack of the currently-pushed composite modes.

This discrepancy and need for allocation can be bridged by changing the Skrifa API to match HarfBuzz’s. Failing that (if there is a good reason to provide the composite_mode during the layer pushing), I suggest Skrifa pass composite_mode to both push_layer and pop_layer. This would obviate the need for this particular vector allocation.

drott:
Skia expects the composite mode at layer creation at least from the API shape point of view, even though the actual blending is only performed later, see `SkCanvas::saveLayer`. So then we could pass the mode in both, or design the callback API to with optional callbacks that the COLRv1 implementation in Skrifa chooses from, depending on whichever is set.

```
void ColorPainter::push_layer(uint8_t compositeMode) {
    SkPaint paint;
    paint.setBlendMode(ToSkBlendMode(compositeMode));
    fCanvas.saveLayer(nullptr, &paint);
}

void ColorPainter::pop_layer() { fCanvas.restore(); }
```

behdad:
Both would be ideal. Thanks.

dfrg:
We could add a new `pop_layer_with_mode` that defaults to calling `pop_layer` without the composition mode. It’s a bit ugly but would avoid a breaking change. 

We also allocate a Vec to store the color stops for every gradient paint. I’d like to replace this with our internal SmallVec type which should make us malloc free for most COLRv1 glyphs.

rsheeter:
> I’d like to replace this with our internal SmallVec type which should make us malloc free for most COLRv1 glyphs.

Sounds like low hanging fruit indeed! Any reason why we shouldn't?

behdad:
[...]

PR #1419, opened 2025-03-21
[Skrifa/COLRv1] Add pop_layer_with_mode()

behdad:
Fixes https://github.com/googlefonts/fontations/issues/1418

### DIFF
--- a/skrifa/src/color/mod.rs
+++ b/skrifa/src/color/mod.rs
@@ -263,7 +263,22 @@ pub trait ColorPainter {
     /// Open a new layer, and merge the layer down using `composite_mode` when
     /// [`pop_layer`](ColorPainter::pop_layer) is called, signalling that this layer is done drawing.
     fn push_layer(&mut self, composite_mode: CompositeMode);
-    fn pop_layer(&mut self);
+
+    /// Merge the pushed layer down using `composite_mode` passed to the matching
+    /// [`push_layer`](ColorPainter::push_layer).
+    fn pop_layer(&mut self) {}
+
+    /// Alternative version of [`push_layer`](ColorPainter::push_layer) where the
+    /// `composite_mode` is also passed to the method. This is useful for
+    /// graphics libraries that need the compositing mode at layer pop time
+    /// and do not want to manually track the mode.
+    ///
+    /// Only one of [`pop_layer`](ColorPainter::pop_layer) or this method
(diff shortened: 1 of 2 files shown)


skrifa/src/color/mod.rs
```diff
@@ -263,7 +263,10 @@ pub trait ColorPainter {
     /// Open a new layer, and merge the layer down using `composite_mode` when
     /// [`pop_layer`](ColorPainter::pop_layer) is called, signalling that this layer is done drawing.
     fn push_layer(&mut self, composite_mode: CompositeMode);
-    fn pop_layer(&mut self);
+    fn pop_layer(&mut self) {}
+    fn pop_layer_with_mode(&mut self, _composite_mode: CompositeMode) {
+        self.pop_layer();
+    }
```
dfrg, line 269: This part looks right. Just need to update all of the call sites to use the new method.

Would also be nice to have a short doc comment mentioning why this is useful and that only one of `pop_layer` or `pop_layer_with_mode` should be implemented.

behdad: > This part looks right. Just need to update all of the call sites to use the new method.

What do you have in mind? traverse calling the new one? NullPen etc should be fine I assume.

behdad:
@dfrg PTAL. Not sure what traverse needs updating.

skrifa/src/color/mod.rs
dfrg: I believe it's just the two `pop_layer` calls here that need to be changed to `pop_layer_with_mode`: https://github.com/googlefonts/fontations/blob/c32c0a6442371d0206f82a4a981b3854f313aaae/skrifa/src/color/traversal.rs#L590

skrifa/src/color/traversal.rs
```diff
@@ -689,6 +689,10 @@ mod tests {
         fn pop_layer(&mut self) {
             // nop
         }
+
+        fn pop_layer_with_mode(&mut self, _composite_mode: CompositeMode) {
```
dfrg, line 693: This one shouldn't be necessary though

behdad: Sure.

skrifa/src/color/mod.rs
behdad: Ah, right. Fixed.

PR #1421, opened 2025-03-21
[skrifa] add 8k and 16k buckets to stack allocator

dfrg:
Let's try to keep some more work off the heap.

Motivated by our excessive allocations for NotoSansCJK-VF as documented by @behdad here: https://docs.google.com/document/d/1S47f3E--yqvFdG7lmmufxRoFi_wMzotC03v8UvS_p54/edit?tab=t.0#heading=h.bfj7urloz3oe

JMM


behdad:
Ah, Rust zeroes stack memory?

dfrg:
> Ah, Rust zeroes stack memory?

Yeah, you can get around it but that requires unsafe :(

PR #1423, opened 2025-03-21
Bypass variation processing for default location

dfrg:
Adds an internal `LocationRef::effective_coords()` method that returns an empty slice when all coordinates are 0. This allows us to check for `coords.is_empty()` to skip processing variations in many places.

This is potentially a nice performance boost when calling `font.axes().location(...)` with default settings.


behdad:
LGTM.

PR #1426, opened 2025-03-21
In URI Template impl build the byte info map at compile time

garretrieger:
- This avoids needing to use OnceLock and removes the run time initialization of the byte info map on first use.
- Also do some refactoring to reduce the amounts of arguments being passed around in the handle_* methods.

### DIFF
--- a/incremental-font-transfer/src/uri_templates.rs
+++ b/incremental-font-transfer/src/uri_templates.rs
@@ -380,23 +380,17 @@ impl ByteInfo {
 
 // This macro generates the byte info array at compile time.
 const NUM_U8S: usize = u8::MAX as usize + 1;
-macro_rules! generate_byte_info_array {
-    () => {{
-        const ARRAY: [ByteInfo; NUM_U8S] = {
-            let mut info = [ByteInfo::Invalid; NUM_U8S];
-            let mut i = 0;
-            while i < NUM_U8S {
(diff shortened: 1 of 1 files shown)


dfrg (approved):
LGTM

incremental-font-transfer/src/uri_templates.rs
```diff
@@ -343,26 +378,26 @@ impl ByteInfo {
...
+            }
+            info
+        };
+        ARRAY
+    }};
 }
 
+/// This maps each possiblue byte (u8) value to an enum which classifies how that value is handled during expansion.
+static BYTE_INFO_MAP: [ByteInfo; NUM_U8S] = generate_byte_info_array!();
+
```
nicoburns, line 400: The macro isn't actually doing anything here (other than outputting the literal source code inside it). And in general, `macro_rules` macros can't/don't evaluate procedural code. This is working with just because of const evaluation.

```suggestion
static BYTE_INFO_MAP: [ByteInfo; NUM_U8S] = {
    let mut info = [ByteInfo::Invalid; NUM_U8S];
    let mut i = 0;
    while i < NUM_U8S {
        info[i] = ByteInfo::new(i as u8);
        i += 1;
    }
    info
};
```

garretrieger: Ah good catch, fixed.

PR #1427, opened 2025-03-21
Closure reachability

cmyr:
This reworks our glyph closure code to much more closely match fonttools.

This is tricky, because fonttools' glyph closure doesn't have unit tests; it's all tested through tests of the general subsetting code, and that doesn't exist yet here.

In any case, this fixes.... a single funny crater issue. yay.

- there's one thing from the python impl that is missing here, and a bit tricky to express, and that's the logic around the 'may_have_non_1to1` method; i'm not sure if this is an optimization or if it's required for correctness, and in any case it will be a bit tricky to express in the rust version (because we only know lookup ids, and not the lookup type, at that point in the code)

In any case, this will likely need to be revisited again somewhat when the subsetting work gets to that stage, so I really don't think it's worth much more additional speculative work at this point.


read-fonts/src/tables/gsub/closure.rs
```diff
@@ -99,52 +235,43 @@ impl Gsub<'_> {
...
-            .try_for_each(|t| t?.add_reachable_glyphs(glyphs))
+    fn add_reachable_glyphs(&self, ctx: &mut ClosureCtx<'_>) -> Result<(), ReadError> {
+        self.iter().try_for_each(|t| t?.add_reachable_glyphs(ctx))
     }
 }
 
 impl GlyphClosure for SingleSubst<'_> {
-    fn add_reachable_glyphs(&self, glyphs: &mut IntSet<GlyphId16>) -> Result<(), ReadError> {
+    fn add_reachable_glyphs(&self, ctx: &mut ClosureCtx<'_>) -> Result<(), ReadError> {
         for (target, replacement) in self.iter_subs()? {
```
qxliu76, line 272: Same as Harfbuzz, we could apply some optimization here: https://github.com/harfbuzz/harfbuzz/blob/d3e4977c708837d32b6d6e8f2ed4ea3045bce09a/src/OT/Layout/GSUB/SingleSubstFormat2.hh#L42

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -241,247 +368,475 @@ impl GlyphClosure for LigatureSubstFormat1<'_> {
...
+            return Ok(());
+        };
+
+        let classdef = self.input_class_def()?;
+        let our_classes = make_class_set(ctx.glyphs(), &classdef);
+        for (class_i, seq) in self
+            .rule_sets()
             .enumerate()
-            .filter_map(|(i, seq)| seq.filter(|_| input_classes.contains(i as u16)))
+            .filter_map(|(i, seq)| seq.map(|seq| (i as u16, seq)))
```
qxliu76, line 703: should we filter ruleSets by coverage first? see: https://github.com/fonttools/fonttools/blob/a6abbac37b2ef272904a9d8134272901138ea8a9/Lib/fontTools/subset/__init__.py#L1217

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -241,247 +368,475 @@ impl GlyphClosure for LigatureSubstFormat1<'_> {
...
-            .chain(self.input_coverages().iter())
-            .chain(self.lookahead_coverages().iter())
-        {
-            if !coverage?.iter().any(|gid| glyphs.contains(gid)) {
-                return Ok(());
-            }
+            .chain(backtrack.into_iter().flat_map(|x| x.iter()))
+            .chain(lookahead.into_iter().flat_map(|x| x.iter()))
+            .all(|cov| {
+                cov.map(|cov| cov.iter().any(|gid| glyphs.contains(gid)))
```
qxliu76, line 771: Suggest to use this API to check if a coverage table intersects a glyph set: https://github.com/googlefonts/fontations/blob/ae80c78ff4a886539e66306962a8a7cb71505e1e/read-fonts/src/tables/layout.rs#L199

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -241,247 +368,475 @@ impl GlyphClosure for LigatureSubstFormat1<'_> {
...
+                        // it with the full current glyph set
+                        None
+                    } else if sequence_idx == 0 {
+                        Some(IntSet::from([coverage.iter().nth(i).unwrap()]))
+                    } else {
+                        Some(IntSet::from([rule.input_sequence()
+                            [sequence_idx as usize - 1]
+                            .get()]))
+                    };
+                    ctx.add_todo(lookup_id, active_glyphs);
```
qxliu76, line 566: After some thing, I would suggest doing recursion here instead of adding new todos and handle them in the final step, because a recursion lookup can keep growing the todos list, which could cause a timeout. By using recursion, we can add a hard limit on recursion depth. If &LookupList is added as a struct member, we should be able to just call glyph_closure on the child Lookup here. See also harfbuzz: https://github.com/harfbuzz/harfbuzz/blob/18ab0f552279159920d2cecbf97052e21fcb8c9a/src/hb-ot-layout-gsubgpos.hh#L92, https://github.com/harfbuzz/harfbuzz/blob/18ab0f552279159920d2cecbf97052e21fcb8c9a/src/hb-ot-layout-gsubgpos.hh#L174

read-fonts/src/tables/gsub/closure.rs
cmyr: hmm this takes `GlyphId` but the current code uses `GlyphId16`, I'll leave this for you to resolve.

cmyr:
@qxliu76 good notes! As you mentioned in chat it sounds like you're going to need this soon for the layout subsetting so I'm going to merge it as-is, and you can patch it up for that work? It will also make sense to make those changes in a context where there is more thorough testing.

Issue #1431, opened 2025-03-22
[read-fonts] Hook up cmap format13 subtables


behdad:
Currently only format4 / format12 are hooked up:
```
    pub fn map_codepoint(&self, codepoint: impl Into<u32>) -> Option<GlyphId> {
        let codepoint = codepoint.into();
        for record in self.encoding_records() {
            if let Ok(subtable) = record.subtable(self.offset_data()) {
                if let Some(gid) = match subtable {
                    CmapSubtable::Format4(format4) => format4.map_codepoint(codepoint),
                    CmapSubtable::Format12(format12) => format12.map_codepoint(codepoint),
                    _ => None,
                } {
                    return Some(gid);
                }
            }
        }
        None
    }
```

PR #1433, opened 2025-03-23
[read-fonts/cmap] Towards supporting format13 subtables

behdad:
Doesn't work, but a start. I don't know why. I also didn't update the tests.

Fixes https://github.com/googlefonts/fontations/issues/1431


dfrg:
This is looking pretty good for reading cmap13 (though I wonder if some of this code can be shared with 12). Skrifa has its own logic for selecting the appropriate subtable rather than just enumerating all records (https://github.com/googlefonts/fontations/blob/e64b757b887b020868c11d85d36e5cee2e30f86e/skrifa/src/charmap.rs#L335) and this will need to be integrated there as well.

behdad:
I see. I'll wait for you to get it to it I think. :)

dfrg:
Added a commit that unifies cmap 12 and 13 code and adds a couple tests for 13. Skrifa integration can be done in a followup.

dfrg (approved):

PR #1435, opened 2025-03-24
[Skrifa] Update documentation for mappings

drott:
Add a comment that mappings may contain glyph ids beyond maxp::num_glyphs.

Compare https://issues.chromium.org/405286754


drott:
CC @bungeman

rsheeter (approved):

PR #1436, opened 2025-03-24
Allow gvar offset size to be upgraded during IFT patch application.

garretrieger:
As mentioned in the spec: https://w3c.github.io/IFT/Overview.html#apply-glyph-keyed a glyph keyed patch application should upgrade the size of offsets in the patched table when needed. Implement this for gvar.

### DIFF
--- a/font-test-data/src/ift.rs
+++ b/font-test-data/src/ift.rs
@@ -3,6 +3,8 @@
 //! Used for incremental font transfer. Specification:
 //! <https://w3c.github.io/IFT/Overview.html>
 
+use std::iter::repeat;
+
 use font_types::{Int24, Tag, Uint24};
 
 use crate::{be_buffer, bebuffer::BeBuffer};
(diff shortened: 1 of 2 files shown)


font-test-data/src/ift.rs
```diff
@@ -3,6 +3,8 @@
 //! Used for incremental font transfer. Specification:
 //! <https://w3c.github.io/IFT/Overview.html>
 
+use std::iter::repeat;
```
cmyr, line 6: Totally absolute nit: I avoid importing free functions directly, and instead import the containing module and then call the functions through the module, because it makes it easier to understand what is being called. (e.g. I'd `use std::iter;` and then call `iter::repeat(0)`).

This used to actually be in the official style guide but I don't see it anymore, so really only make this change if you actually think it makes sense.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -449,18 +461,26 @@ fn patch_offset_array<'a, T: GlyphDataOffsetArray>(
...
 
 /// Classifies the different style of offsets that can be used in a data offset array.
+#[derive(Clone, Copy, PartialEq, Eq)]
 enum OffsetType {
     Long,
     ShortDivByTwo,
 }
 
 impl OffsetType {
+    fn max_representable_size(&self) -> u64 {
```
cmyr, line 477: because this is a copyable type, it's various methods could take `self` by value instead of by reference, but also nbd.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -500,9 +526,12 @@ trait GlyphDataOffsetArray {
         font_builder: &mut FontBuilder<'a>,
         data: impl Into<Cow<'a, [u8]>>,
         offsets: impl Into<Cow<'a, [u8]>>,
+        offset_type: OffsetType,
     ) -> Result<(), PatchingError>;
 }
 
+const LOCA_OFFSET_TYPES: [OffsetType; 2] = [OffsetType::ShortDivByTwo, OffsetType::Long];
```
cmyr, line 533: I'd make this an associated constant on `OffsetType`? (Are we expecting more offset types?)

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -511,6 +540,14 @@ impl GlyphDataOffsetArray for GlyfAndLoca<'_> {
         }
     }
 
+    fn available_offset_types(&self) -> impl Iterator<Item = OffsetType> {
+        // loca cannot change types so the only type available is what this already is.
+        LOCA_OFFSET_TYPES
+            .iter()
+            .copied()
+            .filter(|offset_type| *offset_type == self.offset_type())
+    }
```
cmyr, line 549: why not `std::iter::once(self.offset_type())`?

cmyr (approved):
some nits but overall looks good!

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: since it's only needed inside the available_offset_types() method I moved it into there.

PR #1438, opened 2025-03-25
[int-set] Impl Serialize, Deserialize

cmyr:
This is slightly tricky because we need to ensure on deserialization that the data is valid for the given domain.

We need these impls in order to be able to use int set more broadly.

Also: adding serde_json as a dev dependency to read-fonts caused a bunch of existing code to stop compiling in a way that really looks like a rustc bug; I've changed the offenders here but it makes this a noisier patch than I'd hoped.

### DIFF
--- a/read-fonts/src/collections/int_set/mod.rs
+++ b/read-fonts/src/collections/int_set/mod.rs
@@ -638,39 +638,18 @@ impl<'de, T: Domain> serde::Deserialize<'de> for IntSet<T> {
         D: serde::Deserializer<'de>,
     {
         let members = Membership::deserialize(deserializer)?;
-        // we need to verify that the values are valid for the given domain;
-        // we construct a temporary u32 intset so we can iterate the raw values.
-        let temp = IntSet::<u32>(members, PhantomData);
-        // an inverted set checks the domain during iteration, so invalid values
-        // are guarded against.
-        //
-        // HELPME GARRET: please confirm my understanding is correct
-        if temp.is_inverted() {
-            return Ok(IntSet(temp.0, PhantomData));
-        }
-        if T::is_continuous() {
-            let start = T::ordered_values().next().unwrap_or(0);
-            let end = T::ordered_values().next_back().unwrap_or(0);
+        let bits = match &members {
+            Membership::Inclusive(bit_set) => bit_set,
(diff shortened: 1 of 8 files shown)


read-fonts/src/collections/int_set/mod.rs
```diff
@@ -620,6 +624,56 @@ where
...
+        D: serde::Deserializer<'de>,
+    {
+        let members = Membership::deserialize(deserializer)?;
+        // we need to verify that the values are valid for the given domain;
+        // we construct a temporary u32 intset so we can iterate the raw values.
+        let temp = IntSet::<u32>(members, PhantomData);
+        // an inverted set checks the domain during iteration, so invalid values
+        // are guarded against.
+        //
+        // HELPME GARRET: please confirm my understanding is correct
```
cmyr, line 647: @garretrieger I got tripped up on this and am still not totally confident, let me know if this makes sense to you?

garretrieger: What you want to check is that the underlying bitset within the membership enum only contains values that are members of the domain. You'll want to do this in both the inclusive and exclusive (inverted) case.

read-fonts/src/collections/int_set/mod.rs
```diff
@@ -2536,4 +2624,47 @@ mod test {
...
+    fn serde_inverted() {
+        let mut set = IntSet::all();
+        set.remove_range(0u16..=420);
+        let bytes = serde_json::to_string(&set).unwrap();
+        assert!(bytes.len() < 5000, "sanity check serialization");
+        assert_eq!(roundtrip_json(&set).unwrap(), set)
+    }
+
+    #[test]
+    fn serde_inverted_out_of_domain() {
```
garretrieger, line 2651: Also add an out of domain test on:
- a non-continous domain (like EvenInts) 
- Inverted non-continous domain
- Inverted continous domain (should fail during deserialize with the changes I suggested above)

read-fonts/src/collections/int_set/mod.rs
cmyr: Right okay so I want some way to just iter the bits on the bitset, and check that they're in domain.

garretrieger (approved):

PR #1439, opened 2025-03-25
[IFT] Add CFF patching support to IFT client.

garretrieger:
CFF2 is not implemented yet, will be adding that next.

Context: https://github.com/w3c/IFT/issues/260


font-types/src/uint24.rs
```diff
@@ -55,6 +55,26 @@ impl From<Uint24> for usize {
     }
 }
 
+#[derive(Debug)]
```
cmyr, line 58: All public types should have docs here (we're lax on it elsewhere but we're quite good in font-types)

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -273,15 +284,31 @@ impl WritableOffset for u32 {
     }
 }
 
+impl WritableOffset for Uint24 {
```
cmyr, line 287: Maybe it would be worth taking a look at the overall design for how we do serialization in write-fonts? I'm wondering if it would be simpler for you to have some kind of buffer type (`BeBuffer` is another ref here) and then it would be trivial to have a method there like [`TableData::write`](https://github.com/googlefonts/fontations/blob/bfe1c5554e5281afdaa14f2b1447fe4034a0b69f/write-fonts/src/write.rs#L218), which would let you write any scalar in place.

cmyr (approved):
👍

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: Actually just realizing I can use the klippa serializer here too like I do during table reconstruction. I'll make that update and let's see how it looks.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -442,15 +468,49 @@ fn patch_offset_array<'a, T: GlyphDataOffsetArray>(
     let mut new_data = vec![0u8; total_data_size as usize];
     let mut new_offsets = vec![0u8; offsets_size];
     match new_offset_type {
-        OffsetType::ShortDivByTwo => synthesize_offset_array::<2, u16, _>(
+        // Bias 1 offsets
+        OffsetType::CffOne => synthesize_offset_array::<1, 1, 1, u8, _>(
```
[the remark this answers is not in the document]
garretrieger: Agreed, I've been trying to think of a better way of doing this, but still keeping these values as compile time constants. Another option would be to convert the offset type enum into a single int value and then in synthesize_offset_array convert that back into the enum.

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -264,160 +263,138 @@ fn retained_glyphs_total_size<T: GlyphDataOffsetArray>(
...
-        dest[..4].copy_from_slice(&data);
-    }
-}
-
-impl WritableOffset for Uint24 {
-    fn write_to(self, dest: &mut [u8]) {
-        let data: [u8; 3] = self.to_raw();
-        dest[..3].copy_from_slice(&data);
-    }
+struct OffsetArrayBuilder<'a, T: GlyphDataOffsetArray> {
```
cmyr, line 277: unless it's necessary for some reason (often involving associated types) in general if you can omit a bound on a generic in a type definition you should (since it then ends up being required everywhere else you might impl something for a type, including unrelated traits etc)

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -264,160 +263,138 @@ fn retained_glyphs_total_size<T: GlyphDataOffsetArray>(
...
 }
 
-impl WritableOffset for u16 {
-    fn write_to(self, dest: &mut [u8]) {
-        let data: [u8; 2] = self.to_raw();
-        dest[..2].copy_from_slice(&data);
-    }
-}
+impl<T: GlyphDataOffsetArray> OffsetArrayBuilder<'_, T> {
+    fn synthesize_offset_array<OffsetInfo: OffsetTypeInfo, OffsetType: Scalar + TryFrom<usize>>(
```
cmyr, line 287: at this point I'd probably just call it `build`?

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -540,36 +489,144 @@ enum OffsetType {
     Long,
 }
 
+trait OffsetTypeInfo {
+    fn width() -> usize;
+    fn divisor() -> usize;
+    fn bias() -> u32;
```
cmyr, line 495: maybe just make these into associated consts?

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -540,36 +489,144 @@ enum OffsetType {
...
 
     fn offset_width(&self) -> usize {
         match self {
-            Self::CffOne => 1,
-            Self::CffTwo => 2,
-            Self::CffThree => 3,
-            Self::CffFour => 4,
-            Self::ShortDivByTwo => 2,
-            Self::Long => 4,
+            Self::CffOne => CffOneInfo::width(),
```
cmyr, line 602: could we replace all this with an  `info()` method that returns the relevant `OffsetTypeInfo`?

incremental-font-transfer/src/glyph_keyed.rs
```diff
@@ -803,13 +857,14 @@ impl GlyphDataOffsetArray for Gvar<'_> {
         // is used to recalculate offsets as needed.
         let orig_bytes = self.as_bytes();
         let orig_size = orig_bytes.len();
-        let part2_offsets = offsets.into();
-        let part4_data = data.into();
+        //let part2_offsets = offsets.offset_array.into();
```
cmyr, line 860: noise

cmyr (approved):
this looks like an improvement, a few notes inline

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: good idea, done.

incremental-font-transfer/src/glyph_keyed.rs
garretrieger: info() would need to return a "impl OffsetTypeInfo" which won't work since that can't be used when different concrete types may be returned. I tried to figure out a way to get this to work but wasn't able to. Any ideas on how to do it?

garretrieger: Made this a little bit better by adding a concrete offset info instance into each enum. Still need the matches though.

PR #1440, opened 2025-03-26
[int-set] Remove use of Cell

cmyr:
This was something we'd wanted to do a while ago but now I'm motivated by clippy not liking `Cell` to be used in any type that could be stored in a {hash,btree}set.

### DIFF
--- a/read-fonts/src/collections/int_set/bitpage.rs
+++ b/read-fonts/src/collections/int_set/bitpage.rs
@@ -94,10 +94,12 @@ impl BitPage {
 
     /// Marks `(val % page width)` a member of this set and returns `true` if it is newly added.
     pub(crate) fn insert(&mut self, val: u32) -> bool {
-        let ret = !self.contains(val);
-        *self.element_mut(val) |= elem_index_bit_mask(val);
-        self.length += ret as u32;
-        ret
+        let el_mut = self.element_mut(val);
+        let mask = elem_index_bit_mask(val);
+        let is_new = (*el_mut & mask) == 0;
+        *el_mut |= mask;
+        self.length += is_new as u32;
+        is_new
     }
 
     /// Marks all values `[first, last]` as members of this set.
--- a/read-fonts/src/collections/int_set/bitpage.rs
+++ b/read-fonts/src/collections/int_set/bitpage.rs
@@ -100,21 +100,6 @@ impl BitPage {
         ret
     }
 
-    /// Blindly marks `(val % page width)` a member of this set.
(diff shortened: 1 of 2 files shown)


read-fonts/src/collections/int_set/bitpage.rs
```diff
@@ -97,18 +96,23 @@ impl BitPage {
...
     ///
-    /// This is used to maximize performance in cases where the return value on [`insert()`] is not needed.
+    /// # WARNING:
+    /// After calling this method, the `length` value for this page is no longer
+    /// valid. The caller is responsible for ensuring that the
+    /// `recalculate_length` method is called before the
     ///
     /// [`insert()`]: Self::insert
-    pub(crate) fn insert_no_return(&mut self, val: u32) {
+    pub(crate) fn insert_no_return_and_promise_to_do_your_own_bookkeeping(&mut self, val: u32) {
```
cmyr, line 114: unclear to me if this is still worth it? It's only used for extending, and for small sets having to recompute the values of all pages is probably more expensive than just summing the returned bools 🤷

read-fonts/src/collections/int_set/bitpage.rs
```diff
@@ -97,18 +96,23 @@ impl BitPage {
     pub(crate) fn insert(&mut self, val: u32) -> bool {
         let ret = !self.contains(val);
```
garretrieger, line 97: Was looking at what contains() does and it also grabs the element and applies a bit mask, I wonder if it might be better to just check element_mut & bit_mask inside this function prior to applying it.

read-fonts/src/collections/int_set/bitpage.rs
garretrieger: This is a good question to answer via the benchmarks (https://github.com/googlefonts/fontations/tree/main/read-fonts/benches). They test across a variety of set sizes so you can see the exact tradeoffs being made. Recommend doing a before and after to see what happens.

garretrieger (commented):
I would be nice to do a before and after run of benchmarks for this change. I expect it should likely not cause any significant changes in performance but would be great to confirm.

read-fonts/src/collections/int_set/bitpage.rs
cmyr: okay yea ripping all of this out is actually a win in many cases (larger sets especially) since it saves needing to update _all_ pages at the end.

garretrieger (approved):

PR #1445, opened 2025-03-26
[klippa] subset Coverage table

qxliu76:


klippa/src/layout.rs
```diff
@@ -512,6 +519,211 @@ impl<'a> Serialize<'a> for ClassDefFormat2<'a> {
...
+        _args: Self::ArgsForSubset,
+    ) -> Result<Self::Output, SerializeErrorFlags> {
+        let glyph_count = (self.glyph_count() as usize).min(plan.font_num_glyphs);
+        let Some(glyph_array) = self.glyph_array().get(0..glyph_count) else {
+            return Err(SerializeErrorFlags::SERIALIZE_ERROR_READ_ERROR);
+        };
+
+        let num_bits = 16 - (glyph_count as u16).leading_zeros() as usize;
+        let retained_glyphs: Vec<u32> =
+            if glyph_count > (plan.glyphset_gsub.len() as usize) * num_bits {
```
garretrieger, line 554: can you add a comment that explains the two different if branches do the same thing and we're just picking which approach will be faster depending on the number of glyphs.

klippa/src/layout.rs
```diff
@@ -512,6 +519,211 @@ impl<'a> Serialize<'a> for ClassDefFormat2<'a> {
...
+        plan: &Plan,
+        s: &mut Serializer,
+        _args: Self::ArgsForSubset,
+    ) -> Result<Self::Output, SerializeErrorFlags> {
+        let range_count = self.range_count();
+        if range_count as usize > plan.font_num_glyphs {
+            return Err(s.set_err(SerializeErrorFlags::SERIALIZE_ERROR_READ_ERROR));
+        }
+        let num_bits = 16 - range_count.leading_zeros() as usize;
+        let retained_glyphs: Vec<u32> = if self.population() > plan.glyph_map_gsub.len() * num_bits
```
garretrieger, line 594: same here

garretrieger (approved):

garretrieger (approved):

klippa/src/layout.rs
```diff
@@ -365,14 +368,16 @@ fn classdef_remap_and_serialize(
 }
 
 trait Serialize<'a> {
+    type ArgsForSerialize: 'a;
```
cmyr, line 371: you can probably just call the type `Args`, it will be clear from context what args we're talking about?

klippa/src/layout.rs
```diff
@@ -512,6 +519,215 @@ impl<'a> Serialize<'a> for ClassDefFormat2<'a> {
...
+        plan: &Plan,
+        s: &mut Serializer,
+        _args: Self::ArgsForSubset,
+    ) -> Result<Self::Output, SerializeErrorFlags> {
+        let glyph_count = (self.glyph_count() as usize).min(plan.font_num_glyphs);
+        let Some(glyph_array) = self.glyph_array().get(0..glyph_count) else {
+            return Err(SerializeErrorFlags::SERIALIZE_ERROR_READ_ERROR);
+        };
+
+        let num_bits = 16 - (glyph_count as u16).leading_zeros() as usize;
```
cmyr, line 552: how sure are you that this can't overflow?

read-fonts/src/tables/layout.rs
```diff
@@ -258,6 +270,17 @@ impl RangeRecord {
             GlyphId::from(self.start_glyph_id())..=GlyphId::from(self.end_glyph_id()),
         )
     }
+
+    /// Return the number of glyphs in this table
```
cmyr, line 274:
```suggestion
    /// Return the number of glyphs in this record
```

cmyr (approved):
looks good!

klippa/src/layout.rs
qxliu76: pretty sure, because it's min(u16, usize)

PR #1452, opened 2025-03-28
[write] Split ligature subtables during compilation

cmyr:
This is much easier than splitting them during packing.

This fixes at least one crater crash, and also resolves an error that we encounter when compiling a brand font (although, unfortunately, we seem to hit another unrelated error afterwards)

### DIFF
--- a/write-fonts/src/tables/gsub/builders.rs
+++ b/write-fonts/src/tables/gsub/builders.rs
@@ -2,10 +2,94 @@
 
 use std::{collections::BTreeMap, convert::TryFrom};
 
-use types::GlyphId16;
+use types::{FixedSize, GlyphId16, Offset16};
 
-use crate::tables::{layout::builders::Builder, variations::ivs_builder::VariationStoreBuilder};
+use crate::{
+    tables::{
+        layout::{builders::Builder, CoverageTable},
+        variations::ivs_builder::VariationStoreBuilder,
+    },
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/gsub/builders.rs
```diff
@@ -2,10 +2,95 @@
...
+        let item_size = T::size_for_item(&item);
+        if item_size + self.current_size > Self::MAX_TABLE_SIZE {
+            let current_len = self.current_coverage.len();
+            self.finish_current();
+            let type_ = self.finished.last().unwrap().table_type();
+            log::info!("adding split in {type_} at {current_len}");
+        }
+
+        if self.current_size == 0 {
+            self.finish_current();
```
dfrg, line 69: I would assume that if `self.current_size` is 0 then `finish_current()` would have nothing to do. That _seems_ to be the case based on my reading of the code but I’m not sure if I’m missing something.

dfrg (approved):
LGTM with one tiny bit of confusion.

write-fonts/src/tables/gsub/builders.rs
cmyr: oh yea this is an extra statement; it's a noop because we've already called `finish_current` above, and `finish_current` is a noop if we haven't added any items yet.

PR #1456, opened 2025-03-31
[klippa] subset GDEF table

qxliu76:


klippa/src/offset.rs
```diff
@@ -104,3 +104,41 @@ impl<O: Scalar> SerializeCopy for O {
         )
     }
 }
+
+// this is trait is used to serialize tables that im
```
garretrieger, line 108: Looks like the comment got cut off.

klippa/test-data/expected/layout.gdef/Roboto-Regular.default.303.ttx
```diff
@@ -0,0 +1,308 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="\x00\x01\x00\x00" ttLibVersion="4.56">
```
garretrieger, line 2: Is this ttx file needed?

klippa/src/gdef.rs
```diff
@@ -1,11 +1,488 @@
...
+
+    let mut subset_varstore = false;
+    // TODO: make sure repacker will not move the target subtable before the other children
+    // ref: <https://github.com/harfbuzz/harfbuzz/blob/aad5780f5305f38ef128c61854c5d5a0c4ca3f4f/src/OT/Layout/GDEF/GDEF.hh#L665>
+    // Push var store first (if it's needed) so that it's last in the
+    // serialization order. Some font consumers assume that varstore runs to
+    // the end of the GDEF table.
+    // See: https://github.com/harfbuzz/harfbuzz/issues/4636
+
+    if version.minor >= 3 {
```
garretrieger, line 77: Instead of having subset_varstore be mutable you could do this:

let subset_varstore = if version.minor >= 3 { ... } else { false }

Likewise for all of the other ones below as well.

garretrieger (approved):

klippa/test-data/expected/layout.gdef/Roboto-Regular.default.303.ttx
qxliu76: oops, no, it's now deleted

garretrieger (approved):

klippa/src/offset.rs
```diff
@@ -104,3 +104,41 @@ impl<O: Scalar> SerializeCopy for O {
         )
     }
 }
+
+// this is trait is used to serialize a table that an offset at pos points to
```
cmyr, line 108: Looks like it wants to be a doc comment?

klippa/src/offset.rs
```diff
@@ -104,3 +104,41 @@ impl<O: Scalar> SerializeCopy for O {
...
+pub(crate) trait SerializeSerialize {
+    fn serialize_serialize<'a, T: Serialize<'a>>(
+        s: &mut Serializer,
+        args: T::Args,
+        pos: usize,
+    ) -> Result<(), SerializeErrorFlags>;
+}
```
cmyr, line 115: I don't think this is necessary.

You can add a method to `Serialize`, like this:

```rust
trait Serialize<'a> {
    type Args: 'a;
    /// Serialize this table
    fn serialize(s: &mut Serializer, args: Self::Args) -> Result<(), SerializeErrorFlags>;
    fn serialize_offset<O: Scalar>(
        s: &mut Serializer,
        args: Self::Args,
        pos: usize,
    ) -> Result<(), SerializeErrorFlags> {
        s.push()?;
        match Self::serialize(s, args) {
            Ok(()) => {
                let Some(obj_idx) = s.pop_pack(true) else {
                    return Err(s.error());
                };
                s.add_link(
                    pos..pos + O::RAW_BYTE_LEN,
                    obj_idx,
                    OffsetWhence::Head,
                    0,
                    false,
                )?;
                Ok(())
            }
            Err(e) => {
                s.pop_discard();
                Err(e)
            }
        }
    }
}
```

and then this is available to all types that implement this trait, and you can just do,

```rust
CoverageTable::serialize_offset::<Offset16>(s, &retained_glyphs, coverage_offset_pos)
```

cmyr (approved):
Looks fine, but I wonder if maybe the traits are getting out of control? It feels like there is a lot of overlap between them, and i think there's room for some simplification.

In particular I think you can get rid of `SerializeSerialize`, and I wonder if my same comment wouldn't also apply to `SerializeSubst`, now that I look at it?

It might be worth making a list of all the traits that have `Serialize` in their name, and seeing how much they overlap, and if there might be some room for consolidation?

klippa/src/offset.rs
qxliu76: Thanks for the suggestion! 
So `Serialize` is used for table serialization and` SerializeSerialize` is used for serializing an offset and the table it points to.
Unfortunately I'm unable to consolidate them because not all tables that implement Serialize trait have an offset field.

PR #1457, opened 2025-04-01
Add accessors to common axis values

simoncozens:
AxisValue table formats 1, 2 and 3 are very similar; STAT table consumers often want to know the values (or nominal values), axis indices, etc. for axis value tables, without wanting to match and extract from the inner data structure. This is a handy abstraction to make this more pleasant for consumers.


simoncozens:
Let's see if such a thing is needed more than once before automating it. :-)

PR #1459, opened 2025-04-02
[IFT] Use stored charstring offsets to locate charstring tables

garretrieger:
Context: https://w3c.github.io/IFT/Overview.html#cff

In the IFT mapping tables of fonts with CFF or CFF2 tables the offsets to the charstrings INDEX structures are stored. Parse and utilize these to locate the charstring data structures during patch application instead of pulling the offsets out of the CFF tables.

### DIFF
--- a/resources/codegen_inputs/ift.rs
+++ b/resources/codegen_inputs/ift.rs
@@ -66,9 +66,11 @@ table PatchMapFormat1 {
   /// Patch format number for patches referenced by this mapping.
   patch_format: u8,
 
+  // Offset to the cff charstrings INDEX from the start of the CFF table.
   #[if_flag($field_flags, PatchMapFieldPresenceFlags::CFF_CHARSTRINGS_OFFSET)]
   cff_charstrings_offset: u32,
 
+  // Offset to the cff charstrings INDEX from the start of the CFF2 table.
   #[if_flag($field_flags, PatchMapFieldPresenceFlags::CFF2_CHARSTRINGS_OFFSET)]
   cff2_charstrings_offset: u32,
 }
@@ -165,9 +167,11 @@ table PatchMapFormat2 {
   #[count($uri_template_length)]
   uri_template: [u8],
 
+  // Offset to the cff charstrings INDEX from the start of the CFF table.
   #[if_flag($field_flags, PatchMapFieldPresenceFlags::CFF_CHARSTRINGS_OFFSET)]
   cff_charstrings_offset: u32,
 
+  // Offset to the cff charstrings INDEX from the start of the CFF2 table.
   #[if_flag($field_flags, PatchMapFieldPresenceFlags::CFF2_CHARSTRINGS_OFFSET)]
   cff2_charstrings_offset: u32,
 }
--- a/resources/codegen_inputs/ift.rs
+++ b/resources/codegen_inputs/ift.rs
@@ -137,6 +137,9 @@ table PatchMapFormat2 {
   uri_template_length: u16,
   #[count($uri_template_length)]
   uri_template: [u8],
+
+  #[count(..)]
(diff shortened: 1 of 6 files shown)


resources/codegen_inputs/ift.rs
```diff
@@ -52,6 +52,9 @@ table PatchMapFormat1 {
 
   /// Patch format number for patches referenced by this mapping.
   patch_format: u8,
+
+  #[count(..)]
```
cmyr, line 56: so in _theory_ there are better ways to do this, but the easy ones would require there to be some field in this table that would tell us whether or not this field exists.

For instance, I could imagine using some of those `reserved` bytes up at the head of this table to store flags indicating whether or not these offsets are present.

Alternatively, you could pass in arguments to the table that would be used to determine the presence of these fields; that would look like,

```rust
#[read_args(has_cff: bool, has_cff2: bool)]
table PatchMapFormat1 {
// ..
#[if_true($has_cff)] // if_true doesn't exist but easy to add
cff_charstring_offset: u32,
#[if_true($has_cff2)]
cff2_charstring_offset: u32,
}
```

Barring that, though, this is probably your best bet.

Is the spec finalized? life is a lot easier when this sort of thing is explicit...

write-fonts/generated/generated_ift.rs
```diff
@@ -28,6 +28,7 @@ impl Ift {
         uri_template_length: u16,
         uri_template: Vec<u8>,
         patch_format: u8,
+        optional_charstring_offsets: Vec<u8>,
```
cmyr, line 31: yuck.

A few options:

- you could use something like `#[compile_type(OptionalCharstringOffsets)]` and then add that type to wrtie-fonts; it would just wrap two `Option<u32>` and write them if they were present.
- ... actually that's probably the best option. It feels hacky, though, and I'd prefer trying to fix this in one of the larger ways mentioned above.

but requiring this to be a buffer of pre-encoded bytes is definitely not great API.

cmyr (commented):
As you mentioned I think we can do better in codegen here. I've included some possibilities inline, I'm not sure if any of them are feasible?

resources/codegen_inputs/ift.rs
garretrieger: Changing the spec is possible, after further though I think I'll go that route and add some flag bits.

garretrieger: Changes made (https://github.com/w3c/IFT/pull/266) I'll rework this to match

garretrieger: Done, please take another look.

resources/codegen_inputs/ift.rs
```diff
@@ -52,6 +65,12 @@ table PatchMapFormat1 {
 
   /// Patch format number for patches referenced by this mapping.
   patch_format: u8,
+
+  #[if_flag($field_flags, PatchMapFieldPresenceFlags::CFF_CHARSTRINGS_OFFSET)]
+  cff_charstrings_offset: u32,
```
cmyr, line 70: Yes, this is much nicer.

one final note is that it would be nice to document what this actually is? It's confusing to have a field named `_offset` that is not actually an offset to a subtable. Does this reference data here? Data in some other table?

(if it doesn't reference something that is an offset from this table I also wonder if maybe it wants to have a name that isn't `_offset`? it would be one of only (or very few) places in the spec where a field has a name that ends in 'offset` that is not an offset in the generally understood sense.)

cmyr (approved):
Looks much nicer!

resources/codegen_inputs/ift.rs
garretrieger: Added a comment, these are indeed offsets, but relative to the start of the CFF/CFF2 table.

PR #1460, opened 2025-04-03
[clippy] Roll the rock back up the hill

cmyr:
This makes some changes that will conflict with #1450, but I think it's most important to keep us passing CI for the time-being, while we decide on that patch.

JMM


dfrg (approved):
🤚 🪨 🏔️

PR #1461, opened 2025-04-04
[read-fonts] small perf improvement in IVS

dfrg:
In `compute_scalar`, simply check for `peak == 0` for each region before reading all of the other data. Seems to be a roughly 20% win on fonts with sparse variation regions like Roboto Flex.


behdad:
LGTM.

read-fonts/src/tables/variations.rs
```diff
@@ -1371,13 +1375,17 @@ impl VariationRegion<'_> {
     pub fn compute_scalar_f32(&self, coords: &[F2Dot14]) -> f32 {
         let mut scalar = 1.0;
         for (i, axis_coords) in self.region_axes().iter().enumerate() {
-            let coord = coords.get(i).map(|coord| coord.to_f32()).unwrap_or(0.0);
+            let peak = axis_coords.peak_coord.get().to_f32();
+            if peak == 0.0 {
+                continue;
+            }
```
rsheeter, line 1381: Seems like here and above could share a method to iter potentially active regions? - I'm naively assuming that to_fixed isn't typically producing a different zero'ness than to_f32 I suppose

rsheeter (approved):

read-fonts/src/tables/variations.rs
dfrg: If the intent is just filtering out regions where peak is 0, then we can do the check on the 2.14 value so it’s not dependent on type. Will add a helper method.

PR #1464, opened 2025-04-08
[klippa] layout closure: collect features

qxliu76:


read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,209 @@
...
+        } else {
+            let scripts = scripts.unwrap();
+            for tag in scripts.iter() {
+                let Some(idx) = self.index_for_tag(tag) else {
+                    continue;
+                };
+                let script = script_records[idx as usize].script(font_data)?;
+                script.collect_features(&mut c, languages)?;
+            }
+        }
```
cmyr, line 133: prefer matching to `is_none()` + `unwrap()`:

```suggestion
        match scripts {
            None => {
                for record in script_records {
                    let script = record.script(font_data)?;
                    script.collect_features(&mut c, languages)?;
                }
            }
            Some(scripts) => {
                for idx in scripts.iter().filter_map(|tag| self.index_for_tag(tag)) {
                    let script = script_records[idx as usize].script(font_data)?;
                    script.collect_features(&mut c, languages)?;
                }
            }
        }
```

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,209 @@
...
+        c: &mut CollectFeaturesContext,
+        languages: Option<&IntSet<Tag>>,
+    ) -> Result<(), ReadError> {
+        if c.script_visited(self) {
+            return Ok(());
+        }
+
+        let lang_sys_records = self.lang_sys_records();
+        let font_data = self.offset_data();
+        if languages.is_none() {
```
cmyr, line 150: ditto with the `is_none()`  + `unwrap()`

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,209 @@
...
+            if required_feature_idx != 0xFFFF && !c.feature_indices_limit_exceeded(1) {
+                c.feature_indices.insert(required_feature_idx);
+            }
+
+            if !c.feature_indices_limit_exceeded(self.feature_index_count()) {
+                c.feature_indices
+                    .extend_unsorted(self.feature_indices().iter().map(|f| f.get()));
+            }
+        } else {
+            let filter = c.feature_indices_filter.as_mut().unwrap();
```
cmyr, line 193: ditto :)

cmyr (approved):
looks okay, some optional style nits inline :)

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        let mut indices = IntSet::empty();
+        for (idx, record) in self.feature_list.feature_records().iter().enumerate() {
+            let tag = record.feature_tag();
+            if features.contains(tag) {
+                indices.insert(idx as u16);
+            }
+        }
+        self.feature_indices_filter = Some(indices);
+    }
```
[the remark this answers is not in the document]
qxliu76: Nice! Thank you!

read-fonts/src/tables/layout/closure.rs
qxliu76: oh yeah, it's much better, thanks!

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    pub(crate) fn script_visited(&mut self, s: &Script) -> bool {
+        if self.script_count > MAX_SCRIPTS {
+            return true;
+        }
+
+        self.script_count += 1;
```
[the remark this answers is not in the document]
qxliu76: yeah I think so, this is the limit on operation count

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -1033,4 +1050,43 @@ mod tests {
...
+        let gsub = font.gsub().unwrap();
+
+        let mut script_tags = IntSet::empty();
+        script_tags.insert(Tag::new(b"LATN"));
+
+        let ret = gsub
+            .collect_features(Some(&script_tags), None, None)
+            .unwrap();
+        assert!(ret.is_empty());
+    }
```
garretrieger, line 1091: Can you also add tests that use inverted sets? Or if you remove the Option as I mentioned above then switch the None's to inverted sets would cover that.

read-fonts/src/tables/gpos/closure.rs
```diff
@@ -0,0 +1,23 @@
...
+
+impl Gpos<'_> {
+    /// Return a set of all feature indices underneath the specified scripts, languages and features
+    ///
+    /// if no script is provided, all scripts will be queried
+    /// if no language is provided, all languages will be queried
+    /// if no feature is provided, all features will be queried
+    pub fn collect_features(
+        &self,
+        scripts: Option<&IntSet<Tag>>,
```
garretrieger, line 14: Instead of using Option to indicate matching all you can just rely on using inverted sets. Passing IntSet::<Tag>::all() should result in the same behaviour as None.

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,195 @@
...
+        match scripts {
+            None => {
+                // All scripts
+                for record in script_records {
+                    let script = record.script(font_data)?;
+                    script.collect_features(&mut c, languages)?;
+                }
+            }
+            Some(scripts) => {
+                for idx in scripts.iter().filter_map(|tag| self.index_for_tag(tag)) {
```
garretrieger, line 112: scripts might be inverted in which case this iteration will be over all integer values. If you remove the Option's as suggested above then you can change this to check is_inverted() and if it is inverted loop through script records filtering by the set and if not inverted use this approach.

read-fonts/src/tables/layout/closure.rs
```diff
@@ -0,0 +1,195 @@
...
+                }
+
+                for record in lang_sys_records {
+                    let lang_sys = record.lang_sys(font_data)?;
+                    lang_sys.collect_features(c);
+                }
+            }
+            Some(languages) => {
+                for idx in languages
+                    .iter()
```
garretrieger, line 148: Likewise here.

garretrieger (approved):

PR #1466, opened 2025-04-17
[IFT] allow a custom brotli implementation to be provided for patch application.

garretrieger:
This is needed to support a WASM version of the rust IFT client which can't use the c brotli used by default. As an added bonus this opens the door for increasing fuzzing coverage by allowing the fuzzer to provide a custom implementation that can generate errors.

### DIFF
--- a/shared-brotli-patch-decoder/src/lib.rs
+++ b/shared-brotli-patch-decoder/src/lib.rs
@@ -9,42 +9,77 @@ pub mod decode_error;
 #[cfg(fuzzing)]
 use decode_error::DecodeError;
 
-/// Decodes shared brotli encoded data using the optional shared dictionary.
-///
-/// The shared dictionary is a raw LZ77 style dictionary, see:
-/// <https://datatracker.ietf.org/doc/html/draft-vandevenne-shared-brotli-format-11#section-3.2>
-///
-/// Will fail if the decoded result will be greater then max_uncompressed_length. Any excess data
-/// in encoded after the encoded stream finishes is also considered an error.
-pub fn shared_brotli_decode(
-    encoded: &[u8],
-    shared_dictionary: Option<&[u8]>,
-    max_uncompressed_length: usize,
-) -> Result<Vec<u8>, decode_error::DecodeError> {
-    cfg_if::cfg_if! {
-        if #[cfg(fuzzing)]
-        {
-            // When running under a fuzzer disable brotli decoding and instead just pass through the input data.
-            // This allows the fuzzer to more effectively explore code gated behind brotli decoding.
-            // TODO(garretrieger): instead consider modifying the top level IFT apis to allow a custom brotli decoder
-            //   implementation to be provided. This would allow fuzzing to sub in a custom impl that could return all
-            //   of the possible errors that the standard impl here can generate.
-            return if encoded.len() <= max_uncompressed_length {
-                Ok(encoded.to_vec())
+/// A shared brotli decoder.
+pub trait SharedBrotliDecoder {
+    /// Decodes shared brotli encoded data using the optional shared dictionary.
+    ///
+    /// The shared dictionary is a raw LZ77 style dictionary, see:
+    /// <https://datatracker.ietf.org/doc/html/draft-vandevenne-shared-brotli-format-11#section-3.2>
+    ///
+    /// Will fail if the decoded result will be greater then max_uncompressed_length. Any excess data
+    /// in encoded after the encoded stream finishes is also considered an error.
+    fn decode(
+        &self,
+        encoded: &[u8],
+        shared_dictionary: Option<&[u8]>,
+        max_uncompressed_length: usize,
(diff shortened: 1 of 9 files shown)


shared-brotli-patch-decoder/src/lib.rs
```diff
@@ -9,42 +9,77 @@ pub mod decode_error;
...
-        if #[cfg(fuzzing)]
-        {
-            // When running under a fuzzer disable brotli decoding and instead just pass through the input data.
-            // This allows the fuzzer to more effectively explore code gated behind brotli decoding.
-            // TODO(garretrieger): instead consider modifying the top level IFT apis to allow a custom brotli decoder
-            //   implementation to be provided. This would allow fuzzing to sub in a custom impl that could return all
-            //   of the possible errors that the standard impl here can generate.
-            return if encoded.len() <= max_uncompressed_length {
-                Ok(encoded.to_vec())
+/// A shared brotli decoder.
```
cmyr, line 12: Shared with who? I'd like a bit more info on this trait, since naively it feels like `BrotliDecoder` would be a reasonable name?

shared-brotli-patch-decoder/src/lib.rs
```diff
@@ -9,42 +9,77 @@ pub mod decode_error;
...
+pub trait SharedBrotliDecoder {
+    /// Decodes shared brotli encoded data using the optional shared dictionary.
+    ///
+    /// The shared dictionary is a raw LZ77 style dictionary, see:
+    /// <https://datatracker.ietf.org/doc/html/draft-vandevenne-shared-brotli-format-11#section-3.2>
+    ///
+    /// Will fail if the decoded result will be greater then max_uncompressed_length. Any excess data
```
cmyr, line 19:
```suggestion
    /// Will fail if the decoded result will be greater than `max_uncompressed_length`. Any excess data
```

shared-brotli-patch-decoder/src/lib.rs
```diff
@@ -9,42 +9,77 @@ pub mod decode_error;
...
+    /// in encoded after the encoded stream finishes is also considered an error.
+    fn decode(
+        &self,
+        encoded: &[u8],
+        shared_dictionary: Option<&[u8]>,
+        max_uncompressed_length: usize,
+    ) -> Result<Vec<u8>, decode_error::DecodeError>;
+}
+
+pub struct BuiltInBrotliDecoder;
```
cmyr, line 29: Docs please..

shared-brotli-patch-decoder/src/lib.rs
```diff
@@ -9,42 +9,77 @@ pub mod decode_error;
...
+    fn decode(
+        &self,
+        encoded: &[u8],
+        shared_dictionary: Option<&[u8]>,
+        max_uncompressed_length: usize,
+    ) -> Result<Vec<u8>, decode_error::DecodeError> {
+        self.0
+            .decode(encoded, shared_dictionary, max_uncompressed_length)
+    }
+}
```
cmyr, line 46: this is a bit funny, I'm wondering why we are boxing the inner type instead of using it directly, given that it impls the trait?

shared-brotli-patch-decoder/src/lib.rs
```diff
@@ -9,42 +9,77 @@ pub mod decode_error;
...
+        max_uncompressed_length: usize,
+    ) -> Result<Vec<u8>, decode_error::DecodeError>;
+}
+
+pub struct BuiltInBrotliDecoder;
+
+pub struct ExternalBrotliDecoder(pub Box<dyn SharedBrotliDecoder>);
+
+/// Just passes through the input data, primarily intended for use in fuzzers.
+pub struct NoopBrotliDecoder;
```
cmyr, line 34: if this isn't really intended as public api (as in, we don't expect 'real users' to ever use it) I would add a `#[doc(hidden)]` attribute so it doesn't show up in docs/completion suggestions.

cmyr (approved):
I get the overall idea here, but I wonder if the impl is slightly complicated?

- do we really need the `ExternalBrotliDecoder` wrapper type? I can imagine some cases where we might (if we wanted to hide the trait from an interface for instance) but it isn't obvious that this is happening. Is there a specific rationale, here?

shared-brotli-patch-decoder/src/lib.rs
garretrieger: Shared brotli (https://datatracker.ietf.org/doc/draft-vandevenne-shared-brotli-format/) is the name of a standardized extension to the original brotli specification (specifically this is what adds support for shared dictionaries)

shared-brotli-patch-decoder/src/lib.rs
garretrieger: I expect this may be used elsewhere. It's useful for unit testing and fuzzing. Added a comment to that effect.

PR #1467, opened 2025-04-20
Add support for bitmaps to `skrifa`

LaurenzV:
A continuation of the existing bitmaps branch, in the hope of getting the ball rolling on this again. It's mostly the same as before, but I added some more documentation. I did make a change in the calculation of `x_bearing` and `y_bearing` for sbix glyphs, and this does trigger a small regression in the integration test suite of my library. However, I believe the code before was slightly wrong? But I'm not 100% sure, so it would be good if someone could take a closer look. 

I've also added a couple of TODO notes where I had some uncertainties.

I've added some tests for EBDT, but not sure what to do about sbix. Either I could download https://github.com/simoncozens/test-fonts/blob/master/CFF-and-SBIX.otf and include this in the tests, or I can just give you confirmation that the integration works fine in my library. 😄


skrifa/src/bitmap.rs
```diff
@@ -0,0 +1,472 @@
...
+                // PNG format:
+                // 8 byte header, IHDR chunk (4 byte length, 4 byte chunk type), width, height
+                let reader = FontData::new(png_data);
+                let width = reader.read_at::<u32>(16).ok()?;
+                let height = reader.read_at::<u32>(20).ok()?;
+                Some(BitmapGlyph {
+                    data: BitmapData::Png(glyph.data()),
+                    // TODO: Does that bearing calculation look correct?
+                    bearing_x: x_origin + glyph.origin_offset_x() as f32,
+                    bearing_y: y_origin + glyph.origin_offset_y() as f32,
```
LaurenzV, line 193: This is where I changed the calculation logic for the bearings because I think it was wrong, might be worth comparing it with the initial version to see whether I missed something.

rsheeter:
> I've added some tests for EBDT, but not sure what to do about sbix

Would it help if we were to add sbix examples to https://github.com/googlefonts/color-fonts?

LaurenzV:
Sure, I think it would help!

rsheeter:
Added some in https://github.com/googlefonts/color-fonts/pull/129

skrifa/src/bitmap.rs
dfrg: The original calculations came from here: https://github.com/google/skia/blob/6408d4a595cb49b75e1aa36beb0f73d57b1ea96d/src/ports/fontations/src/bitmap.rs#L170

which is what is currently shipping in Chrome. I'm hesitant to change these without fully understanding the consequences. Note that the values are separate because they are scaled differently here: https://github.com/google/skia/blob/6408d4a595cb49b75e1aa36beb0f73d57b1ea96d/src/ports/SkTypeface_fontations.cpp#L636

Do you have any example fonts where this change leads to more correct rendering?

@drott can you offer any insight on this?

LaurenzV: I can definitely revert that change, it's likely that I got it wrong given that it seems to work fine in Chrome! Though I really don't understand how that code matches what is written in the specification.

behdad:
We should add this to hb-fontations later.

Re sbix x/y bearing, see https://github.com/harfbuzz/harfbuzz/issues/2679

Also: https://github.com/harfbuzz/harfbuzz/issues/4581

LaurenzV:
That’s useful, thanks! If I can find some time I will also try to make experiments with a custom font and hopefully figure out the proper way of doing this.

skrifa/src/bitmap.rs
LaurenzV: Thanks, I will try to get this built and then investigate!

drott:
> Is it intended that the bounding box does not move in the first case but it does in the second (the fontations one)? Is this what the issue you linked is about?

The main thing we would need to see is that the Skrifa SBIX implementation works the same way as the current read-fonts' based SBIX implementation in bitmap.rs in Skia's `src/ports/fontations/src/` - if it's not too much trouble, in your local build, could you adapt the bitmap.rs part to the new Skrifa code in this PR? Without any changes there, Skia will use the SBIX implementation based on read-fonts, not your new one.

I don't remember off-hand whether the bounding box is supposed to move. But the current read-fonts impl based behavior is okay - and was fixed in the linked previous changes.

> If I switch. my local fontations version back to 0.27, I can manage to get it to build with the patched files mentioned. However, with current main, I get errors such as:

This might be hitting https://issues.skia.org/413044303 and I've just talked to Kaylee in the Skia team about the increased importance for addressing this. To resolve this, you could perhaps rebase your patch on skrifa @ 0.27 and test with that, to get the build succeeding? Does that help?

LaurenzV:
Yes to both! Will look into it tomorrow.

skrifa/src/bitmap.rs
```diff
@@ -0,0 +1,496 @@
...
+        };
+        Some(BitmapStrike(kind))
+    }
+
+    /// Returns the best matching glyph for the given size and glyph
+    /// identifier.
+    ///
+    /// In this case, "best" means a glyph of the exact size, nearest larger
+    /// size, or nearest smaller size, in that order.
+    pub fn glyph_for_size(&self, size: Size, glyph_id: GlyphId) -> Option<BitmapGlyph<'a>> {
```
dfrg, line 107: @drott this behavior is slightly different from what currently exists in Skia in that it only considers strikes that actually contain the requested glyph. Since coverage can technically be sparse, I believe this is a more robust solution. We can add a method for selecting strikes without considering glyph coverage if you’d prefer to retain current behavior.

@bungeman I believe you raised a concern about this at some point?

LaurenzV:
Okay, so using the currently pushed version (which is basically the version from Chad's original branch) I get the expected behavior in the slide (and I verified that it is actually picking up my local version by adding a bug and checking whether it changes the behavior). 

So if you are happy with this, I can rebase it to the current main and then it should be good to go.

skrifa/src/bitmap.rs
```diff
@@ -0,0 +1,485 @@
...
+#[derive(Clone)]
+pub struct BitmapStrikes<'a>(StrikesKind<'a>);
+
+impl<'a> BitmapStrikes<'a> {
+    /// Creates a new `BitmapStrikes` for the given font.
+    ///
+    /// This will prefer `sbix`, `CBDT`, and `CBLC` formats in that order.
+    ///
+    /// To select a specific format, use [`with_format`](Self::with_format).
+    pub fn new(font: &(impl TableProvider<'a> + MetadataProvider<'a>)) -> Self {
```
dfrg, line 21: Missed this but let's have this take `&FontRef<'a>` since we've effectively phased out `TableProvider` in skrifa constructors.

skrifa/src/bitmap.rs
```diff
@@ -0,0 +1,485 @@
...
+        }
+        Self(StrikesKind::None)
+    }
+
+    /// Creates a new `BitmapStrikes` for the given font and format.
+    ///
+    /// Returns `None` if the requested format is not available.
+    pub fn with_format(
+        font: &(impl TableProvider<'a> + MetadataProvider<'a>),
+        format: BitmapFormat,
```
dfrg, line 35: Same here with `&FontRef<'a>`.

dfrg (approved):
I'd like to change the two constructor signatures for consistency with the other code but otherwise I think this is ready to land.

If necessary, we can address the `glyph_for_size` question as a follow up.

skrifa/src/bitmap.rs
drott: Works for me as long as we keep the "first larger, then smaller" logic.

drott:
> Yes to both! Will look into it tomorrow.

Cool, thanks for doing that and checking with the slide test. The code that you wrote to connect to Skia from this, would you be ready to share this or upstream this to Skia? If you just have scratch code somewhere and you'd be okay with me integrating that to Skia, I am happy to look into that.

LaurenzV:
Sure, that's the code I used, though I presume the FfiBitmapMetrics struct can be removed completely:

```rs
pub unsafe fn bitmap_glyph<'a>(
    font_ref: &'a BridgeFontRef,
    glyph_id: u16,
    font_size: f32,
) -> Box<BridgeBitmapGlyph<'a>> {
    let glyph_id = GlyphId::from(glyph_id);
    font_ref
        .with_font(|font| {
            let strikes = font.bitmap_strikes();
        
            let g = strikes.glyph_for_size(Size::new(font_size), glyph_id)?;
            
            let data = match g.data {
                bitmap::BitmapData::Png(p) => Some(BitmapPixelData::PngData(p)),
                bitmap::BitmapData::Bgra(_) => None,
                bitmap::BitmapData::Mask(_) => None,
            };
            
            let metrics = FfiBitmapMetrics {
                bearing_x: g.bearing_x,
                bearing_y: g.bearing_y,
                ppem_x: g.ppem_x,
                ppem_y: g.ppem_y,
                placement_origin_bottom_left: g.placement_origin == Origin::BottomLeft,
                inner_bearing_x: g.inner_bearing_x,
                inner_bearing_y: g.inner_bearing_y,
                advance: g.advance.unwrap_or(f32::NAN),
            };

            Some(Box::new(BridgeBitmapGlyph {
                data,
                metrics,
            }))
        })
        .unwrap_or_default()
}
```

LaurenzV:
I just noticed I didn't include the code needed to for the special offset of Apple Color Emoji (https://github.com/linebender/vello/blob/244ed01801816b99c99adcbbedc3612962919040/vello/src/scene/bitmap.rs#L202-L214), do we want this as well or should clients take care of this themselves?

dfrg:
> I just noticed I didn't include the code needed to for the special offset of Apple Color Emoji (https://github.com/linebender/vello/blob/244ed01801816b99c99adcbbedc3612962919040/vello/src/scene/bitmap.rs#L202-L214), do we want this as well or should clients take care of this themselves?

I think we'll leave that hack to clients for now. Maybe we can reconsider in the future.

dfrg:
The future is now?

Just to be clear, the hack adds a -100 font unit vertical offset if the table is sbix, postscript name is “AppleColorEmoji” and both y bearings are 0. I came up with this value empirically by experimenting with CoreText some time ago. I’ll try to replicate the test with modern macOS. 

I’m not opposed to including this but so far we’ve leaned toward not adding font specific hacks directly to skrifa. If we do this, maybe we should add an additional y bearing field so we don’t hide the actual font values?

Issue #1123, opened 2024-08-30
[read-fonts] Issue Errors/Warnings when reading in a wrong Feature table


qxliu76:
Noticed when reading a wrong Feature table, no errors are issued. Below's an example code from gsub.rs in klippa:
 
```
impl<'a> NameIdClosure for Gsub<'a> {
    fn collect_name_ids(&self, plan: &mut Plan) {
        let Ok(feature_list) = self.feature_list() else {
            return;
        };
        for (i, feature_record) in feature_list.feature_records().iter().enumerate() {
            if !plan.gsub_features.contains_key(&(i as u16)) {
                continue;
            }
            let Ok(feature) = feature_record.feature(self.offset_data()) else {
                continue;
            };
            feature.collect_name_ids(plan);
        }
    }
}
```
This line `let Ok(feature) = feature_record.feature(self.offset_data()) ` is reading Feature from a wrong FontData, but I can still get an Ok() result and continue to collect_name_ids from the wrong Feature table. 
We should try to fix this.

rsheeter:
This is a pita that applies in many places. Types would be nice, e.g. if we could make newtypes to make the relationship between the thing that gives you the correct FontData and the thing that consumes it more obvious.

cmyr:
Agree this is a bit annoying. For what it's worth, any method that requires you to pass in the data explicitly should also document where that data is expected to come from ([see docs here](https://docs.rs/read-fonts/latest/read_fonts/tables/layout/struct.FeatureRecord.html#method.feature)) and we can't pass the data ourselves during codegen because the record that contains the offset does not have a reference to the parent table which owns the offset data.

We could get fancier in codegen (as rod says) with adding types to the offset data and that might be worth investigating at some point, but it will add a fair bit of codegen code and might have other complications I'm not foreseeing, so I don't think it's an immediate priority.

rsheeter:
It has burned me and Ginger. I consider that fairly strong evidence of a papercut. Chad had some ideas about making ncier accessors.

dfrg:
The basic idea is that we'd add a new type which I called `RecordWithOffset` (or maybe `RecordRef` to match `TableRef`) that would look something like:

```rust
pub struct RecordRef<'a, T>(FontData<'a>, T);

// Easy access to underlying record
impl Deref for RecordRef<'_, T> {
    type Target = T;
    fn deref(&self) -> &T { &self.1 }
}
```

And codegen would need to generate methods helper `RecordRef` for various types:
[...]

PR #1469, opened 2025-04-22
[read-fonts] parse trak table

dfrg:
Just provides basic parsing of the table data. This required a few hand-written bits because it has nested tables that contain offsets from the root table. Rather than taking a raw `offset_data: FontData` parameter, these methods take a `&Trak` reference. This is a slight deviation from the rest of the `read-fonts` API but offers better type safety? Open to changing this for consistency.

Includes a few simple parsing tests. Also confirmed to pass the AAT trak tests in HarfRuzz.


resources/codegen_inputs/trak.rs
```diff
@@ -0,0 +1,43 @@
+#![parse_module(read_fonts::tables::trak)]
+
+/// The [tracking (trak)](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6trak.html) table.
+#[tag = "trak"]
+table Trak {
+    /// Version number of the tracking table (0x00010000 for the current version).
+    version: MajorMinor,
+    /// Format of the tracking table (set to 0).
+    format: u16,
```
cmyr, line 9: might as well also add `#[compile(0)]` here? (and `#[compile(MajorMinor::VERSION_1_0)]` above?)

rsheeter:
> This is a slight deviation from the rest of the read-fonts API but offers better type safety? 

I've almost always rued this type of inconsistency because it makes future refactors harder when things are inconsistent. Can we follow the normal patern and then, separately, upgrade all of read-fonts to have better type safety in this regard? - which FontData to pass is a long standing papercut.

read-fonts/src/tables/trak.rs
```diff
@@ -0,0 +1,130 @@
...
+        buf = buf.push(0x00000000u32).extend([258u16, 60]);
+        buf = buf.push(0x00010000u32).extend([257u16, 56]);
+        // Size subtable
+        buf = buf.push(0x000C0000u32);
+        buf = buf.push(0x00180000u32);
+        // Per-size tracking data values
+        buf = buf.extend([-15i16, -7, 50, 20, 0, 0]);
+        buf.to_vec()
+    }
+}
```
rsheeter, line 130: Should demonstrate it's ok in the face of malformed and insufficient data

resources/codegen_inputs/trak.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Number of separate tracks included in this table.
+    n_tracks: u16,
+    /// Number of point sizes included in this table.
+    n_sizes: u16,
+    /// Offset from the start of the tracking table to the start of the size subtable.
+    size_table_offset: u32,
+    /// Array of TrackTableEntry records.
+    #[count($n_tracks)]
+    track_table: [TrackTableEntry],
+}
```
rsheeter, line 33: Do we not support sizeTable[nSizes] as a subsequent array? Maybe we should? (could be deferred to an issue referring to this)

dfrg:
> > This is a slight deviation from the rest of the read-fonts API but offers better type safety?
> 
> I've almost always rued this type of inconsistency because it makes future refactors harder when things are inconsistent. Can we follow the normal patern and then, separately, upgrade all of read-fonts to have better type safety in this regard? - which FontData to pass is a long standing papercut.

Sure, I'll change this to take FontData instead.

resources/codegen_inputs/trak.rs
dfrg: We do support that but this table contains an explicit offset to the size table and we've been bitten by this before when assuming we can ignore it. In particular, Apple seems to like tables that are aligned to 4 bytes so it's conceivable that the size table might not directly follow the tracks.

resources/codegen_inputs/trak.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// Single entry in a tracking table.
+record TrackTableEntry {
+    /// Track value for this record.
+    track: Fixed,
+    /// The 'name' table index for this track (a short word or phrase like "loose" or "very tight"). NameIndex has a value greater than 255 and less than 32768.
+    name_index: NameId,
+    /// Offset from the start of the tracking table to per-size tracking values for this track.
+    offset: u16,
```
cmyr, line 42: Is it like this because if you use `Offset16` it generates getters and the data source is ambiguous? I think this makes sense, given it's such a special case. I do now wonder if this should be general pattern, where we have a `#[data_source(SomeTable)]` annotation... ofc it would break a lot of code unless we got clever. (I think we could get clever, though... 🤔)

cmyr (approved):
looks good, I'm nerdsniped a bit by this `pass in the table directly` idea.

resources/codegen_inputs/trak.rs
dfrg: will do!

resources/codegen_inputs/trak.rs
dfrg: Yeah, I think we have encountered this pattern a few times (in bitmaps maybe)? I'm not sure if it's enough to justify supporting this in codegen but if you're sufficiently nerd sniped... :)

rsheeter: As per https://github.com/googlefonts/fontations/issues/1123 this routinely bites people trying to use our stuff so I think it's worth fixing. That said ... probably less urgent than making crater # go up.

PR #1470, opened 2025-04-22
[klippa] closure lookups step 1: implement intersects()

qxliu76:
For simple tables only


read-fonts/src/tables/gpos/closure.rs
```diff
@@ -17,3 +24,230 @@ impl Gpos<'_> {
...
+    fn closure_lookups(
+        &self,
+        c: &mut LookupClosureCtx,
+        lookup_index: u16,
+    ) -> Result<(), ReadError> {
+        if c.lookup_visited(lookup_index) {
+            return Ok(());
+        }
+
+        c.set_lookup_visited(lookup_index);
```
cmyr, line 58: could this be one method, like `HashSet::insert`, that returns a bool indicating if it was already visited or not?

read-fonts/src/tables/gpos/closure.rs
```diff
@@ -17,3 +24,230 @@ impl Gpos<'_> {
...
+                Ok(false)
+            }
+            PositionSubtables::ChainContextual(subtables) => {
+                for t in subtables.iter() {
+                    if t?.intersects(glyph_set)? {
+                        return Ok(true);
+                    };
+                }
+                Ok(false)
+            }
```
cmyr, line 143: instead of all this repetition you can use generics to implement the trait for the `Subtables` type itself, like:


```rust
impl<'a, T, Ext> LookupClosure for Subtables<'a, T, Ext>
where
    T: LookupClosure + FontRead<'a> + 'a,
    Ext: ExtensionLookup<'a, T> + 'a,
{
    fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
        for sub in self.iter() {
            if sub?.intersects(glyph_set)? {
                return Ok(true);
            }
        }
        Ok(false)
    }
}
```

and then you can replace all of the above with,

```rust
    fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
        match self {
            PositionSubtables::Single(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::Pair(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::Cursive(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::MarkToBase(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::MarkToLig(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::MarkToMark(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::Contextual(subtables) => subtables.intersects(glyph_set),
            PositionSubtables::ChainContextual(subtables) => subtables.intersects(glyph_set),
        }
    }
```

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -579,6 +582,211 @@ fn intersect_coverage(
...
+
+    fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
+        self.subtables()?.intersects(glyph_set)
+    }
+}
+
+impl LookupClosure for SubstitutionSubtables<'_> {
+    fn closure_lookups(&self, _c: &mut LookupClosureCtx, _arg: u16) -> Result<(), ReadError> {
+        Ok(())
+    }
```
cmyr, line 633: not necessary right? because of the default impl?

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -579,6 +582,211 @@ fn intersect_coverage(
...
+
+    fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
+        match self {
+            SubstitutionSubtables::Single(subtables) => {
+                for t in subtables.iter() {
+                    if t?.intersects(glyph_set)? {
+                        return Ok(true);
+                    };
+                }
+                Ok(false)
```
cmyr, line 643: this can all be collapsed like in gpos

read-fonts/src/tables/layout/closure.rs
```diff
@@ -194,6 +196,133 @@ impl LangSys<'_> {
     }
 }
 
+#[allow(dead_code)]
+pub struct LookupClosureCtx<'a> {
```
cmyr, line 200: is this actually public? like is it ever constructed in public API?

cmyr (approved):
This looks quite good, I think your rust is coming along nicely :)

read-fonts/src/tables/gsub/closure.rs
qxliu76: oh I'll implement it in later PRs

read-fonts/src/tables/layout.rs
```diff
@@ -255,6 +285,44 @@ impl CoverageFormat2<'_> {
...
+                }
+                let end = record.end_glyph_id();
+                last = end;
+
+                let start = GlyphId::from(start_glyph);
+                if glyphs.contains(start) {
+                    out.insert(start);
+                }
+
+                for g in glyphs.iter_after(start) {
```
garretrieger, line 315: Nothing to do yet, but I'll add an iter_range() method to IntSet that iterates all values within the range (inclusive) that should simplify cases like this throughout this PR by not needing to check the start value separately.

garretrieger (approved):

read-fonts/src/tables/layout.rs
qxliu76: Yeah that will be very useful. Thanks!

PR #1473, opened 2025-04-22
[IFT] Capture more detail for IFT patching errors.

garretrieger:


incremental-font-transfer/src/font_patch.rs
```diff
@@ -74,31 +74,40 @@ impl From<SerializeErrorFlags> for PatchingError {
...
+                "Malformed brotli stream uri = {}, msg = {}",
+                uri, msg
+            )),
+            DecodeError::InvalidDictionary => {
+                PatchingError::InvalidPatch(format!("Malformed dictionary. ({})", uri))
             }
-            DecodeError::IoError(_) => {
-                PatchingError::InvalidPatch("IO error decoding input brotli stream.")
+            DecodeError::MaxSizeExceeded => {
+                PatchingError::InvalidPatch(format!("Max size exceeded. ({})", uri))
```
cmyr, line 91: I will reiterate my general distaste for errors with strings in them, on general principle 🤷

cmyr (approved):
`MyError(String)` is strong code smell to me in most cases (one of the exceptions being in a library API where the user might want to create a custom error message) but I'm not going to block you on that.

garretrieger:
This isn't needed all that much, so I'll just abandon it for now and revisit later if we run into cases where the additional detail is helpful.

PR #1475, opened 2025-04-23
[codegen] Typed data source for offsets in records

### LINKED ISSUE #681: It's not obvious which FontData to give

### LINKED ISSUE #1123: [read-fonts] Issue Errors/Warnings when reading in a wrong Feature table

cmyr:
Take two. This adds a new getter for offset fields in records, where instead of passing in the `FontData` you pass in the appropriate table itself.

- closes #1472 
- closes #1123 
- closes #681


read-fonts/generated/generated_base.rs
```diff
@@ -431,9 +431,23 @@ impl BaseScriptRecord {
     ///
     /// The `data` argument should be retrieved from the parent table
     /// By calling its `offset_data` method.
+    ///
+    /// NOTE: you should prefer to use [`read_base_script`][Self::read_base_script],
+    /// which takes the relevant parent table as input, instead of raw `FontData`.
     pub fn base_script<'a>(&self, data: FontData<'a>) -> Result<BaseScript<'a>, ReadError> {
```
rsheeter, line 437: I think these ones should be `#[deprecated]`

rsheeter (approved):
This is great! Ship it!

read-fonts/generated/generated_base.rs
cmyr: last time I deprecated right away, dominik complained that it broke his build. I agree that this is a bit annoying though because without deprecating it's hard to point people to the new methods? cc @drott

PR #1478, opened 2025-04-24
[read-fonts] parse the morx table 

dfrg:
Confirmed to pass HarfRuzz morx tests.


read-fonts/src/array.rs
```diff
@@ -142,6 +142,13 @@ impl<'a, T: FontRead<'a> + VarSize> VarLenArray<'a, T> {
             }
 
             let item_len = T::read_len_at(data, 0)?;
+            // If the length is 0 then then it's not useful to continue
```
cmyr, line 145: this maybe wants to be in a separate PR with a test case?

read-fonts/src/tables/aat.rs
```diff
@@ -373,40 +375,51 @@ impl<'a> SomeTable<'a> for StateTable<'a> {
     }
 }
 
-pub struct ExtendedStateTable<'a, T = ()> {
-    header: StxHeader<'a>,
+#[derive(Clone)]
+pub struct ExtendedStateTable<'a, T = NoPayload> {
```
cmyr, line 379: Looks like a nice opportunity to add some docs 😏

read-fonts/src/tables/morx.rs
```diff
@@ -0,0 +1,340 @@
...
+            .ok()
+            .map(|size| size as usize)
+    }
+}
+
+impl VarSize for Subtable<'_> {
+    type Size = u32;
+
+    fn read_len_at(data: FontData, pos: usize) -> Option<usize> {
+        data.read_at::<u32>(pos).ok().map(|size| size as usize)
```
cmyr, line 22: can you add a comment explaining why the default impl is inadequate?

read-fonts/src/tables/morx.rs
```diff
@@ -0,0 +1,340 @@
...
+    }
+}
+
+impl<'a> Subtable<'a> {
+    /// If true, this subtable will process glyphs in logical order (or reverse
+    /// logical order, depending on the value of bit 0x80000000).
+    pub fn is_logical(&self) -> bool {
```
cmyr, line 29: these might all be good candidates for `#[inline]` annotations?

resources/codegen_inputs/morx.rs
```diff
@@ -0,0 +1,82 @@
...
+#[tag = "morx"]
+table Morx {
+    /// Version number of the extended glyph metamorphosis table (either 2 or 3).
+    version: u16,
+    /// Not used; set to 0.
+    #[skip_getter]
+    #[compile(0)]
+    unused: u16,
+    /// Number of metamorphosis chains contained in this table.
+    n_chains: u32,
```
cmyr, line 13: this should also be computed at compile time?

resources/codegen_inputs/morx.rs
```diff
@@ -0,0 +1,82 @@
...
+/// A chain in a `morx` table.
+table Chain {
+    /// The default specification for subtables.
+    default_flags: u32,
+    /// Total byte count, including this header; must be a multiple of 4.
+    chain_length: u32,
+    /// Number of feature subtable entries.
+    n_feature_entries: u32,
+    /// The number of subtables in the chain.
+    n_subtables: u32,
```
cmyr, line 27: these three should all be computed at compile time; the first one with like `#[compile(self.compute_chain_length())]` and the other two with `#[compile(array_len($thing)]`


(this is true of a bunch of things here and I appreciate this doesn't include write-fonts support so I'll stop pointing them out; but will note that often when we add write-fonts support we forget to update these annotations, and then need to break the API later)

cmyr (approved):
Some notes but overall looks good!

read-fonts/src/array.rs
dfrg: Oops, I meant to do this but forgot. Will submit a separate PR.

read-fonts/src/tables/aat.rs
dfrg: Done!

read-fonts/src/tables/morx.rs
dfrg: Added a comment noting that the default impl assumes that the length field itself is not included in the computed length. As an aside, this feels like a bit of a footgun-- I wasn't aware that the default impl existed and it would have provided buggy behavior if I tried to use it.

read-fonts/src/tables/morx.rs
dfrg: Good catch! These will be used from a separate crate.

resources/codegen_inputs/morx.rs
dfrg: Added some compile annotations. In general, compiling these AAT tables is going to take some hand written code regardless because most of the arrays are unsized and we simply read to the end of the available data. I'm assuming we won't need to compile these at all but who knows?

PR #1479, opened 2025-04-24
Add public IntSet methods iter_from() and range().

garretrieger:
iter_from() iterates over set members >= to the given value. range() iterates over set members in the given range (inclusive).

### DIFF
--- a/fuzz/fuzz_targets/int_set_op_processor.rs
+++ b/fuzz/fuzz_targets/int_set_op_processor.rs
@@ -815,6 +815,71 @@ where
     }
 }
 
+/* ### Iter From ### */
+
+struct IterFromOp<T>(T);
+
+impl<T> IterFromOp<T>
+where
+    T: SetMember + 'static,
+{
+    fn parse_args(data: &mut Cursor<&[u8]>) -> Option<Box<dyn Operation<T>>> {
+        Some(Box::new(IterFromOp::<T>(read_u32(data)?)))
+    }
+}
+
+impl<T> Operation<T> for IterFromOp<T>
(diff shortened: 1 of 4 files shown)


cmyr (commented):
Okay an annoying observation up front:


could this API be collapsed into something like a single `iter_range` method, but where the argument type was [`impl RangeBounds<T>`](https://doc.rust-lang.org/1.86.0/std/ops/trait.RangeBounds.html)? Then the user could pass `..5` for iter_before, `5..` for iter_after, etc etc. Does this sound like a useful simplification?

garretrieger:
> Okay an annoying observation up front:
> 
> could this API be collapsed into something like a single `iter_range` method, but where the argument type was [`impl RangeBounds<T>`](https://doc.rust-lang.org/1.86.0/std/ops/trait.RangeBounds.html)? Then the user could pass `..5` for iter_before, `5..` for iter_after, etc etc. Does this sound like a useful simplification?

I think it should be possible, I'll try it out and see how it goes.

garretrieger:
> > Okay an annoying observation up front:
> > could this API be collapsed into something like a single `iter_range` method, but where the argument type was [`impl RangeBounds<T>`](https://doc.rust-lang.org/1.86.0/std/ops/trait.RangeBounds.html)? Then the user could pass `..5` for iter_before, `5..` for iter_after, etc etc. Does this sound like a useful simplification?
> 
> I think it should be possible, I'll try it out and see how it goes.

Done, please take another look.

cmyr (approved):
looks good, with one note!

PR #1480, opened 2025-04-26
Ability to create a `PString` programmatically

timboudreau:
This library seems oriented toward *modifying* font files, but some of us are insane enough to want to create one programmatically from scratch.  That requires being able to populate strings programmatically from scratch, but there is no clean way to construct a `PString`.

There is no obvious way to construct a `PString` from a `str` or `String`.  While it is possible to hack around that by putting the desired string in quotes and then using `serde_json` to deserialize it, that is a pretty ugly solution.

This patch simply adds a `From<I : Into<String>>` implementation for it.

### DIFF
--- a/write-fonts/src/tables/post.rs
+++ b/write-fonts/src/tables/post.rs
@@ -11,6 +11,12 @@ include!("../../generated/generated_post.rs");
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct PString(String);
 
+impl<I : Into<String>> From<I> for PString {
+    fn from(value: I) -> Self {
+        Self(value.into())
+    }
+}
+
 impl Post {
     /// Construct a new version 2.0 table from a glyph order.
     pub fn new_v2<'a>(order: impl IntoIterator<Item = &'a str>) -> Self {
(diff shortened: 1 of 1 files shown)


timboudreau:
I have filed a CLA.  Not sure what's needed to retrigger the workflow here.

nicoburns:
> I have filed a CLA. Not sure what's needed to retrigger the workflow here.

There's a "recheck" button if you click through to the CLA details. I've pressed it for you seeing as anyone can press it.

write-fonts/src/tables/post.rs
```diff
@@ -133,4 +139,10 @@ mod tests {
         assert_eq!(loaded.glyph_name(GlyphId16::new(4)), Some("A"));
         assert_eq!(loaded.glyph_name(GlyphId16::new(5)), Some("flarb"));
     }
+    
+    #[test]
+    fn test_pstring_dref() {
+        let string = PString::from("Hello");
+        assert_eq!("Hello", std::ops::Deref::deref(&string));
```
cmyr, line 146: fwiw the deref isn't necessary, we have a partialeq for `&str`.

cmyr (commented):
So this crate _is_ designed to be used for programatically creating fonts (we are building a font compiler on top of it) but we haven't needed a constructor for this type because we expect the user to use the `new_v2` convenience method to build a full valid `post` table.

The advantage of this is that it makes sure the generated table is well-formed.


That said, a constructor is fine! But along side `From` can we also add a `PString::new` method? It can have the same signature.

cmyr:
closing this; @timboudreau feel free to reopen and address comments if you're still interested in this API. :)

PR #1484, opened 2025-05-01
[read-fonts] VarLenArray: stop iteration at zero length item

dfrg:
Since we use the item length to track the iterator state, zero length items will cause the iterator to loop forever.

JMM if happy


cmyr (approved):
looks good!

You can fix the clippy issue by adding `#[skip_constructor]` to the table in the codegen input file.

dfrg:
> looks good!
> 
> You can fix the clippy issue by adding `#[skip_constructor]` to the table in the codegen input file.

thanks! I was just looking for the right attribute to fix this :)

PR #1486, opened 2025-05-05
Support mapping codepoints for cmap0 and cmap6 subtables

LaurenzV:
Also indirectly tested with a real-life font I found in a PDF file.


dfrg (approved):
This looks good, thanks!

PR #1489, opened 2025-05-05
MSRV take two

cmyr:
Let's go ahead and get something merged. This supersedes #1450.

- I'm going with 1.75, which gives us impl trait in traits, and tracks ubuntu LTS.
- I'm turning this on for all production crates, but not yet for other stuff (including klippa)


rsheeter (approved):
Relatively painless so far, nice!

PR #1494, opened 2025-05-12
[read-fonts] parse the `kerx` table

dfrg:
No real attempt made at write-fonts support. I don't think we ever expect to compile this table.


cmyr (approved):
well that's nasty..

dfrg:
> well that's nasty..

quite. Thanks for soldiering through!

PR #1495, opened 2025-05-12
Support preload lists on patch map entries.

garretrieger:
This updates the fontations IFT client to match spec changes made in: https://github.com/w3c/IFT/pull/262

Overview:
- Format 2 entries now have the ability to specify lists of entry id deltas (previously each entry had exactly one delta)
- These delta lists become a list of URIs where the first URI is the patch to be applied as usual and any additional URIs should be preloaded if the entry is selected.
- The format 2 parsing is updated to handle multiple deltas and all of the patch map handling has been updated to correctly handle preloads specified in this way.

### DIFF
--- a/incremental-font-transfer/src/bin/ift_extend.rs
+++ b/incremental-font-transfer/src/bin/ift_extend.rs
@@ -126,13 +126,14 @@ fn main() {
                     }
                 }
 
-                let patch_bytes = std::fs::read(uri_path.clone()).unwrap_or_else(|e| {
-                    panic!(
+                let patch_bytes = match std::fs::read(uri_path.clone()) {
+                    Result::Ok(bytes) => bytes,
+                    Result::Err(e) => panic!(
                         "Unable to read patch file ({}): {:?}",
                         uri_path.display(),
                         e
-                    )
-                });
+                    ),
+                };
 
                 UriStatus::Pending(patch_bytes)
             });
--- a/incremental-font-transfer/src/bin/ift_extend.rs
+++ b/incremental-font-transfer/src/bin/ift_extend.rs
@@ -55,6 +55,18 @@ struct Args {
     /// --design_space="wght@300,wdth@50:100"
     #[arg(short, long, value_delimiter = ',', num_args = 1..)]
     design_space: Vec<String>,
+
+    /// The maximum number of simulated network round trips allowed during extension.
+    ///
+    /// If the count is exceeded this command will fail and return an error.
+    #[arg(long)]
+    max_round_trips: Option<u32>,
+
+    /// The maximum number of fetches trips allowed during extension.
+    ///
+    /// If the count is exceeded this command will fail and return an error.
+    #[arg(long)]
+    max_fetches: Option<u32>,
     // TODO(garretrieger): add feature tags arguments.
 }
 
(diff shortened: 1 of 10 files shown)


incremental-font-transfer/src/bin/ift_extend.rs
```diff
@@ -82,41 +94,78 @@ fn main() {
...
+                    }
+                }
+
+                let patch_bytes = std::fs::read(uri_path.clone()).unwrap_or_else(|e| {
+                    panic!(
+                        "Unable to read patch file ({}): {:?}",
+                        uri_path.display(),
+                        e
+                    )
+                });
```
cmyr, line 135: appreciate this isn't new code but would note that if you want to panic it would be more idiomatic to `match` instead of using `unwrap_or_else`, which is intended to be used to compute a value.

incremental-font-transfer/src/patch_group.rs
```diff
@@ -190,36 +224,73 @@ impl PatchGroup<'_> {
         if let (Some(uri), None) = (&ift_selected_uri, &iftx_selected_uri) {
             no_invalidation_iftx.remove(uri);
         }
-        if let (None, Some(uri)) = (ift_selected_uri, iftx_selected_uri) {
-            no_invalidation_ift.remove(&uri);
+        if let (None, Some(uri)) = (&ift_selected_uri, &iftx_selected_uri) {
+            no_invalidation_ift.remove(uri);
```
cmyr, line 228: also just noticing now but I'd probably combine these both into a match, like

```rust
match (&if_selected_uri, &iftx_selected_uri) {
(Some(uri), None) | (None, Some(uri) => no_invalidation_ift.remove(uri),
_ => (),
}
```

incremental-font-transfer/src/patch_group.rs
```diff
@@ -190,36 +224,73 @@ impl PatchGroup<'_> {
...
 
+        let no_invalidation_ift: BTreeMap<String, NoInvalidationPatch> = match ift_selected_uri {
+            None => Self::extract_preloads(no_invalidation_ift, &mut combined_preload_uris),
+            _ => Default::default(),
+        };
+        let mut no_invalidation_iftx: BTreeMap<String, NoInvalidationPatch> =
+            match iftx_selected_uri {
+                None => Self::extract_preloads(no_invalidation_iftx, &mut combined_preload_uris),
+                _ => Default::default(),
+            };
```
cmyr, line 239: just golf, but:

```suggestion
        let mut no_invalidation_iftx = iftx_selected_uri
            .is_none()
            .then(|| Self::extract_preloads(no_invalidation_iftx, &mut combined_preload_uris))
            .unwrap_or_default();
```

incremental-font-transfer/src/patchmap.rs
```diff
@@ -652,14 +677,77 @@ fn compute_format2_new_entry_index(
...
+    flags: EntryFormatFlags,
+    format_data: &[u8],
+) -> Result<(Option<PatchFormat>, &[u8]), ReadError> {
+    if !flags.contains(EntryFormatFlags::PATCH_FORMAT) {
+        return Ok((None, format_data));
+    }
+
+    let Some(format_byte) = format_data.first() else {
+        return Err(ReadError::OutOfBounds);
+    };
```
cmyr, line 690:
```suggestion
    let format_byte = format_data.first().ok_or(ReadError::OutOfBounds)?;
```

cmyr (approved):
some arbitrary and ignorable nits but overall this looks reasonable!

incremental-font-transfer/src/patch_group.rs
garretrieger: but yeah I think I can still do as a match

PR #1496, opened 2025-05-13
[skrifa] support interpreter-version 35

### LINKED ISSUE #1215: skrifa: Ability to enable horizontal hinting
skrifa currently only rounds the Y coordinate if ROUND_XY_TO_GRID is set:

https://github.com/googlefonts/fontations/blob/40329a215086ab2227e516ad8af125a74a4c4f07/skrifa/src/outline/glyf/mod.rs#L854-L861

However, freetype allows users to also enable X rounding:

```c
        if ( subglyph->flags & ROUND_XY_TO_GRID )
        {
          if ( IS_HINTED( loader->load_flags ) )
          {
            if ( driver->interpreter_version == TT_INTERPRETER_VERSION_35 )
              x = FT_PIX_ROUND( x );

            y = FT_PIX_ROUND( y );
          }
        }
```

This is controlled by the [interpreter-version](https://freetype.org/freetype2/docs/reference/ft2-properties.html#interpreter-version) setting. This setting has no other effects.

Users can set this setting with environment variables:

```
export FREETYPE_PROPERTIES=truetype:interpreter-version=35
```

This is the only way to modify this setting.

Would it be possible to have skrifa also interpret the FREETYPE_PROPERTIES environment variable and enable X rounding as freetype does? I'd be willing to implement this.

Some other way to enable X rounding would also be fine, as long as it is under the control of users running programs using skrifa and does require modifying source code.

mahkoh:
With this change, font rendering appears to be essentially identical to freetype.

Chromium compiled with this patch (and a small skia patch to use `InterpreterVersion::from_environment`):

image

Firefox:

image

There are some small differences but they might be caused by a different component.

I've not added any tests. Let me know if/how this should be tested.

Closes #1215

### DIFF
--- a/skrifa/src/outline/hint.rs
+++ b/skrifa/src/outline/hint.rs
@@ -10,6 +10,10 @@ use super::{
 use crate::alloc::{boxed::Box, vec::Vec};
 use raw::types::Fixed;
 
+#[cfg(feature = "libm")]
+#[allow(unused_imports)]
+use core_maths::CoreFloat;
+
 /// Configuration settings for a hinting instance.
 #[derive(Clone, Default, Debug)]
 pub struct HintingOptions {
@@ -386,10 +390,15 @@ impl HintingInstance {
     pub fn reconfigure<'a>(
         &mut self,
         outlines: &OutlineGlyphCollection,
-        size: Size,
+        mut size: Size,
         location: impl Into<LocationRef<'a>>,
         options: impl Into<HintingOptions>,
     ) -> Result<(), DrawError> {
+        if self.interpreter_version == InterpreterVersion::_35 {
+            if let Some(ppem) = size.ppem() {
+                size = Size::new(ppem.round());
+            }
+        }
         self.size = size;
         self.coords.clear();
         self.coords
--- a/skrifa/src/outline/hint.rs
+++ b/skrifa/src/outline/hint.rs
@@ -87,6 +87,22 @@ impl From<Engine> for HintingOptions {
     }
 }
 
+/// The TrueType interpreter version.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+#[repr(i32)]
+pub enum InterpreterVersion {
+    /// Version 35 corresponds to MS rasterizer v.1.7 as used e.g. in Windows 98;
+    /// only grayscale and B/W rasterizing is supported.
+    _35 = 35,
+    /// Version 40 corresponds to MS rasterizer v.2.1; it is roughly equivalent to
+    /// the hinting provided by DirectWrite ClearType (as can be found, for example,
+    /// in Microsoft's Edge Browser on Windows 10).
+    ///
+    /// This is the value returned by `Default::default`.
+    #[default]
+    _40 = 40,
+}
+
 /// Defines the target settings for hinting.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum Target {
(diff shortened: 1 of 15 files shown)


mahkoh:
It might also be necessary to round the font size somewhere. Otherwise fonts with fractional pixel sizes in chromium don't render correctly.

mahkoh:
I was able to fix the aforementioned issue by rounding fond sizes when the interpreter version is 35. This makes chromium render text with fractional pixel sizes like firefox.

I believe that FreeType performs this rounding and that it does not depend on the interpreter version, but without interpreter-version=35 I could not easily spot the difference whereas interpreter-version=35 makes it obvious.

dfrg:
Thank you for taking this on! I won’t be able to do a thorough review until next week but, at a glance, the code looks good and I don’t foresee any issues with landing this in some form.

LaurenzV:
I just wanted to provide another data point for this: I just encountered a font that actually renders wrongly if hinting is disabled! If I don't enable hinting, the characters look like this, completely garbled:

image

With the current hinting mechanism, they look slightly better but still wrong:
image

With your patch, they look correct:
image

So this would definitely be nice having. :)

dfrg:
> I just wanted to provide another data point for this: I just encountered a font that actually renders wrongly if hinting is disabled! If I don't enable hinting, the characters look like this, completely garbled:

I believe this is #1303. For Skia/Chrome, we check the result of the [require_interpreter()](https://docs.rs/skrifa/latest/skrifa/outline/struct.OutlineGlyphCollection.html#method.require_interpreter) method and set the hinting target to `Mono` when that returns true. The result is very similar to the v35 interpreter.

LaurenzV:
Ah interesting, will try that!

skrifa/src/outline/hint.rs
```diff
@@ -323,10 +385,15 @@ impl HintingInstance {
...
         &mut self,
         outlines: &OutlineGlyphCollection,
-        size: Size,
+        mut size: Size,
         location: impl Into<LocationRef<'a>>,
         options: impl Into<HintingOptions>,
     ) -> Result<(), DrawError> {
+        if self.interpreter_version == InterpreterVersion::_35 {
+            if let Some(ppem) = size.ppem() {
+                size = Size::new(ppem.round());
```
valadaptive, line 399: This `round()` call is why the CI is failing--this method doesn't exist in `no_std` builds. You can fix it by adding:
```rust
#[cfg(feature = "libm")]
#[allow(unused_imports)]
use core_maths::CoreFloat;
```
The existing codebase puts it right under all the imports.

drott:
> For others, who arrive here looking for solution how to make font rendering beautiful again in Chrome 136+, now that the flags used to disable Fontation have been deprecated - you can still use `google-chrome --disable-features=FontationsLinuxSystemFonts`

Note that this option is removed after 139.0.7258 https://chromiumdash.appspot.com/commit/2fc1ae192a45eb6f1716e232dd1626317f8d299e
We no longer intend to carry the FreeType support. For reasoning, see https://developer.chrome.com/blog/memory-safety-fonts#why_replace_freetype

mahkoh:
Then lot's hope this PR gets reviewed and merged until then.

I'll be out of the office during August so ideally that would happen before then.

tarko:
yes please

@drott, moving on from freetype is certainly good idea but don't leave us hanging here :)

mahkoh:
I've merged your change but I don't think that is what has kept this PR from being reviewed.

valadaptive:
https://github.com/googlefonts/fontations/issues/1215 is still open. I believe various API consumers in the Rust text ecosystem also want this hinting style (e.g. https://github.com/dfrg/swash/issues/96).

This PR doesn't seem to introduce *that* much new code to test; it seems to be a few new branches in what is already a straight port of FreeType code. How much would it cost in code review time and ongoing maintenance burden?

rsheeter:
> How much would it cost in code review time and ongoing maintenance burden?

Little but not zero is probably fair?

> Doesn't it automatically fix it for chrome when it lands in fontations?

I am under the impression Chrome would need code to activate it before it would do anything.

mahkoh:
>https://github.com/googlefonts/fontations/issues/1215 is still open. I believe various API consumers in the Rust text ecosystem also want this hinting style (e.g. https://github.com/dfrg/swash/issues/96).

Vscode has also reverted the switch to fontations on their end due to the changed font rendering (linked above). So it's not just other rust code but also downstream from chromium.

tarko:
Yes, please

zaafonin:
Please consider this. Electron apps like VSCode suffer because of new font rendering.

S0yKaf:
Hello? It's been almost a year.

insilications:
Please consider reviewing this change...

wareya:
Ran into some fonts that should be getting hinting not getting hinting while working on a text editor. Please make time for this.

S0yKaf:
@dfrg @wmedrano @cmyr @qxliu76 

Is there anything we can do to help you guys move this along? 

This is the most commented on Pull Request in the repo.
And the most commented Issue.

More and more applications have broken font rendering because this isn't getting merged.

stacygaudreau:
VS Code has been borderline unusable on Linux due to this issue for quite some time. Downgrading to a very old version works but is not ideal. Hope this will get merged soon!

PR #1497, opened 2025-05-19
Add specialized glyf loader for the autohinter

dfrg:
This tries to buy some autohinting performance by removing all unnecessary work when processing outlines from the `glyf` table.

In the current state, this contains a lot of duplicated code.


dfrg:
Will take a different approach

PR #1498, opened 2025-05-20
[gvar] Inline a couple of methods

behdad:
GoogleSansFlex variable HB benchmark-font draw benchmark:

Before: 3.33ms
After: 2.57ms
Speedup: 23%


dfrg (approved):
Nice, LGTM

Issue #1500, opened 2025-05-21
Regression in FreeType/Fontations path equality after c32c0a6442371d0206f82a4a981b3854f313aaae


drott:
See https://issues.skia.org/419049647

In the following set of fonts and tests Skia's path difference testing notices path differences between the FreeType contour and the Fontations contour:
* Noto Sans Arabic [Fontations_NotoSans_ar_Arab]
* Noto Sans Armenian [Fontations_NotoSans_Armn]
* Noto Sans Bengali [Fontations_NotoSans_Beng]
* Noto Sans Devanagari [Fontations_NotoSans_Deva]
* Noto Sans Georgian [Fontations_NotoSans_Geor]
* Noto Sans Gujarati [Fontations_NotoSans_Gujr]
* Noto Sans Gurmukhi [Fontations_NotoSans_Guru]
* Noto Sans JP [Fontations_NotoSans_Jpan]
* Noto Sans Javanese [Fontations_NotoSans_Java]
* Noto Sans KR [Fontations_NotoSans_Kore]
* Noto Sans Kannada [Fontations_NotoSans_Knda]
* Noto Sans Khmer [Fontations_NotoSans_Khmr]
* Noto Sans Malayalam [Fontations_NotoSans_Mlym]
* Noto Sans SC [Fontations_NotoSans_Hans]
* Noto Sans Sinhala [Fontations_NotoSans_Sinh]
* Noto Sans TC [Fontations_NotoSans_Hant]
* Noto Sans Tamil [Fontations_NotoSans_Taml]
* Noto Sans Telugu [Fontations_NotoSans_Telu]
* Noto Sans Thai [Fontations_NotoSans_Thai]
* Noto Sans Vithkuqi [Fontations_NotoSans_Vith]

Example failure:

```
["/Users/drott/dev/skia/third_party/externals/googlefonts_testdata/data/fonts/NotoSansGurmukhi[wdth,wght].ttf"#0 ppem: 144 coords: [0.0, 0.0]] glyph id 7 outline doesn't match:
  MoveTo { x: 82.375, y: 0.0 }
  LineTo { x: 82.375, y: 53.71875 }
  LineTo { x: 84.09375, y: 51.265625 }
  QuadTo { cx0: 79.0625, cy0: 50.109375, x: 74.59375, y: 48.375 }
  QuadTo { cx0: 70.125, cy0: 46.65625, x: 66.15625, y: 44.34375 }
[...]

PR #1501, opened 2025-05-21
Partial revert of "bypass variation processing for default location"

dfrg:
In response to #1500


drott (approved):
If this was tested to pass the fauntlet test as in #1500, then LGTM.

dfrg:
Yes, with this change, we get no differences from fauntlet with Noto Sans Gurmukhi.

PR #1513, opened 2025-05-29
[IFT] Correctly handle patch maps that have duplicated URIs.

garretrieger:
https://w3c.github.io/IFT/Overview.html#remove-entries-format-1 and https://w3c.github.io/IFT/Overview.html#remove-entries-format-2 require that when a patch with some URL is applied all entries in the mapping table with the same URI must be marked as applied. The client implementation was not correctly handling this case.

This PR fixes the behaviour to match the spec requirement, this includes a refactor to the PatchUri, PatchMapEntry, and PatchInfo structs in order to be able to handle the required behaviour. Notably:
- PatchUri becomes the expanded URI template rather then a template + expansion arguments. This is required so we can tell during intersection which PatchUri's are duplicates of each other.
- The various bits of context information on PatchUri are moved up into PatchMapEntry.

Additonally updates patch selection and preload selection to remove any duplicate URIs.


incremental-font-transfer/src/patch_group.rs
```diff
@@ -647,6 +678,12 @@ mod tests {
     const TABLE_1_FINAL_STATE: &[u8] = "hijkabcdeflmnohijkabcdeflmno\n".as_bytes();
     const TABLE_2_FINAL_STATE: &[u8] = "foobarbaz foobarbaz foobarbaz\n".as_bytes();
 
+    impl PatchUri {
+        fn new(uri: &str) -> Self {
+            Self(uri.to_string())
+        }
```
cmyr, line 684: doesn't really matter for public API but in cases where you need an owned type in a fn it's generally better to accept the owned type as the argument:

- the caller might already have the owned type
- it makes the fact that the argument is cloned explicit

(you don't need to change this, it just reminded me of this idea)

incremental-font-transfer/src/patchmap.rs
```diff
@@ -975,60 +1034,25 @@ impl Ord for IntersectionInfo {
...
     }
 
+    fn from_order(order: usize) -> Self {
+        Self {
+            intersecting_codepoints: 0,
+            intersecting_layout_tags: 0,
+            intersecting_design_space: Default::default(),
+            entry_order: order,
+        }
+    }
```
cmyr, line 1054:
```suggestion
    fn from_order(entry_order: usize) -> Self {
        Self {
            entry_order,
            ..Default::default()
        }
    }
```

slightly cleaner?

PR #1515, opened 2025-05-29
Prefer the last component in the tree with USE_MY_METRICS set

Hoolean:
[The spec requires that we select the last component with USE_MY_METRICS set in the case of ambiguity:](https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#point-numbers-and-processing-for-composite-glyphs)

> If any component entry in a composite glyph has the USE_MY_METRICS flag set, then the hinted glyph metrics for the composite as a whole are taken from that component, rather than using the values given for the composite glyph itself. That is, the positions of phantom points for the composite glyph are set to the hinted positions of the component’s phantom points. **If more than one component glyph has this flag set, then the metrics for the composite glyph are taken from the last component that has this flag set.** In a variable font, if a component entry has this flag set, then the phantom point positions for the composite glyph are set to the interpolated and hinted positions of the component’s phantom point, and delta values for the composite glyph’s phantom points are not used.

This deprives us of our early-exit when the flag is set as the components iterator cannot be made reversible, but we can move to the faster  'glyphs and flags' variant in every case as compensation :)


Hoolean:
Marking as a draft as I didn't have a lever to test with, so happy to add something automated to increase our confidence.

Would adding a TTF be the preferred way to do this?

PR #1516, opened 2025-05-29
[IFT] avoid overflow in u16 index calculations.

garretrieger:
Do the calculations in u32 and then max_entry_index check catches anything that ends up larger then a u16.


cmyr (approved):
👍

PR #1518, opened 2025-05-30
[klippa] layout closure: closure lookups

qxliu76:


garretrieger (approved):

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -627,8 +628,12 @@ impl LookupClosure for SubstitutionLookup<'_> {
 }
 
 impl LookupClosure for SubstitutionSubtables<'_> {
-    fn closure_lookups(&self, _c: &mut LookupClosureCtx, _arg: u16) -> Result<(), ReadError> {
-        Ok(())
+    fn closure_lookups(&self, c: &mut LookupClosureCtx, arg: u16) -> Result<(), ReadError> {
+        match self {
+            SubstitutionSubtables::ChainContextual(subtables) => subtables.closure_lookups(c, arg),
```
cmyr, line 633: what about contextual?

qxliu76: oh yes, you're absolutely right! I added tests for chainedContextual only so this bug is not caught.

PR #1519, opened 2025-06-03
[read-fonts] parse kern table

dfrg:
Adds support for basic parsing of the kern table, both version 0 and version 1.0 with all subtable formats.

There's a lot of unfortunate hand parsing here due to the nature of the table (two distinct versions masquerading as a single table) and busted fonts encode pairs beyond the length limit.

This is sufficient for passing all kerning tests in harfruzz.


dfrg:
> Nice, this seems relatively clean all things considered!

It's amazing how far codegen will take us when the tables are even moderately sane. Just not the case here

jenskutilek:
Calibri is a common font with a notoriously weird kern table. Have you tested your code against it? I think it is one of the "busted" fonts ;) But Calibri also has GPOS kerning, so maybe the kern table can be ignored.

PR #1525, opened 2025-06-06
Add public methods to FeatureSet for extending/adding tags.

garretrieger:


rsheeter (approved):
Tests would be nice but lgtm

garretrieger:
> Tests would be nice but lgtm

Added

PR #1526, opened 2025-06-08
Support fonts with unsorted table directories

### LINKED ISSUE #1485: Fallback to linear search in unsorted table directory
I have a font where the tags in the table directory seemingly aren't sorted, and because we use `binary_search_by`, a tag cannot be found this way, even though the table exists in the font. Would it perhaps make sense to fall-back to performing a linear search in case a table cannot be found? Or perhaps we could do a validation pass up-front and store a variable whether the directory is sorted, and depending on that either always use binary search or linear search when requesting a table?

LaurenzV:
Fixes #1485.

### DIFF
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -265,6 +265,12 @@ impl<'a> CollectionRef<'a> {
     }
 }
 
+impl TableDirectory<'_> {
+    fn is_sorted(&self) -> bool {
+        self.table_records().is_sorted()
+    }
+}
+
 /// Reference to an in-memory font.
 ///
 /// This is a simple implementation of the [`TableProvider`] trait backed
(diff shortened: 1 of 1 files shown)


LaurenzV:
Hmm, should we raise the minimum version for this, or should I add a custom `is_sorted` implementation?

dfrg:
I think raising the MSRV for `[T]::is_sorted()` is probably not worth it. A custom implementation should be quite short?

LaurenzV:
I believe that should work, right?

read-fonts/src/lib.rs
```diff
@@ -265,6 +265,14 @@ impl<'a> CollectionRef<'a> {
     }
 }
 
+impl TableDirectory<'_> {
+    fn is_sorted(&self) -> bool {
+        self.table_records()
```
dfrg: Since this is called for every instantiation, I’m slightly concerned about leaning on the optimizer to elide the bounds checks (`array_windows` would be nice to have!)

I’d prefer to just go with the simplest approach. Something like:
```rust
let mut last_tag = Tag::from_bytes(&[0u8; 4]);
for tag in self.table_records().iter().map(|rec| rec.tag()) {
    if tag <= last_tag {
        return false;
    }
    last_tag = tag;
}
true
```

PR #1529, opened 2025-06-10
[klippa] clousre glyphs

qxliu76:
Current unit tests are enough, so no additional unit test is added,


read-fonts/src/tables/gsub/closure.rs
```diff
@@ -253,126 +298,300 @@ impl Gsub<'_> {
...
+        lookup_list: &SubstitutionLookupList,
+        lookup_index: u16,
+    ) -> Result<(), ReadError> {
+        if !ctx.should_visit_lookup(lookup_index) {
+            return Ok(());
+        }
+        self.subtables()?
+            .closure_glyphs(ctx, lookup_list, lookup_index)
+
+        //ctx.flush()
```
garretrieger: If this isn't need I'd just remove the line.

garretrieger (approved):

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -28,214 +27,260 @@ use crate::tables::layout::{
...
+const CLOSURE_MAX_STAGES: u8 = 12;
 impl Gsub<'_> {
     /// Return the set of glyphs reachable from the input set via any substitution.
     pub fn closure_glyphs(
         &self,
-        mut glyphs: IntSet<GlyphId16>,
-    ) -> Result<IntSet<GlyphId16>, ReadError> {
-        // we need to do this iteratively, since any glyph found in one pass
-        // over the lookups could also be the target of substitutions.
+        lookups: &IntSet<u16>,
```
cmyr, line 194: is this all based on  the harfbuzz? if so it would be useful to add some links to the original implementation source...

cmyr (approved):
Looks good... the previous implementation was based pretty closely on the closure in fonttools, so if this one is based on the impl in harfbuzz it might be useful to add some references in places where you are basically transliterating from the C++?

PR #1530, opened 2025-06-11
Provide glyph bounds in VF and CFF cases

### LINKED ISSUE #1528: [skrifa] GlyphMetrics.bounds does not seem to be variation-aware
With this code:

```rust
let normalized = font.axes().location(location);
    let glyph_metrics = font.glyph_metrics(Size::unscaled(), &normalized);
    if let Some(x_box) = font
        .charmap()
        .map('x' as u32)
        .and_then(|gid| glyph_metrics.bounds(gid))
    {
        println!("Bounds of x {:?}\n at {:?}", x_box, normalized);
    }
```

Roboto Serif at opsz=8 gives me:

```
Bounds of x BoundingBox { x_min: 6.0, y_min: 0.0, x_max: 569.0, y_max: 508.0 }
 at Location { coords: [0.0, -1.0, 0.0, 0.0] }
```

Roboto Serif at opsz=144 gives me:

```
Bounds of x BoundingBox { x_min: 6.0, y_min: 0.0, x_max: 569.0, y_max: 508.0 }
 at Location { coords: [0.0, 1.0, 0.0, 0.0] }
```

This is the default master in both cases. opsz=8 y_max should be 565, opsz=144 should be 518.

simoncozens:
Fixes #1528. Adds a ControlBoundsPen structure, stores the FontRef in the GlyphMetrics struct (with appropriate changes to lifetimes everywhere), and uses outlining to get the glyph bounds when the font has variations or a CFF/CFF2 table.

### DIFF
--- a/skrifa/src/metrics.rs
+++ b/skrifa/src/metrics.rs
@@ -30,6 +30,11 @@ use read_fonts::{
     FontRef, TableProvider,
 };
 
+use crate::{
+    outline::{pen::ControlBoundsPen, DrawSettings},
+    MetadataProvider,
+};
+
 use super::instance::{LocationRef, NormalizedCoord, Size};
 
 /// Type for a bounding box with single precision floating point coordinates.
@@ -341,12 +346,13 @@ impl<'a> GlyphMetrics<'a> {
 
     /// Returns the bounding box for the specified glyph.
     ///
-    /// Note that variations are not reflected in the bounding box returned by
-    /// this method.
-    ///
     /// Returns `None` if `glyph_id >= self.glyph_count()`, the underlying font
-    /// data is invalid, or the font does not contain TrueType outlines.
+    /// data is invalid.
     pub fn bounds(&self, glyph_id: GlyphId) -> Option<BoundingBox> {
+        if self.gvar.is_some() || self.font.cff().ok().is_some() || self.font.cff2().ok().is_some()
+        {
+            return self.bounds_from_outline(glyph_id);
+        }
         let (loca, glyf) = self.loca_glyf.as_ref()?;
         Some(match loca.get_glyf(glyph_id, glyf).ok()? {
             Some(glyph) => BoundingBox {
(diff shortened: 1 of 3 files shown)


skrifa/src/metrics.rs
```diff
@@ -213,6 +218,7 @@ impl Metrics {
 /// Glyph specific metrics.
 #[derive(Clone)]
 pub struct GlyphMetrics<'a> {
+    font: &'a FontRef<'a>,
```
dfrg: FontRef is small enough that I’d prefer to just clone it and store by value here to avoid the extra lifetime.

skrifa/src/bitmap.rs
```diff
@@ -18,7 +18,7 @@ impl<'a> BitmapStrikes<'a> {
     /// This will prefer `sbix`, `CBDT`, and `CBLC` formats in that order.
     ///
     /// To select a specific format, use [`with_format`](Self::with_format).
-    pub fn new(font: &FontRef<'a>) -> Self {
+    pub fn new(font: &'a FontRef<'a>) -> Self {
```
dfrg: Cloning the FontRef in GlyphMetrics should avoid the changes in this file.

skrifa/src/metrics.rs
```diff
@@ -370,6 +378,17 @@ impl GlyphMetrics<'_> {
         deltas[1] -= deltas[0];
         Some([deltas[0], deltas[1]].map(|delta| delta.x.to_i32()))
     }
+
+    fn bounds_from_outline(&self, glyph_id: GlyphId) -> Option<BoundingBox> {
+        if let Some(outline) = self.font.outline_glyphs().get(glyph_id) {
+            let settings = DrawSettings::unhinted(Size::unscaled(), self.coords);
```
dfrg: I guess we’ll need to save the size from the constructor as well so we can return an appropriately scaled bbox here.

dfrg (approved):
LGTM assuming we undo the `&'a self` changes and make sure the bounding box is computed at the correct size. Thanks!

skrifa/src/metrics.rs
simoncozens: Maybe. Or can we `self.fixed_scale.apply`?

simoncozens: Ignore that. I'll save the size.

PR #1531, opened 2025-06-11
[IFT] Update URL template expansion implementation to match latest spec

garretrieger:
URL templates were reworked here: https://github.com/w3c/IFT/pull/281

Update the IFT url template expansion to match. Also the spec has standardized on "URL" instead of "URI". Update the IFT code to consistently use URL.


incremental-font-transfer/src/patch_group.rs
```diff
@@ -726,6 +726,10 @@ mod tests {
         font_builder.build()
     }
 
+    const URL_TEMPLATE: &[u8] = &[
+        10, b'/', b'/', b'f', b'o', b'o', b'.', b'b', b'a', b'r', b'/', 128,
+    ];
```
cmyr: maybe deserves explaining the `10` and `128` bytes? but you can more easily express this as,

```suggestion
    const URL_TEMPLATE: &[u8] = b"\x0a//foo.bar/\x80";
```

incremental-font-transfer/src/patchmap.rs
```diff
@@ -1583,74 +1587,69 @@ mod tests {
...
+    fn url_template_substitution() {
         // These test cases are used in other tests.
-        check_uri_template_substitution("//foo.bar/{id}", 1, "//foo.bar/04");
-        check_uri_template_substitution("//foo.bar/{id}", 2, "//foo.bar/08");
-        check_uri_template_substitution("//foo.bar/{id}", 3, "//foo.bar/0C");
-        check_uri_template_substitution("//foo.bar/{id}", 4, "//foo.bar/0G");
-        check_uri_template_substitution("//foo.bar/{id}", 5, "//foo.bar/0K");
+
+        let foo_bar_id = &[
+            10, b'/', b'/', b'f', b'o', b'o', b'.', b'b', b'a', b'r', b'/', 128,
```
cmyr: yea i think all of these would be easier to understand as either a byte string literal or with a macro (to handle the prefix/suffix)

cmyr (approved):
I think this is a case where it would've made a lot of sense to PR the renaming patch first without any other changes, and then PR the changes afterwards; as-is it's very difficult to identify the actual changes amid all of the renames.

That said, nothing jumps out at my except for the fact that byte string literals are nice. 🤷

PR #1533, opened 2025-06-16
[cff] handle implied seac operator

### LINKED ISSUE #1532: CFF Fonts rendering empty glyphs
CFF fonts are rendering as empty glyphs. Expecting: u + umlaut = ü (U+00FC).

Versions:
- (stable) 137.0.7151.68 (Official Build) (64-bit)
- (dev) 139.0.7232.3 (Official Build) dev (64-bit) 

Basic example:

[fontations-cff-issue.zip](https://github.com/user-attachments/files/20723223/fontations-cff-issue.zip) -- html file and font as seen in screenshot below

Image

Related / Similar:
- https://issues.chromium.org/issues/372591157 (In Google Meet, unable to see icons in Material Icons CFF fonts)
- https://github.com/googlefonts/fontations/pull/1185
- https://github.com/googlefonts/fontations/issues/1184

dfrg:
This (deprecated) behavior of the `endchar` operator was left as a TODO in the code but there apparently are still some fonts in the wild that use it.

Adds support for seac (which is basically just a specially encoded "subroutine" call for compositing base and accent characters) and all the infra necessary to support it.

Fixes #1532


dfrg:
Confirmed to match FreeType with a fauntlet run on the font in the issue.

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -195,11 +211,43 @@ where
                 return Ok(false);
             }
             // End the current charstring
-            // TODO: handle implied 'seac' operator
             // Spec: <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf#page=21>
             // FT: <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/psaux/psintrp.c#L2463>
             EndChar => {
-                if !self.stack.is_empty() && !self.have_read_width {
+                if self.stack.len() == 4 || self.stack.len() == 5 && !self.have_read_width {
+                    // handle implied seac operator
```
cmyr: is there a spec document or something we can link to?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -195,11 +211,43 @@ where
...
+                    let cff = Cff::read(FontData::new(self.cff_data))?;
```
cmyr: also would it make senes to have a separate method for this logic?

read-fonts/src/tables/postscript/encoding.rs
```diff
@@ -0,0 +1,22 @@
+//! CFF standard encoding.
+
+/// See "Standard" encoding at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5176.CFF.pdf#page=37>
```
rsheeter, line 8: Despite the comment I have no idea what these magical bytes mean if I just glance at this. The referenced content does not appear to explain what a cff standard endocing is. Let's add a few more words explaining what this is.

skrifa/src/outline/cff/mod.rs
```diff
@@ -871,4 +877,21 @@ mod tests {
...
+        let font = FontRef::new(font_test_data::CHARSTRING_PATH_OPS).unwrap();
+        let glyphs = font.outline_glyphs();
+        let gid = GlyphId::new(3);
+        assert_eq!(font.glyph_names().get(gid).unwrap(), "Scaron");
+        let glyph = glyphs.get(gid).unwrap();
+        let mut pen = SvgPen::new();
+        glyph
+            .draw((Size::unscaled(), LocationRef::default()), &mut pen)
+            .unwrap();
+        assert_eq!(
```
rsheeter: Can you add a comment explainign what this is proving? E.g. if we didn't have seach X would happen, this result proves that Y happened.

rsheeter (approved):

read-fonts/src/tables/postscript/charstring.rs
dfrg: good call... there's an appendix that describes this weird behavior. Added a link

read-fonts/src/tables/postscript/charstring.rs
dfrg: Seems reasonable. Done

read-fonts/src/tables/postscript/encoding.rs
dfrg: Added a brief explanation and a link to the page in the spec where this _is_ actually discussed

skrifa/src/outline/cff/mod.rs
dfrg: Added a comment. Also changed the test to just check the resulting SVG for two `Z`s since we're actually just looking for two contours.

PR #1534, opened 2025-06-17
[klippa] layout closure

qxliu76:


garretrieger (approved):

klippa/src/gpos.rs
```diff
@@ -20,3 +24,161 @@ impl NameIdClosure for Gpos<'_> {
...
+        };
+
+        let feature_indices = self.prune_features(&lookup_indices, feature_indices);
+        let duplicate_feature_index_map =
+            self.find_duplicate_features(&lookup_indices, feature_indices);
+
+        let (script_langsys_map, feature_indices) =
+            self.prune_langsys(&duplicate_feature_index_map, &plan.layout_scripts);
+
+        let _ = std::mem::replace(&mut plan.gpos_lookups, remap_indices(lookup_indices));
```
cmyr: There's no need to use replace if you're just setting? you should be able to just write,

```rust
plan.gpos_lookups = remap_indices(lookup_indices);
```

klippa/src/gsub.rs
```diff
@@ -21,10 +25,88 @@ impl NameIdClosure for Gsub<'_> {
...
+        let feature_indices = self.prune_features(&lookup_indices, feature_indices);
+        let duplicate_feature_index_map =
+            self.find_duplicate_features(&lookup_indices, feature_indices);
+
+        let (script_langsys_map, feature_indices) =
+            self.prune_langsys(&duplicate_feature_index_map, &plan.layout_scripts);
+
+        let _ = std::mem::replace(&mut plan.gsub_lookups, remap_indices(lookup_indices));
+        let _ = std::mem::replace(&mut plan.gsub_features, remap_indices(feature_indices));
+        let _ = std::mem::replace(&mut plan.gsub_script_langsys, script_langsys_map);
```
cmyr: this code looks very nearly identical from the GPOS code. Does it differ in any meaningful way? If not it might be worth combining somehow?

qxliu76: yeah we do glyph closure for GSUB only.

PR #1536, opened 2025-06-18
FontRef API changes

dfrg:
This does a few things:
* Makes all `FontRef` fields private and adds accessor methods. Hopefully will avoid breaking changes on this type in the future.
* Adds a new field/accessor for the ttc index of the font.
* Adds a new `FontRef::all_in(&[u8])` method that returns an iterator over all fonts in the given data (as `Result<FontRef, _>`).

There were some uses of the `FontRef::table_directory` field in other crates. Updates those to use the new method.

### DIFF
--- a/read-fonts/src/lib.rs
+++ b/read-fonts/src/lib.rs
@@ -255,7 +255,11 @@ impl<'a> CollectionRef<'a> {
             .ok_or(ReadError::InvalidCollectionIndex(index))?
             .get() as usize;
         let table_dir_data = self.data.slice(offset..).ok_or(ReadError::OutOfBounds)?;
-        FontRef::with_table_directory(self.data, TableDirectory::read(table_dir_data)?)
+        FontRef::with_table_directory(
+            self.data,
+            TableDirectory::read(table_dir_data)?,
+            Some(index),
+        )
     }
 
     /// Returns an iterator over the fonts in the collection.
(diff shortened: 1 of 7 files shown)


read-fonts/src/lib.rs
```diff
@@ -357,9 +387,23 @@ impl<'a> FontRef<'a> {
             })
     }
 
+    /// Returns an iterator over all of the available fonts in
+    /// the given font data.
+    pub fn all_in(
```
cmyr: maybe... `iter_all`? `iter_from_bytes`?

cmyr (approved):
looks good, my only nit is that `all_in` just... sounds weird to me? but 🤷

read-fonts/src/lib.rs
dfrg: totally open to suggestions on this one. `iter_all` seems fine to me

rsheeter: `all_in` because using fonts is a gamble.

`iter_all` feels a touch redundant. I would expect `iter` to iterate all of them unless I filtered or limited or something.

`fonts(data)`, `iter(data)` ... even `iter_all(datat)` all seem fine.

dfrg: > `all_in` because using fonts is a gamble.

🤣 

> `iter_all` feels a touch redundant. I would expect `iter` to iterate all of them unless I filtered or limited or something.
> 
> `fonts(data)`, `iter(data)` ... even `iter_all(datat)` all seem fine.

I think I'll just go with `fonts(data)`. Seems nice and simple.

drott:
As discussed in chat, `table_directory` field access is used in base.rs in Skia, as part of the Skia/Fontations FFI. To ease rolling, we need a deprecation step in between.

PR #1545, opened 2025-06-24
[skrifa] tthint: round value for MPPEM instruction

### LINKED ISSUE #1544: Hinting at fractional sizes leads to disfigured glyphs
Users report that at fractional font sizes, fonts with hints such a the deprecated set of "web safe fonts" that MS used to distribute, and hinted fonts from ChromeOS such as Tinos or Cousine end up with unintended hinting artifacts and contours appearing incorrectly compresed or overlapping.

https://crbug.com/415453014
https://crbug.com/420947528

crbug.com/420947528 describes this as appearing with Cousine at 12.4px CSS pixels on an external monitor (likely with device pixel ratio of 1).

dfrg:
The MPPEM instruction is expected to produce an integral value and when hinting at fractional sizes we were truncating rather than rounding. This changes our computation to use a fixed point round that matches FreeType.

Also updates fauntlet to support testing at fractional sizes.

Fixes #1544

### DIFF
--- a/fauntlet/src/main.rs
+++ b/fauntlet/src/main.rs
@@ -29,7 +29,7 @@ enum Command {
 #[allow(clippy::explicit_write)]
 fn main() {
     // Pixels per em sizes. A size of 0 means an explicit unscaled comparison
-    let ppem_sizes = [0, 8, 16, 50, 72, 113, 144];
+    let ppem_sizes = [0.0, 8.0, 24.8, 16.0, 50.0, 72.0, 113.0, 144.0];
 
     // Locations in normalized variation space
     let var_locations = [-1.0, -0.32, 0.0, 0.42, 1.0].map(F2Dot14::from_f32);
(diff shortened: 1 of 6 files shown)


fauntlet/src/main.rs
```diff
@@ -29,7 +29,7 @@ enum Command {
 #[allow(clippy::explicit_write)]
 fn main() {
     // Pixels per em sizes. A size of 0 means an explicit unscaled comparison
-    let ppem_sizes = [0, 8, 16, 50, 72, 113, 144];
+    let ppem_sizes = [0.0, 8.0, 24.8, 16.0, 50.0, 72.0, 113.0, 144.0];
```
drott: Perhaps add a comment here why 24.8 is important and why there should be a fractional size in here with pointing to bug or PR?

drott (approved):
LGTM, thanks for the fix.

As follow-up, in addition to checking with fauntlet, would it be possible to build an inline unit test by subsetting from an open source hinted font, so we have this safely regression-tested?

fauntlet/src/main.rs
dfrg: Explained why we added this size with a link to the issue

PR #1549, opened 2025-07-01
[klippa] subset SinglePos

qxliu76:


klippa/src/gpos.rs
```diff
@@ -152,6 +161,402 @@ fn subset_gpos(gpos: &Gpos, plan: &Plan, s: &mut Serializer) -> Result<(), Seria
     Ok(())
 }
 
+fn compute_effective_format(
```
cmyr, line 164: See https://github.com/googlefonts/fontations/blob/e3ef9fa5ae908e2c9a5baff7528d98a36c922af2/write-fonts/src/tables/gpos/value_record.rs#L117 for somewhat similar code, probably not close enough to reuse...

qxliu76: yeah not reusable for subsetting. Thanks for the pointer!

garretrieger (approved):

PR #1550, opened 2025-07-04
Optimize/specialize `outline::path::contour_to_path`

valadaptive:
I'm playing around with a port of ab_glyph to skrifa, and it's a bit slower than ttf_parser currently.

The biggest culprit currently, at least for TrueType outlines, is `contour_to_path`. I *think* the issue is that it creates a peekable iterator for the contour points, and that iterator seems to not be getting fully inlined or optimized.

I've specialized the function into two separate "FreeType-style" and "HarfBuzz-style" functions, and now also pass in the first point. This lets us get rid of the peekable iterator in the FreeType path, as well as removing code paths not taken for the given `PathStyle`, and improves performance by ~13% on an outlining-only benchmark.

Before:
image
image

After:
image
image

### DIFF
--- a/skrifa/src/outline/autohint/outline.rs
+++ b/skrifa/src/outline/autohint/outline.rs
@@ -204,9 +204,13 @@ impl Outline {
             let Some(points) = self.points.get(contour.range()) else {
                 continue;
             };
-            if let Some(last_point) = points.last().map(Point::as_contour_point) {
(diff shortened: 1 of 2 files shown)


dfrg (approved):
Was finally able to run a larger set of tests and this works great. Thanks!

PR #1554, opened 2025-07-05
Mark `dict::parse_int` as inlineable

valadaptive:
This is a very simple change, but provides a ~12% performance boost on outlining CFF glyphs (using ab_glyph's benchmarks; see #1550):
image

Not sure why this wasn't getting inlined before; it's a pretty small function.


dfrg (approved):
Looks good, thanks!

PR #1557, opened 2025-07-07
[fauntlet] add CLI options for hinting modes

dfrg:
Adds new `--hinting-engine` and `--hinting-target` CLI options to select the hinting mode. Allows specifying `all` for each to run the full matrix of modes.


cmyr (approved):
looks good!

PR #1559, opened 2025-07-09
Implement hvgl outlines

### LINKED ISSUE #1369: Support hvgl outlines
The `hvgl` table was introduced by Apple with a new PingFang font family (PingFangUI.ttc). There isn't yet any official documentation of the format, but some [unofficial documentation](https://bungeman.github.io/hvgl.html) now exists.

See also [FreeType issue](https://gitlab.freedesktop.org/freetype/freetype/-/issues/1281) and [Chromium issue](https://issues.chromium.org/issues/380313555).

valadaptive:
Resolves #1369 by implementing support in skrifa for outlining [hvgl](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6hvgl.html) glyphs. The code is mainly ported [from HarfBuzz](https://github.com/harfbuzz/harfbuzz/blob/3c81945/src/hb-aat-var-hvgl-table.cc).

## (Lack of) parsing code

We are currently in the pocket of Big Endian, and cannot read the little-endian `hvgl` or `hvpm` tables. Implementing full little-endian support in `font-types`, `read-fonts`, and codegen is going to be an ordeal (see https://github.com/googlefonts/fontations/pull/1555), so I've handwritten a draft implementation of the hvgl table parser in `read-fonts`.

You can ignore the `font-types` and `read-fonts` changes--I'd like to implement little-endian support before this PR is merged so that we can properly codegen the `hvgl` table (and implement the `hvpm` table).

## Performance

It's kinda slow. I have some very rough relative benchmarks; they're comparing apples to oranges to some extent because there's not currently a way to convert between HVF and other font formats.

Outlining with Noto Sans CJK SC in TrueType takes ~50µs, the same font in CFF2 takes ~100µs, and "PingFang UI Text SC" takes ~215µs. This goes down to ~180µs if I compile for `x86-64-v3`, which enables AVX. Function multiversioning would help here, but I'm not sure how to do that without introducing unsafe code.

The biggest performance sink, aside from applying all the deltas (often a couple dozen axes' worth), is fetching every subpart for a composite glyph. Fetching a subpart requires first looking up its offset in the part table, and then looking up its data based on that offset. My hypothesis is that this results in 2 cache misses in a row; ouch.  To add insult to injury, we need to recurse through subparts twice: once to know how much scratch memory to allocate for applying segment deltas, and once to actually draw the glyph.

It feels like I'm missing something--the composite part table has a "max number of axes in any subpart" field that neither HarfBuzz nor this implementation makes use of. I assume it can be used to preallocate *something*, but I'm not sure what.

## Misc. code changes

`hvgl` is kinda like `loca`/`glyf`, and we can share some of the code for allocating scratch memory. I've moved and renamed some of that code since it now makes sense to share it.

I've moved `Transform` out of the `color` module in order to reuse it for hvgl, and added some methods that hvgl makes use of. Perhaps it would make sense to have an API-consumer-facing `Transform` type and a separate one for internal use; I'm not sure.

## What's left

Before merging, we need little-endian support so we can do real codegen for the `hvgl` table. After that, I'd like to add support for the `hvpm` table (I'm very excited to add another cache miss to every part lookup).

### DIFF
--- a/skrifa/src/outline/hvgl/memory.rs
+++ b/skrifa/src/outline/hvgl/memory.rs
@@ -4,7 +4,7 @@ use super::{Outline, Segment, Transform};
 
 /// Buffers used during glyph scaling.
 pub(crate) struct HvglOutlineMemory<'a> {
-    pub coords: &'a mut [f32],
+    pub coords: &'a mut [f64],
     pub transforms: &'a mut [Transform],
     pub segments: &'a mut [Segment],
 }
--- a/skrifa/src/outline/hvgl/mod.rs
+++ b/skrifa/src/outline/hvgl/mod.rs
@@ -207,7 +207,7 @@ impl<'a> Outlines<'a> {
                 .coords
                 .split_at_mut(coords.len().min(memory.coords.len()));
             for (dst, src) in provided_coords.iter_mut().zip(coords.iter()) {
-                *dst = src.to_f32();
+                *dst = src.to_f32() as f64;
             }
             remaining_coords.fill(0.0);
         }
@@ -272,7 +272,7 @@ impl<'a> Outlines<'a> {
         &self,
         shape: &'a ShapePart<'a>,
         scale: f64,
-        coords: &mut [f32],
+        coords: &mut [f64],
         transforms: &mut [Transform],
         segment_memory: &mut [Segment],
         pen: &mut impl OutlinePen,
@@ -317,7 +317,7 @@ impl<'a> Outlines<'a> {
             } else {
                 &axis_deltas[..total_num_segments * 4]
             };
-            let scalar = coord.abs() as f64;
+            let scalar = coord.abs();
 
             for (segment, delta) in all_segments.iter_mut().zip(axis_deltas.chunks_exact(4)) {
                 // This is autovectorized quite well, but we don't get FMA. I'm
@@ -410,7 +410,7 @@ impl<'a> Outlines<'a> {
         &self,
         composite: &'a CompositePart<'a>,
         scale: f64,
-        coords: &mut [f32],
+        coords: &mut [f64],
         transforms: &mut [Transform],
         segment_memory: &mut [Segment],
         pen: &mut impl OutlinePen,
@@ -478,7 +478,7 @@ impl<'a> Outlines<'a> {
         &self,
         part: &Part<'a>,
         scale: f64,
-        coords: &mut [f32],
+        coords: &mut [f64],
         transforms: &mut [Transform],
         segment_memory: &mut [Segment],
         pen: &mut impl OutlinePen,
@@ -499,8 +499,8 @@ impl<'a> Outlines<'a> {
 /// See <https://github.com/harfbuzz/harfbuzz/blob/3c81945/src/hb-aat-var-hvgl-table.cc#L350-L389>.
 fn composite_apply_to_coords<'a>(
     part: &'a CompositePart<'a>,
-    out_coords: &mut [f32],
-    coords: &[f32],
+    out_coords: &mut [f64],
+    coords: &[f64],
 ) -> Result<(), ReadError> {
     let ecs = part.extremum_column_starts()?;
     let extremum_row_indices = part.extremum_row_indices()?;
@@ -513,7 +513,7 @@ fn composite_apply_to_coords<'a>(
     {
         *out_coords
             .get_mut(row_idx.get() as usize)
-            .ok_or(ReadError::OutOfBounds)? += delta.get();
+            .ok_or(ReadError::OutOfBounds)? += delta.get() as f64;
     }
 
     for (axis_idx, coord) in coords.iter().copied().enumerate() {
@@ -541,7 +541,7 @@ fn composite_apply_to_coords<'a>(
                 .get();
             *out_coords
                 .get_mut(row as usize)
-                .ok_or(ReadError::OutOfBounds)? += delta * scalar;
+                .ok_or(ReadError::OutOfBounds)? += delta as f64 * scalar;
         }
     }
 
@@ -554,7 +554,7 @@ fn composite_apply_to_coords<'a>(
 fn composite_apply_to_transforms<'a>(
     part: &'a CompositePart<'a>,
     transforms: &mut [Transform],
-    coords: &[f32],
+    coords: &[f64],
 ) -> Result<(), ReadError> {
     let master_translation_deltas = part.master_translation_deltas()?;
     let mut extremum_translation_deltas = part.extremum_translation_deltas()?;
@@ -592,7 +592,7 @@ fn composite_apply_to_transforms<'a>(
                 break;
             };
 
-            let scalar = coord.abs() as f64;
+            let scalar = coord.abs();
             *dst = dst.pre_translate(delta.x as f64 * scalar, delta.y as f64 * scalar);
         }
     } else {
@@ -660,7 +660,7 @@ fn composite_apply_to_transforms<'a>(
                 if pos != (coord > 0.0) {
                     continue;
                 }
-                let scalar = coord.abs() as f64;
+                let scalar = coord.abs();
 
                 if extremum_rotation_delta != 0.0 {
                     let mut center_x = extremum_translation_delta.x as f64;
--- a/skrifa/src/outline/hvgl/outline.rs
+++ b/skrifa/src/outline/hvgl/outline.rs
@@ -29,7 +29,7 @@ impl Outline<'_> {
         // One transform per subpart
         size += self.part.num_total_subparts() as usize * size_of::<Transform>();
         // One axis coordinate per axis
-        size += self.part.num_total_axes() as usize * size_of::<f32>();
+        size += self.part.num_total_axes() as usize * size_of::<f64>();
         if size != 0 {
             // If we're given a buffer that is not aligned, we'll need to
             // adjust, so add our maximum alignment requirement in bytes.


behdad:
> It feels like I'm missing something--the composite part table has a "max number of axes in any subpart" field that neither HarfBuzz nor this implementation makes use of. I assume it can be used to preallocate _something_, but I'm not sure what.

HarfBuzz just uses vectors, and reuses the vectors (via the "scratch" pattern), so the realloc cost disappears. We do prealloc the vector for that size. Here:

https://github.com/harfbuzz/harfbuzz/blob/3c819450d469b617dff09949ac080b9f6f9cf433/src/hb-aat-var-hvgl-table.hh#L575
https://github.com/harfbuzz/harfbuzz/blob/3c819450d469b617dff09949ac080b9f6f9cf433/src/hb-aat-var-hvgl-table.hh#L584

behdad:
I skimmed the drawing part. LGTM!

valadaptive:
> HarfBuzz just uses vectors, and reuses the vectors (via the "scratch" pattern), so the realloc cost disappears. We do prealloc the vector for that size. Here:

I'm doing the same here (although my benchmarks use stack memory). The field I was thinking of is [the `maxNumExtremes` field](https://github.com/harfbuzz/harfbuzz/blob/3c819450d469b617dff09949ac080b9f6f9cf433/src/hb-aat-var-hvgl-table.hh#L378), which tracks the maximum of subparts' axis counts instead of the sum like `totalNumAxes` does. It doesn't seem to be used anywhere in the HarfBuzz code (and I don't know what it would be used for).

> For performance, you can compare using HB's perf/benchmark-font suite filtered for `draw_glyph`. Note that the HB impl is in my measurements, 2.5x faster than the Apple libhvf impl.

Looks like 280ms to outline the whole font, compared to HB's 230ms. Definitely more room for me to optimize, but I'm feeling OK about those numbers (especially since they're both much faster than libhvf, as you mentioned).

Thanks for pointing me to the HarfBuzz hvgl implementation, by the way! It was pretty straightforward to port over, and I definitely wouldn't have been able to figure things out from the spec alone.

behdad:
> I'm doing the same here (although my benchmarks use stack memory). The field I was thinking of is [the `maxNumExtremes` field](https://github.com/harfbuzz/harfbuzz/blob/3c819450d469b617dff09949ac080b9f6f9cf433/src/hb-aat-var-hvgl-table.hh#L378), which tracks the maximum of subparts' axis counts instead of the sum like `totalNumAxes` does. It doesn't seem to be used anywhere in the HarfBuzz code (and I don't know what it would be used for).

I believe Apple uses `maxNumExtremes` to pre-allocate a "blend factor" (what we call "scalar") to be computed first for each extreme's negative/positive side. It then does a multiply of that vector with the float matrix of deltas. This was, in at least the first CoreText implementation that appeared, done with libBLAS. At least, that's how they had written the paper / initial implementation. I then suggested to them that this will do at least 50% extra work since at least half of the blend vector will be zeroes, so I think they moved away from that implementation. In HB we calculate the scalar as needed and never store in a vector for sure.

behdad:
@valadaptive Did you happen to find anything that can be improved in the HB impl? Thanks.

valadaptive:
The HB implementation looks good to me; my changes were mostly just using Rust idioms and less pointer manipulation.

behdad:
I can't get `hb-view --font-funcs fontations` to render PingFangUI with this branch. No idea why.

behdad:
To build HB's fontations backend with this change, we just need:
```diff
diff --git a/src/rust/font.rs b/src/rust/font.rs
index e2180a2e0..e7afad243 100644
--- a/src/rust/font.rs
+++ b/src/rust/font.rs
@@ -9,9 +9,7 @@ use std::sync::{Mutex, OnceLock};
 
 use skrifa::charmap::Charmap;
 use skrifa::charmap::MapVariant::Variant;
-use skrifa::color::{
-    Brush, ColorGlyphCollection, ColorPainter, ColorStop, CompositeMode, Extend, Transform,
-};
+use skrifa::color::{Brush, ColorGlyphCollection, ColorPainter, ColorStop, CompositeMode, Extend};
 use skrifa::font::FontRef;
 use skrifa::instance::{Location, NormalizedCoord, Size};
 use skrifa::metrics::{BoundingBox, GlyphMetrics};
@@ -23,7 +21,7 @@ use skrifa::raw::tables::vorg::Vorg;
 use skrifa::raw::tables::vvar::Vvar;
 use skrifa::raw::TableProvider;
 use skrifa::OutlineGlyphCollection;
-use skrifa::{GlyphId, GlyphNames, MetadataProvider};
+use skrifa::{GlyphId, GlyphNames, MetadataProvider, Transform};
 
 // A struct for storing your “fontations” data
 #[repr(C)]
```
Then it builds and rendering works.

I ran:
```
$ perf/hb-draw-compare PingFangUI-mac15.0.1.ttc --hb-info b/util/hb-info --hb-view b/util/hb-view --font-funcs2 fontations
```
for default instance, the two don't match. But I believe that's because of the PingFang peculiarity of the `avar2` table, which HB has adjusted for. We would need the same in fontations. The diffs are really small. Investigating more.

I also ran the perf suite on it to verify @valadaptive's numbers. I get similar results (<20% slower than HB):
```
$ ninja -Cbuild perf/benchmark-font && build/perf/benchmark-font --benchmark_filter=draw PingFangUI-mac15.0.1.ttc 
-----------------------------------------------------------------------------------------------------
Benchmark                                                           Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------
BM_Font/draw_glyph/PingFangUI-mac15.0.1.ttc/ot                    276 ms          275 ms            3
BM_Font/draw_glyph/PingFangUI-mac15.0.1.ttc/fontations            327 ms          325 ms            2
BM_Font/draw_glyph/PingFangUI-mac15.0.1.ttc/var/ot                300 ms          299 ms            2
BM_Font/draw_glyph/PingFangUI-mac15.0.1.ttc/var/fontations        352 ms          350 ms            2
```
However, with `hb-draw-compare` tool, the fontations rendering is *really* slow. I suppose skrifa doesn't cache *something*, I don't know what. I'll investigate.

behdad:
> However, with `hb-draw-compare` tool, the fontations rendering is _really_ slow. I suppose skrifa doesn't cache _something_, I don't know what. I'll investigate.

So, the rendering part for fontations is 3.5x slower than for ot. Which is quite something. It feels like creating a `hb_font_t`'s fontations backend is really slow. Profiling shows that we're loading glyph names repeatedly, which is indeed what HB's fontations backend does *per-font* instead of caching on the face. However, it's not clear to me why names are accessed at all. Anyway, this sounds like something on the HB side than fontations side.

behdad:
> Using `f64` arithmetic for the transforms seems to fix those cases. There are still some very tiny differences that seem to be a result of the HarfBuzz backend rounding one or two coords differently in every glyph. Not sure what's up with that.

I can confirm all the remaining diffs are of `0.00391` value. I might look into more.

behdad:
> Using `f64` arithmetic for the transforms seems to fix those cases. There are still some very tiny differences that seem to be a result of the HarfBuzz backend rounding one or two coords differently in every glyph. Not sure what's up with that.

We do all coords math in `hvgl` in double as well. Is that what you do?

valadaptive:
Ah, that was the issue. Rerunning all the diff tests is slow but I think that's fixed things. Does CoreText do everything in double precision as well? Seems maybe not the best for performance.

behdad:
> Ah, that was the issue. Rerunning all the diff tests is slow but I think that's fixed things. Does CoreText do everything in double precision as well? Seems maybe not the best for performance.

I confirm there's no diff now. And we match CoreText, and the info shared with us requests for doubles as well. I could measure no speed diff.

behdad:
> Rerunning all the diff tests is slow

I'm using this diff for now, until I fix https://github.com/harfbuzz/harfbuzz/issues/5384
```diff
diff --git a/src/rust/font.rs b/src/rust/font.rs
index e2180a2e0..ea2d11ab1 100644
--- a/src/rust/font.rs
+++ b/src/rust/font.rs
@@ -988,6 +988,7 @@ extern "C" fn _hb_fontations_glyph_from_name(
     glyph: *mut hb_codepoint_t,
     _user_data: *mut ::std::os::raw::c_void,
 ) -> hb_bool_t {
+    return false as hb_bool_t; // Not implemented
     let data = unsafe { &mut *(font_data as *mut FontationsData) };
 
     let name = unsafe { std::slice::from_raw_parts(name as *const u8, len as usize) };
```

behdad:
@valadaptive Also, I don't know what your motivation for this work was, but there's also `VARC` that we need to port from HB to fontations. It's unfortunate that we have two formats for basically the same thing now: `VARC` becoming standard, and `hvgl` in the wild...

valadaptive:
Just something I've had on the backburner for a while, which I learned about from a [Parley bug](https://github.com/linebender/parley/issues/242). I thought it'd be a good way to learn more about fontations without realizing just how weird the HVF format is.

valadaptive:
I feel like I'm not very qualified to rant about HVF since I'm missing a lot of context, but there are certainly some design decisions that I'm not sure I fully grasp. Beyond half the coords being floats and the other half being doubles, I'm not sure why they've required the subtables, `hvgl` table, and *entire font file* to be aligned to an 8-byte boundary. Are they planning on removing unaligned loads from the M5 or something? They've tried to lay out everything in a supposedly cache-friendly manner, and then required a double indirection to load each subpart, and that's even before `hvpm` comes into the picture. They've included a `maxNumExtremes` field, but as you pointed out, it's basically useless, and they've omitted any "maximum subpart segments" field so preallocating a fixed block of memory requires two full tree traversals. Were all these odd decisions based around them trying to use matrix operations for the entire thing and just sticking BLAS data structures into an OpenType table?

I'm sure there's something here I'm missing.

behdad:
FWIW, we just merged the HB branch.
https://github.com/harfbuzz/harfbuzz/pull/5379

behdad:
> BM_Font/draw_glyph/PingFangUI-mac15.0.1.ttc/ot                    276 ms          275 ms            3

Note that HB has avx2 codepath for the matrix multiplication. If I enable that, the timing goes down to 210ms.

behdad:
> FWIW, we just merged the HB branch. [harfbuzz/harfbuzz#5379](https://github.com/harfbuzz/harfbuzz/pull/5379)

Also the fonttools branch: https://github.com/fonttools/fonttools/pull/3876

PR #1565, opened 2025-07-17
[skrifa] tthint: round ppem based on head.flags 

dfrg:
When hinting, round ppem prior to computing scale when the font has bit 3 set in `head.flags`. This matches FreeType behavior when scaling at fractional sizes.

Additional fix for #1544, specifically targeting the MS core web fonts

@drott we can't really match this behavior for glyph metrics because we don't know when hinting was requested. This patch adds a new `OutlineGlyphCollection::fractional_size_hinting() -> bool` method. When Skia constructs a `GlyphMetrics` object, it should check the result of this method and round the size passed to ` GlyphMetrics::new()` when hinting is enabled to ensure matching metrics.

### DIFF
--- a/skrifa/src/outline/glyf/mod.rs
+++ b/skrifa/src/outline/glyf/mod.rs
@@ -54,7 +54,7 @@ pub struct Outlines<'a> {
     units_per_em: u16,
     os2_vmetrics: [i16; 2],
     prefer_interpreter: bool,
-    prevent_fractional_scaling: bool,
+    pub(crate) prevent_fractional_scaling: bool,
 }
 
 impl<'a> Outlines<'a> {
--- a/skrifa/src/outline/glyf/mod.rs
+++ b/skrifa/src/outline/glyf/mod.rs
@@ -54,7 +54,7 @@ pub struct Outlines<'a> {
     units_per_em: u16,
     os2_vmetrics: [i16; 2],
     prefer_interpreter: bool,
-    pub(crate) prevent_fractional_scaling: bool,
+    pub(crate) fractional_size_hinting: bool,
 }
 
 impl<'a> Outlines<'a> {
@@ -62,7 +62,7 @@ impl<'a> Outlines<'a> {
         let head = font.head().ok()?;
         // If bit 3 of head.flags is set, then we round ppems when
         // scaling
-        let prevent_fractional_scaling = head.flags() & 0x8 != 0;
+        let fractional_size_hinting = head.flags() & 0x8 == 0;
         let loca = font.loca(Some(head.index_to_loc_format() == 1)).ok()?;
         let glyf = font.glyf().ok()?;
         let glyph_metrics = GlyphHMetrics::new(font)?;
@@ -131,7 +131,7 @@ impl<'a> Outlines<'a> {
             units_per_em: font.head().ok()?.units_per_em(),
             os2_vmetrics,
             prefer_interpreter,
-            prevent_fractional_scaling,
+            fractional_size_hinting,
         })
     }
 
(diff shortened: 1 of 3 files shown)


skrifa/src/outline/glyf/mod.rs
```diff
@@ -54,11 +54,16 @@ pub struct Outlines<'a> {
...
+    pub(crate) prevent_fractional_scaling: bool,
 }
 
 impl<'a> Outlines<'a> {
     pub fn new(font: &FontRef<'a>) -> Option<Self> {
-        let loca = font.loca(None).ok()?;
+        let head = font.head().ok()?;
+        // If bit 3 of head.flags is set, then we round ppems when
+        // scaling
+        let prevent_fractional_scaling = head.flags() & 0x8 != 0;
```
rsheeter: I was expecting to see use of an enum for the flags. https://docs.rs/read-fonts/latest/read_fonts/tables/head/type.Head.html#method.flags helpfully advises me to use "the flags enum" but doesn't link to it. Bleh.

skrifa/src/outline/glyf/mod.rs
```diff
@@ -126,6 +131,7 @@ impl<'a> Outlines<'a> {
             units_per_em: font.head().ok()?.units_per_em(),
             os2_vmetrics,
             prefer_interpreter,
+            prevent_fractional_scaling,
```
rsheeter: Google generally prefers positive flags, e.g. enable fractional scaling. This tends to result in more intuitive use, avoids things like "don't prevent X."

skrifa/src/outline/glyf/mod.rs
dfrg: Happy to "enumize" head.flags in a follow up

skrifa/src/outline/glyf/mod.rs
```diff
@@ -323,7 +336,7 @@ impl<'a> HarfBuzzScaler<'a> {
         coords: &'a [F2Dot14],
     ) -> Result<Self, DrawError> {
         outline.ensure_point_count_limit()?;
-        let (is_scaled, scale) = outlines.compute_scale(ppem);
+        let (is_scaled, scale) = outlines.compute_scale(ppem, false);
```
rsheeter: Can't tell what false is, perhaps an enum makes sense?

skrifa/src/outline/glyf/mod.rs
```diff
@@ -1398,4 +1411,20 @@ mod tests {
         // And we get an error instead of an overflow panic
         assert!(matches!(result, Err(DrawError::TooManyPoints(_))));
     }
+
```
rsheeter, line 1420: Suggest adding a test that just checks the result between fractional scaling on/off differs.

rsheeter (approved):
Lgtm with minor suggestions

dfrg:
> 1. Would this cause (additional?) undesirable rendering changes when used without the combination of `OutlineGlyphCollection::fractional_size_hinting()` and rounding the metrics? - Or would this improve how glyphs look, but not fix or adjust the metrics yet?

The latter-- it should produce better rendering but the scaler does this by rounding the size internally which means that the outline will no longer match the advance width produced by `GlyphMetrics` for fractional sizes. Note that this is not true when you are using the advance width from `AdjustedMetrics`. That one will use the same scale factor.

> 2. Could you make a release for this and [[skrifa] tthint: round value for MPPEM instruction #1545](https://github.com/googlefonts/fontations/pull/1545) ?

Done

drott:
> The latter-- it should produce better rendering but the scaler does this by rounding the size internally which means that the outline will no longer match the advance width produced by `GlyphMetrics` for fractional sizes. Note that this is not true when you are using the advance width from `AdjustedMetrics`. That one will use the same scale factor.

Thanks, so I will roll t his to Chromium (first), then to Skia and work on improvement there afterwards.

drott:
> The latter-- it should produce better rendering but the scaler does this by rounding the size internally which means that the outline will no longer match the advance width produced by `GlyphMetrics` for fractional sizes. Note that this is not true when you are using the advance width from `AdjustedMetrics`. That one will use the same scale factor.

My assumption here was, that this fix is neutral in terms of advances and we can an achievement, but in the roll, I do see advances changing - they might use AdjustedMetrics for hinted fonts, compare:
With the rolls above, I do get advance width changes https://chromium-review.googlesource.com/c/chromium/src/+/6779139?checksRunsSelected=linux-rel,linux_chromium_asan_rel_ng&tab=checks

I do seem them changing however for what I think are not fractional font sizes.

PR #1566, opened 2025-07-17
[read-fonts] change head.flags to an enum

dfrg:
Seems this field didn't get much use.

This probably breaks fontc.


cmyr (approved):
okay looks good, will require minor bumps of all the crates ofc.

PR #1568, opened 2025-07-21
[skrifa] match FT for CFF to autohint outline conversion 

dfrg:
Our CFF path to TrueType-style outline conversion that is used to feed the autohinter was generating a slightly different point configuration leading to tiny autohinting differences. This aims to match FreeType by omitting the last on curve of a contour if it matches the starting position.

Tested against Material Icons and Cantarell VF which now match for all autohinting modes.

### DIFF
--- a/skrifa/src/outline/unscaled.rs
+++ b/skrifa/src/outline/unscaled.rs
@@ -139,11 +139,19 @@ impl UnscaledOutlineRef<'_> {
     }
 }
 
+#[derive(Copy, Clone)]
+enum PendingElement {
+    Line([f32; 2]),
+    Cubic([f32; 6]),
+}
+
 /// Adapts an UnscaledOutlineSink to be fed from a pen while tracking
 /// memory allocation errors.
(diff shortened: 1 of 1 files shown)


dfrg:
Also update `NopFilteringSink` so that the path conversion style comparison tests pass.

skrifa/src/outline/unscaled.rs
```diff
@@ -364,4 +394,51 @@ mod tests {
         assert_eq!(bottom_point_ix, 0);
         assert_eq!(bottom_y, Some(80));
     }
+
+    /// When a contour ends with a line or cubic whose end matches the start
+    /// point, omit the last on curve.
```
drott: Optional nit: Add that we're doing this to match FreeType?

drott (approved):
Great, thanks for solving this. LGTM.

skrifa/src/outline/unscaled.rs
dfrg: Good call. Done.

PR #1569, opened 2025-07-21
Run fauntlet in CI

dfrg:
Adds a fauntlet check to CI along with a small set of test fonts selected to test specific code paths. The reasoning for each is noted in the new fauntlet/test_fonts/README.md

There's at least one remaining bug so this should fail but at least we'll know that the check works properly.


fauntlet/test_fonts/README.md
rsheeter, line 1: It would be nice to not add large binaries, particularly the ones that 100+KB. We often subset to mitigate this. Is that possible here? - subset them, documenting the exact commands in readme.

fauntlet/test_fonts/README.md
```diff
@@ -0,0 +1,23 @@
...
+This is a hand curated selection of fonts to test Skrifa output against
+FreeType on each pull request to prevent obvious regressions.
+
+## Running tests
+
+```shell
+cargo run --release -p fauntlet -- compare-glyphs --hinting-engine all --hinting-target all --exit-on-fail test_fonts/*.*
+```
+
+## Fonts
```
rsheeter, line 12: Suggest noting where you gott each font from in slightly more detail, e.g. repo, path, and commit.

fauntlet/test_fonts/README.md
dfrg: I'll subset them and perhaps drop the layout tables as well.

dfrg:
> LGTM but lets not leave it failing for long, that teaches people to ignore the check extremely quickly.

Makes sense. My intent is to fix the failure before actually merging this so we keep CI clean.

Issue #870, opened 2024-04-02
Picking Family and Style names with WWS


bungeman:
OpenType fonts have at least three family names (1, 16, 21) and three style names (2, 17, 22). The general way to pick these is to prefer the higher numbered ones, unless OS2::fsSelection bit 8 exists and is set in which case 21 and 22 should be ignored by newer interpreters (unless of course the lower numbered ones don't exist).

These rules are non-trivial and non-obvious so it would be nice if the library handled this instead of users needing to know about these interesting rules (which may change again in the future). Since retrieving the family and style names are a common use, it may be best to provide a `FontRef::family_name() -> StringId` (or `FontRef::family_name() -> LocalizedStrings`) and a similar `FontRef::style_name()` which works in a similar way.

This idea was first discussed at on Skia change [[fontations] Modernize family name](https://skia-review.googlesource.com/c/skia/+/833533/1..4/src/ports/fontations/src/ffi.rs#b467), though note that this change does not include a fallback to the nameid 21 when it would normally be ignored but 16 and 1 aren't available.

PR #1576, opened 2025-07-24
feat: get best names

ollimeier:
I often need the family name, subfamily name or both together as the full name. And the preferred one is often not name id 4, but 16 + 17. 

This PR makes it much more easy to get the 'nice' names. I have done something similar for fonttools:
https://github.com/fonttools/fonttools/blob/d2ac9cbf488acc867866297d76a9671fe992947f/Lib/fontTools/ttLib/tables/_n_a_m_e.py#L154

Note: I added `write-fonts` to Cargo.toml because this way I can build a demo font for the unittests.

I hope you find this handy as well. Happy to get feedback :)

### DIFF
--- a/skrifa/src/provider.rs
+++ b/skrifa/src/provider.rs
@@ -32,10 +32,24 @@ pub trait MetadataProvider<'a>: Sized {
     /// given informational string identifier.
     fn localized_strings(&self, id: StringId) -> LocalizedStrings<'a>;
 
+    /// Returns an optional best family name.
+    /// WWS Family Name (NID 21)
+    /// Typographic Family Name (NID 16)
+    /// Family Name (NID 1)
     fn best_family_name(&self) -> Option<String>;
 
+    /// Returns an optional best subfamily name.
+    /// WWS Subfamily Name (NID 22)
+    /// Typographic Subfamily Name (NID 17)
+    /// Subfamily Name (NID 2)
     fn best_subfamily_name(&self) -> Option<String>;
 
+    /// Returns an optional best full name.
+    /// WWS Family Name + WWS Subfamily Name (NID 21 + 22)
+    /// Typographic Family Name + Typographic Subfamily Name (NID 16 + 17)
+    /// Family Name + Subfamily Name (NID 1 + 2)
+    /// Full Name (NID 4)
+    /// PostScript Name (NID 6)
     fn best_full_name(&self) -> Option<String>;
 
     /// Returns the mapping from glyph identifiers to names.
@@ -89,14 +103,28 @@ impl<'a> MetadataProvider<'a> for FontRef<'a> {
         LocalizedStrings::new(self, id)
     }
 
+    /// Returns an optional best family name.
+    /// WWS Family Name (NID 21)
+    /// Typographic Family Name (NID 16)
+    /// Family Name (NID 1)
     fn best_family_name(&self) -> Option<String> {
         get_best_family_name(self)
     }
 
+    /// Returns an optional best subfamily name.
+    /// WWS Subfamily Name (NID 22)
+    /// Typographic Subfamily Name (NID 17)
+    /// Subfamily Name (NID 2)
     fn best_subfamily_name(&self) -> Option<String> {
         get_best_subfamily_name(self)
     }
 
+    /// Returns an optional best full name.
+    /// WWS Family Name + WWS Subfamily Name (NID 21 + 22)
+    /// Typographic Family Name + Typographic Subfamily Name (NID 16 + 17)
+    /// Family Name + Subfamily Name (NID 1 + 2)
+    /// Full Name (NID 4)
+    /// PostScript Name (NID 6)
     fn best_full_name(&self) -> Option<String> {
         get_best_full_name(self)
     }


skrifa/src/provider.rs
```diff
@@ -29,6 +32,12 @@ pub trait MetadataProvider<'a>: Sized {
     /// given informational string identifier.
     fn localized_strings(&self, id: StringId) -> LocalizedStrings<'a>;
 
+    fn best_family_name(&self) -> Option<String>;
+
+    fn best_subfamily_name(&self) -> Option<String>;
+
+    fn best_full_name(&self) -> Option<String>;
+
```
cmyr: these would deserve docs explaining what 'best' means :)

skrifa/src/string.rs
```diff
@@ -597,12 +597,72 @@ const LANGUAGE_ID_TO_BCP47: &[(u16, &str)] = &[
     (0x540a, "es-US"),        //Spanish
 ];
 
+fn _get_best_name(font: &FontRef, name_ids: Vec<StringId>) -> Option<String> {
```
cmyr: you can omit the leading `_`, pub/private is enforced by the compiler.

skrifa/src/string.rs
```diff
@@ -597,12 +597,72 @@ const LANGUAGE_ID_TO_BCP47: &[(u16, &str)] = &[
...
+    for name_id in name_ids {
+        if let Some(name) = LocalizedStrings::new(font, name_id).english_or_first() {
+            return Some(name.to_string());
+        }
+    }
+    None
```
cmyr: you can rewrite this as,

```rust
name_ids
    .iter()
    .find_map(|id| LocalizedStrings::new(font, id).english_or_first()
    .map(|name| name.to_string())
```

skrifa/src/string.rs
```diff
@@ -597,12 +597,72 @@ const LANGUAGE_ID_TO_BCP47: &[(u16, &str)] = &[
...
+}
+
+pub fn get_best_family_name(font: &FontRef) -> Option<String> {
+    let name_ids = vec![
+        StringId::WWS_FAMILY_NAME,
+        StringId::TYPOGRAPHIC_FAMILY_NAME,
+        StringId::FAMILY_NAME,
+    ];
```
cmyr: instead of `vec!` (which allocates) this should just be a const array:

```rust
const NAME_IDS: [StringId; 3] = [
    StringId::WWS_FAMILY_NAME,
    StringId::TYPOGRAPHIC_FAMILY_NAME,
    StringId::FAMILY_NAME,
];

File: skrifa/src/string.rs [hunk 5]
```diff
@@ -597,12 +597,72 @@ const LANGUAGE_ID_TO_BCP47: &[(u16, &str)] = &[
...
+    _get_best_name(font, name_ids)
+}
+
+pub fn get_best_full_name(font: &FontRef) -> Option<String> {
+    let ids = [
+        (21, Some(22)),
+        (16, Some(17)),
+        (1, Some(2)),
+        (4, None), // 4 = Full Name
+        (6, None), // 6 = PostScript Name
```
cmyr: any reason not to have these be `StringId` instead of untyped integers?

cmyr (commented):
Thanks for the PR and sorry for the slow response!

I'll leave it up to @dfrg to make the call on whether or not this API is an appropriate addition to skrifa, but below are the things I would want to see addressed if this was going to be merged. :)

skrifa/src/string.rs
ollimeier: Thanks for the info. I changed it.

skrifa/src/string.rs
ollimeier: Thanks for the recommendation. I changed it.

skrifa/src/string.rs
ollimeier: I changed it as recommended, but I have to convert it via .`to_vec()` when passing to `get_best_name`. This does not feel right. Is this what you would do or is there a better/smarter/more elegant solution?

ollimeier: Please see: https://github.com/googlefonts/fontations/pull/1576/commits/870829775fb444bdd87085cd9750f714242e2302

skrifa/src/string.rs
ollimeier: I found is more user-friendly, because I am so used to name ID numbers. It also made the code have short lines :)
But I get your point. I changed it.

skrifa/src/provider.rs
ollimeier: You're right. I missed that. I added it. Thanks a lot for your valid feedback.

valadaptive:
I believe this would resolve https://github.com/googlefonts/fontations/issues/870

skrifa/src/string.rs
```diff
@@ -597,12 +597,70 @@ const LANGUAGE_ID_TO_BCP47: &[(u16, &str)] = &[
     (0x540a, "es-US"),        //Spanish
 ];
 
+fn get_best_name(font: &FontRef, name_ids: Vec<StringId>) -> Option<String> {
```
valadaptive, line 600: Any reason not to take an `impl IntoIterator<Item = StringId>`?

ollimeier: I am happy to improve it if there is a chance for a merge.

dfrg:
Thanks for the patch! While _something_ like this is desirable to assist users in accessing the common names (and to handle proper searching order, see #870), I think this approach is too opinionated both by assuming the user wants the "english or first" name and by converting to `String` which makes language selection impossible and forces a perhaps unnecessary allocation. 

In addition, these methods should probably exist on some intermediate object rather than being added directly to `MetadataProvider`. Ultimately, we probably want to move this functionality to `read-fonts` as well and I'm not quite sure what that would look like yet.

PR #1577, opened 2025-07-30
[klippa] subset MarkToLig subtable

qxliu76:
And patch in a fix for MarkToBase table to remove base glyphs without defined anchor points after subsetting, see https://github.com/harfbuzz/harfbuzz/pull/5412, https://github.com/harfbuzz/harfbuzz/pull/5434


cmyr (approved):
Looks good, but why is this one commit? Are these two changes related in any way?


Especially for bug fixes it is valuable to have the change encapsulated in a separate commit and PR, to make it easier to explore in the future with `git blame`, or to revert if necessary...

qxliu76:
yeah you're right, I'll avoid combining 2 changes in the future.:)

PR #1584, opened 2025-08-05
[read-fonts] inline all the things

dfrg:
Just adds a bunch of `#[inline]` attributes to codegen'd functions.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -831,6 +831,7 @@ impl Field {
 
         Some(quote! {
             #( #docs )*
+            #[inline]
             pub fn #name(&self) -> #return_type {
                 let range = #range_stmt;
                 #read_stmt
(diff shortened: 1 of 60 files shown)


rsheeter (approved):
LGTM assuming it helps. Would be nice to cite metrics for improvement

behdad:
On Mac I don't see any change whatsoever. On Linux, I actually see 5% slowdown on the ligature-heavy `BM_Shape/SourceSerifVariable-Roman.ttf/react-dom.txt/harfrust` benchmark and 1% slowdown in other benches.

I'm okay merging, but looks like our config `https://github.com/harfbuzz/harfbuzz/blob/main/src/rust/Cargo.toml` already does a good job of inlining.

Perhaps the config should be documented in the HR repo itself, or enabled there by default.

dfrg:
Should still probably emit inline attributes for consumers that don't use lto, but this is way out of date so closing.

PR #1588, opened 2025-08-06
[codegen] capture initial fixed fields in a struct

dfrg:
This is a fairly large update to codegen. It is technically breaking but will only affect users that reference the `TableRef` type explicitly.

The change basically captures the sequence of "fixed" fields for each table into a separate zerocopy struct (fixed is defined as always present and existing at a constant offset). At read time, we parse a reference to this struct and store it as a new field on `TableRef`. Any subsequent access to the fields from the fixed set simply reads from that reference. 

Note that we still continue to validate the trailing conditional/variable size fields at this point. We may want to reconsider that as a next step.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -19,9 +19,12 @@ impl Fields {
             .flat_map(Field::input_fields)
             .collect::<ReferencedFields>();
 
+        let mut is_fixed = true;
         for field in fields.iter_mut() {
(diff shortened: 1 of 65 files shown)


behdad:
Is there an increased setup cost per shape call? I wonder if https://github.com/harfbuzz/harfrust/issues/115 is relevant.

dfrg:
Completely obsolete.

PR #1589, opened 2025-08-07
Add read and write for `DSIG` table

Hoolean:
Hello! One from the DaMa team: this PR adds `DSIG` to `read-fonts` and `write-fonts`.

We have added unit tests that compare the reading and writing against fontTools, for both a dummy table, and a table with a single signature, and all appears to be working as expected.

However, we are opening this as a draft pull request as we would like to confirm some implementation details of the codegen; would we be able to pick your brain on the inline TODOs there? : )


read-fonts/src/tables/dsig.rs
```diff
@@ -0,0 +1,58 @@
...
+include!("../../generated/generated_dsig.rs");
+
+#[cfg(test)]
+mod tests {
+    use font_test_data::bebuffer::BeBuffer;
+
+    use super::{Dsig, PermissionFlags};
+    use crate::{FontData, FontRead};
+
+    /// An empty, dummy DSIG, as inserted by fonttools.
```
madig, line 52: Maybe you can add a `Dsig::DUMMY` variable or constructor? I suppose that most workflows dealing with DSIG are going to use it.

Hoolean: Was tempted but it's less of a special case than I'd worried before looking under the hood; it's just a regular table construction with 0 signature records (as seen in `write-fonts` where we construct it in a type-safe way), and so I don't think we need to hang onto the raw bytes anywhere in the end :+1:

resources/codegen_inputs/dsig.rs
```diff
@@ -0,0 +1,68 @@
...
+/// [Signature Record](https://learn.microsoft.com/en-us/typography/opentype/spec/dsig#table-structure)
+record SignatureRecord {
+    /// Format of the signature
+    // TODO: How do we validate?
+    // TODO: Can we use an enum, even though the format is on the 'outside'?
+    #[compile(1)]
+    format: u32,
+
+    /// Length of signature in bytes
+    // TODO: Can this be derived automatically?
```
cmyr: it's unclear to me why this field even exists; it seems redundant given that the length is also encoded in the table?

resources/codegen_inputs/dsig.rs
```diff
@@ -0,0 +1,68 @@
...
+flags u16 PermissionFlags {
+    /// Bit 0: Cannot be resigned
+    CANNOT_BE_RESIGNED = 0b0000_0000_0000_0001,
+}
+
+/// [Signature Record](https://learn.microsoft.com/en-us/typography/opentype/spec/dsig#table-structure)
+record SignatureRecord {
+    /// Format of the signature
+    // TODO: How do we validate?
+    // TODO: Can we use an enum, even though the format is on the 'outside'?
```
cmyr: given that there is only one format we can leave this as is; we will validate in a custom getter fn for the signature block itself, and we won't bother with an enum (as we currently don't bother with other tables that only have a single format)

resources/codegen_inputs/dsig.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Length of signature in bytes
+    // TODO: Can this be derived automatically?
+    #[compile(self.compute_signature_block_len())]
+    length: u32,
+    
+    /// Offset to the signature block from the beginning of the table
+    // TODO: Do we still need the traverse skip?
+    // TODO: Do we need to factor in the outer length?
+    #[traverse_with(skip)]
```
cmyr: my hunch is that this should be traversable, but we might need to skip traverse on the signature itself..

resources/codegen_inputs/dsig.rs
```diff
@@ -0,0 +1,68 @@
...
+    signature_block_offset: Offset32<SignatureBlockFormat1>,
```
cmyr, line 40: so for validation, I think we want to provide a custom getter for this offset, which will check that format is set to `1` and error otherwise.

you can do this with the attribute, `#[offset_getter(signature_block)]`, and then implement a `fn signature_block` that resolves the offset.

resources/codegen_inputs/dsig.rs
```diff
@@ -0,0 +1,68 @@
...
+    _reserved1: u16,
+
+    /// Reserved for future use; set to zero.
+    #[skip_getter]
+    #[compile(0)]
+    _reserved2: u16,
+
+    /// Length (in bytes) of the PKCS#7 packet in the signature field.
+    // TODO: Can this be derived automatically?
+    #[compile(self.compute_signature_len())]
```
cmyr:
```suggestion
    #[compile(array_len($signature))]
```

resources/codegen_inputs/dsig.rs
Hoolean: Ye, right...

As far as I can tell, [fonttools doesn't even use this when deserialising](https://github.com/fonttools/fonttools/blob/ec716f11851f8d5a04e3f535b53219d97001482a/Lib/fontTools/ttLib/tables/D_S_I_G_.py#L62-L72) (although it does write it). Some applications may do stronger validation though, and so I'm not sure which is most appropriate for us to match here; I guess we should normally err on the side of strong validation for something like a digital signature?

Hoolean: Maybe a remnant of a format that was never added

resources/codegen_inputs/dsig.rs
Hoolean: Nice; out of interest, is there a way to codegen checking such a constant on read? e.g. do we do anything similar for table versions that should always be 1?

resources/codegen_inputs/dsig.rs
Hoolean: Can move; is there a good rule of thumb for this? e.g. is the issue something along the lines of 'a traversable &[u8] would do the unwanted of digging into each `u8`'? 👀

resources/codegen_inputs/dsig.rs
Hoolean: Ideal ❤️

Hoolean: Applied and removed the now-unnecessary manual function

resources/codegen_inputs/dsig.rs
cmyr: in any case this should be computable.

resources/codegen_inputs/dsig.rs
cmyr: no, currently we just ignore the format field if there is only one known format. It wouldn't be difficult to add a check there, though, and perhaps we should...

resources/codegen_inputs/dsig.rs
cmyr: right actually since it's just a raw u8 slice it will also be traversable.

In general you only need to skip fields that have use custom non-codegen'd types, which don't have traversal implementations.

Hoolean: Nice, I've pruned it now

resources/codegen_inputs/dsig.rs
Hoolean: Implemented; I'll see what we can do with the outer length now too

resources/codegen_inputs/dsig.rs
Hoolean: Agreed:
- For writing, I've kept the manual calculation approach.
- For reading, I've now added validation that the inner- and outer-lengths match.
    - This is stricter than fonttools, which does not care, but I think is worth it to ensure that the cryptographic signature is not ambiguous depending on which length is trusted.

resources/codegen_inputs/dsig.rs
Hoolean: I've added the manual approach suggested above; I'll hold off on validating `compile(...)` fields (e.g. versions) on the assumption that it's a wider change, or is there something you'd recommend we add here?

cmyr: oops missed this! no, if we're doing that it should be a separate patch.

cmyr (approved):
looks good!

PR #1591, opened 2025-08-08
[Coverage/ClassDef] Add a cost() method

behdad:
Needed in HarfRust.

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -160,6 +160,10 @@ impl FeatureTableSubstitutionRecord {
     }
 }
 
+fn bit_storage(v: u32) -> u32 {
+    u32::BITS - v.leading_zeros()
+}
+
 impl<'a> CoverageTable<'a> {
     pub fn iter(&self) -> impl Iterator<Item = GlyphId16> + 'a {
         // all one expression so that we have a single return type
@@ -209,6 +213,14 @@ impl<'a> CoverageTable<'a> {
         match self {
             CoverageTable::Format1(sub) => sub.population(),
             CoverageTable::Format2(sub) => sub.population(),
+        }
+    }
+
+    /// Return the cost of looking up a glyph in this table
+    pub fn cost(&self) -> u32 {
+        match self {
+            CoverageTable::Format1(sub) => sub.cost(),
+            CoverageTable::Format2(sub) => sub.cost(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/layout.rs
```diff
@@ -160,6 +160,15 @@ impl FeatureTableSubstitutionRecord {
     }
 }
 
+fn bit_storage(len: usize) -> u32 {
+    let bits = usize::BITS;
+    if bits == 0 {
+        0
+    } else {
+        bits - len.leading_zeros()
+    }
```
dfrg: I'm guessing this comes from templated code in HB? I think we could just do `usize::BITS - len.leading_zeros()` without the check? I assume if `usize::BITS` is 0 then we're on a fairly useless platform :)

read-fonts/src/tables/layout.rs
```diff
@@ -262,6 +279,11 @@ impl CoverageFormat1<'_> {
     pub fn population(&self) -> usize {
         self.glyph_count() as usize
     }
+
+    /// Return the cost of looking up a glyph in this table
+    pub fn cost(&self) -> u32 {
+        bit_storage(self.glyph_array().len())
```
dfrg: if we do `bit_storage(self.glyph_count() as usize)` then we avoid the cost of reading/validating the array here

read-fonts/src/tables/layout.rs
```diff
@@ -344,6 +366,11 @@ impl CoverageFormat2<'_> {
             .iter()
             .fold(0, |acc, record| acc + record.population())
     }
+
+    /// Return the cost of looking up a glyph in this table
+    pub fn cost(&self) -> u32 {
+        bit_storage(self.range_records().len())
```
dfrg: same here with `self.range_count() as usize`

read-fonts/src/tables/layout.rs
```diff
@@ -537,6 +569,11 @@ impl<'a> ClassDefFormat2<'a> {
             .fold(0, |acc, record| acc + record.population())
     }
 
+    /// Return the cost of looking up a glyph in this table
+    pub fn cost(&self) -> u32 {
+        bit_storage(self.class_range_records().len())
```
dfrg: and here: `self.class_range_count() as usize`

dfrg (approved):
Generally LGTM with a few suggestions to avoid unnecessary work

read-fonts/src/tables/layout.rs
behdad: It actually came from CoPilot :)). I thought it's checking that the input is zero. I had a wrong memory that leading_zeros() is only defined for non-zero numbers. The HB code handles the 0 case separately so I was going to copy it. But reading rust docs, there's no need for a conditional whatsoever. So I'm going to change it to just `usize::BITS - v.leading_zeros()`.

The rust docs say there's also `ilog2` that does mostly the same, except that, that one is undefined for 0. So I think I keep this using leading_zeros.

behdad: I also changed the into to u32. No need to pay for ctz cost of u64 even if it's a single op. :))

PR #1593, opened 2025-08-08
[klippa] subset Context/ChainContext Format2 tables

qxliu76:
This PR depends on #1585

### DIFF
--- a/klippa/src/gsubgpos.rs
+++ b/klippa/src/gsubgpos.rs
@@ -1,7 +1,7 @@
 //! impl subset() for Sequence Context/Chained Sequence Context tables
 use crate::{
-    layout::intersected_glyphs_and_indices,
-    offset::SerializeSerialize,
+    layout::{intersected_glyphs_and_indices, ClassDefSubsetStruct},
+    offset::{SerializeSerialize, SerializeSubset},
     offset_array::SubsetOffsetArray,
     serialize::{SerializeErrorFlags, Serializer},
     Plan, SubsetTable,
@@ -9,11 +9,13 @@ use crate::{
 use fnv::FnvHashMap;
 use write_fonts::{
     read::tables::layout::{
-        ChainedSequenceContext, ChainedSequenceContextFormat1, ChainedSequenceRule,
-        ChainedSequenceRuleSet, CoverageTable, SequenceContext, SequenceContextFormat1,
-        SequenceLookupRecord, SequenceRule, SequenceRuleSet,
+        ChainedClassSequenceRule, ChainedClassSequenceRuleSet, ChainedSequenceContext,
+        ChainedSequenceContextFormat1, ChainedSequenceContextFormat2, ChainedSequenceRule,
+        ChainedSequenceRuleSet, ClassSequenceRule, ClassSequenceRuleSet, CoverageTable,
+        SequenceContext, SequenceContextFormat1, SequenceContextFormat2, SequenceLookupRecord,
+        SequenceRule, SequenceRuleSet,
     },
-    types::{BigEndian, GlyphId, GlyphId16, Offset16},
+    types::{BigEndian, FixedSize, GlyphId, GlyphId16, Offset16},
 };
 
 impl<'a> SubsetTable<'a> for SequenceContext<'_> {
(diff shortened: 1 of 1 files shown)


klippa/src/gsubgpos.rs
```diff
@@ -0,0 +1,1001 @@
...
+        let mut rule_set_count = 0_u16;
+
+        let mut new_cov_glyphs = Vec::with_capacity(cov_glyphs.len());
+        // seq rulesets offsets
+        let rule_sets = self.seq_rule_sets();
+        for (g, idx) in cov_glyphs.iter().zip(rule_sets_idxes.iter()) {
+            match rule_sets.subset_offset(idx as usize, s, plan, lookup_map) {
+                Ok(()) => {
+                    new_cov_glyphs.push(*g);
+                    rule_set_count += 1;
```
cmyr: is it necessary to keep track of this on its own? Can't we just use `new_cov_glyphs.len()`?

klippa/src/gsubgpos.rs
```diff
@@ -0,0 +1,1001 @@
...
+    sequence: &[BigEndian<u16>],
+    class_map: &FnvHashMap<u16, u16>,
+    s: &mut Serializer,
+) -> Result<(), SerializeErrorFlags> {
+    for c in sequence {
+        if let Some(new_c) = class_map.get(&c.get()) {
+            s.embed(*new_c)?;
+        } else {
+            return Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY);
+        }
```
cmyr: Alternatively,

```suggestion
        let new_c = classs_map.get(&c.get()).ok_or(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY)?;
        s.embed(*new_c);
```

klippa/src/gsubgpos.rs
```diff
@@ -0,0 +1,1001 @@
...
+) -> Result<u16, SerializeErrorFlags> {
+    let mut seq_lookup_count = 0_u16;
+    for lookup in lookup_records {
+        match lookup.subset(plan, s, lookup_map) {
+            Ok(()) => seq_lookup_count += 1,
+            Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY) => (),
+            Err(e) => return Err(e),
+        }
+    }
+    Ok(seq_lookup_count)
```
cmyr: and just for fun, this could also easily be an iterator:

```suggestion
    lookup_records.iter().try_fold(0u16, |acc, record| {
        match record.subset(plan, s, lookup_map) {
            Ok(_) => Ok(acc + 1),
            Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY) => Ok(acc),
            Err(e) => Err(e),
        }
    })
```

klippa/src/gsubgpos.rs
```diff
@@ -0,0 +1,1001 @@
...
+    type Output = ();
+    fn subset(
+        &self,
+        plan: &Plan,
+        s: &mut Serializer,
+        lookup_map: Self::ArgsForSubset,
+    ) -> Result<(), SerializeErrorFlags> {
+        let glyph_map = &plan.glyph_map_gsub;
+        // backtrack glyph count
+        s.embed(self.backtrack_glyph_count())?;
```
cmyr: I'm not finding that these particular comments are telling me anything that isn't fairly clear from the code. I think it's useful if you're embedding literal values (`s.embed(0)`) but when you're embedding values directly from the previous table using the getter, it's clear what it is.

klippa/src/gsubgpos.rs
```diff
@@ -0,0 +1,1001 @@
...
+    }
+}
+
+impl<'a> SubsetTable<'a> for ChainedClassSequenceRuleSet<'_> {
+    type ArgsForSubset = (
+        &'a FnvHashMap<u16, u16>,
+        &'a FnvHashMap<u16, u16>,
+        &'a FnvHashMap<u16, u16>,
+        &'a FnvHashMap<u16, u16>,
+    );
```
cmyr: I think maybe this deserves a typedef or a newtype or something, it's so easy to get confused about what order these should be passed in...

klippa/src/gsubgpos.rs
qxliu76: RuleSet could become empty even the coverage glyph will be retained,so we need to track non-empty RuleSet separately,

PR #1599, opened 2025-08-12
[klippa] subset Context/ChainedContext Format3 tables

qxliu76:
This PR depends on #1585 and #1593


cmyr:
If the PR is based on an earlier PR, can you set the `base` of the PR to that branch, here? That will make it so that we only see the relevant changes, instead of all the changes from `main`.

PR #1601, opened 2025-08-13
Write gvar table shared tuples in the expected place

belluzj:
fontc/fontations was writing the gvar components in a wrong order, it was pushing the Shared Tuples to the end of the table, while the spec mandates that they are between the table header and the Glyph Variation Data.

image

https://learn.microsoft.com/en-us/typography/opentype/spec/gvar#shared-tuples-array
> The shared tuples array follows the GlyphVariationData offsets array at the end of the 'gvar' header.

We believe this fixes the Windows 10 crashes of the Google Sans Code release: https://github.com/googlefonts/googlesans-code/issues/22

This also fixes unpacking the gvar table in OT Master.

We tested the previous release versus a TTF built with fontc using this fontations branch, and the new version didn't crash or mess up the system as the official release did.

### DIFF
--- a/write-fonts/src/tables/gvar.rs
+++ b/write-fonts/src/tables/gvar.rs
@@ -146,7 +146,7 @@ impl Gvar {
         self.glyph_variation_data_offsets.len().try_into().unwrap()
     }
 
-    fn compute_data_array_offset(&self) -> u32 {
+    fn compute_shared_tuples_offset(&self) -> u32 {
         const BASE_OFFSET: usize = MajorMinor::RAW_BYTE_LEN
             + u16::RAW_BYTE_LEN // axis count
             + u16::RAW_BYTE_LEN // shared tuples count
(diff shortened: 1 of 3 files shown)


write-fonts/src/tables/gvar.rs
```diff
@@ -1240,6 +1258,7 @@ mod tests {
 
         let bytes = sink.into_data().bytes;
         let expected_len = (n_glyphs + 1) as usize * offset_len // offsets
+                             + 8 // shared tuples TODO remove magic number
```
cmyr, line 1261: this seems fine?

cmyr (approved):
good catch, and the fix here looks reasonable.

write-fonts/src/tables/gvar.rs
belluzj: I'm OK with it if you're OK with it :)

PR #1602, opened 2025-08-13
[read-fonts] new Value type and PairPos2 fast path

dfrg:
Adds a new `Value` type that is a fully resolved version of `ValueRecord`. This can be generated from `ValueRecord` or be read directly to avoid the conversion cost.

In addition, the `PairPosFormat2` gets a new `values()` method that is essentially a fast path for reading the position adjustments for a pair of classes.

This shows a measurable improvement in shaping performance. I'll send a harfrust PR containing the other half of this and let @behdad confirm on the HB side.

Still needs more tests before merging but the code itself should be ready for review.

### DIFF
--- a/read-fonts/src/tables/gpos.rs
+++ b/read-fonts/src/tables/gpos.rs
@@ -16,7 +16,7 @@ pub use super::layout::{
     Lookup, ScriptList,
 };
 use super::layout::{ExtensionLookup, LookupFlag, Subtables};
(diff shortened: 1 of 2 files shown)


behdad:
> This shows a measurable improvement in shaping performance. I'll send a harfrust PR containing the other half of this and let @behdad confirm on the HB side.

7% in Roboto benchmark, 14% in SourceSansPro

PR #1603, opened 2025-08-13
Rename BitSet -> U32Set and make it public.

### LINKED ISSUE #1600: Make BitSet and all its variants public
I like to use them in HarfRust.

garretrieger:
This is a lower level set container that only works with u32's and unlike IntSet is not invertible. However, this means there's less overhead for most of the basic operations. This set will be useful in performance critical areas where the custom int types and invertibility aren't needed.

Fixes #1600

### DIFF
--- a/read-fonts/src/collections/int_set/bitset.rs
+++ b/read-fonts/src/collections/int_set/bitset.rs
@@ -1,4 +1,19 @@
-//! A fast & efficient ordered set for unsigned integers.
+//! A fast, efficient, sparse, & ordered unsigned integer (`u32`) bit set.
+//!
+//! There are a couple of differences with IntSet:
+//! - This set is not invertible and can only record the set of integers which are members.
+//! - This set works only with `u32` values, unlike IntSet which supports custom integer types.
+//!
+//! When dealing with only `u32`'s and invertibility is not needed then this set is slightly faster
+//! than the more generic IntSet.
+//!
+//! The bitset is implemented using fixed size pages which allows it to compactly
+//! represent sparse membership. However, the set excels when set members are typically
+//! clustered together. For example when representing glyph id or unicode codepoint values
+//! in a font.
+//!
+//! When constructing a new [`U32Set`] from an existing list of integer values the most efficient
+//! way to create the set is to initialize it from a sorted list of values via the extend() method.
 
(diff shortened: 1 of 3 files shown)


read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -1,4 +1,19 @@
-//! A fast & efficient ordered set for unsigned integers.
+//! A fast, efficient, sparse, & ordered unsigned integer (u32) bit set.
+//!
+//! There are a couple of differences with IntSet:
+//! - This set is not invertible and can only record the set of integers which are members.
+//! - This set works only with "u32" values, unlike IntSet which supports custom integer types.
```
cmyr:
```suggestion
//! - This set works only with `u32` values, unlike IntSet which supports custom integer types.
```

read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -1,4 +1,19 @@
...
+//!
+//! When dealing with only u32's and invertibility is not needed then this set is slightly faster
+//! than the more generic IntSet.
+//!
+//! The bitset is implemented using fixed size pages which allows it to compactly
+//! represent sparse membership. However, the set excels when set members are typically
+//! clustered together. For example when representing glyph id or unicode codepoint values
+//! in a font.
+//!
+//! When constructing a new [`U32Set`] from an existing lists of integer values the most efficient
```
cmyr:
```suggestion
//! When constructing a new [`U32Set`] from an existing list of integer values the most efficient
```

read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -11,29 +26,29 @@ use std::ops::RangeInclusive;
 // log_2(PAGE_BITS)
 const PAGE_BITS_LOG_2: u32 = PAGE_BITS.ilog2();
 
-/// An ordered integer (u32) set.
+/// A fast, efficient, sparse, & ordered u32 set. Not invertible.
```
cmyr: I'm not sure 'not invertible' is that meaningful, really? If it is, I would prefer to say something like, "for a higher-level API that supports inversion, use [`IntSet`]".

cmyr (approved):
Looks good!

PR #1604, opened 2025-08-14
[read-fonts] small fix for cmap10 

dfrg:
Just a codegen input update. We have an appropriate size for `glyph_id_array` so use it.

JMM


cmyr (approved):
Looks good!

PR #1606, opened 2025-08-14
[read-fonts] add some cmap helpers

dfrg:
This is a first small step towards unifying some duplicated code between skrifa and harfrust.

Adds some utility methods to `Cmap` and related types that assist in selecting appropriate subtables for symbol, unicode and variation sequences.

### DIFF
--- a/read-fonts/src/tables/cmap.rs
+++ b/read-fonts/src/tables/cmap.rs
@@ -7,6 +7,19 @@ use crate::collections::IntSet;
 use crate::{FontRef, TableProvider};
 use std::ops::Range;
 
+// See <https://docs.microsoft.com/en-us/typography/opentype/spec/cmap#windows-platform-platform-id--3>
+const WINDOWS_SYMBOL_ENCODING: u16 = 0;
+const WINDOWS_UNICODE_BMP_ENCODING: u16 = 1;
+const WINDOWS_UNICODE_FULL_ENCODING: u16 = 10;
+
+// See <https://docs.microsoft.com/en-us/typography/opentype/spec/name#platform-specific-encoding-and-language-ids-unicode-platform-platform-id--0>
+const UNICODE_1_0_ENCODING: u16 = 0;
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/cmap.rs
```diff
@@ -46,6 +59,81 @@ impl Cmap<'_> {
...
 
+    /// Returns the index, encoding record and subtable for the most
+    /// comprehensive mapping available.
+    ///
+    /// Comprehensive means that tables capable of mapping the Unicode full
+    /// repertoire are chosen over those that only support the basic
+    /// multilingual plane. The exception is that symbol mappings are
+    /// preferred above all others
+    /// (see <https://github.com/harfbuzz/harfbuzz/issues/1918>).
+    pub fn best_subtable(&self) -> Option<(u16, EncodingRecord, CmapSubtable<'a>)> {
```
rsheeter, line 70: `find(platform, encoding)` - which I think is find first on the off chance there are many - seems unopinionated and general purpose, entirely reasonable to add to read-fonts.

best seems like it might be more subjective. Is this definitely the only reasonable definition of best or is it an opinionated best?

dfrg: I'd say it's best in the sense that the result will satisfy 100% of users who want to map a Unicode codepoint to a glyph id.

However, this is a really good question that has a broader scope which is: how do we move a good chunk of skrifa functionality into read-fonts so that we can support harfrust without being opinionated? I'll hold off on merging this and raise the issue for discussion on Monday.

rsheeter: > it's best in the sense that the result will satisfy 100% of users who want to map a Unicode codepoint to a glyph id

That seems to imply there is no known alternate definition of best and that on that basis one could claim it to be unopinionated, this is just the correct way to do it?

behdad: I believe there's no good reason to prefer some other approach.

PR #1607, opened 2025-08-14
Bump MSRV to 1.80

cmyr:
JMM


dfrg (approved):
Please yes 🙇‍♂️

nicoburns:
Not sure if fontations uses floats much (/at all), but a note that a lot of the Linebender crates (including `color` which serves a similarly fundemental role in the ecosystem) are now using an MSRV of `1.82` so that they can use floating point operations in `const fn`. And I think it would be entirely reasonably for fontations to follow that if desired.

PR #1608, opened 2025-08-14
[klippa] subset SingleSubst table

qxliu76:


klippa/src/gsub/single_subst.rs
```diff
@@ -0,0 +1,310 @@
...
+        let cap = cov_glyphs.len() as usize;
+        let mut retained_glyphs = Vec::with_capacity(cap);
+        let mut sub_glyphs = Vec::with_capacity(cap);
+        for (new_g, new_sub_g) in cov_glyphs
+            .iter()
+            .map(|g| (g, GlyphId::from(g.to_u32().wrapping_add_signed(delta))))
+            .filter_map(|(g, sub_g)| {
+                let new_g = glyph_map.get(&g)?;
+                let new_sub_g = glyph_map.get(&sub_g)?;
+                Some((*new_g, *new_sub_g))
```
cmyr, line 64: no need to change, but there is a fun `zip` method on `Option`, which would let you write,


```suggestion
                glyph_map.get(&g).copied().zip(glyph_map.get(&sub_g).copied())
```

cmyr (approved):
looks good!

garretrieger (approved):

klippa/src/gsub/single_subst.rs
qxliu76: thanks!

PR #1612, opened 2025-08-14
[IntSet] speedup IntSet::contains() by caching the last page index.

garretrieger:
Copies the approach used in harfbuzz. By caching the last page index we can avoid a binary search in subsequent contains() calls when the val resides on the same page as the previous call. This is common in IntSet use cases due to lookups commonly being done in sorted order and high levels of locality typical in shaping and subsetting use cases.

This does add some additional overhead to non-local lookups, but those are small and outweighed by the significant speedups on local lookups. In the benchmarks:
- for the completely unordered (0 locality) cases there's at most a 7% regression.
- for the ordered lookups there's a 25-80% speedup.

### DIFF
--- a/read-fonts/src/collections/int_set/bitset.rs
+++ b/read-fonts/src/collections/int_set/bitset.rs
@@ -18,6 +18,7 @@
 use super::bitpage::BitPage;
 use super::bitpage::RangeIter;
 use super::bitpage::PAGE_BITS;
+use core::sync::atomic::AtomicUsize;
 use std::cmp::Ordering;
 use std::hash::Hash;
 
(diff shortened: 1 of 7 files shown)


read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -614,6 +614,45 @@ impl Extend<u32> for U32Set {
     }
 }
 
+/// This helper accelerates set membership checks by caching the last page found.
+///
+/// If membership checks are in sorted order, or are typically clustered together then
+/// this will yield performance improvements over the standard U32Set::contains() call.
+pub struct U32SetContainsAccelerator<'a> {
+    set: &'a U32Set,
+    last_page_index: usize,
```
behdad: I think this might need to be atomic with relaxed access.

read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -614,6 +614,45 @@ impl Extend<u32> for U32Set {
...
+    last_major_value: u32,
```
behdad: In HB we just keep the last_page_index and lookup the major. Not sure how much overhead that has.

garretrieger: It's not too much, only about a 5% difference and by not storing the major we can support the set changing since the cached index was stored so going to follow the hb approach.

behdad:
LGTM. Thanks. Clippy unhappy though.

garretrieger:
> LGTM. Thanks. Clippy unhappy though.

Seems to be unrelated to my changes.

read-fonts/src/collections/int_set/bitset.rs
```diff
@@ -29,13 +30,27 @@ const PAGE_BITS_LOG_2: u32 = PAGE_BITS.ilog2();
...
 /// For a higher-level API that supports inversion and generic int types, use [`super::IntSet`]
-#[derive(Clone, Debug, Default)]
+#[derive(Debug, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct U32Set {
     // TODO(garretrieger): consider a "small array" type instead of Vec.
     pages: Vec<BitPage>,
     page_map: Vec<PageInfo>,
     length: u64,
+    last_page_map_index: AtomicUsize,
```
cmyr, line 43: you have custom impls of Hash/PartialEq/PartialOrd etc? because this is what clippy is complaining about; because `IntSet` is used as a key in some hashmaps in various places, the fact that it now has interior mutability could in theory cause headaches, since the key could be mutated in such a way as to violate the invariants of the collection types.

cmyr (approved):
So the clippy failure _is_ related to this PR, and I'm not sure what the best move is. Looking at https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type, it seems like there is a configuration option we can use to specify that this type is actually okay? Configs are documented at https://doc.rust-lang.org/clippy/configuration.html, so maybe try adding that file?

read-fonts/src/collections/int_set/bitset.rs
garretrieger: Ah makes sense, yeah I'll add custom impls of all of those that ignore the atomic.

garretrieger: So we actually do already have custom impls for all of those that don't factory in the atomic, so we should be good. I imagine we just need to silence the clippy checks then?

garretrieger: Updated to add the clippy allows as needed. Also flagged the interior mutable field not to be serialized.

garretrieger: Switched to configuring the U32Set type to ignore interior mutability checks and removed allow of the clippy allows.

PR #1625, opened 2025-08-20
[klippa] subset LigatureSubst table

qxliu76:


klippa/src/gsub/ligature_subst.rs
```diff
@@ -0,0 +1,215 @@
...
+            intersected_glyphs_and_indices(&coverage, glyph_set, &plan.glyph_map_gsub);
+
+        if cov_glyphs.is_empty() {
+            return Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY);
+        }
+
+        let lig_sets = self.ligature_sets();
+
+        // we need to serialize retained coverage glyphs first so it'll be packed after the LigatureSet and Ligature tables
+        // ref: <https://github.com/harfbuzz/harfbuzz/blob/0a257b0188ce8b002b51d9955713cd7136ca4769/src/OT/Layout/GSUB/LigatureSubstFormat1.hh#L155>
```
cmyr, line 47: This code doesn't use the repacker, correct? You are doing your own packing?

qxliu76: currently no call to repacker, but we'd expect that repacker provides a method that makes sure one table will be always packed after another.:)

qxliu76: this is in serializer packing, but if there's an overflow in the end we'll call repakcer

garretrieger: If repacker gets called then it will possibly modify the ordering (moving Coverage so it's no longer last) this can cause problems in specific versions of windows 7. In harfbuzz we use virtual links to instruct the repacker not to reorder Coverage. Do we have a similar concept available in the fontations repacker and/or serializer?

qxliu76: we don't have virtual links in fontation's serializer, we could add something similar to the serializer but that depends on what fontations's repacker needs as input

garretrieger: @cmyr does the fontations repacker have virtual link support?

garretrieger:
Let's merge this as is without the virtual link stuff since we aren't invoking the repacker as of yet, so this is technically correct in it's current form. Once we integrate the repacker we can come back and add virtual links as needed.

PR #1627, opened 2025-08-21
[closure] Add failing test case

cmyr:
This is reduced from a case I encountered in fontc. I'm not entirely sure what's going on, but it looks like we are in some cases failing to consider the actual input glyphs when computing closure over a contextual lookup?

This PR is just an illustrative test case, to make sure that I'm not misunderstanding the API or expectations.

@qxliu76 does this look right to you?


qxliu76:
I think this is expected, because in `fn compute_closure()` your lookup input is `IntSet::all()`, so 'D', 'E', 'F' are added from `lookup index =0`

qxliu76:
> Isn't the `IntSet::all()` referring to which _features_ to use to collect the initial lookups? It shouldn't include lookup `0`, which isn't directly included in any feature?

No, you need to use https://github.com/googlefonts/fontations/blob/3995f2aa04f3776d56a8e7e6597555ce9e687acb/read-fonts/src/tables/gsub/closure.rs#L235 to get lookups from all features(IntSet::all())

qxliu76:
The input of `glyph_closure()` is lookup indices not feature indices.

cmyr:
sorry I'm still not following.

`compute_closure`, which is used in the test cases, gets the lookups here, and it should not include lookup 0?

https://github.com/googlefonts/fontations/blob/1a6bcf142b358238e237b7d217409e194db9acd9/read-fonts/src/tables/gsub/closure.rs#L1054

qxliu76:
ah sorry, you're right. Lookup 0 is not included here.
Based on debugging result, 'd', 'e','f' is passed into lookup 0 for glyph closure, I'll look into this further.

qxliu76:
ah it seems there's a bug in glyph closure code, I'll have a PR out soon. Thanks a lot for spotting it!

qxliu76:
So the fix is in this PR: https://github.com/googlefonts/fontations/pull/1628, I'm closing this one. Thank you!

PR #1630, opened 2025-08-22
Add `Tag::into_u32` method

nicoburns:
Motivation is exactly the same as the `from_u32` method: it allows interop with systems that represent tag in this format (HarfBuzz, CoreText, etc).

### DIFF
--- a/font-types/src/tag.rs
+++ b/font-types/src/tag.rs
@@ -75,6 +75,14 @@ impl Tag {
         Self::from_be_bytes(src.to_be_bytes())
     }
 
+    /// Convert a `Tag` into a big-endian `u32`
+    ///
+    /// This is provided as a convenience method for interop with code that
+    /// stores tags as big-endian u32s.
+    pub const fn into_u32(self) -> u32 {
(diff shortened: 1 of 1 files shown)


font-types/src/tag.rs
```diff
@@ -75,6 +75,14 @@ impl Tag {
         Self::from_be_bytes(src.to_be_bytes())
     }
 
+    /// Convert a `Tag` into a big-endian `u32`
```
dfrg, line 78:
```suggestion
    /// Convert a `Tag` into a big-endian `u32`.
```

font-types/src/tag.rs
```diff
@@ -75,6 +75,14 @@ impl Tag {
...
+    ///
+    /// This is provided as a convenience method for interop with code that
+    /// stores tags as big-endian u32s.
+    pub const fn into_u32(self) -> u32 {
```
dfrg, line 82: I know a conversion consuming `self` _should_ have an `into_` prefix but the vast majority of the methods in this crate use `to_` so I'd prefer `to_u32` here for consistency.

PR #1632, opened 2025-08-24
Add CPAL interface in skrifa

### LINKED ISSUE #1382: Add nice CPAL interface to skrifa
see #1379

valadaptive:
Involves some read-fonts changes too: when the `NameId` type was added, `palette_entry_labels_array` was updated to use it but `palette_labels_array` was not. I've changed the latter to use `NameId` as well. I've also adjusted some doc comments.

Resolves #1382. This should be a lot nicer than roughing it with read-fonts.

This adds a `color_palettes` method to `MetadataProvider` which returns the new high-level `ColorPaletteCollection`. Each palette can then be individually fetched from it, wrapped in the `ColorPalette` type.

I'm assuming a usage pattern where the API consumer fetches one palette and then fetches a bunch of colors from it (probably via painter callbacks), and I've optimized accordingly for that use case.

Some questions I have about the API design:
- Should `ColorPalette::palette_type` return `cpal::PaletteType::empty()` if the palette type array is missing entirely? Currently it returns an `Option<PaletteType>`, and returns `None` in that case.
- Should `ColorPalette::label` and `ColorPaletteCollection::color_label` return `NameId(0xFFFF)` or `None` if the name ID is `0xFFFF`? I think the latter makes more sense so that's what I went with.
- Should Skrifa define its own color type or just use `ColorRecord`? I can't think of any fancy methods to add to such a type.
- Should `ColorPaletteCollection` have a method for returning a color directly based on a palette index *and* color index? I prefer the approach of "get palette, then get color" to avoid redundant bounds checks and ensure the indices don't get mixed up, but it might be nice to provide a convenience method.

### DIFF
--- a/skrifa/src/color/mod.rs
+++ b/skrifa/src/color/mod.rs
@@ -42,7 +42,11 @@ mod traversal;
 #[cfg(test)]
 mod traversal_tests;
 
-use raw::{tables::colr, FontRef};
+use raw::{
+    tables::{colr, cpal},
+    types::{BigEndian, NameId},
+    FontRef,
+};
 #[cfg(test)]
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
     }
 }
 
+/// A single color palette.
+pub struct ColorPalette<'a> {
+    cpal: cpal::Cpal<'a>,
+    /// Preparsed subarray of color records for just this palette.
+    sub_array: &'a [cpal::ColorRecord],
```
dfrg: I agree that we don't really need to wrap `ColorRecord`. Still, the name is a bit ugly and we shouldn't force users to dig into `read-fonts` for this type so let's pull it into this module with something like:
```rust
#[doc(inline)]
pub use cpal::ColorRecord as Color;
```

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+    /// This palette's index in the CPAL table.
+    index: u16,
+}
+
+impl<'a> ColorPalette<'a> {
+    /// Returns the color at this palette index.
+    pub fn get(&self, color_index: u16) -> Option<cpal::ColorRecord> {
+        self.sub_array.get(usize::from(color_index)).copied()
+    }
```
dfrg: I'm a bit conflicted on this one. If we're not wrapping `ColorRecord` then I think we should just do
```rust
/// Returns the set of colors for this palette.
pub fn colors(&self) -> &'a [Color]
```

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+    pub fn palette_type(&self) -> Option<cpal::PaletteType> {
+        self.cpal
+            .palette_types_array()?
+            .ok()?
+            .get(usize::from(self.index))
+            .map(|p| p.get())
+    }
+
+    /// Returns this palette's label/name, if present.
+    pub fn label(&self) -> Option<NameId> {
```
dfrg:
```suggestion
    pub fn label(&self) -> Option<StringId> {
```
Skrifa generally refers to "localized strings" rather than "names" because not everything in the name table is a name so the API uses `StringId` 🤷

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+    }
+
+    /// Returns this palette's index in the CPAL table.
+    pub fn index(&self) -> u16 {
+        self.index
+    }
+}
+
+/// Collection of color palettes for color glyphs.
+pub struct ColorPaletteCollection<'a> {
```
dfrg:
```suggestion
pub struct ColorPalettes<'a> {
```

I've avoided the `Collection` suffix in recent additions (`GlyphNames` and `BitmapStrikes`) so let's do the same here.

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+        Self {
+            cpal: font.cpal().ok(),
+        }
+    }
+
+    /// Returns the total number of palettes in this collection (0 if this collection's font has no
+    /// CPAL table).
+    pub fn num_palettes(&self) -> u16 {
+        self.cpal.as_ref().map_or(0, |cpal| cpal.num_palettes())
+    }
```
dfrg: We've been slightly inconsistent here. We have `BitmapStrikes::len()` but `GlyphNames::num_glyphs()`. I think the pattern is that we only use `num_x` over `len` when `x` is `glyphs` so let's change this one to `len`.

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+
+    /// Returns the number of colors per palette (all palettes have the same number of colors).
+    pub fn num_colors(&self) -> u16 {
+        self.cpal
+            .as_ref()
+            .map_or(0, |cpal| cpal.num_palette_entries())
+    }
```
dfrg: Adding `ColorPalette::colors()` makes this one unnecessary.

skrifa/src/color/mod.rs
```diff
@@ -449,6 +453,108 @@ impl<'a> ColorGlyphCollection<'a> {
...
+        Some(ColorPalette {
+            cpal,
+            sub_array,
+            index,
+        })
+    }
+
+    /// Returns the label/name for a given color, if present (labels are per-color, but shared
+    /// across all palettes).
+    pub fn color_label(&self, color_index: u16) -> Option<NameId> {
```
dfrg:
```suggestion
    pub fn color_label(&self, color_index: u16) -> Option<StringId> {
```

skrifa/src/color/mod.rs
valadaptive: I do like that it lets you know how many colors are in each palette without having to actually fetch some arbitrary palette, and that it informs you that all palettes have the same number of colors, but I agree it isn't necessary

valadaptive:
I think I've addressed all the review feedback.

skrifa/src/color/mod.rs
```diff
@@ -457,15 +461,15 @@ impl<'a> ColorGlyphCollection<'a> {
...
+    sub_array: &'a [Color],
     /// This palette's index in the CPAL table.
     index: u16,
 }
 
 impl<'a> ColorPalette<'a> {
     /// Returns the color at this palette index.
-    pub fn get(&self, color_index: u16) -> Option<cpal::ColorRecord> {
-        self.sub_array.get(usize::from(color_index)).copied()
+    pub fn colors(&self) -> &[Color] {
```
dfrg, line 471: Looks like the doc comment for this needs an update now.

dfrg (commented):
One last nit about a doc comment and some clippy lints to fix in the tests and this one should be good a go. 

A much needed improvement, thanks!

PR #1634, opened 2025-08-26
[read-fonts] Avoid timeout when computing array len

dfrg:
When malicious or corrupted font data contains a variable sized array with a large count and a 0 length element, we will spin on parsing that element until we exhaust the count. This checks for 0 length elements and bails early returning the size up to that point.

Fixes https://github.com/harfbuzz/harfrust/issues/203


behdad:
Lgtm.

PR #1635, opened 2025-08-26
[U32Set] Implement intersects()

behdad:


garretrieger (approved):
Fixed a bug in the implementation, added tests, and updated to utilize this new method from IntSet::intersects_set(). lgtm now.

PR #1637, opened 2025-08-27
[read-fonts/aat] Make StateMachine header public

behdad:
Needed in HarfRust.

### DIFF
--- a/read-fonts/src/tables/aat.rs
+++ b/read-fonts/src/tables/aat.rs
@@ -321,7 +321,6 @@ where
 ///
 /// The input to the state machine consists of the current state
 /// and a glyph class. The output is an [entry](StateEntry) containing
-
 /// the next state and a payload that is dependent on the type of
 /// layout action being performed.
 ///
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/kern.rs
```diff
@@ -58,6 +58,10 @@ impl<'a> Iterator for Subtables<'a> {
     type Item = Result<Subtable<'a>, ReadError>;
 
     fn next(&mut self) -> Option<Self::Item> {
+        if self.n_tables == 0 || self.data.is_empty() {
+            return None;
+        }
+        self.n_tables -= 1;
```
dfrg, line 64: This change looks good but the condition below should now check for `self.n_tables == 0` to correctly catch the last table.

read-fonts/src/tables/aat.rs
```diff
@@ -321,15 +321,15 @@ where
 ///
 /// The input to the state machine consists of the current state
 /// and a glyph class. The output is an [entry](StateEntry) containing
-
+///
```
dfrg: Not sure why this line was added? Seems like it would break the doc comment.

PR #1647, opened 2025-08-29
Don't recompute cmap12 limits

taj-p:
It looks like we're recomputing the `cmap12_limits` when building a charmap from a `MappingIndex` despite it already being calculated within `MappingSelection`:

https://github.com/taj-p/fontations/blob/07df8391a0b33f2402cf8ab78f5bf01c90c3a647/skrifa/src/charmap.rs#L398

This PR uses the precomputed `cmap12_limits` instead of recomputing its value.


dfrg (approved):
Looks good, thanks!

PR #1650, opened 2025-09-05
Bump kurbo to 0.12.0

anthrotype:
https://github.com/linebender/kurbo/releases/tag/v0.12.0

fontc needs this to make those -77 diffs in fontc_crater go away:

> Improved cubic to quadratic conversion handling for degenerate cubic curves with 3-4 consecutive equal control points.

Since both write-fonts and fontc use kurbo, we need to bump it in here first, then bump both write-fonts and kurbo in fontc...


anthrotype:
@dfrg said in chat:

> Yes, let’s go ahead and bump it

PR #1658, opened 2025-09-13
[read-fonts] Add per-table feature flags

nicoburns:
This dramatically reduces compile times for `read-fonts` in cases where only a few tables are needed. Examples of cases where this applies include `fontdb` and `fontique` which only need to read high-level metadata and don't need to access details like glyph outlines, shaping, or hinting data.

Notes:
- This was also surprisingly easy to do (the tables' implementation were *mostly* already independent of each other.
- It would be relatively easy to further reduce the dependencies between modules: many of the dependencies are "trivial" dependencies that only need one or two types from the module they depend on. We could pull these types out into a shared module to remove the dependency
- I have added a CI job using "cargo hack" that runs `cargo check` with each feature flag to ensure that they all compile independently of each other.


behdad:
Any reason not to include `gdef` in `layout`?

nicoburns:
> Any reason not to include `gdef` in `layout`?

It's perhaps not obvious from the names, but "layout" is actually support code tables that deal with layout rather than the group of tables that deal with layout (the feature names match file/module names the `read-fonts/tables` directory).

Perhaps we should make that clearer by suffixing the names with `_support` or `_base` or similar. And consider adding actual meta-features for table groups.

behdad:
> It's perhaps not obvious from the names, but "layout" is actually support code tables that deal with layout rather than the group of tables that deal with layout (the feature names match file/module names the `read-fonts/tables` directory).

I suspected that. Thanks for clarifying. I think it's fine.

dfrg:
I feel like this isn’t going to be much of a win in practice. Doing anything useful with a font is going to require at least one of skrifa or harfrust, both of which need all of the expensive-to-compile tables.

nicoburns:
With `traversal`:

Screenshot 2025-09-19 at 13 47 01

Without `traversal`:

Screenshot 2025-09-19 at 13 47 11

rsheeter:
> Examples of cases where this applies include fontdb and fontique which only need to read high-level metadata and don't need to access details like glyph outlines, shaping, or hinting data.

I'm curious, are there users that _only_ need these? - the consumers that come to mind for me would also typically need at least one of shaping, subsetting, and access to outlines.

PR #1660, opened 2025-09-15
Move `Setting`, `Weight`, `Stretch`, and `Style` types to `font-types`

nicoburns:
Once `font-types` hits 1.0, these will be super-useful as interop types for the Rust crates ecosystem.

We may also wish to consider some renames:

- `Weight` -> `FontWeight`
- `Style` -> `FontStyle`
- `Stretch` -> `FontStretch` (or possibly `FontWidth` - matching CSS)

### DIFF
--- a/font-types/src/attribute.rs
+++ b/font-types/src/attribute.rs
@@ -1,0 +1,166 @@
+//! Primary attributes typically used for font classification and selection.
+
+/// Visual width of a font-- a relative change from the normal aspect
+/// ratio, typically in the range 0.5 to 2.0.
+///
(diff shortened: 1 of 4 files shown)


cmyr:
@nicoburns do you have a concrete use case/motivation for these changes? In general we want to keep changes in font-types minimal (specifically because it uses some `unsafe`, which means we potentially have to go through security review when making changes) and we'd be wary of scope creep (like wanting more CSS-style attributes.)

Put another way, would moving this API to font-types be useful to you, even if we were to never significantly add to the existing API?

PR #1661, opened 2025-09-15
WIP: avoid function calling to compute offsets

rsheeter:


nicoburns:
Unfortunately I'm also not seeing any improvement from this

PR #1663, opened 2025-09-17
Add with overflow in ColrInstance::var_deltas

dfrg:
Ref https://oss-fuzz.com/testcase-detail/5180237819478016 and https://g-issues.oss-fuzz.com/issues/439498857

First commit contains failing test.

### DIFF
--- a/skrifa/src/color/instance.rs
+++ b/skrifa/src/color/instance.rs
@@ -595,3 +595,21 @@ pub fn resolve_paint<'a>(
         },
     })
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use raw::{FontRef, TableProvider};
+
+    /// OSS Fuzz caught add with overflow when computing delta indices.
+    /// See <https://oss-fuzz.com/testcase-detail/5180237819478016>
+    /// and <https://g-issues.oss-fuzz.com/issues/439498857>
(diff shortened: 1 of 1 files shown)


drott (approved):
LGTM, thanks for the fix.

skrifa/src/color/instance.rs
```diff
@@ -59,9 +59,11 @@ impl<'a> ColrInstance<'a> {
         {
             return deltas;
         }
+        // Avoid overflow if var_index_base + N > u32::MAX
+        let actual_count = ((u32::MAX - var_index_base) as usize).min(N);
         let var_store = self.var_store.as_ref().unwrap();
         if let Some(index_map) = self.index_map.as_ref() {
-            for (i, delta) in deltas.iter_mut().enumerate() {
+            for (i, delta) in deltas.iter_mut().enumerate().take(actual_count) {
                 let var_index = var_index_base + i as u32;
```
cmyr, line 67: other option would be to make this checked_add, with an `if let` or something? either works.

dfrg: yeah, I'm trying to be more careful about performance in hot code (which this is). I was leaning toward just `wrapping_add` but `min` + `take` seemed like a decent compromise.

PR #1666, opened 2025-09-19
[skrifa] Don't enable `traversal` feature by default

nicoburns:
Improves compile times of `read-fonts` by around 25% (8s -> 6s on my machine) for cases where the `traversal` feature is not required.


dfrg (approved):
I believe keeping this enabled by default was an oversight. LGTM

PR #1679, opened 2025-10-21
[klippa-repacker] graph/vertex construction + sort/update shortest_distance and priority_queue

qxliu76:
Simple overflow resolution including serialization + more vertex tests will be in another PR.

### DIFF
--- a/klippa/src/graph.rs
+++ b/klippa/src/graph.rs
@@ -1,0 +1,552 @@
+//! Define a graph struct that represents a serialized table
+//! Implement methods to modify and reorder the graph
+
+use crate::{
+    priority_queue::PriorityQueue,
+    serialize::{Link, LinkWidth, ObjIdx, Object, Serializer},
+};
+use fnv::FnvHashMap;
+use write_fonts::read::collections::IntSet;
+
+#[derive(Debug, Copy, Clone, PartialEq, Default)]
+pub(crate) struct RepackErrorFlags(u16);
+
+impl RepackErrorFlags {
+    pub(crate) const ERROR_NONE: Self = Self(0x0000);
+    pub(crate) const GRAPH_ERROR_ORPHANED_NODES: Self = Self(0x0001);
+    pub(crate) const GRAPH_ERROR_INVALID_OBJ_INDEX: Self = Self(0x0002);
+    pub(crate) const GRAPH_ERROR_INVALID_LINK_POSITION: Self = Self(0x0004);
+    pub(crate) const GRAPH_ERROR_CYCLE_DETECTED: Self = Self(0x0008);
+    #[allow(dead_code)]
+    pub(crate) const GRAPH_ERROR_INVALID_ROOT: Self = Self(0x0010);
+    #[allow(dead_code)]
+    pub(crate) const REPACK_ERROR_SPLIT_SUBTABLE: Self = Self(0x0020);
+    #[allow(dead_code)]
+    pub(crate) const REPACK_ERROR_EXT_PROMOTION: Self = Self(0x0040);
+    #[allow(dead_code)]
+    pub(crate) const REPACK_ERROR_NO_RESOLUTION: Self = Self(0x0080);
+}
+
+impl std::ops::BitOrAssign for RepackErrorFlags {
+    /// Adds the set of flags.
+    #[inline]
+    fn bitor_assign(&mut self, other: Self) {
+        self.0 |= other.0;
+    }
+}
+
+impl std::ops::Not for RepackErrorFlags {
+    type Output = bool;
+    #[inline]
+    fn not(self) -> bool {
+        self == RepackErrorFlags::ERROR_NONE
+    }
+}
+
+#[derive(Default, Debug)]
+struct Vertex {
+    head: usize,
+    tail: usize,
+    // real_links: link position-> Link mapping
+    // real links are associated with actual offsets
+    real_links: FnvHashMap<u32, Link>,
+    // virtual links not associated with actual offsets,
+    // they exist merely to enforce an ordering constraint.
+    virtual_links: Vec<Link>,
+
+    distance: u64,
+    space: u32,
+    priority: u8,
+    //start: usize,
+    //end: usize,
+    incoming_edges: usize,
+    has_incoming_virtual_edges: bool,
+    parents: FnvHashMap<ObjIdx, u16>,
+}
+
+impl Vertex {
+    fn from_object(obj: &Object) -> Self {
+        Self {
+            head: obj.head(),
+            tail: obj.tail(),
+            real_links: obj.real_links(),
+            virtual_links: obj.virtual_links(),
+            ..Default::default()
+        }
+    }
+
+    fn table_size(&self) -> usize {
+        self.tail - self.head
+    }
+
+    fn reset_parents(&mut self) {
+        self.incoming_edges = 0;
+        self.has_incoming_virtual_edges = false;
+        self.parents.clear();
+    }
+
+    fn add_parent(&mut self, parent_idx: ObjIdx, is_virtual: bool) {
+        self.has_incoming_virtual_edges |= is_virtual;
+        self.parents
+            .entry(parent_idx)
+            .and_modify(|c| *c += 1)
+            .or_insert(1);
+
+        self.incoming_edges += 1;
+    }
+
+    fn link_positions_valid(&self, num_objs: usize) -> bool {
+        let table_size = self.table_size();
+        let mut assigned_bytes = IntSet::empty();
+        for (pos, l) in &self.real_links {
+            if l.obj_idx() >= num_objs {
+                return false;
+            }
+
+            let width = l.link_width() as u8;
+            if width < 2 {
+                return false;
+            }
+
+            let start = *pos;
+            let end = start + width as u32 - 1;
+            if end as usize >= table_size {
+                return false;
+            }
+
+            if assigned_bytes.intersects_range(start..=end) {
+                return false;
+            }
+            assigned_bytes.insert_range(start..=end);
+        }
+        true
+    }
+
+    #[allow(dead_code)]
+    fn has_max_priority(&self) -> bool {
+        self.priority >= 3
+    }
+
+    fn modified_distance(&self, order: u32) -> i64 {
+        let prev_dist = self.distance as i64;
+        let table_size = (self.tail - self.head) as i64;
+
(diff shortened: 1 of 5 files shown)


klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+    serialize::{Link, LinkWidth, ObjIdx, Object, Serializer},
+};
+use fnv::FnvHashMap;
+use write_fonts::read::collections::IntSet;
+
+#[derive(Default, Debug)]
+struct Obj {
+    head: usize,
+    tail: usize,
+    // real links are associated with actual offsets
```
garretrieger, line 51: The key in this map is the position of the link in the object data right? Could you add a comment stating that.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+    real_links: FnvHashMap<u32, Link>,
+    // virtual links not associated with actual offsets,
+    // they exist merely to enforce an ordering constraint.
+    virtual_links: FnvHashMap<u32, Link>,
+}
+
+#[derive(Default, Debug)]
+struct Vertex {
+    obj: Obj,
```
garretrieger: In hb repacker we have both vertex and object since the object structure gets reused from the serializer. Since we're not doing that here it should be fine to put all of the Obj fields into Vertex and not have the Obj structure.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+    fn incoming_edges(&self) -> usize {
+        self.incoming_edges
+    }
+}
+
+//TODO: add support for space assignment and splitting
+#[derive(Default, Debug)]
+pub(crate) struct Graph {
+    vertices: Vec<Vertex>,
+    ordering: Vec<usize>,
```
garretrieger, line 156: Can you add a comment explaining what the ordering array is storing.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+    ordering_scratch: Vec<usize>,
+    num_roots_for_space: Vec<usize>,
+    #[allow(dead_code)]
+    data: Vec<u8>,
+
+    // graph state flags
+    parents_invalid: bool,
+    distance_invalid: bool,
+    positions_invalid: bool,
+    successful: bool,
```
garretrieger: Might be helpful to turn this into an error enum so we can track the different types of errors, something like
```
state: Option<RepackerError>
```

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+        let mut i = 0;
+        this.ordering.resize_with(count, || {
+            i += 1;
+            count - i
+        });
+        Some(this)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn in_error(&self) -> bool {
```
garretrieger, line 206: With an error enum this could then return a Result<(), RepackerError>

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+
+        self.update_parents();
+
+        queue.push((self.root().modified_distance(0), self.root_idx()));
+        let mut order = 1_u32;
+        let mut pos = 0;
+
+        let new_ordering = &mut self.ordering_scratch;
+        while let Some((_, next_id)) = queue.pop() {
+            if pos >= v_count {
```
garretrieger, line 237: This should set the repacker into an Error state.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+        });
+        Some(this)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn in_error(&self) -> bool {
+        !self.successful
+    }
+
+    pub(crate) fn sort_shortest_distance(&mut self) {
```
garretrieger: I would have this return a Result. In general when porting from hb if there are methods that can set the repacker into an error state we should have them also return a result to force the caller to check for errors.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+                    queue.push((child_v.modified_distance(order), child_idx));
+                    order += 1;
+                }
+            }
+        }
+
+        std::mem::swap(&mut self.ordering, new_ordering);
+        if pos != v_count {
+            self.set_err();
+            //TODO: add print_orphaned_nodes()
```
garretrieger, line 264: Since in rust we'll have better error handling support, we can just return an error enum that indicates there's an orphaned node and then the high level user can called handling calling print orphaned nodes.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+            return;
+        }
+
+        for v in self.vertices.iter_mut() {
+            v.reset_parents();
+        }
+
+        let count = self.vertices.len();
+        let mut real_links_idxes = Vec::with_capacity(count);
+        let mut virtual_links_idxes = Vec::with_capacity(count);
```
garretrieger, line 284: Is this temp storage needed because of borrow checker? You could get away with just having one, process the real links first, clear, then process the virtual links.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+            self.set_err();
+            //TODO: add print_orphaned_nodes()
+        }
+    }
+
+    fn root(&self) -> &Vertex {
+        &self.vertices[self.root_idx()]
+    }
+
+    pub(crate) fn update_parents(&mut self) {
```
garretrieger: Return Result<...>

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+        }
+        let root_idx = self.root_idx();
+        self.vertices[root_idx].distance = 0;
+
+        let count = self.vertices.len();
+        let mut queue = PriorityQueue::with_capacity(count);
+        queue.push((0_u64, root_idx));
+
+        let mut visited = vec![false; count];
+        let mut distance_map = FnvHashMap::default();
```
garretrieger: Since key is an object index and we expect one value per vertex, you can use a presized vector as a map which will be a little bit faster.

klippa/src/graph.rs
```diff
@@ -0,0 +1,515 @@
...
+                };
+                v.add_parent(idx, true);
+            }
+        }
+    }
+
+    // Finds the distance to each object in the graph from the root node
+    // Uses Dijkstra's algorithm to find all of the shortest distances.
+    // ref: <https://github.com/harfbuzz/harfbuzz/blob/3f70b6987830bba1e4922cad03028cdd9d78b3a1/src/graph/graph.hh#L1512C13-L1512C72>
+    pub(crate) fn update_distances(&mut self) {
```
garretrieger: Return Result<...>

klippa/src/repack.rs
```diff
@@ -0,0 +1,11 @@
+//! Implement repacking algorithm to resolve offset overflows
+//! ref:<https://github.com/harfbuzz/harfbuzz/blob/9cfb0e6786ecceabaec7a26fd74b1ddb1209f74d/src/hb-repacker.hh#L454>
+
+use crate::{graph::Graph, serialize::Serializer};
+//TODO: add more functionality, serialize output etc.
+pub(crate) fn resolve_overflows(s: &Serializer) {
```
garretrieger: Return Result<...>

klippa/src/graph.rs
qxliu76: seems we still need two temp vectors because we have to save all link obj_idxes first, then we're done with immutable borrows.

garretrieger: sounds good.

garretrieger (approved):

klippa/src/graph.rs
```diff
@@ -0,0 +1,559 @@
...
+
+use crate::{
+    priority_queue::PriorityQueue,
+    serialize::{Link, LinkWidth, ObjIdx, Object, Serializer},
+};
+use fnv::FnvHashMap;
+use write_fonts::read::collections::IntSet;
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub(crate) struct RepackErrorFlags(u16);
```
cmyr, line 12: would it make sense to just use [bitflags](https://docs.rs/bitflags/latest/bitflags/) for this? Are we trying to avoid dependencies?

At a higher level I assume you're just trying to port the CPP as literally as possible, and are not concerned with being idiomatic?

klippa/src/graph.rs
```diff
@@ -0,0 +1,559 @@
...
+    #[allow(dead_code)]
+    pub(crate) const REPACK_ERROR_EXT_PROMOTION: Self = Self(0x0040);
+    #[allow(dead_code)]
+    pub(crate) const REPACK_ERROR_NO_RESOLUTION: Self = Self(0x0080);
+}
+
+impl Default for RepackErrorFlags {
+    #[inline]
+    fn default() -> Self {
+        Self::ERROR_NONE
```
cmyr: this can be derived.

klippa/src/graph.rs
```diff
@@ -0,0 +1,559 @@
...
+
+        let distance = match self.priority {
+            0 => prev_dist,
+            1 => prev_dist - table_size / 2,
+            2 => prev_dist - table_size,
+            _ => 0,
+        }
+        .clamp(0, 0x7FFFFFFFFFF_i64);
+
+        (distance << 18) | (0x003FFFF & order as i64)
```
cmyr, line 142: what are these magic numbers? can we define `consts` and give them names?

klippa/src/serialize.rs
```diff
@@ -364,10 +399,12 @@ impl Serializer {
...
         self.packed_map.clear();
 
-        while self.current.is_some() {
-            let o = self.current.unwrap();
-            self.current = self.object_pool.next_idx(self.current.unwrap());
-            self.object_pool.release(o);
+        loop {
+            let Some(current) = self.current else {
+                break;
+            };
```
cmyr: could this be,

```rust
while let Some(current)  = self.current.take() {
..
}
```

klippa/src/graph.rs
qxliu76: For this errorflag, we don't need that many features bit flags provides so I implemented it myself. We do want to be idiomatic as much as possible when porting from C++, so any comments are welcome:)

klippa/src/graph.rs
qxliu76: It's bit manipulation, I'd like to keep it as is.

PR #1684, opened 2025-10-30
Handle NO_VARIATION_INDEX case when computing deltas

simoncozens:
Fixes #1720. See there for explanation.


simoncozens:
No pressure from me, I worked around it in my own code.

PR #1688, opened 2025-11-05
[read-fonts+skrifa] CFF: clear stem hints for seac components

### LINKED ISSUE #1680: Helvetica rendered incorrectly (missing dots)
In Vivaldi and Chromium, as of around 7 months ago, when I believe they switched to fontations, they stopped rendering characters with dots in them, including `ij.:;` I'm using Linux Mint 21.2, based on Ubuntu 22.04. I have a Helvetica font installed on my computer, when most people probably don't. It is the same as this file: https://github.com/stef-gelsomino/technotes/blob/a63f33120a902e19b4f1a371d8738d40b6c631a8/helvetica.otf
Older versions of Chromium used to render the font correctly, and Firefox on my computer still renders it correctly (when Firefox devtools says it is using Helvetica).

Many sites use Helvetica if it's installed, like for example GitHub. Here's a screenshot of the fontations README:

Image

dfrg:
When processing the implied seac operator, each component should have its own set of stem hints. This patch adds a new `clear_hints()` method to `CommandSink` which is called at the appropriate times. The hinter in skrifa then clears the relevant state.

With #1686 and #1687: fixes #1680

### DIFF
--- a/read-fonts/src/tables/postscript/charstring.rs
+++ b/read-fonts/src/tables/postscript/charstring.rs
@@ -497,7 +497,7 @@ where
         // FreeType calls cf2_interpT2CharString for each component
         // which uses a fresh set of stem hints. Since our hinter is in
         // a separate crate, we signal this through the sink. Also
-        // reset our own stem count so we need to correct number of
+        // reset our own stem count so we read the correct number of
         // bytes for each hint mask instruction.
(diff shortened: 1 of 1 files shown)


cmyr (approved):
looks good!

PR #1697, opened 2025-11-11
Fix documentation build

drott:
Replace `doc_auto_cfg` with `doc_cfg`, see
https://github.com/rust-lang/rust/pull/138907

Fixes issues with docs.rs builds of Fontations crates.


DJMcNab:
I think this patch release would still be helpful.

PR #1698, opened 2025-11-14
WIP: Reproducing tests for F26Dot6 multiply and divide bug

Hoolean:
Hello!

We spotted some funky behaviour when using `HarfBuzzScaler`, and think we have narrowed it down to `F26Dot6` multiplication and division. On the other hand, as the `FreeTypeScaler` and hinting virtual machine are having no such issues, there may be something else at play...

This PR adds failing tests for the `F26Dot6` results that seem suspect, as well as the original `HarfBuzzScaler` issue that led us to begin investigating : )

(As there's no fix yet, it's not ready to be merged as-is)

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -508,4 +508,12 @@ mod tests {
         // Just don't panic with overflow
         let _ = a.mul_div(Fixed::ONE, b);
     }
+
+    /// These identities should always hold; added as a failing test to
+    /// reproduce a bug.
+    #[test]
+    fn f26dot6_div_identity() {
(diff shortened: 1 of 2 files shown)


Hoolean:
At a glance, it looks like `fixed_mul_div` may need constants made generic to work for both `Fixed` and `F26Dot6` simultaneously, but I'm afraid I'm unable to investigate further right now

cmyr:
@dfrg what's the status of this? Do we want to fix?

dfrg:
Yes, I think we want to fix it. But first we need to find all the places that might depend on this behavior and update them. 

The fix might just be removing the mul/div operators from F26Dot6.

cmyr:
thanks! this is fixed in #1880

PR #1699, opened 2025-11-17
[klippa-repacker] extension promotion

qxliu76:


klippa/src/graph.rs
```diff
@@ -750,6 +773,30 @@ impl Graph {
...
+        &self,
+        obj_idx: ObjIdx,
+        visited: &mut IntSet<u32>,
+        max_depth: u16,
+    ) -> usize {
+        if !visited.insert(obj_idx as u32) {
+            return 0;
+        }
+
+        assert!(obj_idx < self.vertices.len());
```
wmedrano, line 786: The Index operator (`[]`) does bounds checking and panics if out of range.

https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=968505ac5d9fdb728b82d0c569fe1c69

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+    bytes: &'a mut [u8],
+}
+
+impl<'a> ExtensionSubtable<'a> {
+    const FORMAT_BYTE_LEN: usize = 2;
+    const LOOKUP_TYPE_LEN: usize = 2;
+    fn from_graph(graph: &'a mut Graph, obj_idx: ObjIdx) -> Result<Self, RepackErrorFlags> {
+        let ext = graph
+            .obj_data(obj_idx)
+            .map(move |data| Self { bytes: data })
```
wmedrano: I don't think `move` is needed here. Only `data` is used, and that is taken as a parameter and not captured from the environment.

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+            .ok_or(RepackErrorFlags::GRAPH_ERROR_INVALID_OBJ_INDEX)?;
+
+        if !ext.sanitize() {
+            return Err(RepackErrorFlags::REPACK_ERROR_EXT_PROMOTION);
+        }
+        Ok(ext)
+    }
+
+    fn sanitize(&self) -> bool {
```
wmedrano, line 48: Would better name be `is_sanitized` or is there a plan to have this do more?

klippa/src/graph.rs
qxliu76: Thanks! Fixed.

klippa/src/graph/layout.rs
qxliu76: good catch! Fixed

klippa/src/graph/layout.rs
qxliu76: No, we sanitize only once. Let me keep this name convention in subsetter.:)

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
+//! Read layout tables in a graph
+//!
```
cmyr: stray?

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+        self.bytes
+            .get_mut(0..Self::FORMAT_BYTE_LEN)
+            .unwrap()
+            .copy_from_slice(&1_u16.to_be_bytes());
+
+        self.bytes
+            .get_mut(Self::FORMAT_BYTE_LEN..Self::FORMAT_BYTE_LEN + Self::LOOKUP_TYPE_LEN)
+            .unwrap()
+            .copy_from_slice(&lookup_type.to_be_bytes());
+    }
```
cmyr: so I do think all of this code would be a lot simpler to read and write if you had some mechanism such as what we use in the `TableData` struct in write-fonts: https://github.com/googlefonts/fontations/blob/38e7270af551f36428b090cecc1a4eeee21d2fad/write-fonts/src/write.rs#L189

basically you could have this same API but it's wrapping the `&mut [u8]`, and then you would only need to write,

```rust
self.buf.write_at(0, 1_u16);
self.buf.write_at(Self::FORMAT_BYTE_LEN, lookup_type);
```

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+
+pub(crate) struct Lookup<'a> {
+    bytes: &'a mut [u8],
+}
+
+impl<'a> Lookup<'a> {
+    const LOOKUP_MIN_SIZE: usize = 6;
+    const LOOKUP_TYPE_POS: usize = 0;
+    const LOOKUP_FLAG_POS: usize = 2;
+    const NUM_SUBTABLE_BYTE_POS: usize = 4;
```
cmyr, line 66: something interesting is that all of these values are probably known by codegen, and could maybe be exposed?

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+            + if self.use_mark_filtering_set() { 2 } else { 0 };
+        self.bytes.len() >= size
+    }
+
+    pub(crate) fn num_subtables(&self) -> u16 {
+        let num_subtable_bytes = self
+            .bytes
+            .get(Self::NUM_SUBTABLE_BYTE_POS..Self::NUM_SUBTABLE_BYTE_POS + 2)
+            .unwrap();
+        u16::read(num_subtable_bytes).unwrap()
```
cmyr: ditto, we have all of this code in read-fonts that simplifies this, and could at least be copied? It would let you just do,

```rust
let num_subtable_bytes: u16 = self.bytes.read().unwrap();
```

klippa/src/graph/layout.rs
```diff
@@ -0,0 +1,197 @@
...
+        &mut self,
+        lookup_idx: ObjIdx,
+        subtable_idx: ObjIdx,
+        lookup_type: u16,
+        idx_map: &mut FnvHashMap<usize, usize>,
+    ) -> Result<usize, RepackErrorFlags> {
+        let ext_idx = if let Some(idx) = idx_map.get(&subtable_idx) {
+            let Some(subtable_v) = self.vertices.get_mut(subtable_idx) else {
+                return Err(RepackErrorFlags::GRAPH_ERROR_INVALID_OBJ_INDEX);
+            };
```
cmyr: could just be,


```suggestion
            let subtable_v = self.vertices.get_mut(subtable_idx)
                .ok_or(RepackErrorFlags::GRAPH_ERROR_INVALID_OBJ_INDEX)?;
```

klippa/src/repack.rs
```diff
@@ -64,10 +77,147 @@ pub(crate) fn resolve_graph_overflows(
...
+    impl LookupSize {
+        fn cmp(&self, other: &LookupSize) -> std::cmp::Ordering {
+            let bytes_per_subtable_a = self.subtable_count as f64 / self.subgraph_size as f64;
+            let bytes_per_subtable_b = other.subtable_count as f64 / other.subgraph_size as f64;
+            if bytes_per_subtable_b < bytes_per_subtable_a {
+                Ordering::Less
+            } else if bytes_per_subtable_b > bytes_per_subtable_a {
+                Ordering::Greater
+            } else {
+                Ordering::Equal
```
cmyr, line 111: be aware that this will return `Equal` in unexpected cases, such as if either side is a NaN.

cmyr (approved):
looks reasonable, some nits and observations inline :)

klippa/src/graph/layout.rs
qxliu76: Thanks! This is very helpful. 
I'm going to wrap the data bytes into a single struct that provides both read and write APIs, so it could be reused by other tables for table splitting.

klippa/src/graph/layout.rs
qxliu76: Thanks, it seems codegen only provides byte ranges API and would need multiple reads for a specific field. e.g. for subtable count field, `lookup_flag_byte_range() ` needs to be called :
https://github.com/googlefonts/fontations/blob/38e7270af551f36428b090cecc1a4eeee21d2fad/read-fonts/generated/generated_layout.rs#L846

klippa/src/graph.rs
```diff
@@ -201,6 +212,15 @@ impl Vertex {
             self.priority = 3;
         }
     }
+
+    pub(crate) fn add_link(&mut self, width: LinkWidth, child_idx: ObjIdx, position: u32) {
+        let link = Link::new(width, child_idx, position);
+        self.real_links.insert(position, link);
+    }
+
+    pub(crate) fn child_idxes(&self) -> Vec<ObjIdx> {
```
garretrieger, line 221: This could return an iterator and then the caller can collect to vector if desired. That way callers can avoid allocating a vector if they don't need one. Looking at the calls to this later the vector isn't needed.

klippa/src/graph.rs
```diff
@@ -750,6 +773,32 @@ impl Graph {
...
+        obj_idx: ObjIdx,
+        visited: &mut IntSet<u32>,
+        max_depth: u16,
+    ) -> Result<usize, RepackErrorFlags> {
+        if !visited.insert(obj_idx as u32) {
+            return Ok(0);
+        }
+
+        assert!(obj_idx < self.vertices.len());
+        let Some(v) = self.vertices.get(obj_idx) else {
```
garretrieger: Can use the new vertex() method.

klippa/src/graph.rs
qxliu76: I tried but ran into borrow checker error. It seems we cannot avoid vector allocations in all similar situations, which might be one source where rust is slower than C++. Error details:

    |           for subtable_idx in self
    |                               ----
    |                               |
    |  _____________________________immutable borrow occurs here
    | |
    | |             .vertex(lookup_idx)
    | |             .ok_or(RepackErrorFlags::GRAPH_ERROR_INVALID_OBJ_INDEX)?
    | |             .child_idxes()
    | |__________________________- a temporary with access to the immutable borrow is created here ...
    |           {
    |               self.make_subtable_extension(lookup_idx, subtable_idx, lookup_type, idx_map)?;
    |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
    |           }
    |           - ... and the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `impl Iterator<Item = usize> + '_`

garretrieger: Ah got it, no problem to leave this as is then.

PR #1710, opened 2026-01-05
[klippa-repacker] split ligature subst table

qxliu76:
plus change RepackerErrorFlags from bitflags to enum

### DIFF
--- a/klippa/src/graph.rs
+++ b/klippa/src/graph.rs
@@ -8,42 +8,22 @@ use crate::{
 use fnv::FnvHashMap;
 use write_fonts::{read::collections::IntSet, types::Uint24};
 
+mod coverage_graph;
 pub(crate) mod layout;
+pub(crate) mod ligature_graph;
 
-#[derive(Debug, Copy, Clone, PartialEq, Default)]
-pub(crate) struct RepackErrorFlags(u32);
-
-impl RepackErrorFlags {
-    pub(crate) const ERROR_NONE: Self = Self(0x0000);
-    pub(crate) const GRAPH_ERROR_ORPHANED_NODES: Self = Self(0x0001);
-    pub(crate) const GRAPH_ERROR_INVALID_OBJ_INDEX: Self = Self(0x0002);
-    pub(crate) const GRAPH_ERROR_INVALID_LINK_POSITION: Self = Self(0x0004);
-    pub(crate) const GRAPH_ERROR_CYCLE_DETECTED: Self = Self(0x0008);
-    #[allow(dead_code)]
-    pub(crate) const GRAPH_ERROR_INVALID_ROOT: Self = Self(0x0010);
-    pub(crate) const REPACK_ERROR_SERIALIZE: Self = Self(0x0020);
-    #[allow(dead_code)]
-    pub(crate) const REPACK_ERROR_SPLIT_SUBTABLE: Self = Self(0x0040);
-    #[allow(dead_code)]
-    pub(crate) const REPACK_ERROR_EXT_PROMOTION: Self = Self(0x0080);
-    #[allow(dead_code)]
-    pub(crate) const REPACK_ERROR_NO_RESOLUTION: Self = Self(0x0100);
-}
-
-impl std::ops::BitOrAssign for RepackErrorFlags {
-    /// Adds the set of flags.
-    #[inline]
-    fn bitor_assign(&mut self, other: Self) {
-        self.0 |= other.0;
-    }
-}
-
-impl std::ops::Not for RepackErrorFlags {
-    type Output = bool;
-    #[inline]
-    fn not(self) -> bool {
-        self == RepackErrorFlags::ERROR_NONE
-    }
+#[derive(Debug)]
+pub(crate) enum RepackError {
+    GraphErrorOrphanedNodes,
+    GraphErrorInvalidObjIndex,
+    GraphErrorInvalidLinkPosition,
+    GraphErrorCycleDetected,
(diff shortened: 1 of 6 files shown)


klippa/src/graph/coverage_graph.rs
```diff
@@ -0,0 +1,49 @@
...
+        .ok_or(RepackErrorFlags::GraphErrorInvalidObjIndex)?;
+
+    let coverage_table = CoverageTable::read(FontData::new(coverage_data))
+        .map_err(|_| RepackErrorFlags::RepackErrorReadTable)?;
+
+    Ok(coverage_table.iter().map(GlyphId::from).collect())
+}
+
+// Make a coverage table at the specified coverage vertex
+pub(crate) fn make_coverage(
```
cmyr, line 30: Would this make more sense as a method on `Graph`?

klippa/src/graph.rs
```diff
@@ -8,42 +8,24 @@ use crate::{
...
-impl std::ops::Not for RepackErrorFlags {
-    type Output = bool;
-    #[inline]
-    fn not(self) -> bool {
-        self == RepackErrorFlags::ERROR_NONE
-    }
+pub(crate) mod ligature_graph;
+
+#[derive(Default, Debug, PartialEq)]
+pub(crate) enum RepackErrorFlags {
```
cmyr: These are no longer 'flags' so maybe `RepackError` would be a better name?

klippa/src/graph.rs
```diff
@@ -8,42 +8,24 @@ use crate::{
...
+    #[default]
+    GraphErrorNone,
```
cmyr: Does this serve any purpose? it seems confusing and counter intuitive.

klippa/src/graph/layout.rs
```diff
@@ -1,46 +1,59 @@
...
 
-struct DataBytes<'a> {
+pub(super) struct DataBytes<'a> {
     bytes: &'a mut [u8],
 }
 
-impl DataBytes<'_> {
+impl<'a> DataBytes<'a> {
+    pub(super) fn from_graph(
+        graph: &'a mut Graph,
```
cmyr: we don't seem to actually mutate this at all... is this just because there is a `vertex_data_mut` fn but not `vertext_data` fn?

klippa/src/graph/layout.rs
```diff
@@ -192,4 +213,91 @@ impl Graph {
         lookup_table.set_lookup_type(ext_type);
         Ok(())
     }
+
+    // Make a Lookup table at the specified lookup vertex
```
cmyr: might as well make this a doc comment with `///` because then it will show up on hover in IDEs etc.

klippa/src/graph.rs
```diff
@@ -90,6 +72,18 @@ impl Vertex {
         }
     }
 
+    fn duplicate(other_v: &Vertex) -> Self {
```
cmyr: why not make this take `&self`? I think it's also worth at least a brief explanation of why this is different from `Clone`..

klippa/src/graph.rs
```diff
@@ -951,33 +969,41 @@ impl Graph {
...
         for idx in child_idxes {
-            self.duplicate_subgraph(idx, index_map);
+            self.duplicate_subgraph(idx, index_map)?;
         }
+        Ok(())
     }
 
-    // Creates a copy of the specified obj and returns the new obj_idx.
-    fn duplicate_obj(&mut self, obj_idx: ObjIdx) -> ObjIdx {
+    // Creates a copy of the specified vertex and returns the new vertex idx.
```
cmyr: noticing this is a pattern, but I don't see any reason for this to not also be a doc comment.

klippa/src/repack.rs
```diff
@@ -80,15 +83,109 @@ pub(crate) fn resolve_graph_overflows(
...
+        _ => false,
     }
 }
 
 fn promote_extensions_if_needed(
     graph: &mut Graph,
     lookup_list_idx: ObjIdx,
     lookups: &[ObjIdx],
-    table_tag: Tag,
+    table_tag: &Tag,
```
cmyr: this is 4 bytes, is there any advantage to using a borrow?

klippa/src/graph/ligature_graph.rs
```diff
@@ -0,0 +1,568 @@
...
+    },
+    serialize::{LinkWidth, ObjIdx},
+};
+use write_fonts::{
+    read::collections::IntSet,
+    types::{FixedSize, GlyphId, Offset16},
+};
+
+// output only contains new subtable indices
+pub(crate) fn split_ligature_subst(
```
cmyr, line 18: if this is based off of a specific implementation in harfbuzz it would be good to include a link to the original. The same goes for any other code here...

klippa/src/graph/ligature_graph.rs
```diff
@@ -0,0 +1,568 @@
...
+                    + LigatureSet::MIN_SIZE
+                    + lig_size;
+            }
+        }
+    }
+    Ok(out)
+}
+
+#[allow(clippy::too_many_arguments)]
+fn clone_range(
```
cmyr, line 113: ditto: if this is adapted from harfbuzz, can we get a link to the original?

klippa/src/graph/coverage_graph.rs
qxliu76: it's about serializing a coverage table and adding it to the graph, I'd keep it this way.

PR #1711, opened 2026-01-06
Fix fuzzer found addition overflow.

garretrieger:
Fixes: https://oss-fuzz.com/testcase-detail/6652412011282432

### DIFF
--- a/incremental-font-transfer/src/patchmap.rs
+++ b/incremental-font-transfer/src/patchmap.rs
@@ -234,7 +234,11 @@ fn intersect_format1_feature_map<const RECORD_INTERSECTION: bool>(
 
     let max_entry_index = map.max_entry_index();
     let max_glyph_map_entry_index = map.max_glyph_map_entry_index();
-    let field_width = if max_entry_index < 256 { 1u16 } else { 2u16 };
+    let entry_map_record_size = if max_entry_index < 256 {
+        2usize
+    } else {
+        4usize
+    };
 
     // We need to check up front there is enough data for all of the listed entry records, this
     // isn't checked by the read_fonts generated code. Specification requires the operation to fail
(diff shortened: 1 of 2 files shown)


incremental-font-transfer/src/patchmap.rs
```diff
@@ -249,7 +249,7 @@ fn intersect_format1_feature_map<const RECORD_INTERSECTION: bool>(
     };
     let mut record_it = feature_map.feature_records().iter().peekable();
 
-    let mut cumulative_entry_map_count = 0;
+    let mut cumulative_entry_map_count: usize = 0;
```
cmyr, line 256: is there a reasonable max value for this, per the spec? Would it make more sense to have it be an explicit `u32` or `u64`?

incremental-font-transfer/src/patchmap.rs
```diff
@@ -305,7 +309,7 @@ fn intersect_format1_feature_map<const RECORD_INTERSECTION: bool>(
         let entry_count = record.entry_map_count().get();
 
         for i in 0..entry_count {
-            let index = i as usize + cumulative_entry_map_count as usize;
+            let index = i as usize + cumulative_entry_map_count;
```
cmyr, line 316: could this not also plausibly overflow?

incremental-font-transfer/src/patchmap.rs
garretrieger: There is no explicit max, this value is derived from the sum of u16's. Ultimately this value is used to index into the byte array so usize makes the most sense.

incremental-font-transfer/src/patchmap.rs
garretrieger: Line 242 would out of bounds long before this could overflow (i is a u16 here).

garretrieger: Looks like there was a potential overflow in the function which checks total size, so fixed that as well.

PR #1709, opened 2025-12-30
[PoC] Speed up reading

behdad:
This is the result of a few days vibe-coding with codex. This moves access checks from table construction time, to each individual fields's getter method.

If there's out of bounds access, this code crashes. We can see if we can codegen `sanitize` methods ala HB to address that and reject at font table level bad data.

Accompanying HR PR: https://github.com/harfbuzz/harfrust/pull/314

Experiment details:
https://docs.google.com/document/d/1LjYFjZj8Kw8zyhqsfZg0_VgzHf2zhYUsksPCkZL7GJI/edit

PR #1715, opened 2026-01-16
[read] Add unsafe read for scalars and arrays

cmyr:
And use it in codegen. This is intended as a proof of concept to explore performance options for harfrust.

This is similar to the first couple commits of #1709 except this is more narrowly scoped (only focused on the two hot-path methods) and is in theory intended to be mergeable if we decide to go this route.


cmyr:
may revisit this later but doesn't need to sit open right now

PR #1716, opened 2026-01-19
[write-fonts] VARC table

simoncozens:
This adds a new MultiItemVariationStoreBuilder (abstracting out common functionality between it and the VariationStoreBuilder), and implements write-fonts support for the VARC table.

The `varc.rs` tests provide some examples of how to use this to build VARC tables.

### DIFF
--- a/write-fonts/src/tables/varc.rs
+++ b/write-fonts/src/tables/varc.rs
@@ -368,18 +368,6 @@ mod tests {
 
     use super::*;
 
-    // fn dump_test_ttf(varc: &Varc) {
-    //     let mut fontbuilder = crate::FontBuilder::new();
-    //     fontbuilder
-    //         .add_table(&crate::tables::maxp::Maxp {
-    //             num_glyphs: 255,
-    //             ..Default::default()
-    //         })
-    //         .unwrap();
-    //     fontbuilder.add_table(varc);
-    //     std::fs::write("debug.ttf", fontbuilder.build()).unwrap()
-    // }
-
     #[test]
     fn test_write_uint32var() {
         let mut writer = TableWriter::default();


write-fonts/src/tables/postscript.rs
```diff
@@ -0,0 +1,67 @@
+//! Postscript (CFF/CFF2) table related code.
+
+include!("../../generated/generated_postscript.rs");
+
+impl Index2 {
+    /// Construct an `Index2` from a list of byte items.
+    pub fn from_items(items: Vec<Vec<u8>>) -> Self {
+        if items.is_empty() {
+            return Index2::new(0, 1, vec![1], vec![]);
```
behdad, line 9: The OT spec has the following note:
> An empty INDEX is represented by a count field with a 0 value and no additional fields. Thus, the total size of an empty INDEX is 4 bytes.

Is this implemented?

https://learn.microsoft.com/en-us/typography/opentype/spec/cff2#common-formats-dict-and-index

write-fonts/src/tables/varc.rs
```diff
@@ -1,6 +1,601 @@
...
+    }
+}
+
+pub struct VarComponent {
+    pub reset_unspecified_axes: bool,
+    pub gid: GlyphId,
+    pub condition_index: Option<VarcVariationIndex>,
+    pub axis_values: Option<BTreeMap<u16, f32>>,
+    pub axis_values_var_index: Option<VarcVariationIndex>,
+    pub transform: DecomposedTransform,
```
behdad, line 154: It came as a surprise to me that `DecomposedTransform` has optional fields. Maybe the name should reflect that?

behdad:
This looks great to me! Thanks for working on it.

write-fonts/src/tables/postscript.rs
simoncozens: No, but that's not my bug. Writing the Index2 is handled by the codegen. Another reason why it may be better to manually implement it as an array type.

behdad: Write. Thanks for confirming. https://github.com/googlefonts/fontations/issues/1719

write-fonts/src/tables/variations/mivs_builder.rs
```diff
@@ -0,0 +1,416 @@
+//! Building the MultiItemVariationStore for VARC
+//!
+//! See also [VariationStoreBuilder](crate::tables::variations::ivs_builder::VariationStoreBuilder). Where Item Variation Stores stores a
+//! single scalar delta per region, [`MultiItemVariationStore`] stores a tuple
+//! of deltas per region, and uses a sparse representation of regions
+//! (only active axes are stored).
```
cmyr: I personally find the docs easier to read when the links are put at the end..

```suggestion
//! See also [`VariationStoreBuilder`]. Where Item Variation Stores stores a
//! single scalar delta per region, [`MultiItemVariationStore`] stores a tuple
//! of deltas per region, and uses a sparse representation of regions
//! (only active axes are stored).
//!
//! [`VariationStoreBuilder`]: crate::tables::variations::ivs_builder::VariationStoreBuilder
```

write-fonts/src/tables/variations/mivs_builder.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// Sorted by region index.
+    deltas: Vec<(u16, Vec<i32>)>,
+}
+
+impl MultiDeltaSet {
+    fn new(tuple_len: usize, mut deltas: Vec<(u16, Vec<i32>)>) -> Self {
+        // Sort by region index
+        deltas.sort_by_key(|(idx, _)| *idx);
+        // Verify all tuples have the expected length
+        debug_assert!(
```
cmyr: Any reason to not make this a full `assert!`? In general I would avoid `debug_assert!` except in cases where there is a very specific rationale, most often performance-related.

In general I think it's fine to enforce our invariants with a real assert, and make those invariants explicit in a doc comment. This is private API, so it's reasonable to expect the caller to uphold some explicit invariants...

write-fonts/src/tables/variations/mivs_builder.rs
```diff
@@ -0,0 +1,416 @@
...
+fn encode_packed_deltas(packed: &PackedDeltas, output: &mut Vec<u8>) {
+    use crate::write::FontWrite;
+
+    // Use a TableWriter to get the bytes
+    let mut writer = crate::write::TableWriter::default();
+    packed.write_into(&mut writer);
+
+    // Extract the bytes from the writer's internal data
+    let data = writer.into_data();
+    output.extend(&data.bytes);
```
cmyr, line 330: slightly annoying that this involves writing all the bytes twice, and we _could_ improve on that with a manual reimplementation of the (simple) `FontWrite` impl for `PackedDeltas` but doesn't necessarily feel worth it..

write-fonts/src/tables/gpos.rs
```diff
@@ -6,14 +6,16 @@ include!("../../generated/generated_gpos.rs");
 
 use std::collections::HashSet;
 
+use crate::tables::layout::VariationIndex;
```
cmyr: the neat-freak in my would toss this in with the other imports, below

write-fonts/src/tables/varc.rs
```diff
@@ -0,0 +1,601 @@
...
+
+        self.transform.write_into(writer);
+
+        // Technically we are supposed to process and discard one uint32var
+        // per each set bit in RESERVED_MASK. But we explicitly set bits in
+        // RESERVED_MASK to zero, as per the spec. So we just do nothing here.
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
cmyr, line 243: maybe worth linking https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#uint32var, here?

write-fonts/src/tables/varc.rs
```diff
@@ -0,0 +1,601 @@
...
+mod tests {
+    use crate::{
+        dump_table,
+        tables::{layout::CoverageFormat1, variations::mivs_builder::SparseRegion},
+        write::TableWriter,
+    };
+
+    use super::*;
+
+    // fn dump_test_ttf(varc: &Varc) {
```
cmyr: do we want to keep this around for something? if yes might be worth a comment.

write-fonts/src/tables/variations/mivs_builder.rs
rsheeter: No asserts, use an error, as per discussion of how asserts kill servers.

PR #1718, opened 2026-01-19
[read-fonts] Add initial CffFontRef type

dfrg:
Adds a new `CffFontRef` type that encapsulates CFF and CFF2 functionality for both raw blobs and OpenType tables. This only includes data that is parsed from the base blob and the top dict so far.

Based on #1717 which should land first.


cmyr (approved):
looks good!

PR #1721, opened 2026-01-30
[klippa-repacker] split markbase_pos table

qxliu76:
Also use treemap instead of hashmap for real_links, so that topological sorting is deterministic

### DIFF
--- a/klippa/src/graph.rs
+++ b/klippa/src/graph.rs
@@ -6,11 +6,13 @@ use crate::{
     serialize::{Link, LinkWidth, ObjIdx, Object, OffsetWhence, SerializeErrorFlags, Serializer},
 };
 use fnv::FnvHashMap;
+use std::collections::BTreeMap;
 use write_fonts::{read::collections::IntSet, types::Uint24};
 
 mod coverage_graph;
 pub(crate) mod layout;
 pub(crate) mod ligature_graph;
+pub(crate) mod markbasepos_graph;
 
 #[derive(Debug)]
 pub(crate) enum RepackError {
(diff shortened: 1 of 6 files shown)


klippa/src/graph.rs
```diff
@@ -1323,15 +1329,15 @@ impl Graph {
     }
 
     // Moves the child of old_parent_idx pointed to by old_offset to a new vertex at the new_offset.
-    // Returns the idx of the child that was moved.
+    // Returns the idx of the child that was moved
```
cmyr, line 1332: what does it mean if we do not return an index?

klippa/src/graph/markbasepos_graph.rs
```diff
@@ -0,0 +1,578 @@
...
+        class_mark_indices
+            .get_mut(mark_class as usize)
+            .ok_or(RepackError::ErrorReadTable)?
+            .insert(i as u16);
+        *class = mark_class;
+    }
+    Ok((class_mark_indices, mark_classes))
+}
+
+fn compute_split_points(
```
cmyr, line 134: I think it's useful for anything like this to link back to the original source where possible? I assume in this case that would be

https://github.com/harfbuzz/harfbuzz/blob/aba63bb5f/src/graph/markbasepos-graph.hh#L239

klippa/src/graph/markbasepos_graph.rs
```diff
@@ -0,0 +1,578 @@
...
+            accumulated = base_size + delta;
+            partial_coverage_size = 4 + 2 * num_marks_for_class;
+            visited.clear();
+        }
+    }
+    Ok(out)
+}
+
+// Create a new MarkBasePos that has all of the data for classes from [start, end).
+fn clone_range(
```
cmyr, line 195: https://github.com/harfbuzz/harfbuzz/blob/aba63bb5/src/graph/markbasepos-graph.hh#L412 ?

klippa/src/graph.rs
qxliu76: meaning there's no child link at the offset position(offset is null)

garretrieger (approved):

PR #1723, opened 2026-02-02
[read-fonts] add subfont type for new CffFontRef

dfrg:
Adds Font and Private DICT parsing along with a `CffSubfont` type to wrap it all up.

### DIFF
--- a/read-fonts/src/tables/postscript/font/cff.rs
+++ b/read-fonts/src/tables/postscript/font/cff.rs
@@ -263,6 +263,16 @@ enum CffFontKind<'a> {
 }
 
 /// Metadata for a CFF subfont.
+///
+/// A subfont is the collection of data read from both the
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/font/cff.rs
```diff
@@ -141,6 +143,97 @@ impl<'a> CffFontRef<'a> {
                 .map_or(Some(0), |fds| fds.font_index(gid)),
         }
     }
+
+    /// Returns the subfont with the given index and normalized variation
+    /// coordinates.
```
cmyr, line 148: it would be nice if this could link to some documentation explaining what exactly a CFF subfont is?

dfrg: Good point. The gist is that a "subfont" is our representation of the data in the Font DICT and Private DICT. This data controls transforms and hinting and the font can associate different configurations to sets of glyphs. The mapping from glyph ids is provided by the "FDArray" operator. It's all lots of fun.

Will add some comments explaining this.

PR #1724, opened 2026-02-03
[read-fonts] add some ops to PS stack

dfrg:
Adds div and exch operators to the PostScript stack. Required for Type1 support.


read-fonts/src/tables/postscript/stack.rs
```diff
@@ -84,6 +84,19 @@ impl Stack {
         self.value_is_fixed[..self.top].reverse();
     }
 
+    /// Swaps the top two elements.
+    #[allow(unused)]
```
cmyr, line 88: i guess MSRV prevents us from using [`#[expect(unused)]`](https://doc.rust-lang.org/1.93.0/reference/attributes/diagnostics.html#the-expect-attribute), which I would now prefer in most cases to 'allow'...

dfrg: I thought about that but wasn't sure how it interacts with tests? In any case, I'll make sure to remove these attributes in a coming PR.

PR #1725, opened 2026-02-03
[read-fonts] add context for charstring eval

dfrg:
CFF/Type1 have slightly different logic around seac components and subroutines. This introduces a new `CharstringContext` trait abstract those differences.

### DIFF
--- a/read-fonts/src/tables/postscript/charstring.rs
+++ b/read-fonts/src/tables/postscript/charstring.rs
@@ -13,6 +13,50 @@ use crate::{
 /// <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf#page=33>
 pub const NESTING_DEPTH_LIMIT: u32 = 10;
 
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -13,6 +13,50 @@ use crate::{
 /// <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf#page=33>
 pub const NESTING_DEPTH_LIMIT: u32 = 10;
 
+/// Trait that provides access to resources for charstring evaluation.
```
cmyr, line 16: is this exclusively used in type-1? if so I would note that in the docs.

dfrg: It's used for both CFF and Type1 so that the charstring eval code can be agnostic, at least with regard to resource resolution. There will be some nice `if self.is_type1()` code coming that ruins this for the various operators :)

PR #1728, opened 2026-02-04
[codegen] Special-case CFF INDEX struct size

behdad:
I don't know if hacking codegen like this is kosher, but I think we need some code to this effect. CFF INDEXes are self-contained structs and know their own length. Please advise.

### DIFF
--- a/font-codegen/src/table.rs
+++ b/font-codegen/src/table.rs
@@ -998,7 +998,40 @@ impl Table {
     }
 
     fn iter_field_validation_stmts(&self) -> impl Iterator<Item = TokenStream> + '_ {
-        self.fields.iter().map(Field::field_parse_validation_stmts)
+        let table_name = self.raw_name().to_string();
+        let is_index = table_name == "Index1" || table_name == "Index2";
+        self.fields.iter().map(move |field| {
+            if is_index && field.name == "data" {
+                let len_field_name = field.shape_byte_len_field_name();
+                quote! {
+                    let #len_field_name = {
+                        let count: usize = count as usize;
+                        if count == 0 {
+                            0usize
+                        } else {
+                            let off_size: usize = off_size as usize;
+                            let offsets_end = cursor.position()?;
+                            let last_offset_pos = offsets_end
+                                .checked_sub(off_size)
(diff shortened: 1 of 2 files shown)


font-codegen/src/table.rs
```diff
@@ -998,7 +998,40 @@ impl Table {
     }
 
     fn iter_field_validation_stmts(&self) -> impl Iterator<Item = TokenStream> + '_ {
-        self.fields.iter().map(Field::field_parse_validation_stmts)
+        let table_name = self.raw_name().to_string();
+        let is_index = table_name == "Index1" || table_name == "Index2";
```
cmyr, line 1002: This is a funny little case hm? In general we don't like to have this kind of code be actually defined in codegen. Currently we have special labels for the different special patterns for computing a count (such as the `add_multiply`, used elsewhere in `Index1` and `Index2` and so following that pattern we would have some annotation like, `#[count(cff_index_data($offsets)]`, and maybe we would also introduce a type for the offsets themselves that would abstract the variable length so we could use a ComputedArray, which would simplify getting the last element... let me try putting that patch together as a first step.

behdad: Thanks.

cmyr: thinking about this more, is this patch motivated by a specific concern? it is true that the raw data array here will potentially include extra bytes belonging to another table, but given that we know the start/end offsets of the actual objects, is there any risk of us misinterpreting the data?

behdad: Well one issue that current code is incorrect is that if count=0, we shouldn't expect offSize at all.

As for consuming the rest of the bytes, I noticed it, because I'm working on VARC table rendering, which uses Index2 in a couple of different ways. So I wasn't sure if the current code will work for me.

dfrg: We could just mark the count field as a version and then set the others to min version of 1. Hacky but should work?

cmyr: we have a general concept of a field being conditional on something, so that's fine: the annoying bit is just that it means `off_size` would become `Option<u8>`.

PR #1731, opened 2026-02-04
[klippa] fix gvar padding bug

qxliu76:
ref: https://github.com/harfbuzz/harfbuzz/pull/5716

This PR is dependent on: https://github.com/googlefonts/fontations/pull/1732

### DIFF
--- a/otexplorer/src/main.rs
+++ b/otexplorer/src/main.rs
@@ -117,6 +117,7 @@ fn get_some_table<'a>(
         tables::maxp::Maxp::TAG => font.maxp().map(|x| Box::new(x) as _),
         tables::name::Name::TAG => font.name().map(|x| Box::new(x) as _),
         tables::post::Post::TAG => font.post().map(|x| Box::new(x) as _),
+        tables::varc::Varc::TAG => font.varc().map(|x| Box::new(x) as _),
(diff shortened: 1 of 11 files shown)


cmyr:
if I understand your concern: the 'subtables' API  on the specific GPOS/GSUB lookup enums is intended explicitly as an abstraction that hides whether not a lookup is an extension or not. You can avoid this by actually matching on the lookup and calling `subtables` there.

read-fonts/src/tables/gpos.rs
```diff
@@ -87,6 +87,24 @@ impl<'a> PositionLookup<'a> {
         self.of_unit_type().lookup_type()
     }
 
+    /// If lookup type is extension, return lookup type of extension subtable
+    /// otherwise return lookup type
+    pub fn ext_lookup_type(&self) -> Result<u16, ReadError> {
```
cmyr: if this is unrelated to the gvar padding issue it should be in its own PR :)

qxliu76: it's needed for the new test to pass. I'll try to make them in 2 separate commits but still in the same PR

cmyr: or just fix the other bug first, and then this patch can go in after that?

qxliu76: Done

PR #1732, opened 2026-02-04
[klippa]fix lookup type for extension lookups

qxliu76:
use ext subtable type instead of lookup type.
Since there's no enum for extension subtables in GSUB/GPOS, see: https://github.com/googlefonts/fontations/blob/0cf460a37a19117d0fd2a40d7fc168f75d13dec7/read-fonts/src/tables/gsub.rs#L41, we have to demote all extension subtables during subsetting, repacker should be able to do extension promotion when an overflow occurs. Maybe we could do the same demotion in harfbuzz.

### DIFF
--- a/klippa/src/gpos.rs
+++ b/klippa/src/gpos.rs
@@ -214,14 +214,25 @@ impl<'a> SubsetTable<'a> for PositionLookup<'_> {
         s: &mut Serializer,
         args: Self::ArgsForSubset,
     ) -> Result<(), SerializeErrorFlags> {
-        s.embed(self.lookup_type())?;
-        let lookup_flag = self.lookup_flag();
-        let lookup_flag_pos = s.embed(lookup_flag)?;
-        let lookup_count_pos = s.embed(0_u16)?;
-
         let subtables = self
             .subtables()
             .map_err(|_| s.set_err(SerializeErrorFlags::SERIALIZE_ERROR_READ_ERROR))?;
+
+        let lookup_type: u16 = match subtables {
+            PositionSubtables::Single(_) => 1,
+            PositionSubtables::Pair(_) => 2,
+            PositionSubtables::Cursive(_) => 3,
+            PositionSubtables::MarkToBase(_) => 4,
+            PositionSubtables::MarkToLig(_) => 5,
+            PositionSubtables::MarkToMark(_) => 6,
+            PositionSubtables::Contextual(_) => 7,
+            PositionSubtables::ChainContextual(_) => 8,
+        };
+        s.embed(lookup_type)?;
+
+        let lookup_flag = self.lookup_flag();
+        let lookup_flag_pos = s.embed(lookup_flag)?;
+        let lookup_count_pos = s.embed(0_u16)?;
         let lookup_count = subtables.subset(plan, s, args)?;
         s.copy_assign(lookup_count_pos, lookup_count);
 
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/gpos.rs
```diff
@@ -87,6 +87,24 @@ impl<'a> PositionLookup<'a> {
...
+            return Ok(raw_type);
+        }
+        let first_offset = raw_lookup
+            .subtable_offsets()
+            .first()
+            .ok_or(ReadError::OutOfBounds)?
+            .get();
+
+        let first_ext: ExtensionPosFormat1<()> = first_offset.resolve(raw_lookup.offset_data())?;
+        Ok(first_ext.extension_lookup_type())
```
cmyr: we could also reuse the logic in `subtable` below, which already does this:


```suggestion
        self.subtables().map(|sub| match sub {
            PositionSubtables::Single(_) => 1,
            PositionSubtables::Pair(_) => 2,
            PositionSubtables::Cursive(_) => 3,
            PositionSubtables::MarkToBase(_) => 4,
            PositionSubtables::MarkToLig(_) => 5,
            PositionSubtables::MarkToMark(_) => 6,
            PositionSubtables::Contextual(_) => 7,
            PositionSubtables::ChainContextual(_) => 8,
        })
```

read-fonts/src/tables/gsub.rs
```diff
@@ -58,6 +58,25 @@ impl<'a> SubstitutionLookup<'a> {
...
+        }
+        let first_offset = raw_lookup
+            .subtable_offsets()
+            .first()
+            .ok_or(ReadError::OutOfBounds)?
+            .get();
+
+        let first_ext: ExtensionSubstFormat1<()> =
+            first_offset.resolve(raw_lookup.offset_data())?;
+        Ok(first_ext.extension_lookup_type())
```
cmyr:
```suggestion
        self.subtables().map(|sub| match sub {
            SubstitutionSubtables::Single(_) => 1,
            SubstitutionSubtables::Multiple(_) => 2,
            SubstitutionSubtables::Alternate(_) => 3,
            SubstitutionSubtables::Ligature(_) => 4,
            SubstitutionSubtables::Contextual(_) => 5,
            SubstitutionSubtables::ChainContextual(_) => 6,
            SubstitutionSubtables::Reverse(_) => 8,
        })
```

qxliu76:
> Should there be a test case for this problem specifically?

the gvar padding fix contains a testcase that has this issue, that's why I combined both fixes into one PR

read-fonts/src/tables/gpos.rs
qxliu76: Seems the `subtables()` has additional steps that construct subtables that are not needed.

cmyr: yes, it is going to be performing some amount of additional work, and I leave it to you to decide if the overhead of that work is worth paying or not. (the overhead should not be significant, we're constructing types but there's no allocation or validation or anything else really)

qxliu76: As you suggested, I pushed a change that matches lookup type from `subtables()`, no read-fonts change is needed. Can you help take another look?

cmyr: This patch is already marked as approved, my notes were just suggestions. Feel free to merge!

PR #1733, opened 2026-02-05
Render `VARC` fonts

behdad:
This hooks up `VARC` fonts. I have tested it with HB cmdline tools and seems to work nicely, although, 25% slower than HB on my test CJK font.

This breaks the VARC table API in read-fonts / write-fonts. But nothing else (hopefully).

### DIFF
--- a/read-fonts/src/tables/varc.rs
+++ b/read-fonts/src/tables/varc.rs
@@ -45,7 +45,7 @@ impl Varc<'_> {
     }
 }
 
-impl<'a> SparseVariationRegion<'a> {
+impl SparseVariationRegion<'_> {
     /// Computes a floating point scalar value for this sparse region and the
     /// specified normalized variation coordinates.
     pub fn compute_scalar_f32(&self, coords: &[F2Dot14]) -> f32 {
--- a/skrifa/src/outline/varc/mod.rs
+++ b/skrifa/src/outline/varc/mod.rs
@@ -316,6 +316,7 @@ impl<'a> Outlines<'a> {
         Ok(self.glyph_metrics.advance_width(outline.glyph_id, coords))
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn draw_glyph(
         &self,
         glyph_id: GlyphId,
@@ -455,6 +456,7 @@ impl<'a> Outlines<'a> {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn component_coords(
         &self,
         component: &VarcComponent<'_>,


read-fonts/src/tables/varc.rs
```diff
@@ -279,9 +352,9 @@ impl DecomposedTransform {
         // TODO: this produces very small floats for rotations, e.g. 90 degree rotation a basic scale
         // puts 1.2246467991473532e-16 into [0]. Should we special case? Round?
 
-        // Python: t = t.rotate(math.radians(self.rotation))
+        // Python: t = t.rotate(self.rotation * math.pi)
         if self.rotation != 0.0 {
-            let (s, c) = (self.rotation).to_radians().sin_cos();
+            let (s, c) = (self.rotation * core::f64::consts::PI).sin_cos();
```
behdad: @simoncozens These probably break your babelfont code, yeah?

skrifa/src/outline/glyf/deltas.rs
```diff
@@ -60,9 +60,6 @@ where
     if iup_buffer.len() < glyph.points.len() || glyph.points.len() < PHANTOM_POINT_COUNT {
         return Err(ReadError::InvalidArrayLen);
     }
-    for delta in deltas.iter_mut() {
-        *delta = Default::default();
-    }
```
behdad, line 65: This is an independent optimization. The next line `compute_deltas_for_glyph` below already does the same.

resources/codegen_inputs/varc.rs
```diff
@@ -42,7 +42,7 @@ table SparseVariationRegionList {
 table SparseVariationRegion {
     region_axis_count: u16,
     #[count($region_axis_count)]
-    region_axis_offsets: [SparseRegionAxisCoordinates],
+    region_axes: [SparseRegionAxisCoordinates],
```
behdad, line 45: This was wrongly named, since there's no offsets involved.

behdad:
Using `hb-draw-compare`, I'm seeing some correctness issues that I'm debugging.

behdad:
The rabbithole also led me to file this issue, which doesn't seem to have a significant effect on this dataset, but a bug nonetheless: https://github.com/googlefonts/fontations/issues/1738

behdad:
> I spent quite a few hours digging this rabbit hole

For my own reference: Gemini session: https://gemini.google.com/share/54b90c1c1d1f

behdad:
Tests added.

read-fonts/src/tables/varc.rs
```diff
@@ -94,29 +93,30 @@ impl<'a> VarcComponent<'a> {
...
     ///  * HarfBuzz [VarComponent::get_path_at](https://github.com/harfbuzz/harfbuzz/blob/0c2f5ecd51d11e32836ee136a1bc765d650a4ec0/src/OT/Var/VARC/VARC.cc#L132)
-    // TODO: do we want to be able to parse into an existing glyph to avoid allocation?
     fn parse(table: &Varc, cursor: &mut Cursor<'a>) -> Result<Self, ReadError> {
         let raw_flags = cursor.read_u32_var()?;
         let flags = VarcFlags::from_bits_truncate(raw_flags);
         // Ref https://github.com/harfbuzz/boring-expansion-spec/blob/main/VARC.md#variable-component-record
 
         // This is a GlyphID16 if GID_IS_24BIT bit of flags is clear, else GlyphID24.
-        let gid = if flags.contains(VarcFlags::GID_IS_24BIT) {
+        let gid = if raw_flags & VarcFlags::GID_IS_24BIT.bits != 0 {
```
dfrg: Any reason for avoiding `contains` here (and below)?

read-fonts/src/tables/varc.rs
```diff
@@ -373,15 +463,78 @@ mod tests {
         }
     }
 
-    fn round6(v: f64) -> f64 {
+    fn round6(v: f32) -> f32 {
         (v * 1_000_000.0).round() / 1_000_000.0
     }
 
+    fn sfnt_table_range(data: &[u8], tag: [u8; 4]) -> (usize, usize) {
```
dfrg, line 527: This should probably just make use of  `FontRef`. Something like:
```rust
FontRef::new(data).unwrap().table_directory().table_records().iter().find(|rec| rec.tag() == tag).unwrap().map(|rec| (rec.offset() as usize, rec.length() as usize))
```

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+        Ok(Some(Outline {
+            glyph_id,
+            coverage_index,
+            max_component_memory,
+        }))
+    }
+
+    /// Lightweight coverage lookup without computing max_component_memory.
+    fn coverage_index(&self, glyph_id: GlyphId) -> Result<Option<u16>, ReadError> {
+        let coverage = self.varc.coverage()?;
```
dfrg: Would it be worth preloading the coverage table and storing it in `Outlines` in the constructor?

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+        path_style: PathStyle,
+        pen: &mut impl OutlinePen,
+    ) -> Result<(), DrawError> {
+        let mut font_coords = CoordVec::new();
+        expand_coords(&mut font_coords, self.axis_count, coords);
+        let mut stack = GlyphStack::new();
+        let pen: &mut dyn OutlinePen = pen;
+        let coverage = self.varc.coverage()?;
+        let var_store = self.varc.multi_var_store().transpose()?;
+        let regions = var_store.as_ref().map(|s| s.region_list()).transpose()?;
```
dfrg: Actually storing all of coverage, var_store and regions on `Outlines` might help. At the very least, it would reduce the number of parameters for all the function calls below.

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+        coords: &[F2Dot14],
+        pen: &mut impl OutlinePen,
+    ) -> Result<i32, DrawError> {
+        let size = Size::unscaled();
+        self.draw(outline, buf, size, coords, PathStyle::default(), pen)?;
+        Ok(self.glyph_metrics.advance_width(outline.glyph_id, coords))
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn draw_glyph(
```
dfrg, line 320: We pass around a lot of context. Maybe add a `VarcScaler` (or `VarDrawContext`) type to wrap this up and simplify some of the function calls?

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+        //println!("reset {}", flags.contains(VarcFlags::RESET_UNSPECIFIED_AXES) as u32);
+        // //print axis coords as integer
+        //print!("axis[");
+        for (i, _coord) in current_coords.iter().copied().enumerate() {
+            if i != 0 {
+                //print!(", ");
+            }
+            //print!("{}:{}", i, (coord.to_bits()) as i32);
+        }
+        //println!("]");
```
dfrg: Some left over debugging stuff?

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+
+        self.axis_values(component, num_axes, &mut scratch.axis_values)?;
+        //print!("bef [");
+        for (i, _value) in scratch.axis_values.iter().copied().enumerate() {
+            if i != 0 {
+                //print!(", ");
+            }
+            //print!("{}", (value * 16384.0).round() as i32);
+        }
+        //println!("]");
```
dfrg: And some more :)

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+        }
+        // //print axis values as integer
+        //print!("aft [");
+        for (i, _value) in scratch.axis_values.iter().copied().enumerate() {
+            if i != 0 {
+                //print!(", ");
+            }
+            //print!("{}", (value * 16384.0).round() as i32);
+        }
+        //println!("]");
```
dfrg: ^

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+    }
+
+    fn eval_condition(
+        condition: &Condition<'a>,
+        coords: &[F2Dot14],
+        var_store: &MultiItemVariationStore<'a>,
+        regions: &SparseVariationRegionList<'a>,
+        scalar_cache: &mut ScalarCache,
+        scratch: &mut Scratchpad,
+    ) -> Result<bool, DrawError> {
```
dfrg, line 667: Should we have a max recursion depth check in this function?

skrifa/src/outline/varc/mod.rs
```diff
@@ -0,0 +1,1410 @@
...
+                skip = 0;
+            }
+            deltas.add_to_f32_scaled(out_slice, scalar)?;
+        }
+    }
+
+    Ok(())
+}
+
+fn compute_sparse_region_scalar(region: &SparseVariationRegion<'_>, coords: &[F2Dot14]) -> f32 {
```
dfrg: This should be a method on `SparseVariationRegion` in read-fonts, probably `compute_scalar_f32` to match `VariationRegion`.

skrifa/src/outline/varc/mod.rs
behdad: Because offsets are positive, a loop is not possible in Condition graph.

skrifa/src/outline/varc/mod.rs
behdad: I tried it before, and tried it now. Somehow it slows down the benchmark consistently.

skrifa/src/outline/varc/mod.rs
behdad: I did this but measuring a slight slowdown as well. Donno. WDYT?

behdad: I'm trying an in-between version as well.

behdad: Still slower. 740ms instead of 733ms, so, 1% slower. Maybe that's fine...

behdad: This and caching the objects, gave my performance back. Sweet.

PR #1734, opened 2026-02-05
[read-fonts] add T1 charstring ops

dfrg:
Extend charstring evaluation to support a minimal set of Type1 specific operators.

### DIFF
--- a/read-fonts/src/tables/postscript/charstring.rs
+++ b/read-fonts/src/tables/postscript/charstring.rs
@@ -1398,31 +1398,83 @@ mod tests {
     }
 
     #[test]
-    fn op_sbw() {
+    fn op_div() {
         let mut commands = CaptureCommandSink::default();
-        let mut eval = Evaluator::new(&NullContext(CharstringKind::Type1), None, &mut commands);
+        let mut eval = Evaluator::new(&NullContext(CharstringKind::Type2), None, &mut commands);
         let mut cursor = FontData::new(&[]).cursor();
-        eval.stack.push(Fixed::from_f64(42.5)).unwrap();
-        eval.stack.push(0).unwrap();
-        eval.stack.push(501).unwrap();
-        eval.stack.push(1000).unwrap();
-        eval.evaluate_operator(Operator::Sbw, &mut cursor, 0)
+        eval.stack.push(Fixed::from_f64(512.5)).unwrap();
+        eval.stack.push(2).unwrap();
+        eval.evaluate_operator(Operator::Div, &mut cursor, 0)
             .unwrap();
-        assert_eq!(eval.sbx, Fixed::from_f64(42.5));
-        assert_eq!(eval.wx, Fixed::from_f64(501.0));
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -1112,16 +1384,106 @@ mod tests {
...
+        eval.stack.push(Fixed::from_f64(42.5)).unwrap();
+        eval.stack.push(501).unwrap();
+        eval.evaluate_operator(Operator::Hsbw, &mut cursor, 0)
+            .unwrap();
+        assert_eq!(eval.sbx, Fixed::from_f64(42.5));
+        assert_eq!(eval.wx, Fixed::from_f64(501.0));
+    }
+
+    #[test]
+    fn op_callothersubr_flex() {
```
drott, line 1481: Would it be possible to exercise a few of the additional new operations be covered in tests,  VStem3, HStem3, for example?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -13,8 +13,24 @@ use crate::{
...
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum CharstringKind {
+    /// Type1 charstring.
+    ///
+    /// See reference at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/T1_SPEC.pdf>.
+    Type1,
+    /// Type2 charstring.
+    ///
+    /// See reference at <https://adobe-type-tools.github.io/font-tech-notes/pdfs/5177.Type2.pdf>.
+    Type2,
```
drott, line 26: Is it possible to cover acceptance/rejection of CharString kind somehow in tests - or exercise usage of this enum?

drott (commented):
Impressive level of deep-diving into CFF. Big thanks for digging in!

read-fonts/src/tables/postscript/charstring.rs
dfrg: The stem operators aren't implemented in this PR so there's not much to test. Since PDFium doesn't enable hinting for Type1, my inclination is to ignore this for now unless some need pops up in the future. I can however add some tests for `div` and `setcurrentpoint`. The only remaining (functional) operator is `seac` which can't really be tested without additional Type1 infrastructure because we need access to the charstrings.

read-fonts/src/tables/postscript/charstring.rs
dfrg: Unfortunately, there's not much we can do since we recently removed operator validation to handle a broken font (and match FT) and these basically control which operators are applicable (or adjust behavior).

I can add another set of tests for `sbw` and `hsbw` that assert no changes to state in Type2 mode.

drott: Thanks for the explanations. I am ok either way then.

drott (approved):
LGTM

PR #1746, opened 2026-02-11
[read-fonts] raw type1 dict data

dfrg:
Implements the first portion of type1 parsing:
- Handle both pfb (printer font binary) and pfa (printer font ascii) files
- find the base dict bytes
- compute the private dict bytes
  - for ascii, these might be hex encoded
  - for binary, these might be stored in multiple segments
  - both ascii and binary require decryption

This is exposed through the `RawDicts` struct which just unifies the two formats and handles all top level assembly/decryption so that we can perform actual parsing.

Also adds a couple test fonts (well one but in both pfb and pfa formats)

### DIFF
--- a/read-fonts/src/tables/postscript/font/type1.rs
+++ b/read-fonts/src/tables/postscript/font/type1.rs
@@ -77,7 +77,7 @@ fn decode_pfb_tag(data: &[u8], start: usize) -> Option<(u16, u32)> {
 }
 
 /// Returns an iterator over the sequence of PFB binary segments.
-fn decode_pfb_binary_segments(data: &[u8]) -> impl Iterator<Item = [u8]> + '_ {
+fn decode_pfb_binary_segments(data: &[u8]) -> impl Iterator<Item = &[u8]> + '_ {
     let mut pos = 0usize;
     core::iter::from_fn(move || {
         let (tag, len) = decode_pfb_tag(data, pos)?;


font-test-data/test_data/README.md
```diff
@@ -132,6 +132,18 @@ Describes the provenance, usage and generation procedures for font data used for
   * font: Hand rolled cmap 10 subtable
   * usage: testing cmap format 10 mapping and iteration
 
+* type1/notoserif-regular.subset.pfa
+  * font: Noto Serif Regular
+  * source: https://fonts.google.com/noto/specimen/Noto+Serif
```
drott, line 137: Interesting, I did not know these exist. Did you convert Noto into a pfa? Or is this part of the Noto releases?

dfrg: I converted them myself. Turns out that FontForge is still a useful tool :)

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -0,0 +1,369 @@
...
+struct RawDicts<'a> {
+    /// Data containing the base dicitionary.
+    base: &'a [u8],
+    /// Data containing the decrypted private dictionary.
+    private: Vec<u8>,
+}
+
+impl<'a> RawDicts<'a> {
+    fn new(data: &'a [u8]) -> Option<Self> {
+        if let Some((PFB_TEXT_SEGMENT_TAG, base_size)) = decode_pfb_tag(data, 0) {
```
cmyr, line 13: so pfb have a leading tag, but pfa do not?

cmyr (approved):
I trust you've learned a bunch of new and fascinating things through this work...

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Yeah, pfa is pure ASCII but pfb consists of a series of segments that can be ASCII or binary. FT assumes that the first is ASCII and the remainder are binary so I’ve done the same here.

PR #1749, opened 2026-02-12
codegen: minimal validate

cmyr:
This is an attempt to implement 'minimal validation': this means reducing the work that we do at parse time (it becomes a single check of the minimum possible table size.) Bounds are now computed and checked when a field is accessed, instead of immediately.

Because this stuff moves to access time, this changes some of the behaviour of the API (although the API itself remains mostly the same.) Specifically, _we can no longer guarantee by construction that data is present for certain fields_.

For a very simple example:

```rust
table MyTable {
    count: u16,
    #[count($count)]
    doodads: [u16],
}
```

When we read this table, we will now only check that it contains at least 2 bytes, for the `count` field. When a user accesses the `doodads` array, we will read the `count` field and do a checked read of the data. Previously we would guarantee that if a table had been successfully constructed, this array read would always succeed. Now, it can fail; _in the case of a failed read, we will return `Default::default()`_, or in this case the empty array.

This is a potential complication for existing code that assumed tables were always well formed. I'm not sure how significant a problem this is, right now, but I'm bringing it up so that we can discuss it.

The other slightly-user-facing change here is that the various `byte_range` fns on tables that return the location of fields in the underlying data have a slightly different convention: previously these would return `Option<Range<usize>>` for fields that might be missing, whereas now they return `Range<usize>`, and if a field is missing then the range has zero length. This simplifies codegen because it means we can always treat the end of the previous range as the start of the next field, even if that range is missing.

Other than that, this feels... okay? It's possible that I've overlooked some of the implications of these changes, and introduced new codepaths that could cause crashes on some inputs, and this might be a good motivation for us to start fuzzing read-fonts more aggressively.

Also after this patch there are a pair of failing tests in IFT that I haven't investigated.

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -517,6 +517,13 @@ impl Field {
     }
 
     pub(crate) fn known_min_size_stmt(&self) -> Option<TokenStream> {
+        if let Some(lit) = self.attrs.count.as_deref().and_then(Count::lit_int) {
+            if let FieldType::Array { inner_typ } = &self.typ {
+                if let FieldType::Scalar { typ } = inner_typ.as_ref() {
+                    return Some(quote!( #typ :: RAW_BYTE_LEN * #lit ));
+                }
+            }
+        }
         match &self.typ {
             _ if self.is_conditional() => None,
             FieldType::Offset { typ, .. } | FieldType::Scalar { typ } => {
@@ -757,7 +764,7 @@ impl Field {
         } else if is_var_array {
             quote!( self.data.split_off(range.start).and_then(|d| VarLenArray::read(d).ok()) #maybe_unwrap_or_def )
         } else if is_array {
-            quote!(self.data.read_array(range).ok() #maybe_unwrap_or_def)
+            quote!(self.data.read_array(range).ok() #maybe_unwrap #maybe_unwrap_or_def)
         } else {
             quote!(self.data.read_at(range.start).ok() #maybe_unwrap #maybe_unwrap_or_def)
         };
--- a/font-codegen/src/parsing.rs
+++ b/font-codegen/src/parsing.rs
@@ -1594,6 +1594,14 @@ impl Count {
         }
     }
 
+    pub(crate) fn lit_int(&self) -> Option<&syn::LitInt> {
+        if let Count::SingleArg(CountArg::Literal(int)) = self {
+            Some(int)
+        } else {
+            None
+        }
+    }
+
     pub(crate) fn iter_referenced_fields(&self) -> impl Iterator<Item = &syn::Ident> {
         let (one, two) = match self {
             Self::SingleArg(CountArg::Field(ident)) => (Some(ident), None),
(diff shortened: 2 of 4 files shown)


font-codegen/src/fields.rs
```diff
@@ -531,13 +516,19 @@ impl Field {
...
+        match &self.typ {
+            _ if self.is_conditional() => None,
+            FieldType::Offset { typ, .. } | FieldType::Scalar { typ } => {
+                Some(quote!(#typ :: RAW_BYTE_LEN))
+            }
+            FieldType::Array { .. }
+            | FieldType::ComputedArray(_)
+            | FieldType::VarLenArray(_)
+            | FieldType::Struct { .. } => Some(Default::default()),
+            //FieldType::Struct { typ } => Some(quote!( #typ :: MIN_SIZE  )),
```
dfrg: Would love to see this working at some point

font-codegen/src/table.rs
```diff
@@ -105,47 +84,45 @@ pub(crate) fn generate(item: &Table) -> syn::Result<TokenStream> {
...
+            #phantom_decl
+        }
 
         #[allow(clippy::needless_lifetimes)]
         impl<'a, #generic> #raw_name<'a, #generic> {
+            pub const MIN_SIZE: usize = #min_valid_size;
+
+            basic_table_impls!(impl_the_methods);
 
+            #( #field_byte_range_fns )*
```
dfrg: Can we move the `field_byte_range_fns` to a separate impl or at least to the bottom of this one so that they don't appear at the top of the docs?

font-codegen/src/table.rs
```diff
@@ -105,47 +84,45 @@ pub(crate) fn generate(item: &Table) -> syn::Result<TokenStream> {
...
             #( #table_ref_getters )*
 
         }
 
+        //impl<'a, #generic> FontTable<'a> for #raw_name<'a, #generic> {
+            //fn offset_data(&self) -> FontData<'a> {
+                //self.data
+            //}
+        //}
```
dfrg: dead code?

read-fonts/src/font_data.rs
```diff
@@ -8,7 +8,7 @@ use types::{BigEndian, FixedSize, Scalar};
 
 use crate::array::ComputedArray;
 use crate::read::{ComputeSize, FontReadWithArgs, ReadError};
-use crate::table_ref::TableRef;
+//use crate::table_ref::TableRef;
```
dfrg: dead code?

font-codegen/src/table.rs
cmyr: oh yes, good call

font-codegen/src/fields.rs
cmyr: is there a concrete use case? My sense is that most of the time this exist (record directly in table) it's when the record has something fancy going on that means its length can't be known?

dfrg: It might just be the bitmap codegen that you mentioned in chat. Would be nice to support this but definitely not a priority.

PR #1750, opened 2026-02-21
[read-fonts] type1 tokenization

dfrg:
Adds a new parser for breaking type1 font data into tokens for easier processing.

Based on #1746

### DIFF
--- a/font-codegen/src/fields.rs
+++ b/font-codegen/src/fields.rs
@@ -8,8 +8,7 @@ use syn::spanned::Spanned;
 
 use super::parsing::{
     logged_syn_error, Attr, Condition, Count, CountArg, CustomCompile, Field, FieldReadArgs,
-    FieldType, FieldValidation, Fields, IfTransform, NeededWhen, OffsetTarget, Phase, Record,
-    ReferencedFields,
+    FieldType, FieldValidation, Fields, NeededWhen, OffsetTarget, Phase, Record, ReferencedFields,
 };
 
 impl Fields {
@@ -214,22 +213,6 @@ impl Fields {
                                 }
                             }
                         }
-                        Condition::IfCond { xform } => match xform {
-                            IfTransform::AnyFlag(_, _) => {
-                                let condition_not_set_message =
-                                    format!("if_cond is not satisfied but '{name}' is present.");
-                                let condition_set_message =
-                                    format!("if_cond is satisfied by '{name}' is not present.");
-                                quote! {
-                                    if !(#condition) && self.#name.is_some() {
-                                        ctx.report(#condition_not_set_message);
-                                    }
-                                    if (#condition) && self.#name.is_none() {
-                                        ctx.report(#condition_set_message);
-                                    }
-                                }
-                            }
-                        },
                     }
                 });
 
@@ -302,24 +285,11 @@ impl Fields {
     }
 }
 
-fn if_expression(xform: &IfTransform, add_self: bool) -> TokenStream {
-    match xform {
-        IfTransform::AnyFlag(field, flags) => {
-            if add_self {
-                quote!(self.#field.intersects(#(#flags)|*))
-            } else {
-                quote!(#field.intersects(#(#flags)|*))
-            }
-        }
-    }
-}
-
 impl Condition {
     fn condition_tokens_for_read(&self) -> TokenStream {
         match self {
             Condition::SinceVersion(version) => quote!(version.compatible(#version)),
             Condition::IfFlag { field, flag } => quote!(#field.contains(#flag)),
-            Condition::IfCond { xform } => if_expression(xform, false),
         }
     }
 
@@ -327,7 +297,6 @@ impl Condition {
         match self {
             Condition::SinceVersion(version) => quote!(version.compatible(#version)),
             Condition::IfFlag { field, flag } => quote!(self.#field.contains(#flag)),
-            Condition::IfCond { xform } => if_expression(xform, true),
         }
     }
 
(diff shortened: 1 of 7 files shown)


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -366,4 +602,107 @@ mod tests {
...
+                Token::Raw(b"]"),
+                Token::Raw(b"readonly"),
+                Token::Raw(b"def"),
+                Token::Name(b"FontName"),
+                Token::Name(b"NotoSerif-Regular"),
+            ]
+        );
+    }
+
+    fn check_tokens(source: &str, expected: &[Token]) {
```
wmedrano, line 726: Nit: I like adding `#[track_caller]` so that the blamed line is within the unit test.

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -366,4 +602,107 @@ mod tests {
...
+                Token::Int(-5),
+                Token::Int(252),
+            ],
+        );
+    }
+
+    #[test]
+    fn parse_strings() {
+        check_tokens(
+            "(string (nested) 1) % and a hex string:\n <DEAD BEEF>",
```
wmedrano, line 642: What about unterminated strings, so like?

`(string (nested) 1 % and a hex string:\n <DEAD BEEF>`

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -366,4 +602,107 @@ mod tests {
...
+                Token::LitString(b"string (nested) 1"),
+                Token::HexString(b"DEAD BEEF"),
+            ],
+        );
+    }
+
+    #[test]
+    fn parse_procs() {
+        check_tokens(
+            "{a {nested 20} proc} % and a\n {simple proc}",
```
wmedrano: Similar to above, what about unterminated procs?

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Sounds good to me. Done.

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Good catch. Despite the overall lax parsing, FreeType does add an explicit check for this. I've added the same.

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Fixed as above.

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -199,6 +199,251 @@ fn is_whitespace(c: u8) -> bool {
...
+            let start = self.pos;
+            let c = self.next_byte()?;
+            match c {
+                // Line comment
+                b'%' => {
+                    while let Some(c) = self.next_byte() {
+                        if c == b'\n' || c == b'\r' {
+                            break;
+                        }
+                    }
```
cmyr: just a thought, but I find the big loop with all the inline logic a bit annoying, it looks like this could generally be reworked so that we're (at least sometimes?) calling methods on self? Like this could just be,


```rust
match c {
    b'%' => return self.eat_line_comment(),
   _ => (),
}
```

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -199,6 +199,251 @@ fn is_whitespace(c: u8) -> bool {
...
+                }
+                // Procedures
+                b'{' => {
+                    while self.next_byte()? != b'}' {
+                        // This handles nested procedures
+                        self.next()?;
```
cmyr: so this is advancing the cursor but not actually handling anything? Which I guess makes sense, was a bit confused at first.

read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -199,6 +199,251 @@ fn is_whitespace(c: u8) -> bool {
...
+        if hash_idx == 1 || hash_idx == 2 {
+            // It's a radix number, like 8#40.
+            let radix_str = s.get(0..hash_idx)?;
+            let number_str = s.get(hash_idx + 1..)?;
+            let radix = radix_str.parse::<u32>().ok()?;
+            if (2..=36).contains(&radix) {
+                i64::from_str_radix(number_str, radix).ok()
+            } else {
+                None
+            }
```
cmyr: this made me go check the docs. Just golf, but I'd probably do one of,

```rust
let radix = radix_str.parse::<u32>().ok().filter(|n| (2..-36).contains(n))?;
```

or,

```rust
let radix = radix_str.parse::<u32>().ok()?;
let radix = radix.max(2).min(36); // invariant of from_str_radix
i64::from_str_radix(number_str, radix).ok()

```

read-fonts/src/tables/postscript/font/type1.rs
dfrg: I gambled that this might be short enough to avoid triggering your loop length detector… and lost :) Will break these out.

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Yeah, we don’t actually care about the content of the procedures since we’re not writing a postscript interpreter. Just trying to skip them accurately.

read-fonts/src/tables/postscript/font/type1.rs
dfrg: First one lgtm!

PR #1752, opened 2026-02-27
[read-fonts] type1: read subroutines

dfrg:
Just collect the subroutines for later processing.

### DIFF
--- a/read-fonts/src/tables/postscript/font/type1.rs
+++ b/read-fonts/src/tables/postscript/font/type1.rs
@@ -518,10 +518,7 @@ impl Parser<'_> {
         self.expect(Token::Raw(b"array"))?;
         let mut is_dense = true;
         // The pattern for each subroutine is `dup <subr_num> <data>`
-        loop {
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -456,6 +497,62 @@ impl Parser<'_> {
...
+                self.expect(Token::Raw(b"]"))?;
+                return Some(subrs);
+            }
+            Token::Int(n) => n.try_into().ok()?,
+            _ => return None,
+        };
+        self.expect(Token::Raw(b"array"))?;
+        let mut is_dense = true;
+        // The pattern for each subroutine is `dup <subr_num> <data>`
+        loop {
```
wmedrano: `while self.accept(Token::Raw(b"dup"))`?

dfrg: cleans that up nicely, thanks!

PR #1754, opened 2026-03-03
[read-fonts] add Type1Font struct

dfrg:
To combine results of reading matrix, subroutines and charstrings. Also updates font matrix parsing to extract adjusted upem when the y scale component is `!= 1/1000`

Based on #1753 which must land first

### DIFF
--- a/read-fonts/src/tables/postscript/font/type1.rs
+++ b/read-fonts/src/tables/postscript/font/type1.rs
@@ -464,7 +464,7 @@ impl<'a> Parser<'a> {
                     return Some(Token::Raw(data));
                 }
                 _ => {
-                    let count = self.skip_until(|b| is_whitespace(b) || is_special(b));
+                    let count = self.skip_until(is_special_or_whitespace);
                     let content = self.data.get(start..start + count)?;
                     // Look for numbers but don't try to parse fractional
                     // values since we want to handle those with special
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -559,7 +633,9 @@ impl Parser<'_> {
     /// Like FreeType, this is designed assuming a upem of 1000 and produces
     /// an identity matrix in that case. This is, the result is scaled such
     /// that 0.001 yields a value of 1.0.
-    fn read_font_matrix(&mut self) -> Option<FontMatrix> {
+    ///
+    /// See <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/type1/t1load.c#L1403>
+    fn read_font_matrix(&mut self) -> Option<ScaledFontMatrix> {
         let mut components = [Fixed::ZERO; 6];
         // skip [
         self.next()?;
```
wmedrano: Should this be `self.expect(Token::Raw('['))`?

Similar for `]` case.

wmedrano (approved):

read-fonts/src/tables/postscript/font/type1.rs
dfrg: Indeed, FT does check the opening brace (and also allows `{`) but hits an early exit before validating the closing brace so I've changed the behavior here to match.

PR #1758, opened 2026-03-11
[read-fonts] parse Type1 encodings

dfrg:
Parses all forms of the encoding field in the Type1 font dict. Also expands our encoding tables to include expert and latin-1 mappings to string ids along with adding a new `PredefinedEncoding` enum that provides a public API for these.

### DIFF
--- a/read-fonts/src/tables/postscript/font/type1.rs
+++ b/read-fonts/src/tables/postscript/font/type1.rs
@@ -859,7 +859,7 @@ impl Parser<'_> {
             Token::Int(count) => {
                 // We're limited to 256 character codes
                 // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/type1/t1load.c#L1518>
-                let count: usize = count.min(256).max(0) as usize;
+                let count: usize = count.clamp(0, 256) as usize;
                 let mut map = Vec::new();
                 // Start with all glyphs mapped to notdef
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -28,12 +29,19 @@ impl Type1Font {
...
         let mut parser = Parser::new(raw_dicts.base);
         while let Some(token) = parser.next() {
-            if let Token::Name(b"FontMatrix") = token {
-                font.matrix = Some(parser.read_font_matrix()?);
+            match token {
+                Token::Name(b"FontMatrix") => font.matrix = Some(parser.read_font_matrix()?),
+                // Simply save the encoding offset. We'll parse it after
+                // we have read charstrings so we have an accurate mapping
+                // if we've synthesized or remapped a notdef glyph
+                Token::Name(b"Encoding") => encoding_offset = Some(parser.pos),
```
cmyr, line 43: is it meaningful/a concern if we have multiple encoding offsets, and are ignoring earlier ones?

behdad: That's generally possible in CFF as well. It's fine, it would be invalid anyway to do so.

dfrg: yeah, we just take the latest definition... this is what FreeType does as well. That reminds me that FreeType retains the _earliest_ definition for charstrings if more than one exists (and this apparently happens in some fonts) so I need to fix that.

PR #1760, opened 2026-03-11
[read-fonts] Type1 parsing fixes

dfrg:
Just a couple changes to match FreeType's handling of malformed fonts:
- When duplicate subroutines or charstrings are encountered, only keep the first definition
- When searching for the `eexec` token which marks the start of the private dictionary, use the parser to avoid matching "eexec" in comments and strings

### DIFF
--- a/read-fonts/src/tables/postscript/font/type1.rs
+++ b/read-fonts/src/tables/postscript/font/type1.rs
@@ -25,6 +25,10 @@ impl Type1Font {
 
     fn new_impl(data: &[u8]) -> Option<Self> {
         let raw_dicts = RawDicts::new(data)?;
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -1151,6 +1171,7 @@ mod tests {
         check_hex_decode(b"743", &[116, 48]);
     }
 
+    #[track_caller]
```
cmyr, line 1174: what's this for?

dfrg: Will suggested adding this to test helper functions so that failures are reported at the call site in the actual test rather than inside the helper and that made sense to me.

PR #1765, opened 2026-03-19
[read-fonts] collection of small type1 fixes 

dfrg:
These get us to a matching state with FT for a large set of type1 fonts.

- (h)sbw operators accumulate side bearings to handle seac
- ignore some additional operators in type1
- make sure paths start with a move
- actually store the width (and return it) rather than just marking it "read"
- only apply initial tangent for hhcruveto and vvcurveto if we have enough points on the stack
- handle metrics adjustments in seac operator
- add support for transforming horizontal metrics so that the user can scale advance widths properly

Also updates fauntlet to check advance widths for type1 fonts.

Some of the API around transforms here is likely to change but I wanted to land the functional fixes since this puts us in a good state wrt matching FT.

### DIFF
--- a/fauntlet/src/compare_glyphs.rs
+++ b/fauntlet/src/compare_glyphs.rs
@@ -58,22 +58,17 @@ pub fn compare_glyphs(
         // only for TrueType and Type1 glyphs)
         if let Some(skrifa_advance) = maybe_skrifa_advance {
             if ft_advance != skrifa_advance {
-                let have_mismatch = if let Some((hvar_adv, gvar_adv)) =
-                    skrifa_instance.hvar_and_gvar_advance_deltas(gid)
-                {
-                    // Some fonts have slight discrepancies between HVAR and
-                    // gvar advance deltas.
-                    // If these _are_ the same but the scaler computed advance
-                    // is different then we definitely have a bug.
-                    // If the difference is greater than 1 then we might have
-                    // a bug.
-                    if hvar_adv == gvar_adv || (ft_advance - skrifa_advance).abs() > 1.0 {
-                        true
-                    } else {
-                        false
+                // Some fonts have slight discrepancies between HVAR and
+                // gvar advance deltas.
+                // If these _are_ the same but the scaler computed advance
+                // is different then we definitely have a bug.
+                // If the difference is greater than 1 then we might have
+                // a bug.
+                let have_mismatch = match skrifa_instance.hvar_and_gvar_advance_deltas(gid) {
+                    Some((hvar_adv, gvar_adv)) => {
+                        hvar_adv == gvar_adv || (ft_advance - skrifa_advance).abs() > 1.0
                     }
-                } else {
-                    true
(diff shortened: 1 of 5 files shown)


fauntlet/src/compare_glyphs.rs
```diff
@@ -65,7 +68,15 @@ pub fn compare_glyphs(
...
                     // a bug.
                     if hvar_adv == gvar_adv || (ft_advance - skrifa_advance).abs() > 1.0 {
-                        writeln!(
+                        true
+                    } else {
+                        false
+                    }
+                } else {
+                    true
+                };
```
cmyr: I'd personally express this as a match, instead of a nested conditional? but 🤷 

```rust
let have_mismatch = match skrifa_instance.hvar_and_gvar_advance_deltas(gid) {
    _ if (ft_advance - skrifa_advance).abs() > 1.0 => true,
    Some((hvar_adv, gvar_adv)) => hvar_adv == gvar_adv,
    None => true,
};
```

read-fonts/src/tables/postscript/font/cff.rs
```diff
@@ -216,7 +216,11 @@ impl<'a> CffFontRef<'a> {
         };
         let charstring_data = charstrings.get(gid.to_u32() as usize)?;
         let ctx = (self.data, &charstrings, &self.global_subrs, &subrs);
-        super::super::charstring::evaluate(&ctx, blend, charstring_data, sink)
+        if let Some(width) = charstring::evaluate(&ctx, blend, charstring_data, sink)? {
+            Ok(Some(width + subfont.nominal_width))
+        } else {
+            Ok(Some(subfont.default_width))
```
cmyr: the return type is `Result<Option<_>,_>`  but do we ever return `None`? Is this just to make it easier at the call site?

read-fonts/src/tables/postscript/charstring.rs
```diff
@@ -253,6 +257,7 @@ where
     ) -> Result<bool, Error> {
         use Operator::*;
         use PointMode::*;
+        // println!("{operator:?}");
```
cmyr: leftovers

fauntlet/src/compare_glyphs.rs
dfrg: well the whole `if expr { true } else { false }` thing is kinda embarrassing :) Fixed.

read-fonts/src/tables/postscript/font/cff.rs
dfrg: I think my concern was around the `DefaultWidth` field being missing in which case we "guarantee" an advance width that is always 0. Changed to make `default_width` optional so that this return type makes sense and the user can decide what to do with a missing advance width.

read-fonts/src/tables/postscript/charstring.rs
dfrg: oof, deleted

PR #1772, opened 2026-03-22
[read-fonts] small type1 fixes

dfrg:
Fixes some parsing bugs that were very helpfully pointed out by @LaurenzV.


read-fonts/src/tables/postscript/font/type1.rs
```diff
@@ -357,13 +357,15 @@ fn find_eexec_data(data: &[u8]) -> Option<usize> {
...
-                b' ' | b'\t' | b'\n' => {}
+                b' ' | b'\t' => {}
+                b'\n' => seen_linefeed = true,
                 b'\r' => {
-                    // Only stop at \r if it is not followed by \n
-                    if data.get(start + 1) != Some(&b'\n') {
-                        start += 1;
+                    // If we've already seen \n or there is not a \n later
+                    // in the data, then stop at this \r
+                    if seen_linefeed || !data[start..].contains(&b'\n') {
```
LaurenzV: Isn't it very expensive to check from start to the whole end for `\n` in every iteration?

dfrg: yeah... a few kb of `\r`s could definitely produce a nice DoS. Made the check lazy and cached the result so we only do the search once if necessary.

PR #1777, opened 2026-03-25
[read-fonts] CFF dict/charstring robustness fixes

dfrg:
A handful of small changes to dict and charstring parsing that relax various requirements to make parsing more robust when dealing with malformed fonts.

Based on #1776 which should land first.


read-fonts/src/tables/postscript/dict.rs
```diff
@@ -325,6 +325,12 @@ pub fn entries<'a>(
         }
         let token = match parse_token(&mut cursor) {
             Ok(token) => token,
+            Err(Error::InvalidDictOperator(_)) => {
```
wmedrano, line 328: Is clearing the stack in line with freetype?

dfrg: Indeed. I added a comment mentioning this with links to the relevant FT code (https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/cff/cffparse.c#L1328)

PR #1779, opened 2026-03-26
[read-fonts] parse CFF encodings 

dfrg:
And expose them in similar fashion to type1.

This uses minimal codegen because encodings have a format field that is masked and some optional fields depending on the high bit of format and I don't believe we have the machinery to handle that.

### DIFF
--- a/read-fonts/src/tables/postscript/encoding.rs
+++ b/read-fonts/src/tables/postscript/encoding.rs
@@ -78,8 +78,8 @@ impl<'a> Encoding<'a> {
     /// expert encodings, respectively.
     pub fn new(data: &'a [u8], offset: usize) -> Result<Self, ReadError> {
         match offset {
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/postscript/encoding.rs
```diff
@@ -5,6 +5,171 @@
...
+    fn code_to_sid(&self, code: u8) -> Option<StringId> {
+        let code = code as usize;
+        let sid = match self {
+            Self::Standard => STANDARD_ENCODING[code] as u16,
+            Self::Expert => EXPERT_ENCODING[code],
+            _ => return None,
+        };
+        Some(StringId::new(sid))
+    }
+}
```
dfrg, line 63: Pardon the code shuffling here. I rearranged the file to match charset.rs which had a more logical structure (tables at the bottom)

PR #1780, opened 2026-03-27
[read-fonts] handle Adobe glyph list

dfrg:
Adds support for mapping between glyph names and characters using the Adobe glyph list.

Required to replace functionality that PDFium uses in FreeType and will allow us to generate a Unicode cmap for Type1/CFF fonts.

Uses a python script to generate a Rust data structure containing an optimized trie which packs the glyph list into about 55k. The code handles mapping in both directions including the u and uni prefixes and components separated by underscores. Borrows some code and all tests from `fontdrasil` with the hope that this will be a suitable replacement for the AGL (but not the AGLFN) there.

All this lives in a new top level `ps` module where I intend to eventually move all of the PostScript code with some better organization to reflect the fact that we now support PS fonts outside of the sfnt container.

### DIFF
--- a/read-fonts/Cargo.toml
+++ b/read-fonts/Cargo.toml
@@ -31,6 +31,8 @@ spec_next = []
 default = ["std"]
 serde = ["dep:serde", "font-types/serde"]
 libm = ["dep:core_maths"]
+# Support for mapping glyph names <-> unicode via the Adobe glyph list
+agl = []
 
 [dependencies]
 font-types = { workspace = true, features = ["bytemuck"] }
--- a/read-fonts/src/ps.rs
+++ b/read-fonts/src/ps.rs
@@ -1,3 +1,4 @@
 //! PostScript fonts.
 
+#[cfg(feature = "agl")]
 pub mod agl;


.github/workflows/rust.yml
```diff
@@ -119,6 +119,8 @@ jobs:
 
       - name: run codegen
         run: cargo run --bin=codegen resources/codegen_plan.toml
+      - name: build agl
+        run: python read-fonts/scripts/gen_agl.py
       - name: ensure no unstaged changes
```
dfrg, line 124: I added the generated AGL code to `read-fonts/generated/generated_agl.rs` which the codegen check happily deletes and then the check complains. My "fix" was to regenerate the AGL here but I'm open to suggestions.

dfrg:
Good catch. I definitely intend to feature gate this.

.github/workflows/rust.yml
cmyr: hmm yea that's annoying. Ideally this would be a build script or something, but I don't hate this 🤷

dfrg: I thought about using a build script but we already have some complaints about compile times and this should be regenerated basically never so I just went with the python approach

behdad:
This can benefit from the same algorithm that subroutinizes CFF charstrings! I'll think about it and see if my agents can produce something smaller.

PR #1782, opened 2026-03-31
Fix fixed-point division overflow using unsigned_abs()

wahidfayad:
### Problem
The fixed-point division implementation manually handled signs, which can overflow when dividing i32::MIN, causing incorrect results.

### Solution
Replaced manual sign-handling with `unsigned_abs()`, which correctly converts i32::MIN to a u32 before division. This prevents overflow and ensures correct behavior for all edge cases.

### Changes
- Updated `impl Div for $ty` in `font-types/src/fixed.rs` within `fixed_mul_div!` macro.
- Simplified logic using `unsigned_abs()` and XOR for sign computation.

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -196,25 +196,16 @@ macro_rules! fixed_mul_div {
 
         impl Div for $ty {
             type Output = Self;
-            #[inline(always)]
-            fn div(self, other: Self) -> Self::Output {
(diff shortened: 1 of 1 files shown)


wahidfayad:
Ready for review -- fixed the missing type `Output = Self;`, reformatted with `cargo fmt`, and added a test case covering `i32::MIN` division overflow.

PR #1786, opened 2026-04-01
[read-fonts] build cmap from glyph names/ids + AGL

dfrg:
Of course, this couldn't be a simple collect list of `(agl_char(name), gid)`, sort and bsearch. That would be too easy.

Builds on #1783 which should land first.

### DIFF
--- a/read-fonts/src/ps/charmap.rs
+++ b/read-fonts/src/ps/charmap.rs
@@ -117,6 +117,7 @@ impl Charmap {
 }
 
 /// Iterator for a character map.
+#[derive(Clone)]
 pub struct Iter<'a>(core::iter::Copied<core::slice::Iter<'a, (u32, GlyphId)>>);
 
 impl Iterator for Iter<'_> {
--- a/read-fonts/src/ps.rs
+++ b/read-fonts/src/ps.rs
@@ -3,6 +3,7 @@
 #[cfg(feature = "agl")]
 pub mod agl;
 pub mod cff;
+#[cfg(feature = "std")]
 pub mod charmap;
 pub mod cs;
 pub mod encoding;


read-fonts/src/ps/charmap.rs
```diff
@@ -0,0 +1,235 @@
...
+        result
+    }
+
+    pub fn iter(&self) -> Iter<'_> {
+        Iter(self.mapping.iter().copied())
+    }
+}
+
+/// Iterator for a character map.
+pub struct Iter<'a>(core::iter::Copied<core::slice::Iter<'a, (u32, GlyphId)>>);
```
cmyr, line 121: what's the rationale here? It doesn't seem like we're doing anything fancy? (I would've expected impl Iterator?)


In any case I feel like the chad I know would want this to be at least `Clone`..

read-fonts/src/ps/charmap.rs
```diff
@@ -0,0 +1,235 @@
...
+
+#[cfg(test)]
+#[cfg(feature = "agl")]
+mod tests {
+    use super::super::type1::Type1Font;
+    use super::*;
+
+    #[test]
+    fn cmap() {
+        let cmap = Charmap::from_glyph_names(
```
cmyr, line 158: `cargo test -p read-fonts --no-default-features --features=agl` is unhappy..

read-fonts/src/ps/charmap.rs
dfrg: I’d like to potentially add a charmap API that covers both type1 and opentype and handling iteration across the different possible table types requires named iterator types (see the current code in skrifa). 

And yes, I wish for this to be 'Clone`, thanks :)

read-fonts/src/ps/charmap.rs
dfrg: 😢 will fix

PR #1789, opened 2026-04-01
[read-fonts] always use alloc

dfrg:
This makes read-fonts require a heap allocator but we no longer lose any functionality in no_std builds since we only actually depend on alloc.

Revisit and add an alloc feature if somebody absolutely needs heapless read-fonts.


qxliu76 (approved):
LGTM

PR #1793, opened 2026-04-02
Move OutlinePen to read-fonts

dfrg:
- reexports from skrifa to prevent breakage in Skia
- adds some `draw` helper methods to `Type1Font` and `CffFontRef` that output directly to `OutlinePen` which covers the vast majority of use cases
- demonstrates the simplified code in fauntlet

### DIFF
--- a/font-types/src/lib.rs
+++ b/font-types/src/lib.rs
@@ -24,7 +24,6 @@ mod longdatetime;
 mod matrix;
 mod name_id;
 mod offset;
-pub mod pen;
 mod point;
 mod raw;
 mod tag;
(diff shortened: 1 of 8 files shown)


font-types/src/lib.rs
```diff
@@ -24,6 +24,7 @@ mod longdatetime;
 mod matrix;
 mod name_id;
 mod offset;
+pub mod pen;
```
cmyr: what's the advantage of having this here, versus in read-fonts?

dfrg: as discussed in chat, will move this to read-fonts

PR #1794, opened 2026-04-02
[skrifa] Allow for recording autohinters

simoncozens:
These changes (mainly hidden behind a feature flag) allow for *non*-runtime auto hinters to record hinting decisions in ways that can be serialised to TrueType bytecode.

### DIFF
--- a/skrifa/Cargo.toml
+++ b/skrifa/Cargo.toml
@@ -16,7 +16,8 @@ repository.workspace = true
 all-features = true
 
 [features]
-default = ["std", "autohint_shaping"]
+default = ["std", "autohint_shaping", "autohinter"]
+autohinter = []
 std = ["read-fonts/std"]
 traversal = ["std", "read-fonts/experimental_traverse"]
 # Enables extended shaping support for the autohinter. Enabled by default.
--- a/skrifa/Cargo.toml
+++ b/skrifa/Cargo.toml
@@ -16,8 +16,7 @@ repository.workspace = true
 all-features = true
 
 [features]
-default = ["std", "autohint_shaping", "autohinter"]
-autohinter = []
+default = ["std", "autohint_shaping"]
 std = ["read-fonts/std"]
 traversal = ["std", "read-fonts/experimental_traverse"]
 # Enables extended shaping support for the autohinter. Enabled by default.
--- a/skrifa/generated/generated_autohint_styles.rs
+++ b/skrifa/generated/generated_autohint_styles.rs
@@ -3,7 +3,7 @@
 // Use ../scripts/gen_autohint_scripts.py to regenerate.
 
 #[rustfmt::skip]
-pub(super) const SCRIPT_CLASSES: &[ScriptClass] = &[
+pub const SCRIPT_CLASSES: &[ScriptClass] = &[
     ScriptClass {
         name: "Adlam",
         group: ScriptGroup::Default,
@@ -828,7 +828,7 @@ pub(super) const SCRIPT_CLASSES: &[ScriptClass] = &[
 }
 
 #[rustfmt::skip]
-pub(super) const STYLE_CLASSES: &[StyleClass] = &[
+pub const STYLE_CLASSES: &[StyleClass] = &[
     StyleClass { name: "Adlam", index: 0, script: &SCRIPT_CLASSES[0], feature: None },
     StyleClass { name: "Arabic", index: 1, script: &SCRIPT_CLASSES[1], feature: None },
     StyleClass { name: "Armenian", index: 2, script: &SCRIPT_CLASSES[2], feature: None },
--- a/skrifa/src/outline/autohint/hint/mod.rs
+++ b/skrifa/src/outline/autohint/hint/mod.rs
@@ -6,6 +6,7 @@ mod outline;
 use super::{
     metrics::{scale_style_metrics, Scale, UnscaledStyleMetrics},
     outline::Outline,
+    recorder::HintsRecorder,
     style::{GlyphStyle, ScriptGroup},
     topo::{self, Axis},
 };
@@ -39,6 +40,45 @@ pub(crate) fn hint_outline(
     scale: &Scale,
     glyph_style: Option<GlyphStyle>,
 ) -> HintedMetrics {
+    hint_outline_impl(outline, metrics, scale, glyph_style, None).hinted_metrics
+}
+
+pub(crate) struct HintedPlan {
+    pub hinted_metrics: HintedMetrics,
+    #[allow(dead_code)]
+    pub vertical_axis: Option<Axis>,
+}
+
+#[allow(dead_code)] // used in tests
+#[cfg(feature = "autohinter")]
+pub(crate) fn hint_outline_with_recorder(
+    outline: &mut Outline,
+    metrics: &UnscaledStyleMetrics,
+    scale: &Scale,
+    glyph_style: Option<GlyphStyle>,
+    recorder: Option<&mut HintsRecorder>,
+) -> HintedMetrics {
+    hint_outline_impl(outline, metrics, scale, glyph_style, recorder).hinted_metrics
+}
+
+#[cfg(feature = "autohinter")]
+pub(crate) fn hint_outline_with_plan(
+    outline: &mut Outline,
+    metrics: &UnscaledStyleMetrics,
+    scale: &Scale,
+    glyph_style: Option<GlyphStyle>,
+    recorder: Option<&mut HintsRecorder>,
+) -> HintedPlan {
+    hint_outline_impl(outline, metrics, scale, glyph_style, recorder)
+}
+
+fn hint_outline_impl(
+    outline: &mut Outline,
+    metrics: &UnscaledStyleMetrics,
+    scale: &Scale,
+    glyph_style: Option<GlyphStyle>,
+    mut recorder: Option<&mut HintsRecorder>,
+) -> HintedPlan {
     let scaled_metrics = scale_style_metrics(metrics, *scale);
     let scale = &scaled_metrics.scale;
     let mut axis = Axis::default();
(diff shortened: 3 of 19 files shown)


skrifa/generated/generated_autohint_styles.rs
```diff
@@ -828,7 +828,7 @@ pub(super) const SCRIPT_CLASSES: &[ScriptClass] = &[
 }
 
 #[rustfmt::skip]
-pub(super) const STYLE_CLASSES: &[StyleClass] = &[
+pub const STYLE_CLASSES: &[StyleClass] = &[
```
dfrg, line 831: Is is necessary to make these public or would accessing the classes through glyph and metrics suffice?

skrifa/src/outline/autohint/hint/mod.rs
```diff
@@ -39,6 +40,45 @@ pub(crate) fn hint_outline(
...
+    metrics: &UnscaledStyleMetrics,
+    scale: &Scale,
+    glyph_style: Option<GlyphStyle>,
+    recorder: Option<&mut HintsRecorder>,
+) -> HintedMetrics {
+    hint_outline_impl(outline, metrics, scale, glyph_style, recorder).hinted_metrics
+}
+
+#[cfg(feature = "autohinter")]
+pub(crate) fn hint_outline_with_plan(
```
dfrg: Suggest renaming this `hint_outline_with_recorder` and dropping the other method with that name as it doesn't really do anything useful. Since this would always take a recorder, the `Option` wrapping `&mut HintsRecorder` seems unnecessary.

skrifa/src/outline/autohint/metrics/mod.rs
```diff
@@ -19,6 +19,169 @@ use std::sync::{Arc, RwLock};
 pub(crate) use blues::{BlueZones, ScaledBlue, ScaledBlues, UnscaledBlue, UnscaledBlues};
 pub(crate) use scale::{compute_unscaled_style_metrics, scale_style_metrics};
 
+#[cfg(feature = "autohinter")]
+/// Public snapshot of a single unscaled blue zone.
+#[derive(Clone, Debug, Default)]
+pub struct ExportedUnscaledBlue {
```
dfrg: These types seem a bit redundant. If possible, we should make the fields of the actual types `pub(crate)` and add some methods for accessing the necessary data and then expose those.

skrifa/src/outline/autohint/metrics/blues.rs
```diff
@@ -27,7 +27,7 @@ const BLUE_STRING_MAX_LEN: usize = 51;
 /// Defines the zone(s) that are associated with a blue value.
 #[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
 #[repr(transparent)]
-pub(crate) struct BlueZones(u16);
+pub struct BlueZones(u16);
```
dfrg, line 30: Is this actually exposed publicly anywhere?

skrifa/src/outline/autohint/metrics/scale.rs
```diff
@@ -156,14 +156,23 @@ fn scale_default_axis_metrics(
...
         }
     }
-    // Now scale the widths
+    // Now scale the widths. FreeType ensures there is always at least one
+    // width entry (the standard width), even if width extraction found none.
     axis.width_metrics = width_metrics;
-    for unscaled_width in widths {
-        let scaled = fixed_mul(axis.scale, *unscaled_width);
+    if widths.is_empty() {
+        let scaled = fixed_mul(axis.scale, axis.width_metrics.standard_width);
```
dfrg, line 187: I remember copying some bugs for FreeType around missing widths. This change makes sense to me but I'll need to do some testing to make sure we're not breaking compatibility.

skrifa/src/outline/autohint/instance.rs
```diff
@@ -47,6 +47,35 @@ impl GlyphStyles {
...
+    #[cfg(feature = "autohinter")]
+    /// Returns the skrifa-internal style index for `glyph_id`, if the glyph
+    /// has been assigned to a style.  The index is into the `STYLE_CLASSES`
+    /// array.  Returns `None` for unassigned glyphs.
+    pub fn style_index(&self, glyph_id: u32) -> Option<usize> {
+        use raw::types::GlyphId;
+        self.0
+            .style(GlyphId::new(glyph_id))
+            .and_then(|s| s.style_index().map(|v| v as usize))
+    }
```
dfrg: I'd rather just expose the `GlyphStyle` type and add a `get` method for it.

skrifa/src/outline/autohint/mod.rs
```diff
@@ -4,12 +4,300 @@ mod hint;
...
 pub(crate) use instance::Instance;
+#[cfg(feature = "autohinter")]
+pub use metrics::{
+    compute_scaled_style_metrics_exported, compute_unscaled_style_metrics_exported,
+    ExportedScaledBlue, ExportedScaledStyleMetrics, ExportedScaledWidth, ExportedUnscaledBlue,
+    ExportedUnscaledStyleMetrics,
+};
+
+#[cfg(feature = "autohinter")]
+pub use style::{SCRIPT_CLASSES, STYLE_CLASSES};
```
dfrg, line 21: We expose the arrays but not `ScriptClass` or `StyleClass`.

skrifa/src/outline/autohint/mod.rs
```diff
@@ -4,12 +4,300 @@ mod hint;
...
+#[cfg(feature = "autohinter")]
+use crate::{FontRef, MetadataProvider};
+#[cfg(feature = "autohinter")]
+use alloc::vec::Vec;
+#[cfg(feature = "autohinter")]
+use raw::types::{F2Dot14, GlyphId};
+
+#[cfg(feature = "autohinter")]
+#[derive(Copy, Clone, PartialEq, Eq, Debug, Default)]
+pub struct ExportedHintRecord {
```
dfrg: These `Exported` types are all fairly opaque and are visible in the top level `outline` module which adds quite a bit of noise. I think we can better organize these for a public API.

skrifa/src/outline/autohint/mod.rs
```diff
@@ -4,12 +4,300 @@ mod hint;
...
+
+    Some(ExportedHintPlan {
+        records,
+        segments,
+        edges,
+    })
+}
+
+#[cfg(feature = "autohinter")]
+pub fn compute_hint_records_exported(
```
dfrg: This doesn't really do anything useful beyond `compute_hint_plan_exported` so we should drop it.

skrifa/src/outline/autohint/mod.rs
```diff
@@ -4,12 +4,300 @@ mod hint;
...
+            is_non_base,
+            is_digit,
+            ppem,
+        )?
+        .records,
+    )
+}
+
+#[cfg(feature = "autohinter")]
+fn export_records(records: Vec<recorder::HintRecord>) -> Vec<ExportedHintRecord> {
```
dfrg: Seems like we have a nice enum here that gets flattened into a struct with the union of expected fields. If this is the shape of the data expected by the autohinter, I think this mapping should happen there instead.

skrifa/src/outline/autohint/mod.rs
```diff
@@ -4,12 +4,300 @@ mod hint;
...
+        has_blue,
+        blue_scaled,
+        blue_fitted,
+        blue_ix,
+        blue_is_shoot,
+    }
+}
+
+#[cfg(feature = "autohinter")]
+fn action_code(action: recorder::Action) -> u8 {
```
dfrg: Similar to above, we should publicly expose the action type and handle the mapping to ttfautohint values in the other crate. Can these actions apply to both points and edges or are the sets exclusive?

skrifa/src/outline/autohint/outline.rs
```diff
@@ -451,21 +451,10 @@ impl Outline {
...
-            x + ((3 * y) >> 3)
-        } else {
-            y + ((3 * x) >> 3)
-        }
-    }
-    let d_in = hypot(in_x, in_y);
-    let d_out = hypot(out_x, out_y);
-    let d_hypot = hypot(ax, ay);
-    (d_in + d_out - d_hypot) < (d_hypot >> 4)
+    let d_in = in_x.abs() + in_y.abs();
```
dfrg, line 469: This might indeed be a better calculation but I'd prefer to avoid breaking compatibility with the runtime autohinter at the moment. I'm not opposed to adding a parameter to control this if you'd like to keep this behavior for the recorder.

skrifa/Cargo.toml
```diff
@@ -16,7 +16,8 @@ repository.workspace = true
 all-features = true
 
 [features]
-default = ["std", "autohint_shaping"]
+default = ["std", "autohint_shaping", "autohinter"]
```
dfrg: I don't think the feature is necessary. Particularly since it only really controls whether the public symbols are exposed and is enabled by default.

skrifa/src/outline/mod.rs
```diff
@@ -507,6 +517,30 @@ impl<'a> OutlineGlyph<'a> {
         }
     }
 
+    #[cfg(feature = "autohinter")]
+    /// Pass a raw scaled outline to a callback with direct point access.
+    pub fn with_scaled_glyf_outline<R>(
```
dfrg, line 511: This is exposing internal types (which should have admittedly been marked with less visibility). I don't have a great alternative for this at the moment so I think we'll just roll with it to avoid blocking. Let's at least mark it `#[doc(hidden)]` and add a comment that this is unstable and subject to change/removal at any time.

dfrg (commented):
The logic here mostly LGTM and nothing is blocking except for the parts that break FT compatibility.

I do have some concerns about the public API. I understand that this is primarily meant to feed the AOT autohinter but the types and fields seem a bit too raw and shouldn't be visible in the root `outline` module.

I have some ideas in mind for improving the API. I can provide more detailed notes here for you to play with or I can cook something up on top of this PR and we can iterate on that, which might save a few cycles for both of us.

skrifa/Cargo.toml
simoncozens: Sorry - it shouldn't be enabled by default!

dfrg:
I got a bit nerd sniped and started #1856 that builds on this but makes the internal types public with some accessors. I think it's only missing segment/edge types at this point. PTAL

simoncozens:
So long as it works with https://github.com/simoncozens/tilvisan I'm happy...

dfrg:
Great, give me until tomorrow to finish this up then I'll send a PR for the changes to tilvisan and we can get this all landed.

simoncozens:
I'm happy with this if you are.

PR #1796, opened 2026-04-02
[read-fonts] a few small cff/type1 fixes

dfrg:
- expose the string index for CFF
- fail early when loading a CFF font with a private dict that is out of range
- expose Type1 glyph remapping info


read-fonts/src/ps/type1.rs
```diff
@@ -37,8 +37,8 @@ impl Type1Font {
         Self::from_dicts(raw_dicts.base, &raw_dicts.private)
     }
 
-    fn from_dicts(base: &[u8], private: &[u8]) -> Option<Self> {
-        let mut font = Type1Font {
+    fn empty() -> Self {
```
cmyr, line 41: do we also want `Default`?

dfrg: I waffled on whether to add it but an empty Type1Font is rather useless so I decided to keep the "default" constructor private.

PR #1797, opened 2026-04-02
Use custom brotli c bindings

wmedrano:
- The current `brotlic-sys` can't be imported into Chromium until the library exports a LICENSE file
  - Would happen after https://github.com/AronParker/brotlic/pull/5 is released
  - This PR unblocks importing into Chromium
- Changes
  - `sys.rs` contains bindings that are used in our shared-brotli-patch-decoder
  - `build.rs` downloads and compiles `brotli`. This won't be used in Chromium as we will have to re-implement it with Chromium's build system and brotli implementation

### DIFF
--- a/shared-brotli-patch-decoder/Cargo.toml
+++ b/shared-brotli-patch-decoder/Cargo.toml
@@ -13,14 +13,16 @@ all-features = true
 
 [features]
 default = ["c-brotli"]
-c-brotli = ["dep:brotlic-sys"]
+c-brotli = ["cc"]
 rust-brotli = ["dep:brotli-decompressor"]
 
 
 [dependencies]
-brotlic-sys = {version = "0.2.2", optional = true}
 brotli-decompressor = {version = "5.0.0", optional = true}
 cfg-if = "1.0.0"
+
+[build-dependencies]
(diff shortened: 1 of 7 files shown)


shared-brotli-patch-decoder/src/c_brotli.rs
```diff
@@ -1,11 +1,7 @@
 use crate::decode_error::DecodeError;
-use brotlic_sys::{
+use crate::sys::{
     BrotliDecoderAttachDictionary, BrotliDecoderCreateInstance, BrotliDecoderDecompressStream,
-    BrotliDecoderDestroyInstance, BrotliDecoderResult_BROTLI_DECODER_RESULT_ERROR,
```
wmedrano, line 4: Use the same API

shared-brotli-patch-decoder/Cargo.toml
```diff
@@ -13,14 +13,20 @@ all-features = true
 
 [features]
 default = ["c-brotli"]
-c-brotli = ["dep:brotlic-sys"]
```
wmedrano, line 16: Use git submodule

PR #1798, opened 2026-04-03
[Skera] Variations PR 1: Preliminaries and non-variation related

simoncozens:
This gathers together small additional fixes which I made as part of progress towards passing Harfbuzz integration suite. They don't relate to variations. The Cargo.toml dependencies will be used in subsequent variations PRs.

* Added support for subsetting extension lookup types in both GPOS (type 9) and GSUB (type 7) by introducing the `ExtensionSubtable` wrapper and helper functions to wrap subtables as extensions, ensuring correct serialization and lookup type assignment.
* Updated the logic for subsetting pair positioning subtables to correctly track and use the set of glyphs actually retained after subsetting, fixing a bug with coverage table generation.
* Added a `STRICT_HARFBUZZ_COMPATIBILITY` flag to control optimizations in coverage table subsetting, ensuring output can match HarfBuzz's behavior exactly when required.
* Fixed a bug in `find_duplicate_features` where the wrong set was checked for emptiness, preventing incorrect early returns.
* Introduced the `SUBSET_FLAGS_FORCE_LONG_LOCA` flag, allowing callers to force the use of long format for the 'loca' table even when short format would suffice, and updated the relevant logic in `glyf_loca.rs` to honor this flag
* Corrected handling of name record serialization to avoid panics if `pop_pack` returns `None`.
* Improved Unicode range calculation in the OS/2 table for codepoints outside the BMP.


simoncozens:
Integration test failure is a progression - test expectations are wrong.

qxliu76:
I don't know why I cannot add comment on dependency changes:  @wmedrano is working on reducing dependencies recently, I'd prefer not to add additional dependencies unless necessary(log, num-traits), for font-types: write-fonts depends on font-types so you can access font-types through wite_fonts::types

qxliu76:
It's really weird I cannot add comments: for commit [Better Harfbuzz compatibility while subsetting Coverage](https://github.com/googlefonts/fontations/pull/1798/changes/84e52da33dd07d1f184f0f7fe0354f8de359f5bd): Here we're doing a better job than HB by removing duplicate glyphs, I feel there's no need to add a switch just to match HB 100%. Or is it needed in any user case?

qxliu76:
It's annoying I got this error: Error Failed to save comment: An internal error occurred, please try again..I might wait for some time and come back on this.

qxliu76:
For commit: [[Not variations] Fix GDEF subsetting: Varstore offsets are 32 bit, not 16](https://github.com/googlefonts/fontations/pull/1798/changes/96e85d69ce637d929e4d72cdc4dff3f2ca5a0757), I believe this bug is fixed already in latest codebase

wmedrano:
> I don't know why I cannot add comment on dependency changes: @wmedrano is working on reducing dependencies recently, I'd prefer not to add additional dependencies unless necessary(log, num-traits), for font-types: write-fonts depends on font-types so you can access font-types through wite_fonts::types

The Chromium team would prefer to keep the bloat down. `log`, `num-traits`, `font-types` are already included, see https://source.chromium.org/chromium/chromium/src/+/main:third_party/rust/chromium_crates_io/Cargo.toml

`regex` is not yet included and is kind of heavier than you would think. Is there a way to avoid it? If its test only than having it as a `dev-dependency` suffices. I skimmed the PR and couldn't actually find where regex is used.

simoncozens:
> It's really weird I cannot add comments: for commit [Better Harfbuzz compatibility while subsetting Coverage](https://github.com/googlefonts/fontations/pull/1798/changes/84e52da33dd07d1f184f0f7fe0354f8de359f5bd): Here we're doing a better job than HB by removing duplicate glyphs, I feel there's no need to add a switch just to match HB 100%. Or is it needed in any user case? 

My thinking was that we want to match Harfbuzz temporarily because that allows us to directly use the HB integration tests on the variable subsetting. After we have parity with HB on the things we care about, then we can remove strict compatibility; that's why there's a switch for now.

qxliu76:
For commit: [[Not variations] Pass through extension lookups, subset their subtables](https://github.com/googlefonts/fontations/pull/1798/changes/70a85748555973e2b973b0e6f67d5f74178bfa39), we intentionally removed extension layer in skera, which doesn't change functionality and could reduce file size. In case of any overflows, we'd rely on the repacker to add extension promotion later. So there's no need to add an extension layer back.

qxliu76:
> > It's really weird I cannot add comments: for commit [Better Harfbuzz compatibility while subsetting Coverage](https://github.com/googlefonts/fontations/pull/1798/changes/84e52da33dd07d1f184f0f7fe0354f8de359f5bd): Here we're doing a better job than HB by removing duplicate glyphs, I feel there's no need to add a switch just to match HB 100%. Or is it needed in any user case?
> 
> My thinking was that we want to match Harfbuzz temporarily because that allows us to directly use the HB integration tests on the variable subsetting. After we have parity with HB on the things we care about, then we can remove strict compatibility; that's why there's a switch for now.

I think we do not want to add code that just makes tests passing. 
I do agree that integration tests might be a headache for now. Sorry I didn't find a good way to pass an argument into `cargo test test_name`. I plan to split generating expected test result and running tests so that might be easier to update expected result.

qxliu76:
I'm still unable to add comments. I'll try again in the afternoon.

qxliu76:
Good catch on [[not variations] Correctly compute PairPos1 coverage](https://github.com/googlefonts/fontations/pull/1798/changes/6b49ad86a11e62fb9b39fe943ce2dc675729fdf9) and [[Not variations] typo was breaking Khmer](https://github.com/googlefonts/fontations/pull/1798/changes/12fa2a1c0f23831591db3275f0de59bd71bc7637)! I have a couple of pending bug fixes as well. Considering recent code changes, would you mind if I combine these 2 commits with mine and make them into one PR? This way you don't need to rebase and update tests

simoncozens:
> I do agree that integration tests might be a headache for now. Sorry I didn't find a good way to pass an argument into cargo test test_name. 

Oh, I made that work in the final PR using libtest_mimic: https://github.com/googlefonts/fontations/blob/abe7d3b8225fcb8ceff03205c7d079c0181a8ea3/skera/tests/integration_test.rs#L832-L845

qxliu76:
[[Not variations] Fix for OS/2 codepage ranges: high codepoints fell through the cracks](https://github.com/googlefonts/fontations/pull/1798/changes/4a599a3805537910705bbd22c7f5ffec8c5e47d8): Thanks! This is one of my pending fix as well.

qxliu76:
This one I'm not sure: [[Not variations] pop_pack may fail when subsetting name table, don't die](https://github.com/googlefonts/fontations/pull/1798/changes/44355763634e1a615d87baa6fb4b9604969a1958), I think we would like to know why pop_pack() failed instead of bypassing the error

qxliu76:
[[not variations] Support FORCE_LONG_LOCA in head table](https://github.com/googlefonts/fontations/pull/1798/changes/2bc4498b8674273571f4023c8c79fd130b2158e1): you could add a force_long_loca boolean field in the plan,  I'd like to avoid adding a flag that doesn't exist in HB.

qxliu76:
[Support no-layout and ift-requirements test profiles](https://github.com/googlefonts/fontations/pull/1798/changes/1643277a3afa3786970628be416aa1d2ae1eb463): there was a big change in this file recently, `drop-tables` is already supported. And your code for iftb_requirements probably doesn't work now. I can help add iftb_requirements support for you.

I guess this is likely related with making tests passing, iftb_requirement flag is used by incremental transfer so I assume it's not critial for instancing

simoncozens:
> I guess this is likely related with making tests passing, iftb_requirement flag is used by incremental transfer so I assume it's not critial for instancing

Exactly; as I said at the top, the point of this PR is to establish a baseline with the Harfbuzz integration tests. With these changes, most of the remaining fails are VF related.

qxliu76:
> > I guess this is likely related with making tests passing, iftb_requirement flag is used by incremental transfer so I assume it's not critial for instancing
> 
> Exactly; as I said at the top, the point of this PR is to establish a baseline with the Harfbuzz integration tests. With these changes, most of the remaining fails are VF related.

Understood, thank you! It's probably not best practice for us to merge code that's just aiding tests

simoncozens:
> For commit: [[Not variations] Pass through extension lookups, subset their subtables](https://github.com/googlefonts/fontations/pull/1798/changes/70a85748555973e2b973b0e6f67d5f74178bfa39), we intentionally removed extension layer in skera, which doesn't change functionality and could reduce file size. In case of any overflows, we'd rely on the repacker to add extension promotion later. So there's no need to add an extension layer back.

I do not think this is correct behaviour. To subset a font which has extensions, you have to descend into those extensions and subset their subtables.

simoncozens:
OK, I've pulled out the good commits from this into separate PRs. Will move onto variations stuff next.

qxliu76:
> > For commit: [[Not variations] Pass through extension lookups, subset their subtables](https://github.com/googlefonts/fontations/pull/1798/changes/70a85748555973e2b973b0e6f67d5f74178bfa39), we intentionally removed extension layer in skera, which doesn't change functionality and could reduce file size. In case of any overflows, we'd rely on the repacker to add extension promotion later. So there's no need to add an extension layer back.
> 
> I do not think this is correct behaviour. To subset a font which has extensions, you have to descend into those extensions and subset their subtables.

skera does subset tables inside extension tables, we just removed extension table itself which is considered just a wrapper of a subtable. Extension type is handled in read-fonts that only inner subtables are returned: https://github.com/googlefonts/fontations/blob/b0c1eb2a2ed424e054fcb3e33ad1952070380ef6/read-fonts/src/tables/gpos.rs#L100, thus skera doesn't need to process extension type

qxliu76:
> OK, I've pulled out the good commits from this into separate PRs. Will move onto variations stuff next.

thank you!

PR #1799, opened 2026-04-03
[Skera] Variations PR 2: Basics

simoncozens:
This PR begins the variations support.

It adds the variations solver, ported from fontTools. Some parts of this are not currently used, and will be used later; so we should expect clippy lint warnings for this which will naturally go away as other PRs are merged.

We add the requested coordinate ranges to the plan, just as Harfbuzz does, add a parser for the command-line instancing spec, and apply this spec to the plan. We begin subsetting proper with two fairly simple tables, `fvar` and `avar`.

We also add instancing support to the integration tests. These integration tests are expected to fail at this stage, because (a) expectations don't expect us to subset fvar and avar, (b) subsetting `fvar` and `avar` without then subsetting other tables which use variations will necessarily cause trouble. We will fix it up in subsequent PRs.


qxliu76:
Thank you! A minor suggestion on this PR, if possible can we split it into 2:
PR 1: 
- include only centralized changes. e.g parse instancing spec, changes to the plan, solver, normalize axis values etc. 
- Tests: Nice to see those solve tests in HB are also ported:) I'd add unit tests for instancing spec parsing as well. HB example:https://github.com/harfbuzz/harfbuzz/blob/main/util/test-hb-subset-parsing.c 

PR2:  avar/fvar changes with unit tests added for each table, at this point, no integration test would succeed but we'd write unit tests for them (though we don't have a unit test for them in HB) and make sure tests are clean

Another thing is I plan to clean up all subsetter tests first before we merge any instancing code. Hopefully we can start work on this soon.

simoncozens:
>  unit tests added for each table, at this point, no integration test would succeed but we'd write unit tests for them (though we don't have a unit test for them in HB) and make sure tests are clean

Sorry, but this feels like you are asking me to do more work just for the sake of doing work. The HB integration tests pass with the full set of changes, and they test that we can subset fvar and avar. I don't see the point of writing another set of unit tests to test the exact same thing.

qxliu76:
I'm sorry to make you feel like that way, I know HB doesn't have unit tests, and since the final PR passes integration tests, adding them here might feel like extra busywork. It is completely natural that you feel protective of your time. 

My main concern is protecting your intermediate PRs. Since our integration tests only validate the final wired-up state, unit tests give us the confidence to merge these early PRs safely because we prove the building blocks work in isolation. This also saves us from interconnected debugging headaches at the finish line. 

The goal of unit test is not to test the exact the same thing. Unit tests are fast, stable, and easy to debug. We'll be able to know immediately if we introduced a bug. e.g later on when we add support for avar2, we know immediately if the new code breaks the avar1 behavior.

I actually added a lot of unit tests in skera that don't exist in HB. e.g: https://github.com/googlefonts/fontations/blob/7c9d875992c42d0bda6d0c7f807c25222e863490/skera/src/gsub/alternate_subst.rs#L112, 

If you'd like, I can help on that.

simoncozens:
> since the final PR passes integration tests, adding them here might feel like extra busywork.

Yes, exactly.

> My main concern is protecting your intermediate PRs. 

But the only reason we have intermediate PRs is because you asked me to break the final PR into intermediate PRs! You *are* asking me for extra busywork; I'll do it, because I want to see these PRs landed, but I would feel happier if you accepted that.

qxliu76:
I'd like to emphasize that unit tests are a fundamental and essential part of the software development life cycle. When an integration test fails but unit tests do not, it signals a specific issue within the interaction of multiple components. By ensuring unit tests pass first, we can more efficiently rule out basic errors and concentrate our investigation on these complex interactions.

Furthermore, it is concerning that your final PR reportedly passed all HB integration tests, as this should have been impossible **without** my latest subsetter bug fixes. This is highly suspicious and raises critical questions regarding how AI is running these integration tests. 

Thus to land your changes, we need to break down those large PRs into smaller, independent ones with corresponding tests. This approach ensures that code logic is verified at every step, reduces the risk of side effects, and ensures that all tests remain passing all the time.

qxliu76:
Hi Simon, I apologize if my previous message wasn't clear. Please know that my intention isn't to add extra busywork to your plate. I just wanted to clarify that unit tests are essential, and the scope of testing is determined entirely by the code changes themselves. This applies regardless of the PR's size, whether it's broken down into smaller pieces, or whether the code was written by hand or AI. We’ve been making solid progress improving our test coverage in Skera compared to HB, and I’d just like to ensure we maintain that standard.

simoncozens:
Hi Ginger. I completely agree that more unit testing is always a good thing. I apologise for my frustration above - I've already spend a lot of time working on this and I'm eager to start using this code, and I'm just generally an impatient person. Sorry for taking that out on you.

I've added a couple of unit tests for fvar and avar subspacing in [bed05b7](https://github.com/googlefonts/fontations/pull/1799/commits/bed05b78bb35137f51440e4895673863ec0153de) above. If you want to test more examples (I think probably I should add tests for the instance spec parsing...) I'd be happy to add them, but I am not sure when I will be able to get around to it; I have a number of other projects coming due this month and next.

As for other stuff, we should probably talk about it off-line; I'll try to make a Monday meeting.

simoncozens:
I've rebased this and added instancing spec tests. Please let me know if you need anything else for this PR.

skera/src/parsing_util.rs
```diff
@@ -1,9 +1,25 @@
...
+    read::collections::{int_set::Domain, IntSet},
+    types::{GlyphId, NameId, Tag},
+};
 
 use crate::SubsetError;
 
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct InstancingSpec {
+    pub pin_all_axes_to_default: bool,
+    pub axes: BTreeMap<Tag, AxisSpec>,
```
qxliu76, line 14: There's no need to use BTreeMap here, a FnvHashMap should be enough

skera/src/parsing_util.rs
```diff
@@ -118,6 +134,93 @@ pub fn parse_name_languages(input_str: &str) -> Result<IntSet<u16>, SubsetError>
...
+
+        if raw_axis == "*" {
+            if raw_value != "drop" {
+                return Err(SubsetError::InvalidId(raw_value.to_owned()));
+            }
+            spec.pin_all_axes_to_default = true;
+            continue;
+        }
+
+        let axis_tag = parse_axis_tag(raw_axis)?;
```
qxliu76, line 160: we should check and err out if this axis doesn't exist in fvar

skera/src/parsing_util.rs
```diff
@@ -118,6 +134,93 @@ pub fn parse_name_languages(input_str: &str) -> Result<IntSet<u16>, SubsetError>
...
+
+        let (min, def, max) = parse_axis_range_from_string(raw_value)?;
+        spec.axes
+            .insert(axis_tag, AxisSpec::Range { min, def, max });
+    }
+
+    Ok(spec)
+}
+
+fn parse_axis_tag(raw: &str) -> Result<Tag, SubsetError> {
```
qxliu76, line 174: no need to write a fn, we can just use https://github.com/googlefonts/fontations/blob/1fa9efef7cdfc3ef7b047191a4210e89e7f11135/font-types/src/tag.rs#L43

skera/src/parsing_util.rs
```diff
@@ -1,9 +1,25 @@
...
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct InstancingSpec {
+    pub pin_all_axes_to_default: bool,
+    pub axes: BTreeMap<Tag, AxisSpec>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum AxisSpec {
+    PinToDefault,
+    Range { min: f32, def: f32, max: f32 },
```
qxliu76, line 20: No need to use enum, a Triple should be enough.

qxliu76:
can you split this PR into smaller ones. First PR is just about instancing spec parsing and according tests.  Also we should check and err out when the user input contains an axis that doesn't exist in `fvar` table, a SubsetError type should be added accordingly for instancing parsing. Can you add code for these?

qxliu76:
Harfbuzz ref for different error types that we should handle: https://github.com/harfbuzz/harfbuzz/blob/511df88b82e697cd2a0f1f0635787aa0b18bddbb/util/helper-subset.hh#L90

skera/src/parsing_util.rs
```diff
@@ -326,3 +326,70 @@ fn test_parse_name_languages() {
     assert!(output.contains(2));
     assert!(output.contains(5));
 }
+
+#[test]
+fn test_parse_instancing_spec_tag_validation_and_wildcard() {
```
qxliu76, line 331: good to have additional tests, but we'd like to port all HB tests into skera: https://github.com/harfbuzz/harfbuzz/blob/main/util/test-hb-subset-parsing.c

skera/src/parsing_util.rs
```diff
@@ -326,3 +326,70 @@ fn test_parse_name_languages() {
...
+    let spec = parse_instancing_spec("wght=100:400:700").unwrap();
+    assert_eq!(spec.axes.len(), 1);
+    assert!(spec.axes.contains_key(&Tag::new(b"wght")));
```
qxliu76, line 334: besides checking for existence of an axis tag, we also need to check if the min/def/max values are parsed correctly

skera/src/parsing_util.rs
```diff
@@ -326,3 +326,70 @@ fn test_parse_name_languages() {
...
+    let spec = parse_instancing_spec("wght=100").unwrap();
+    assert!(matches!(
+        spec.axes.get(&Tag::new(b"wght")),
+        Some(AxisSpec::Range { min, def, max }) if *min == 100.0 && *def == 100.0 && *max == 100.0
+    ));
+
+    let spec = parse_instancing_spec("wght=100:700").unwrap();
+    assert!(matches!(
+        spec.axes.get(&Tag::new(b"wght")),
+        Some(AxisSpec::Range { min, def, max }) if *min == 100.0 && def.is_nan() && *max == 700.0
```
qxliu76, line 364: We should set def to axis default value during parsing instead of setting it to NaN

simoncozens: Here is what Harfbuzz does:

```
  if (count == 2)
  {
    *axis_min_value = values[0];
    *axis_def_value = NAN;
    *axis_max_value = values[1];
    return true;
  }
```

Would you like to do something different?

qxliu76: See here:https://github.com/harfbuzz/harfbuzz/blob/590a72b249c833571eb718abaa6e418aa2e9055c/util/helper-subset.hh#L100, HB also sets def to axis default value during parsing, and we'd like to do the same thing in skera

qxliu76: and here: https://github.com/harfbuzz/harfbuzz/blob/590a72b249c833571eb718abaa6e418aa2e9055c/util/helper-subset.hh#L86

simoncozens:
Thank you for this review. I can see you have a very clear idea of how the code should look and behave, and I respect that. I agree this will lead to better and more stable code in the future. However, I want to be mindful of both of our time here. Contributing to this project was intended to help *save* both of us time. But I think I've become convinced that landing any kind of change would actually take more time from both of us than would be required for you to simply make the changes you want yourself. There is a lot of code to land; this is the simplest PR, and others will be even more complex. I don't want to keep burdening you with more and more reviews which cause frustration for both of us. Let's continue to work on our own projects in our own ways.

qxliu76:
Thanks very much for all the time and effort you put into this, I really appreciate you trying to help us out.
Yeah, I totally agree that getting AI-generated large code changes up to the testing and quality bar is a huge headache and requires a ton of time investment. It makes total sense to close this if it's taking up too much of your time. 
Thanks again for giving it a shot, and all the best with your own projects.

simoncozens:
> Yeah, I totally agree that getting AI-generated large code changes up to the testing and quality bar is a huge headache

I didn't say that. This was my code.

PR #1812, opened 2026-04-07
[write] Abstract out TVS construction, write cvar

simoncozens:
- **Move common TVS construction routines from gvar into variations**
- **Support writing cvar tables**

### DIFF
--- a/write-fonts/src/tables/cvar.rs
+++ b/write-fonts/src/tables/cvar.rs
@@ -7,6 +7,8 @@
 use read_fonts::TopLevelTable;
 use types::{MajorMinor, Offset16, Tag};
 
+include!("../../generated/generated_cvar.rs");
+
 use crate::{
     table_type::TableType,
     tables::variations::{Deltas, TupleVariationStoreInputError},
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/variations.rs
```diff
@@ -469,6 +628,59 @@ where
     }
 }
 
+/// An error representing invalid input when building a tuple variation store
+#[derive(Clone, Debug)]
+pub enum TupleVariationStoreInputError<T: std::fmt::Display> {
```
cmyr: I don't think the trait bound isn't necessary here?

write-fonts/src/tables/cvar.rs
```diff
@@ -0,0 +1,338 @@
...
+
+use super::variations::{
+    compute_shared_points, compute_tuple_variation_count, compute_tuple_variation_data_offset,
+    PackedDeltas, PackedPointNumbers, Tent, Tuple, TupleVariationCount, TupleVariationHeader,
+};
+
+/// Writer-side representation of the `cvar` table.
+#[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Cvar {
```
cmyr: I appreciate much of the compilation logic would need to be manually implemented, but my preference would still be that we use codegen to make the actual type? It's nice to try and have as much as possible be in codegen, since it means less manual updating in the future if we change anything there.

cmyr (approved):
looks broadly reasonable, with a couple little notes inline.

PR #1826, opened 2026-04-10
[write-fonts] Fix double-count bug in promotion logic

### LINKED ISSUE #1938: Unnecessary Extension promotion in NotoSansGrantha GPOS: normalize away?
I don't know if this is a one-off or more widespread, but I noticed while looking at #1046 that in `NotoSansGrantha.glyphs` ([notofonts/grantha) fontc decides to wrap a GPOS ChainContextPos lookup (index 103) inside ExtensionPos (type 9) whereas fontmake/fonttools uses the inner lookup type directly. The inner subtable content is identical at the TTX level, the only structural difference is the Extension wrapper itself.

This appears to be a borderline case where minor encoding size differences elsewhere in the GPOS table tip fontc's promotion decision differently from fonttools.

The difference is of course semantically a no-op, as Extension promotion doesn't change behavior. But it shows up as a diff in ttx_diff output, adding noise when triaging other real differences. The GPOS table size difference is also insignificant (20 bytes).

I have a local branch in which I experimented with adding an `--unwrap_extensions` flag to ttx_diff.py to optionally strip Extension wrappers before comparing, which makes the GPOS table compare identical for this font.

I haven't PR'ed this yet because I am not 100% sure whether ttx-diff should always ignore differences like this, or if doing this unconditionally would mask legitimate diffs uncovering potential bugs in either compiler's extension promotion.

cmyr:
In our select_promotions logic we were calculating lookup sizes that were frequently greater (and never less than) those produced by harfbuzz. This was caused by us double-counting subtables that were referenced by multiple parents, in some cases.

With this patch, NotoSansGrantha's GPOS table now matches on ttx_diff.

I'm curious to see how this interacts with #1823, which seems like an unrelated patch but which addresses the same symptom...

fixes https://github.com/googlefonts/fontc/issues/1938

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -1088,7 +1088,9 @@ impl Graph {
         visited.clear();
         while !queue.is_empty() {
             let next = queue.pop_front().unwrap();
-            visited.insert(next);
+            if !visited.insert(next) {
+                continue;
+            }
             let obj = self.objects.get(&next).unwrap();
             size += obj.bytes.len();
             queue.extend(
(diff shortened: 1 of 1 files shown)


anthrotype (approved):
good catch! LGTM

Regarding interaction with my #1823, these are fully orthogonal fixes though both address the same symptom (excessive Extension promotion).

Mine fixes how nodes are sorted (Dijkstra ordering), and reduces Extension promotion by avoiding some overflows in the first place, whereas yours fixes how lookup sizes are estimated for the promotion decision, preventing double-counting shared subtables.

PR #1827, opened 2026-04-10
[write-fonts] use min-heap for Dijkstra

anthrotype:
Colin asked spin this one off from #1823, with a failing reproducer demonstrating the bug.

It will be followed by the acutal fix after CI goes red.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -546,11 +546,14 @@ impl Graph {
         self.nodes
             .values_mut()
             .for_each(|node| node.distance = u32::MAX);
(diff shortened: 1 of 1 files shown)


anthrotype:
repro https://github.com/googlefonts/fontations/actions/runs/24255878973/job/70827200694?pr=1827#step:8:450

anthrotype:
this targets main and should be merged first, #1828 is based off the tip of this branch and can be merged second, github should be smarter to point it at main

cmyr (approved):
excellent, ty!

I want these separate so I can look at their independent impact on crater :)

PR #1828, opened 2026-04-10
[write-fonts] Add offset-width penalty to update_distances

anthrotype:
second spin off from https://github.com/googlefonts/fontations/pull/1823, with a failing reproducer to be followed by the actual fix.

The failing test demonstrates that when `update_distances()` uses only `child_size` as the Dijkstra edge weight, ignoring offset width, this produces a depth-first ordering where a small deep node sorts before a larger shallow sibling.

HarfBuzz and skera add  `+ (1 << (link_width * 8))` to the edge weight, producing breadth-first ordering that keeps 16-bit-linked children close to their parents.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -106,7 +106,7 @@ pub struct Graph {
 #[derive(Debug)]
 struct Node {
     size: u32,
-    distance: u32,
+    distance: u64,
     /// overall position after sorting
     position: u32,
     space: Space,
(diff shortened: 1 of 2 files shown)


write-fonts/src/graph.rs
```diff
@@ -1383,6 +1399,49 @@ mod tests {
...
+    // Without penalty: A.dist=10, A1.dist=15, B.dist=100 → A1 before B
+    // With penalty:    A.dist=65546, A1.dist=131087, B.dist=65636 → B before A1
+    #[test]
+    fn offset_width_penalty_breadth_first() {
+        let [root, a, a1, b] = make_ids::<4>();
+        let sizes = [10, 10, 5, 100];
+        let mut graph = TestGraphBuilder::new([root, a, a1, b], sizes)
+            .add_link(root, a, OffsetLen::Offset16)
+            .add_link(root, b, OffsetLen::Offset16)
+            .add_link(a, a1, OffsetLen::Offset16)
```
cmyr, line 1428: is this correct? All of the offsets are the same width?

cmyr (commented):
one qq inline..

write-fonts/src/graph.rs
anthrotype: I think so yes.

I tried a mixed-offset test but it doesn't work because the `assign_space_0` already separates Offset32-reachable nodes into a different space. So the penalty's effect is within a single space.

Do you want me to add a comment or is this good?

cmyr (approved):
okay yes makes sense, I'd misunderstood. The real logic here is in adding a distance penalty that accounts for the offset itself (as explained in the test text)

PR #1829, opened 2026-04-10
[subset] update tests and minor bug fixes

qxliu76:
updated testing code and expected tests(instancing options are ignored)
bug fixes: 
1. offset to VarStore should be  32bit
2. collect GPOS variation idxes when GDEF varstore is non-empty after subsetting
3. fix delta set idx mapping, keep idxes that map to NO_VAR_INDEX

features:
1. added support for no-layout-closure flag
2. pass through some known tables and drop unknown tables


skera/src/gsub.rs
```diff
@@ -97,13 +97,17 @@ impl LayoutClosure for Gsub<'_> {
...
-        let Ok(_) = self.closure_glyphs(&lookup_indices, &mut plan.glyphset_gsub) else {
-            return;
-        };
+        if !plan
+            .subset_flags
+            .contains(SubsetFlags::SUBSET_FLAGS_NO_LAYOUT_CLOSURE)
+        {
+            let Ok(_) = self.closure_glyphs(&lookup_indices, &mut plan.glyphset_gsub) else {
+                return;
+            };
```
cmyr: instead of using a  let/else with no binding this would probably be slightly clearer as just an if statement?

```suggestion
            if self.closure_glyphs(&lookup_indices, &mut plan.glyphset_gsub).is_err() {
                return;
            };
```

skera/src/lib.rs
```diff
@@ -190,6 +194,30 @@ pub static DEFAULT_LAYOUT_FEATURES: &[Tag] = &[
     Tag::new(b"blwm"),
 ];
 
+//ref: <https://github.com/harfbuzz/harfbuzz/blob/5cbce05a7cfb5b3423e32feb19341730fd5ef363/src/hb-subset-input.cc#L46>
```
cmyr: I prefer to shorten these so that they don't get split across multiple lines in my editor, which makes them hard to click on (I acknowledge this is a me problem)

```suggestion
//ref: <https://github.com/harfbuzz/harfbuzz/blob/5cbce05a7c/src/hb-subset-input.cc#L46>
```

skera/src/lib.rs
```diff
@@ -232,13 +260,16 @@ impl SubsetFlags {
...
     //If set don't perform glyph closure on layout substitution rules (GSUB)
-    //This flag is UNIMPLEMENTED yet
     pub const SUBSET_FLAGS_NO_LAYOUT_CLOSURE: Self = Self(0x0200);
 
     //If set perform IUP delta optimization on the remaining gvar table's deltas.
     //This flag is UNIMPLEMENTED yet
     pub const SUBSET_FLAGS_OPTIMIZE_IUP_DELTAS: Self = Self(0x0400);
 
+    //If set do not pull mirrored versions of input codepoints into the subset.
+    //This flag is UNIMPLEMENTED yet
```
cmyr: appreciate this is matching what is already here, but these would all make sense as doc comments instead of normal comments.

skera/src/lib.rs
```diff
@@ -865,16 +899,21 @@ fn remap_delta_set_indices(
...
 
-        let Some((new_var_idx, delta)) = varidx_delta_map.get(var_idx) else {
-            continue;
+        let (new_var_idx, delta) = if var_idx == NO_VARIATION_INDEX {
+            (var_idx, 0)
+        } else {
+            let Some((new_idx, d)) = varidx_delta_map.get(&var_idx) else {
+                continue;
+            };
+            (*new_idx, *d)
```
cmyr, line 909: not blocking, but you could reduce the nesting and make this a bit more linear:

```suggestion
        let (new_var_idx, delta) = if var_idx == NO_VARIATION_INDEX {
            (var_idx, 0)
        } else if let Some((new_idx, d)) = varidx_delta_map.get(&var_idx) {
            (*new_idx, *d)
        } else {
            continue;
        };
```

skera/src/lib.rs
```diff
@@ -1326,7 +1370,18 @@ fn subset_table<'a>(
...
+
+        _ => {
+            if plan
+                .subset_flags
+                .contains(SubsetFlags::SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED)
+            {
+                passthrough_table(tag, font, s)
+            } else {
+                Ok(())
+            }
```
cmyr: also just a formatting preference, but you can simplify slightly by having the condition be inline of the match arm, like:


```suggestion
        _ if plan
            .subset_flags
            .contains(SubsetFlags::SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED) =>
        {
            passthrough_table(tag, font, s)
        }
        _ => Ok(()),
```

cmyr (approved):
a few little syntax ideas but nothing blocking!

garretrieger (approved):

PR #1830, opened 2026-04-10
[codegen] Generate default impls for tables

cmyr:
This generates impls of the Default trait for many font tables.

A table is skipped if it has a 'format' field with an expected  value other than `0` or `1`; in this case it is not expected that any API we generate parses this table directly, and it should only ever be read through a 'format group' enum.

If a table has format 1, we ensure that the first two bytes of the provided value would represent the value '1' as either a u8 or u16, depending on the format of the field. Otherwise the values are all zeros.

### DIFF
--- a/font-codegen/src/format_group.rs
+++ b/font-codegen/src/format_group.rs
@@ -85,6 +85,11 @@ pub(crate) fn generate(item: &TableFormat, items: &Items) -> syn::Result<TokenSt
         min_byte_arms.push(quote!(Self::#var_name(item) => item.min_byte_range(), ));
         min_table_byte_arms.push(quote!(Self::#var_name(item) => item.min_table_bytes(), ));
     }
+    let first_variant = item.variants.first().expect("format group needs variants");
+    // this attribute is currently used only once and we expect it to only ever
+    // be the last variant, but let's be defensive
+    assert!(first_variant.attrs.write_only.is_none(), "sanity check");
+    let first_var_name = &first_variant.name;
 
     Ok(quote! {
         #( #docs )*
@@ -92,6 +97,13 @@ pub(crate) fn generate(item: &TableFormat, items: &Items) -> syn::Result<TokenSt
         pub enum #name<'a> {
             #( #variants ),*
         }
+
+        impl Default for #name<'_> {
+            fn default() -> Self {
+                Self::#first_var_name(Default::default())
+            }
+        }
+
 
         #getters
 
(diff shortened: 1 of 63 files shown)


behdad:
Lgtm.

font-codegen/src/format_group.rs
```diff
@@ -85,6 +85,8 @@ pub(crate) fn generate(item: &TableFormat, items: &Items) -> syn::Result<TokenSt
         min_byte_arms.push(quote!(Self::#var_name(item) => item.min_byte_range(), ));
         min_table_byte_arms.push(quote!(Self::#var_name(item) => item.min_table_bytes(), ));
     }
+    let first_variant = item.variants.iter().find(|v| v.attrs.write_only.is_none());
```
rsheeter: the rationale for this feels a little obtuse, perhaps worthy of a comment explaining _why_ this makes sense?

font-codegen/src/parsing.rs
```diff
@@ -52,6 +52,21 @@ impl Table {
...
+        let fld = self.fields.iter().find(|fld| fld.attrs.format.is_some())?;
+        let format = fld.attrs.format.as_ref().unwrap().base10_parse().ok()?;
+        let fld_tokens = fld.typ.cooked_type_tokens();
+        let width = if fld_tokens == "u8" {
+            1
+        } else if fld_tokens == "u16" {
+            2
+        } else {
+            panic!("only expect format fields to be u8 or u16");
+        };
```
rsheeter, line 67: maybe this is too broad for the current PR but would love it if cooked type (still find that terminology strange, I translate it to roughly resolved in my head) was an enum here and it knew it's own size.

read-fonts/generated/generated_ankr.rs
```diff
@@ -95,6 +95,16 @@ impl<'a> Ankr<'a> {
     }
 }
 
+const _: () = assert!(Ankr::MIN_SIZE <= NULL_POOL_SIZE);
```
wmedrano: These checks seem a bit arbitrary. How about:

`const _: () = assert!(FontData::supports_default_with_size(Ankr::MIN_SIZE))`

cmyr: agreed, I don't need to be following HB blindly here and can definitely make this a bit more legible

font-codegen/src/parsing.rs
cmyr: There's a nice idea in there: we could definitely change the `Scalar` type to just be an enum of the scalars we actually support, that would be a nice little improvement? can take a look at that afterwards. (https://github.com/googlefonts/fontations/issues/1890)

wmedrano (approved):

PR #1831, opened 2026-04-12
[read-fonts] t1: be less strict when parsing metrics

dfrg:
FT parses these as integers but simply _ignores_ the possible fractional bits that may be present in some files. We now do the same instead of failing to load the entire font.


cmyr (approved):
makes sense!

PR #1837, opened 2026-04-14
Fix build for docs.rs

wmedrano:
Fix build for docs.rs

- shared-brotli-patch-decoder fails to build on docs.rs since there is no compiler tooling.
- We skip compiling brotli. It is fine to not link in the symbols to generate docs.


wmedrano:
Failed build log: https://docs.rs/crate/shared-brotli-patch-decoder/latest/builds

dfrg (approved):

PR #1838, opened 2026-04-14
[CONCEPT] Make kurbo optional

wmedrano:


rsheeter:
To ask the obvious: Why?

jqwv:
I think it would be good to make it optional. Not everyone wants to use kurbo. I have my own Curve and Path format, before calling `from_bezpath` I need to convert all Paths to kurbo paths. Which is rather unexpected.

A good solution can be seen in `ttf_parser`: instead of requiring one specific external data type, the caller is expected to implement a Trait with methods like `move_to`, and so on.

It would be good to have an equivalent for `from_bezpath` before making it behind a feature, otherwise importing kurbo is of course still required to create the paths.

wmedrano:
Closing since I reused this branch for another mechanism of reducing deps

PR #1839, opened 2026-04-15
[Skrifa] Add bitmap mask decoding for EBDT

P-DOX:
**Description:**
Adds a decode method ( _**MaskData::decode(width, height) -> Option<Vec<u8>>**_) to MaskData that unpacks raw packed bitmap data into a flat buffer of 8-bit-per-pixel values, following the OpenType EBDT (Embedded Bitmap Data Table) bitmap data layout rules ([OpenType EBDT spec](https://learn.microsoft.com/en-us/typography/opentype/spec/ebdt)).

It handles both layouts defined by OpenType bitmap formats:

- Byte-aligned (non-packed): each row padded to a byte boundary

- Bit-aligned (packed): pixels tightly packed across rows with no padding

 **Goal**:  Add non-PNG CBDT/CBLC bitmap support in Chromium via SkTypeface_Fontations (PNG already supported). Skrifa already has full CBLC/CBDT table parsing and glyph lookup; this PR adds the missing bitmap pixel decoding. 

Skia WIP CL : [Support CBDT/CBLC in Chromium](https://skia-review.googlesource.com/c/skia/+/1155916)

**Testing:** 8 unit tests covering all bpp × packing mode combinations, each constructing a MaskData with hand-crafted bit patterns and verifying exact decoded output.

### DIFF
--- a/skrifa/src/bitmap.rs
+++ b/skrifa/src/bitmap.rs
@@ -1,5 +1,7 @@
 //! Bitmap strikes and glyphs.
 
+use alloc::vec::Vec;
+
 use super::{instance::Size, metrics::GlyphMetrics, MetadataProvider};
 use crate::prelude::LocationRef;
 use raw::{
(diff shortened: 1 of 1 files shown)


skrifa/src/bitmap.rs
```diff
@@ -370,6 +372,85 @@ pub struct MaskData<'a> {
...
 
+impl MaskData<'_> {
+    /// Decodes the raw packed bitmap data into 8-bit-per-pixel values.
+    ///
+    /// Returns a `Vec<u8>` of `width * height` bytes, with each pixel
+    /// value scaled to the 0–255 range.
+    pub fn decode(&self, width: u32, height: u32) -> Option<Vec<u8>> {
+        let w = width as usize;
+        let h = height as usize;
+        let total_pixels = w.checked_mul(h)?;
```
dfrg, line 383: Since we must pass the width and height (perhaps I could have better designed this type), it would be nice to do some validation here to ensure that `self.data.len()` matches the expected size for the given dimensions, packing and bpp. That way we can return `None` rather than panicking while accessing the data below.

In fact, let's change this to return a `Result` with a `MaskDataDecodeError` error type. Perhaps just an enum with variants for `SizeOverflow` and `InvalidDimensions`.

skrifa/src/bitmap.rs
```diff
@@ -370,6 +372,85 @@ pub struct MaskData<'a> {
...
+        if total_pixels == 0 {
+            return None;
+        }
```
dfrg, line 401: If we change the method to return a `Result`, it should be fine to return `Ok(Vec::new())` here.

dfrg (commented):
The logic LGTM and the detailed explanations in the tests are fantastic. I requested a couple changes to improve validation but there are no major blockers.

We do typically pair methods like this that allocate with an additional method that outputs to a user provided buffer. In this case, something like `decode_to_slice(&self, width: u32, height: u32, buf: &mut [u8]) -> Result<(), MaskDataDecodeError>` and then the allocating function simply calls into that one. Entirely up to you whether you'd like to add that to this PR. If not, I'll do it as a quick follow up.

Happy to see this, thanks!

skrifa/src/bitmap.rs
P-DOX: Thanks for your comment. 
That makes sense as well to have `MaskDataDecodeError`. Updated with the changes.

skrifa/src/bitmap.rs
P-DOX: Done, in case of `total_pixels == 0`, just returning `Ok(())` and not touching the buffer.

P-DOX:
Thanks @dfrg for reviewing the PR.
Could you please also help with enabling the `Workflow Security Scan`? It’s currently awaiting admin approval.

dfrg (approved):
This all looks good now, thanks!

PR #1840, opened 2026-04-15
[write] Better handling of zero value deltas

cmyr:
This addresses an obscure issue where we would fail to correctly merge delta sets were equivalent, but where one had an extra zero (such as from an extra intermediate layer.)


anthrotype:
before I review this, I wonder does this affect any specific font on fontc crater?

cmyr:
yes:

```
python3 -m ttx_diff 'https://github.com/notofonts/latin-greek-cyrillic?c4a321e123#sources/NotoSerif.glyphspackage'
```

anthrotype (approved):

PR #1842, opened 2026-04-17
[subset] add null checks to avoid err when an null offset is encountered

qxliu76:
add checks so that we do not emit ReadError in case of null offsets.
currently we see below repeated code patterns in a lot of places because we have nullable offset arrays and non-nullable offset arrays, once sanitize changes landed, I'll try to make a generic fn or trait for them. 
```
            match offset::serialize_subset() {
                Ok(()) => rule_set_count += 1,
                Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY) => (),
                Err(e) => return Err(e),
            }
```

I'll also remove those `iter_as_nullable()` methods once sanitize changes landed

### DIFF
--- a/read-fonts/src/offset_array.rs
+++ b/read-fonts/src/offset_array.rs
@@ -85,6 +85,18 @@ where
                 .map(|off| off.get().resolve_with_args(data, &args))
         })
     }
+
+    /// Iterate over all of the offset targets.
+    ///
+    /// Offset is treated as nullable and each offset will be resolved as it is encountered.
+    pub(crate) fn iter_as_nullable(
+        &self,
+    ) -> impl Iterator<Item = Option<Result<T, ReadError>>> + 'a {
+        self.iter().map(|off| match off {
+            Err(ReadError::NullOffset) => None,
+            other => Some(other),
+        })
+    }
 }
 
 impl<'a, T, O> ArrayOfNullableOffsets<'a, T, O>
--- a/read-fonts/src/tables/colr/closure.rs
+++ b/read-fonts/src/tables/colr/closure.rs
@@ -1,5 +1,4 @@
 //! computing closure for the colr table
-
 use font_types::{GlyphId, GlyphId16};
 
 use crate::{collections::IntSet, tables::variations::NO_VARIATION_INDEX, ResolveOffset};
@@ -27,6 +26,10 @@ impl Colr<'_> {
         let Some(Ok(records)) = self.base_glyph_records() else {
             return;
         };
+
+        let Some(Ok(layers)) = self.layer_records() else {
+            return;
+        };
         for glyph_id in glyph_set.iter() {
             let Ok(glyph_id) = glyph_id.try_into() else {
                 continue;
@@ -38,8 +41,8 @@ impl Colr<'_> {
             let start = record.first_layer_index() as usize;
             let end = start + record.num_layers() as usize;
             for layer_index in start..end {
-                if let Ok((_gid, palette_id)) = self.v0_layer(layer_index) {
-                    palette_indices.insert(palette_id);
+                if let Some(layer) = layers.get(layer_index) {
+                    palette_indices.insert(layer.palette_index());
                 }
             }
         }
@@ -62,16 +65,15 @@ impl Colr<'_> {
         if let Some(Ok(base_glyph_list)) = self.base_glyph_list() {
             let base_glyph_records = base_glyph_list.base_glyph_paint_records();
             let offset_data = base_glyph_list.offset_data();
-            for paint_record in base_glyph_records {
-                let gid = paint_record.glyph_id();
+            for record in base_glyph_records {
+                let gid = record.glyph_id();
                 if !glyph_set.contains(GlyphId::from(gid)) {
                     continue;
                 }
-                if let Ok(paint) = paint_record.paint(offset_data) {
+                if let Ok(paint) = record.paint(offset_data) {
                     c.dispatch(&paint);
                 }
             }
-
             glyph_set.union(&c.glyph_set);
         }
 
@@ -93,6 +95,9 @@ impl Colr<'_> {
         let Some(Ok(records)) = self.base_glyph_records() else {
             return;
         };
+        let Some(Ok(layers)) = self.layer_records() else {
+            return;
+        };
         for glyph_id in glyph_set.iter() {
             let Ok(glyph_id) = glyph_id.try_into() else {
                 continue;
@@ -104,8 +109,8 @@ impl Colr<'_> {
             let start = record.first_layer_index() as usize;
             let end = start + record.num_layers() as usize;
             for layer_index in start..end {
-                if let Ok((gid, _palette_id)) = self.v0_layer(layer_index) {
-                    glyphset_colrv0.insert(GlyphId::from(gid));
+                if let Some(layer) = layers.get(layer_index) {
+                    glyphset_colrv0.insert(GlyphId::from(layer.glyph_id()));
                 }
             }
         }
@@ -166,9 +171,8 @@ impl<'a> Colrv1ClosureContext<'a> {
         false
     }
 
-    fn add_layer_indices(&mut self, first_layer_index: u32, last_layer_index: u32) {
-        self.layer_indices
-            .insert_range(first_layer_index..=last_layer_index);
+    fn add_layer_index(&mut self, layer_index: u32) {
+        self.layer_indices.insert(layer_index);
     }
 
     fn add_palette_index(&mut self, palette_index: u16) {
@@ -270,14 +274,16 @@ impl PaintColrLayers<'_> {
         };
         let first_layer_index = self.first_layer_index();
         let last_layer_index = first_layer_index + num_layers as u32 - 1;
-        c.add_layer_indices(first_layer_index, last_layer_index);
 
         let offset_data = layer_list.offset_data();
+        let paint_offsets = layer_list.paint_offsets();
         for layer_index in first_layer_index..=last_layer_index {
-            if let Some(paint_offset) = layer_list.paint_offsets().get(layer_index as usize) {
-                if let Ok(paint) = paint_offset.get().resolve::<Paint>(offset_data) {
-                    c.dispatch(&paint);
-                }
+            let Some(paint_offset) = paint_offsets.get(layer_index as usize) else {
+                break;
+            };
+            if let Ok(paint) = paint_offset.get().resolve::<Paint>(offset_data) {
+                c.add_layer_index(layer_index);
+                c.dispatch(&paint);
             }
         }
     }
@@ -308,8 +314,8 @@ impl PaintVarLinearGradient<'_> {
     fn v1_closure(&self, c: &mut Colrv1ClosureContext) {
         if let Ok(var_colorline) = self.color_line() {
             var_colorline.v1_closure(c);
+            c.add_variation_indices(self.var_index_base(), 6);
         }
-        c.add_variation_indices(self.var_index_base(), 6);
     }
 }
 
@@ -325,8 +331,8 @@ impl PaintVarRadialGradient<'_> {
     fn v1_closure(&self, c: &mut Colrv1ClosureContext) {
         if let Ok(var_colorline) = self.color_line() {
             var_colorline.v1_closure(c);
+            c.add_variation_indices(self.var_index_base(), 6);
         }
-        c.add_variation_indices(self.var_index_base(), 6);
     }
 }
 
@@ -342,15 +348,15 @@ impl PaintVarSweepGradient<'_> {
     fn v1_closure(&self, c: &mut Colrv1ClosureContext) {
         if let Ok(var_colorline) = self.color_line() {
             var_colorline.v1_closure(c);
+            c.add_variation_indices(self.var_index_base(), 4);
         }
-        c.add_variation_indices(self.var_index_base(), 4);
     }
 }
 
 impl PaintGlyph<'_> {
     fn v1_closure(&self, c: &mut Colrv1ClosureContext) {
-        c.add_glyph_id(self.glyph_id());
         if let Ok(paint) = self.paint() {
+            c.add_glyph_id(self.glyph_id());
             c.dispatch(&paint);
         }
     }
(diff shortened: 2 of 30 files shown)


read-fonts/src/tables/gpos.rs
```diff
@@ -140,6 +140,97 @@ impl<'a> PositionLookup<'a> {
             other => Err(ReadError::InvalidFormat(other as _)),
         }
     }
+
+    /// Return the subtables(nullable) for this lookup.
+    ///
+    /// This method handles both extension and non-extension lookups, and saves
+    /// the caller needing to dig into the `PositionLookup` enum itself.
+    pub fn subtables_nullable(&self) -> Result<Option<PositionSubtables<'a>>, ReadError> {
```
cmyr: So I think the fact that the current `subtables` method will return an error if there is an extension lookup with no subtables is just incorrect, and should be fixed.

I've put up a patch here: https://github.com/googlefonts/fontations/pull/1849

This PR fixes that issue, and also includes your logic for continuing to look at subtables beyond the first, in order to determine the lookup type.

I believe that with that patch merged, we don't need to add this new method. Want to take a look and see if anything is missing? The one difference is that when you iterate the subtables, you will have to manually check the return type and handle the null offset error as needed.

read-fonts/src/tables/colr/closure.rs
```diff
@@ -62,16 +65,15 @@ impl Colr<'_> {
...
-            for paint_record in base_glyph_records {
-                let gid = paint_record.glyph_id();
-                if !glyph_set.contains(GlyphId::from(gid)) {
+            for record in base_glyph_records {
+                let gid = record.glyph_id();
+                if !glyph_set.contains(GlyphId::from(gid)) || record.paint_offset().is_null() {
                     continue;
                 }
-                if let Ok(paint) = paint_record.paint(offset_data) {
+                if let Ok(paint) = record.paint(offset_data) {
```
cmyr: maybe i'm confused but this looks like it isn't actually changing any logic? if the offset is null then `if let Ok(paint)` will not match, and we won't call `dispatch`, which means we'll continue to the next record?

read-fonts/src/tables/gpos/closure.rs
```diff
@@ -129,12 +148,18 @@ impl Intersect for SinglePos<'_> {
 
 impl Intersect for SinglePosFormat1<'_> {
     fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
+        if self.coverage_offset().is_null() {
+            return Ok(false);
+        }
```
cmyr, line 149: we could have a little helper macro for a lot of these; you would invoke it with the offset getter method and then the return value in the null case, like:


```rust
// declared once in lib.rs or something:
macro_rules! bail_if_null {
    ($getter:expr, $null_ret:expr) => {
        match $getter {
            Ok(thing) => thing,
            Err(ReadError::NullOffset) => return Ok($null_ret),
            Err(e) => return Err(e),
        }
    };
}
```

and then this fn would become,

```rust
    fn intersects(&self, glyph_set: &IntSet<GlyphId>) -> Result<bool, ReadError> {
        let coverage = bail_if_null!(self.coverage(), false);
        Ok(coverage.intersects(glyph_set))
    }
```

read-fonts/src/tables/layout/closure.rs
```diff
@@ -475,10 +502,22 @@ pub(crate) enum Format1Rule<'a> {
...
+                    Some(table.coverage())
+                }
+            }
+            ContextFormat1::Chain(table) => {
+                if table.coverage_offset().is_null() {
+                    None
+                } else {
+                    Some(table.coverage())
+                }
+            }
```
cmyr:
```suggestion
            ContextFormat1::Plain(table) if !table.coverage_offset().is_null() => Some(table.coverage()),
            ContextFormat1::Chain(table) if !table.coverage_offset().is_null() => Some(table.coverage()),
            _ => None,
```

read-fonts/src/tables/layout/closure.rs
```diff
@@ -604,17 +643,41 @@ pub(crate) enum Format2Rule<'a> {
 }
 
 impl ContextFormat2<'_> {
-    pub(crate) fn coverage(&self) -> Result<CoverageTable<'_>, ReadError> {
+    pub(crate) fn coverage(&self) -> Option<Result<CoverageTable<'_>, ReadError>> {
         match self {
-            ContextFormat2::Plain(table) => table.coverage(),
-            ContextFormat2::Chain(table) => table.coverage(),
+            ContextFormat2::Plain(table) => {
```
cmyr: see above; you can inline the null check and simplify this significantly.

read-fonts/src/tables/layout/closure.rs
```diff
@@ -604,17 +643,41 @@ pub(crate) enum Format2Rule<'a> {
...
+            }
         }
     }
 
-    pub(crate) fn input_class_def(&self) -> Result<ClassDef<'_>, ReadError> {
+    pub(crate) fn input_class_def(&self) -> Option<Result<ClassDef<'_>, ReadError>> {
         match self {
-            ContextFormat2::Plain(table_ref) => table_ref.class_def(),
-            ContextFormat2::Chain(table_ref) => table_ref.input_class_def(),
+            ContextFormat2::Plain(table_ref) => {
```
cmyr: ditto

read-fonts/src/tables/gsub.rs
```diff
@@ -121,4 +121,95 @@ impl<'a> SubstitutionLookup<'a> {
             other => Err(ReadError::InvalidFormat(other as _)),
         }
     }
+
+    /// Return the subtables(nullable) for this lookup.
+    ///
+    /// This method handles both extension and non-extension lookups, and saves
+    /// the caller needing to dig into the `SubstitutionLookup` enum itself.
+    pub fn subtables_nullable(&self) -> Result<Option<SubstitutionSubtables<'a>>, ReadError> {
```
cmyr: ditto, with the patch I linked above I believe we can avoid adding this.

read-fonts/src/tables/layout.rs
```diff
@@ -103,6 +103,25 @@ impl<'a, T: FontRead<'a> + 'a, Ext: ExtensionLookup<'a, T> + 'a> Subtables<'a, T
...
+                    Some(Err(e)) => Some(Err(e)),
+                    Some(Ok(ext)) => Some(ext.extension()),
+                })),
+            ),
+        };
+
+        left.into_iter()
+            .flatten()
+            .chain(right.into_iter().flatten())
+    }
```
cmyr: this isn't necessary, you can just use `map` on the normal iterator? Something like,

```rust
let subtables = lookup.subtables()?;
let mut null_iter = subtables.iter().map(|table| match table {
    Err(ReadError::NullOffset) => None,
    other => Some(other),
});
```

read-fonts/src/offset_array.rs
```diff
@@ -85,6 +85,30 @@ where
...
+        let mut iter = self.offsets.iter();
+        let args = self.args;
+        let data = self.data;
+        std::iter::from_fn(move || {
+            iter.next().map(|off| {
+                off.get()
+                    .non_null()
+                    .map(|_| off.get().resolve_with_args(data, &args))
+            })
+        })
```
cmyr: again I don't think we need this as public API, and if we _did_ it would be simplest to implement it as a `map` on the existing iter.

cmyr (changes requested):
Okay this looks reasonable, there are a few suggestions inline; the main one is to look at https://github.com/googlefonts/fontations/pull/1849 and try to use that instead of `subtables_nullable`.

read-fonts/src/tables/gpos.rs
qxliu76: Thanks! See my comments in that PR.

read-fonts/src/tables/gpos/closure.rs
qxliu76: This a good idea! With sanitize changes, we won't have a ReadError. It seems not easy to get a helper for checking offsets?

qxliu76:
@cmyr can you help review the PR again and merge it if it looks good. Currently merging is blocked because you requested changes.

read-fonts/src/tables/gpos.rs
cmyr: okay I merged that PR, so you can rebase on `main` and remove `subtables_nullable()` now :)

garretrieger: @qxliu76 if you'd like to stick with the subtables_nullable() approach in this PR I think it should be possible to make that a non-public package scoped helper method. That should address concerns about this being added to the public API of read-fonts.

cmyr: to be clear, I just think subtables_nullable is a bad design, that is adding a ton of unnecessary code to do what can easily be accomplished with a `.map` call during iteration of the current `subtables()` result.

If you want to create a helper inside skera to add this functionality so you don't need to repeat the same code in multiple places, that's fine, but it definitely doesn't belong in read-fonts.

qxliu76: I have addressed review comments. can I have a LGTM to unblock this PR

read-fonts/src/offset_array.rs
```diff
@@ -85,6 +85,24 @@ where
...
         })
     }
+
+    /// Return the offset at specified index
+    pub fn get_offset(&self, idx: usize) -> Result<O, ReadError> {
+        self.offsets
+            .get(idx)
+            .ok_or(ReadError::InvalidCollectionIndex(idx as _))
+            .map(|o| o.get())
+    }
```
cmyr: okay so now I think this is also unnecessary, since it seems you're only using this for null checks? As mentioned before, you can check null just as easily on the result of `get()`.

there's lots of code where you have,

```rust
if lookup_offsets.get_offset(i as usize)?.is_null() {
    continue;
}
let lookup = lookup_offsets.get(i as usize)?;
```

and this can just as easily be,

```rust
let lookup = match lookup_offsets.get(i as usize) {
    Err(ReadError::NullOffset) => continue,
    other => other,
}?;
```

Or you can simplify it further by using a macro:

```rust
// at the top of gsub/closure.rs
macro_rules! continue_if_null {
    ($expr:expr) => {
        match $expr {
            Err(ReadError::NullOffset) => continue,
            other => other,
        }
    };
}
```

and then it's a one-liner:

```rust
let lookup = continue_if_null!(lookup_offsets.get(i as usize))?;
```

read-fonts/src/offset_array.rs
```diff
@@ -85,6 +85,24 @@ where
...
+
+    /// Iterate over all of the offset targets.
+    ///
+    /// Offset is treated as nullable and each offset will be resolved as it is encountered.
+    pub fn iter_as_nullable(&self) -> impl Iterator<Item = Option<Result<T, ReadError>>> + 'a {
+        self.iter().map(|off| match off {
+            Err(ReadError::NullOffset) => None,
+            other => Some(other),
+        })
+    }
```
cmyr, line 99: I'm a bit more sympathetic to this, because I see how it saves having to rewrite the iter()/map() in a whole bunch of places.

Instead of adding it to read-fonts, though, you can implement it in skera using a trait:


```rust
trait IterNullableHelper<'a> {
    fn iter_as_nullable(&self) -> impl Iterator<Item=Option<Result<T, ReadError>>> + 'a;
}

impl<'a, T, O> IterAsNullableHelper for ArrayOfOffsets<'a, T, O>
// these bounds have to match what is in the read-fonts impl block that has the normal `iter` method
where
    O: Scalar + Offset,
    T: ReadArgs + FontReadWithArgs<'a>,
    T::Args: Copy + 'static,
 {
    fn iter_as_nullable(&self) -> impl Iterator<Item=Option<Result<T, ReadError>>> + 'a {
        // your impl here
    }
}
```

skera/src/gpos.rs
```diff
@@ -230,12 +236,15 @@ impl<'a> SubsetTable<'a> for PositionLookup<'_> {
             PositionSubtables::MarkToMark(_) => 6,
             PositionSubtables::Contextual(_) => 7,
             PositionSubtables::ChainContextual(_) => 8,
-            PositionSubtables::EmptyExtension => 9, // should we just skip this?
+            PositionSubtables::EmptyExtension => {
+                return Err(s.set_err(SerializeErrorFlags::SERIALIZE_ERROR_OTHER))
+            }
```
cmyr, line 241: this probably isn't an error condition? It just means that there was an extension lookup type that did not have any subtables, which is valid (although you probably want to drop it?)

skera/src/gsub.rs
```diff
@@ -229,7 +235,9 @@ impl<'a> SubsetTable<'a> for SubstitutionLookup<'_> {
             SubstitutionSubtables::Contextual(_) => 5,
             SubstitutionSubtables::ChainContextual(_) => 6,
             SubstitutionSubtables::Reverse(_) => 8,
-            SubstitutionSubtables::EmptyExtension => 9, // should we just skip this? probably?
+            SubstitutionSubtables::EmptyExtension => {
+                return Err(s.set_err(SerializeErrorFlags::SERIALIZE_ERROR_OTHER))
```
cmyr, line 239: see above

skera/src/gpos.rs
```diff
@@ -287,7 +296,7 @@ impl CollectVariationIndices for Gpos<'_> {
         let lookups = lookups.lookups();
         for i in plan.gpos_lookups.keys() {
             let Ok(lookup) = lookups.get(*i as usize) else {
```
cmyr, line 298: so this is `continuing` on any error condition, including both a nullable offset but also malformed data. Is that correct?

skera/src/gpos.rs
cmyr: Similarly, this is a place where you might want to be checking null and just `continuing`, instead? Like,


```suggestion
let subtables = match lookup.subtables() {
    Ok(subs) => subs,
    Err(ReadError::NullOffset => continue,
    Err(_) => return, // if we don't want to report the error
};
```

cmyr (commented):
Okay just a few more little things and this should be good to go :)

read-fonts/src/offset_array.rs
qxliu76: implemented helper in skera as suggested, but kept this API as `pub(crate)` in read-fonts for closure code usage

skera/src/gpos.rs
qxliu76: it's an error condition: empty lookup indices will be removed during lookup closure, if we ever need to subset an empty lookup table, then something went wrong in the planning phase.

qxliu76:
I need an approval to unblock this PR

cmyr (approved):
Thanks!

PR #1847, opened 2026-04-25
[types] Add f64 conversion fns for all fixed types

cmyr:
I initially opted to only have f32 conversions for the lower-precision types (specifically those types where the full range of values could round-trip losslessly via f32) but this overlooked the fact that f64-f32 conversions could introduce rounding errors that would change the result.

This now adds from_f64 and to_f64 methods for the types that didn't have them. The f32 conversion methods are unchanged.

This would let us address https://github.com/googlefonts/fontc/issues/1966


dfrg (approved):
🚢

PR #1849, opened 2026-04-30
[read] Make subtables() helper handle empty extensions

cmyr:
As written, this method would incorrectly return a ReadError in the case where an extension lookup had no subtables, as well as the case where the first subtable offset of an extension lookup was malformed, including if it was null.

This addresses this by adding a new variant, 'EmptyExtension', that is returned in the case where either there are no subtable offsets, or they are all invalid.

This was motivated by ongoing subsetting work, where this failure mode was not acceptable.

@qxliu76 I believe this will remove the need for the `subtables_nullable` method in #1842, although during iteration if you get a `ReadError` for a particular subtable you'll need to check if the variant is `::NullOffset`.

### DIFF
--- a/read-fonts/src/tables/gpos.rs
+++ b/read-fonts/src/tables/gpos.rs
@@ -75,6 +75,8 @@ pub enum PositionSubtables<'a> {
     MarkToMark(PosSubtables<'a, MarkMarkPosFormat1<'a>>),
     Contextual(PosSubtables<'a, PositionSequenceContext<'a>>),
     ChainContextual(PosSubtables<'a, PositionChainContext<'a>>),
+    /// An extension lookup did not have any subtables
+    EmptyExtension,
 }
 
 impl<'a> PositionLookup<'a> {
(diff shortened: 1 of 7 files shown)


qxliu76:
Thanks for this PR! Considering the upcoming sanitize changes, I assume subtables() API would be infallible, so we might not need to add an extra enum type`PositionSubtables::EmptyExtension` fow now. I will also revisit and remove those iter_as_nullable() APIs then. Does that make sense to you?

cmyr:
Even if the API becomes infallible, it is not strictly invalid to have An extension lookup that contains no subtables, and we'd need to handle that somehow? This case should not produce an error, and currently it does.

qxliu76:
I assume we might return Default() in case it's empty or maybe not. Since we're not sure how the API signature would look like, I'd prefer to keep it as it for now. We could make this change after sanitize changes.

qxliu76:
Also `EmptyExtension` type doesn't comply with the spec, what do you think?

cmyr:
Sorry, I don't think I was clear enough here.

- I do not think that `subtables_nullable` makes sense as public API in read-fonts. The existing `subtables` API allows you to check, both for the first call to `subtables()` as well for each individual subtable returned via `iter()` or `get(idx)`, whether or not the result was null (because in this case, and only this case, we return `ReadError::NullOffset`.) Unless I have missed something, Adding an entire new set of types and new API for this is overkill.
- That said, the `subtables_nullable` implementation _does_ point out a genuine bug in the existing `subtables()` API, and a second additional limitation: the bug is that we return an `Error` in the entirely legal case where there is an extension lookup with no subtables; and the limitation is that we return too early in the case where an extension lookup's first subtable offset is null (or otherwise malformed) but a subsequent subtable offset is valid.

This patch fixes the second two issues, and so means that we do not need `subtables_nullable`.

If you think I have overlooked something, or if you do not understand how to make the existing `subtables()` API work for your purposes, I would be happy to show you how I think it should work; otherwise let's merge this PR and then you can rebase #1842 to remove `subtables_nullable` and use this API.

If this does not work for some _technical reason_ please explain what it is, and then we can figure out if it can be addressed somehow, or if we genuinely need a duplicate version of the API.

cmyr:
> Also `EmptyExtension` type doesn't comply with the spec, what do you think?

The `subtables` api and `Subtables` struct is not intended to represent anything in the spec. It is intended as a simpler way of accessing subtables, without needing to manually handle extension lookups. You are always free to use the standard API, which involves matching on the specific `Substitution`/[`PositionLookup`](https://docs.rs/read-fonts/latest/read_fonts/tables/gpos/enum.PositionLookup.html) itself, and then iterating the subtables directly via the inner [`Lookup`](https://docs.rs/read-fonts/latest/read_fonts/tables/layout/struct.Lookup.html) struct. In the case of an extension lookup, you will then also need to match on the [`ExtensionSubtable`](https://docs.rs/read-fonts/latest/read_fonts/tables/gpos/enum.ExtensionSubtable.html) enum, and handle each case manually.

qxliu76:
I'd choose to keep subtables_nullable() API until sanitize changes landed  and I'll remove it.

cmyr:
sorry, maybe I wasn't clear: I am not going to approve a PR containing `subtables_nullable` unless there is a strong technical reason that it is necessary.

PR #1851, opened 2026-04-30
[colr closure] add a binary search branch in closure

qxliu76:
depends on #1842.
Compare the cost of doing linear search and binary search, and choose the better path to do closure.
Plus a tiny change to use int_set intersect_range() API


cmyr (approved):
looks good!

garretrieger (approved):

PR #1853, opened 2026-05-04
Add safety annotations to shared-brotli-patch-decoder

wmedrano:
- Adds "SAFTEY:" before all calls to `unsafe`.
- Splits up the function to make the lifetime of `BrotliDecoderCreateInstance` more obvious.
- `sys` comments populated based on [brotli/include/brotli/decode.h](https://source.chromium.org/chromium/chromium/src/+/main:third_party/brotli/include/brotli/decode.h?q=brotli%2Fdecode.h&ss=chromium)

### DIFF
--- a/shared-brotli-patch-decoder/src/sys.rs
+++ b/shared-brotli-patch-decoder/src/sys.rs
@@ -8,6 +8,7 @@
 
 use std::os::raw::{c_int, c_void};
 
+/// Opaque structure that holds decoder state.
 pub type BrotliDecoderState = c_void;
 
 // From brotli/c/include/brotli/decode.h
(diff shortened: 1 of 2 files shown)


shared-brotli-patch-decoder/src/sys.rs
```diff
@@ -57,5 +159,15 @@ extern "C" {
         total_out: *mut usize,
     ) -> BrotliDecoderResult;
 
+    /// Deinitializes and frees a [`BrotliDecoderState`] instance.
+    ///
+    /// # Safety
+    ///
+    /// The `state` pointer must be either `null` or a pointer to a valid
```
wmedrano, line 166: BrotliDecoderDestroyInstance has a null check built in, see https://source.chromium.org/chromium/chromium/src/+/main:third_party/brotli/dec/decode.c;l=109;drc=91861f7ddfab07ec80ffea4b5968350a4170e8b6;bpv=1;bpt=1

garretrieger (approved):

PR #1858, opened 2026-05-12
Update rand and rstest dev-dependencies

musicinmybrain:
Update `rand` dev-dependencies in `read-fonts` (for benchmarks) and `write-fonts` (for tests) from 0.8 to 0.10. This required some source-code changes. I verified that `cargo test` and `cargo bench` still worked with and without `--all-features`.

Update the `rstest` dev-dependency in `write-fonts` from 0.18 to 0.26. This didn’t require any source-code changes; `rstest` rarely has significant API changes despite its frequent SemVer breaks.

### DIFF
--- a/read-fonts/Cargo.toml
+++ b/read-fonts/Cargo.toml
@@ -51,7 +51,7 @@ once_cell = { version = "1.21.4", default-features = false, features = [
 read-fonts = { path = ".", features = ["experimental_font_api"] }
 font-test-data = { workspace = true }
 criterion = { workspace = true }
-rand = "0.8.5"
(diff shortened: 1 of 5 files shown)


cmyr:
sorry to let this sit; happy to get it in, but it needs to be rebased. (I tried to resolve conflicts via the web UI but that didn't work apparently, feel free to discard my commit)

musicinmybrain:
> sorry to let this sit; happy to get it in, but it needs to be rebased. (I tried to resolve conflicts via the web UI but that didn't work apparently, feel free to discard my commit)

Ok, rebased and force-pushed!

PR #1861, opened 2026-05-12
[skera] Correctly compute PairPos1 coverage

simoncozens:
Splitting #1798 into smaller PRs.


qxliu76:
I noticed a conflict with main and found out that I merged a larger PR that happened to cover this bug. So we won't need to merge this patch. I'm going to close this PR. Thanks so much for your fix!

PR #1862, opened 2026-05-12
[skera] Typo finding duplicate features

simoncozens:
Splitting #1798 into smaller PRs.

### DIFF
--- a/skera/src/layout.rs
+++ b/skera/src/layout.rs
@@ -880,7 +880,7 @@ pub(crate) fn find_duplicate_features(
     feature_indices: IntSet<u16>,
 ) -> FnvHashMap<u16, u16> {
     let mut out = FnvHashMap::default();
-    if lookup_indices.is_empty() {
+    if feature_indices.is_empty() {
         return out;
     }
 


qxliu76:
I really appreciate your help in splitting the PR!
After thinking it twice: I think it's probably not a typo. The major change in this fn happened inside a for loop: https://github.com/googlefonts/fontations/blob/ae29901fbe8c329879aa4a3b20f5953037db0d45/skera/src/layout.rs#L889, if `feature_indices` is empty, the whole `for `loop will be skipped. If `lookup_indices` is empty, we cannot skip the loop.  So the check on `lookup_indices` is needed to avoid unnecessary work.

simoncozens:
But there are situations where you should not skip the loop. The code above this says "never drop feature pref" (because it is needed for correct processing of Khmer).

https://github.com/googlefonts/fontations/blob/ae29901fbe8c329879aa4a3b20f5953037db0d45/skera/src/layout.rs#L850-L856

Without this patch, we *do* drop the pref feature; with this patch, we don't drop it. See `harfbuzz/test/subset/data/tests/layout.khmer.tests`.

qxliu76:
> But there are situations where you should not skip the loop. The code above this says "never drop feature pref" (because it is needed for correct processing of Khmer).
> 
> https://github.com/googlefonts/fontations/blob/ae29901fbe8c329879aa4a3b20f5953037db0d45/skera/src/layout.rs#L850-L856
> 
> Without this patch, we _do_ drop the pref feature; with this patch, we don't drop it. See `harfbuzz/test/subset/data/tests/layout.khmer.tests`.

oh yeah you're right! It's much clearer to have a test for a bug fix:)

qxliu76 (approved):

simoncozens:
This is why I run skera on the Harfbuzz test suite. :-)

qxliu76:
yeah, I'm currently working on porting all tests from Harfbuzz. We don't aim to match HB 100% in some cases but I'll try to get all tests clean:)

PR #1864, opened 2026-05-13
Pin `brotlic` to working commit

wmedrano:


wmedrano:
- I'm pretty sure this (treating empty dictionary as `None`) is the fix, but looking more into the brotli package to see when it happened.
- I'll follow up by pinning the version of brotlic we use. Ideally a bot would send out an update so that we catch regressions and can keep up to date.

wmedrano:
I updated the build script to pin to a working commit. I'll address the issue of behavior around an empty dictionary soon since it needs more investigation.

PR #1865, opened 2026-05-13
Update to latest brotlic

wmedrano:
- Update the commit for brotlic
- Update `brotli_decode_without_shared_dict` to match new behavior
  - See https://github.com/google/brotli/pull/1479

### DIFF
--- a/shared-brotli-patch-decoder/build.rs
+++ b/shared-brotli-patch-decoder/build.rs
@@ -36,7 +36,7 @@ mod c_brotli {
             }
 
             let status = Command::new("git")
-                .args(["checkout", "af041f55e7bfa3e1bc502f234538e1d6dd6282bd"])
+                .args(["checkout", "1736fe1d50b205f718281a94035fd66f14d72ef5"])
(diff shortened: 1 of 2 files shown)


shared-brotli-patch-decoder/build.rs
```diff
@@ -36,7 +36,11 @@ mod c_brotli {
             }
 
             let status = Command::new("git")
-                .args(["checkout", "af041f55e7bfa3e1bc502f234538e1d6dd6282bd"])
+                .args([
+                    "checkout",
+                    // https://github.com/google/brotli/commit/1736fe1d50b205f718281a94035fd66f14d72ef5
```
cmyr: channeling my inner rod, this comment seems a bit superfluous?

cmyr (approved):
I don't really know how any of this works, but if you're happy I'm happy? One note inline..

wmedrano:
> I don't really know how any of this works, but if you're happy I'm happy? One note inline..

Same, but with the `brotli` people. This is just a slightly oxidized version of their C API.

PR #1866, opened 2026-05-14
[ift] Run clippy in CI

cmyr:
This wasn't running previously and there were a few clippy lints, which were causing my pre-push hook to constantly fail.

@garretrieger this also fixes the existing lints; one of these fixes involves ignoring an error. Do we want to be handling it?

JMM


garretrieger:
> @garretrieger this also fixes the existing lints; one of these fixes involves ignoring an error. Do we want to be handling it?

Yes, I would recommend changing that to propagate the error. There's some examples in that function of converting to a PatchingError::SerializationError

garretrieger (approved):

PR #1869, opened 2026-05-15
[subset] bug fix for postV2 subset

qxliu76:
glyph name indices that >=258 in glyphNameIndex array are not in sorted order. So this is wrong: https://github.com/googlefonts/fontations/blob/03b333678faecebea15413c2f1fc572b44bcf2b2/skera/src/post.rs#L97
This PR re-implements the whole postV2 subset method.


cmyr (approved):
👍

garretrieger (approved):

PR #1870, opened 2026-05-15
Add some additional info to the output of ift_extend:

garretrieger:
- Breakdown of the bytes transferred.
- Set of glyphs that were added to the font by the operation.


cmyr (approved):
looks good, sorry for letting this sit around! please ping me anytime if I've neglected a review.

PR #1871, opened 2026-05-18
[subset] remove invalid lookup indices before closure_lookups

qxliu76:
This avoids err out in case of invalid input lookup index.
This PR depends on #1869


cmyr (approved):
makes sense!

garretrieger (approved):

PR #1873, opened 2026-05-20
Create lightweight font builder crate

wmedrano:


wmedrano:
Closing in favor of feature guarding write-fonts

PR #1874, opened 2026-05-20
[subset] close unicodes over bidi mirror variants

qxliu76:
1. Add unicode_closure() as default behavior and support no-bidi-closure flag in testing.
2. Plus a few lines of docs update.

Sorry this PR depends on #1871, dependency is mainly due to tests update.


skera/src/unicode_closure.rs
```diff
@@ -0,0 +1,469 @@
...
+            if unicodes.contains(cp) {
+                out.insert(mirrored_cp);
+            }
+        }
+    }
+    out
+}
+// implement [Unicode Bidi Mirroring](https://unicode.org/reports/tr44/#BidiMirroring.txt) property detection.
+// Derived from: <http://www.unicode.org/Public/UNIDATA/BidiMirroring.txt>
+static MIRRORED: [(u32, u32); 428] = [
```
dfrg, line 34: This works nicely. I expect the data doesn’t change often, but suggest adding the Unicode version (looks like 17.0.0?) to the comment so we can keep track and update if necessary.

dfrg (approved):
LGTM

skera/src/unicode_closure.rs
qxliu76: Done. Also added 2 consts accordingly.

garretrieger (approved):

PR #1877, opened 2026-05-21
Implement hasher directly in Skera

wmedrano:
- Inlining was requested by chromium in https://chromium-review.git.corp.google.com/c/chromium/src/+/7814023
- Makes sense since it is as tiny as the infamous [left-pad](library)
- Forking is ok since as fnv follows MIT + Apache 2.0 license.

### DIFF
--- a/skera/src/fnv.rs
+++ b/skera/src/fnv.rs
@@ -1,4 +1,5 @@
-//! A fork of the `fnv` crate from <https://github.com/servo/rust-fnv>.
+//! A fork of the `fnv` crate from <https://github.com/servo/rust-fnv>. The code is forked since it
+//! is small and to reduce the maintenance overhead of importing into other monorepos.
 //!
 //! FNV is an implementation of the Fowler–Noll–Vo hash function.
 //!
--- a/skera/src/fnv.rs
+++ b/skera/src/fnv.rs
@@ -1,0 +1,93 @@
+//! A fork of the `fnv` crate from <https://github.com/servo/rust-fnv>.
+//!
+//! FNV is an implementation of the Fowler–Noll–Vo hash function.
+//!
+//! ## About
+//!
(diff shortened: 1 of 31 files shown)


qxliu76 (approved):

qxliu76:
I think we might need to include LICENSE-APACHE and LICENSE-MIT files? otherwise LGTM

garretrieger:
> I think we might need to include LICENSE-APACHE and LICENSE-MIT files? otherwise LGTM

Looks like fontations has both of those licenses in the root, so should be good to go.

garretrieger (approved):

skera/src/fnv.rs
```diff
@@ -0,0 +1,93 @@
+//! A fork of the `fnv` crate from <https://github.com/servo/rust-fnv>.
```
cmyr: I would like just a bit more explanation here about why this is inlined: is it because we need to make changes, or just because we want to remove a dependency? Both? Some other thing?

cmyr (approved):
one little note but otherwise 👍

nicoburns:
Out of interest, is there a technical reason why Skera uses specifically `fnv` as opposed to a newer fast hash function like [foldhash](https://github.com/orlp/foldhash)? Or is it just a case of familiarity?

qxliu76:
> Out of interest, is there a technical reason why Skera uses specifically `fnv` as opposed to a newer fast hash function like [foldhash](https://github.com/orlp/foldhash)? Or is it just a case of familiarity?

I think fnv fits our subsetting needs very well: compact code, fast for integers(we only use integer keys in subsetting), and no need to worry about collision attacks. I actually did some early experiments with fxhash/ahash/fnvhash, FnvHash is the fastest one(I didn't try foldhash though). 
Of course we could change hashing functions if foldHash is proved to be faster in our benchmarking run later.

PR #1880, opened 2026-05-22
Fix 26.6 mul/div operators 

dfrg:
Actually perform operations in 26.6 Also reworks scaling a bit in skrifa to account for this. Checked fauntlet against the Noto library to confirm this doesn't break anything.

Supersedes #1698

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -216,7 +216,6 @@ impl Div for Fixed {
 impl Mul for F26Dot6 {
     type Output = Self;
 
-    #[expect(clippy::suspicious_arithmetic_impl)]
     #[inline(always)]
     fn mul(self, other: Self) -> Self::Output {
         let ab = self.0 as i64 * other.0 as i64;
@@ -227,7 +226,6 @@ impl Mul for F26Dot6 {
 impl Div for F26Dot6 {
     type Output = Self;
 
-    #[expect(clippy::suspicious_arithmetic_impl)]
     fn div(self, other: Self) -> Self {
         let sign = (self.0 < 0) ^ (other.0 < 0);
         let au = self.0.unsigned_abs() as u64;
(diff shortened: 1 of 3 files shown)


font-types/src/fixed.rs
```diff
@@ -137,96 +137,120 @@ macro_rules! fixed_impl {
...
+        } else {
+            q as i32
+        })
+    }
+}
+
+impl Mul for F26Dot6 {
+    type Output = Self;
+
+    #[allow(clippy::suspicious_arithmetic_impl)]
```
cmyr: I'd suggest using `#[expect]` instead of `#[allow]` in basically all cases now? it has the nice advantage of letting us know if the lint has gone away or a future revision makes it unnecessary.

font-types/src/fixed.rs
```diff
@@ -523,4 +549,24 @@ mod tests {
...
+        );
+        assert_eq!(
+            F26Dot6::from_f64(0.5) / F26Dot6::from_f64(2.0),
+            F26Dot6::from_f64(0.25)
+        );
+        assert_eq!(
+            F26Dot6::from_f64(0.5) / F26Dot6::from_f64(-2.4),
+            F26Dot6::from_f64(-0.20833333333333334)
+        );
+    }
```
cmyr, line 588: might as well also include the test from the original PR?

https://github.com/daltonmaag/fontations/blob/e08bf25a03b5851e0ae79fa0dc98c6f6afb56ae1/font-types/src/fixed.rs#L515

cmyr (approved):
thanks, will be nice to get this tided up.

font-types/src/fixed.rs
dfrg: Makes sense, will do in the future. I ended up changing the code so the semantics match Fixed and that removed the need for the clippy lints.

font-types/src/fixed.rs
dfrg: Copied the identity tests from that PR and added a few more to capture some edge cases.

PR #1882, opened 2026-05-23
Create "tables" feature for write-fonts

wmedrano:
# Motivation

Chromium doesn't want to include `kurbo` and its transitive dependencies, especially since they are not used in skera and incremental-font-transfer. This change removes some of the dependency while also improving compile times for skera only use cases.

# Changes

Without the crates feature:

- Most of the codegen goes away. Skera CLI compile time goes from `~18s` to `~16s`
  - The `write-fonts` portion goes from `~10s` to `~1s`
- Can still use the bytes based API
- Removes kurbo and indexmap dependencies

# Before

image

# After

image

### DIFF
--- a/write-fonts/src/object.rs
+++ b/write-fonts/src/object.rs
@@ -1,0 +1,26 @@
+use std::{collections::HashMap, sync::atomic::AtomicU64};
+
+use crate::write::TableData;
+
+static OBJECT_COUNTER: AtomicU64 = AtomicU64::new(0);
+
+/// An identifier for an object in the compilation graph.
+#[derive(Debug, Clone, Copy, PartialOrd, Ord, Hash, PartialEq, Eq)]
+pub struct ObjectId(pub(crate) u64);
+
+impl ObjectId {
+    pub fn next() -> Self {
+        ObjectId(OBJECT_COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed))
+    }
+}
+
+#[derive(Debug, Default)]
+pub(crate) struct ObjectStore {
+    pub(crate) objects: HashMap<TableData, ObjectId>,
+}
+
+impl ObjectStore {
+    pub(crate) fn add(&mut self, data: TableData) -> ObjectId {
+        *self.objects.entry(data).or_insert_with(ObjectId::next)
+    }
+}
--- a/write-fonts/src/offsets.rs
+++ b/write-fonts/src/offsets.rs
@@ -7,6 +7,7 @@ pub const WIDTH_16: usize = 2;
 /// The width in bytes of an Offset24
 pub const WIDTH_24: usize = 3;
 /// The width in bytes of an Offset32
+#[allow(dead_code)]
 pub const WIDTH_32: usize = 4;
 
 /// An offset subtable.
@@ -67,7 +68,7 @@ impl<T, const N: usize> AsMut<T> for OffsetMarker<T, N> {
 
 // NOTE: we don't impl AsRef/AsMut for NullableOffsetMarker, since it is less
 // useful than the Option::as_ref and Option::as_mut methods available through deref
-
+#[cfg_attr(not(feature = "tables"), allow(dead_code))]
 impl<const N: usize, T> OffsetMarker<T, N> {
     /// Create a new marker.
     pub fn new(obj: T) -> Self {
@@ -85,6 +86,7 @@ impl<const N: usize, T> OffsetMarker<T, N> {
     }
 }
 
+#[cfg_attr(not(feature = "tables"), allow(dead_code))]
 impl<const N: usize, T> NullableOffsetMarker<T, N> {
     /// Create a new marker.
     pub fn new(obj: Option<T>) -> Self {
--- a/write-fonts/src/offsets.rs
+++ b/write-fonts/src/offsets.rs
@@ -10,6 +10,36 @@ pub const WIDTH_24: usize = 3;
 #[allow(dead_code)]
 pub const WIDTH_32: usize = 4;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[repr(u8)]
+pub enum OffsetLen {
+    Offset16 = 2,
+    Offset24 = 3,
+    Offset32 = 4,
+}
+
+impl OffsetLen {
+    /// The maximum value for an offset of this length.
+    #[cfg(feature = "tables")]
+    pub const fn max_value(self) -> u32 {
+        match self {
+            Self::Offset16 => u16::MAX as u32,
+            Self::Offset24 => (1 << 24) - 1,
+            Self::Offset32 => u32::MAX,
+        }
+    }
+}
+
+impl std::fmt::Display for OffsetLen {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            Self::Offset16 => write!(f, "Offset16"),
+            Self::Offset24 => write!(f, "Offset24"),
+            Self::Offset32 => write!(f, "Offset32"),
+        }
+    }
+}
+
 /// An offset subtable.
 ///
 /// The generic const `N` is the width of the offset, in bytes.
--- a/write-fonts/src/offsets.rs
+++ b/write-fonts/src/offsets.rs
@@ -98,7 +98,6 @@ impl<T, const N: usize> AsMut<T> for OffsetMarker<T, N> {
 
 // NOTE: we don't impl AsRef/AsMut for NullableOffsetMarker, since it is less
 // useful than the Option::as_ref and Option::as_mut methods available through deref
-#[cfg_attr(not(feature = "tables"), allow(dead_code))]
 impl<const N: usize, T> OffsetMarker<T, N> {
     /// Create a new marker.
     pub fn new(obj: T) -> Self {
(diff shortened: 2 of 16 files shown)


write-fonts/src/write.rs
```diff
@@ -214,45 +214,56 @@ impl TableData {
...
-    ) -> Result<T, ReadError> {
-        let data = FontData::new(&self.bytes);
+    ) -> Result<T, read_fonts::ReadError> {
+        let data = read_fonts::FontData::new(&self.bytes);
         T::read_with_args(data, args)
     }
 
     /// A helper function to read a value out of this data.
-    pub(crate) fn read_at<T: Scalar>(&self, pos: usize) -> Option<T> {
+    pub(crate) fn read_at<T: font_types::FixedSize + font_types::Scalar>(
```
cmyr: why does this signature change?

write-fonts/src/write.rs
```diff
@@ -214,45 +214,56 @@ impl TableData {
...
+    pub(crate) fn reparse<'a, T: read_fonts::FontRead<'a>>(
+        &'a self,
+    ) -> Result<T, read_fonts::ReadError> {
+        let data = read_fonts::FontData::new(&self.bytes);
         T::read(data)
     }
 
     // see above
-    pub(crate) fn reparse_with_args<'a, A, T: FontReadWithArgs<'a, Args = A>>(
+    pub(crate) fn reparse_with_args<'a, A, T: read_fonts::FontReadWithArgs<'a, Args = A>>(
```
cmyr: just for the sake of optimizing for a smaller/more intelligible diff, I'd leave the module import in place?

write-fonts/src/write.rs
```diff
@@ -214,45 +214,56 @@ impl TableData {
         let placeholder = PLACEHOLDER_BYTES.get(..width.min(4)).unwrap();
         self.write_bytes(placeholder);
     }
+}
 
-    pub(crate) fn write<T: Scalar>(&mut self, value: T) {
+#[cfg(feature = "tables")]
```
cmyr: okay so instead of annotating specific methods you've grouped them into impl blocks and feature gated the one? I think that's a good pattern in general, but in this particular case it has the unfortunate consequence of making the diff harder to read 😒

cmyr (commented):
Some comments inline, but before I go too deep maybe let's clarify some of the higher-level things?

- what exactly are we trying to enable/disable with this feature? From my first read-through it seems to be mostly just validation, but it doesn't make sense to me that we would turn off validation if we aren't also turning off compilation, via `FontWrite`?
- my expectation from the name of the feature is that this would just fully remove the owned table types from write-fonts, to the point where I imagined it might fully disable the `tables` module. Is an approach like that feasible?
- if we _do_ want to go this route, i'd like to look carefully and try to figure out the most surgical way of applying the feature, so that we don't end up with a ton of annotations.


So I guess first let's figure out exactly what we would need to accomplish, and from there we can figure out where draw the feature boundaries?

write-fonts/src/write.rs
wmedrano: Added back `use::FixedSize`

`FixedSize` is auto-implemented on `Scalar` but this stopped applying when I stopped importing `font_types::Scalar` module wide.

write-fonts/src/write.rs
wmedrano: Done, module imports in place. I'll try to leave the diff destroying changes before submission.

Any suggestions on

- Use imports and deal with the "unused" warnings
- Use imports guarded by `#[cfg(feature = "tables)]` to prevent "unused" warnings
- Use qualified paths on feature = "tables" only paths

write-fonts/src/write.rs
wmedrano: Ack, will leave this moving around as last minute cleanups before submission instead of review time

wmedrano:
> * what exactly are we trying to enable/disable with this feature? From my first read-through it seems to be mostly just validation, but it doesn't make sense to me that we would turn off validation if we aren't also turning off compilation, via `FontWrite`?

`FontWrite` is still used. We get rid of most tables, but keep `generated_font.rs` which has `FontWrite` on `TableDirectory`. This is still needed by the `add_raw` and `build` method of making fonts.

> * my expectation from the name of the feature is that this would just fully remove the owned table types from write-fonts, to the point where I imagined it might fully disable the `tables` module. Is an approach like that feasible?

Yes, the `tables` module was fully removed. The `Validate` trait was unused. I reverted the change to `Validate` since it added a lot of noise to the diff.

> * if we _do_ want to go this route, i'd like to look carefully and try to figure out the most surgical way of applying the feature, so that we don't end up with a ton of annotations.

This was my attempt at it, hence the moving of stuff to error, split of utils, split of graph.rs. `graph` (and error because `error -> graph`) was tricky since they use a lot of `crate::tables::` stuff to perform table rewrites

> So I guess first let's figure out exactly what we would need to accomplish, and from there we can figure out where draw the feature boundaries?

So "I + Chromium" want to remove `kurbo` as a dependency since it is not needed and managing deps in a monorepo is pretty tedious. I previously had to go on a sidequest to upgrade dependencies within Google's monorepo since we reached 4 versions of `kurbo` which is passed the allowed threshold and triggers some process. FYI: Implementation for creating "kurbo" feature https://github.com/googlefonts/fontations/compare/wm-deps2

The change here goes a bit further and removes stuff. As is, I think it makes sense since skera/subsetting is a common use case and it would be nice to not pay the compilation toll. The (now reverted) exclusion of `trait Validate` is a bit overkill.

nicoburns:
Would a principled way of looking at what this PR is trying to achieve be:

"functionality for writing the SFNT container format without table-specific details"

?

That seems like a generally useful subset of functionality (and the "read' equivalent would also be useful), which could either be supported with a feature flag, or by splitting the core "sfnt container" functionality out into it's own crate(s) (`read-fonts-core`/`write-fonts-core`? or perhaps a combined `sfnt` crate?) that only depends on `font-types`.

I actually have something along those lines here: https://github.com/nicoburns/oaty which I'm using to split a font from a TTC out into a standalone TTF (because Skia on macOS can only access the first font in a TTC if you're not using Skia's built-in font enumeration), which didn't seem to justify pulling in `read-fonts` and `write-fonts` in a crate (`anyrender_skia`) that doesn't otherwise depend on Fontations.

wmedrano:
@nicoburns Yeah, that's a good way to think about it. So far:

This will have `FontBuilder` with
- `new`
- `add_raw` to add tag + bytes
- `copy_missing_tables` to copy from `read_fonts::FontRef` (can be feature guarded)
- `contains_tag`, `ordered_tags`
- `build` to materialize the Vec<u8>

There will still be a dependency on `read_fonts`. Though maybe we can guard against it as long as `"tables"` is disabled. You also have to do something about `read_fonts::tables::compute_checksum` and maybe another odd case or 2.

Could be useful in case you no longer want oaty, but it won't benefit `skera` as `skera` already depends on `read_fonts`

wmedrano:
@cmyr This is ready for another look. Only `FontBuilder` is publicly exported when "tables" is disabled.

image

image


Maybe this might make it easier to review:

- https://github.com/googlefonts/fontations/pull/1882/changes/e7848121f0d4ccb16a66aa6569a8b5e18e8090ee adds the feature guard in a brute force manner
- https://github.com/googlefonts/fontations/pull/1882/changes/c3bbdca9f47aa121a70ff09dbad7fbc5154783e4 splits graph.rs + utils.rs so we can guard entire `mod` instead of each individual struct and function

nicoburns:
> There will still be a dependency on read_fonts. Though maybe we can guard against it as long as "tables" is disabled. You also have to do something about read_fonts::tables::compute_checksum and maybe another odd case or 2.

What might be nice is if we could add a similar feature flag to `read-fonts` that turns off all the tables when disabled. We'd likely need a `RawTableProvider` trait with the `data_for_tag` method but without the provided methods (this could be a supertrait of `TableProvider` to avoid duplication). `read_fonts::tables::compute_checksum` might need to be moved to another module, but is otherwise straightforward I think.

But just this PR is pretty nice by itself: we're depending on Skera in Blitz where we already depend on `read-fonts` but do not otherwise have a dependency on `write-fonts` (I guess that's basically the same situation as Chrome).

write-fonts/src/write.rs
cmyr: I'm okay with options 2 & 3, depending on what makes sense for a given case: if an import is only used in a few places then just full paths makes sense, and if it's used all over then it can be feature gated.

write-fonts/src/object.rs
```diff
@@ -0,0 +1,26 @@
+use std::{collections::HashMap, sync::atomic::AtomicU64};
```
cmyr, line 1: i'm curious about the rationale for splitting these out of `graph`? they really are internal implementation details of the graph logic?

Ideally if we do want to split them out they could at least live in src/graph/...

write-fonts/src/offsets.rs
```diff
@@ -10,8 +10,39 @@ pub const WIDTH_16: usize = 2;
 #[allow(dead_code)] // will be used one day :')
 pub const WIDTH_24: usize = 3;
 /// The width in bytes of an Offset32
+#[allow(dead_code)]
```
cmyr, line 10: I'd prefer to just feature gate these I think. I really dislike `#[allow(dead_code)]`, and am preferring `#[expect(dead_code)]` whenever i'm just trying to temporarily disable something.

It would be nice if this whole module was feature gated, but i've traced through the logic a bit and can see why this is annoying...

cmyr (approved):
okay, i do see how this is annoying.

one important thing: we should make sure that CI checks --no-default-features for this crate, because it's going to be easy to break.

additionally I'd like to minimize the use of `#[cfg_attr(not(feature = "tables"), allow(dead_code))]` where possible, in favour of just feature gating things?

a few other notes inline, but broadly happy to merge this.

write-fonts/src/object.rs
wmedrano: I feature blocked the entire `graph` module. These were still used for table writer internally, even in the `add_raw` path.

write-fonts/src/offsets.rs
wmedrano: Mostly done. I left a few cases that would be large like `Validate` trait. Can guard it later tho.

PR #1884, opened 2026-05-26
Show an error when rustfmt fails

wmedrano:
I used an inferior computing environment and was confused as to why the code generated code looked terrible.


cmyr (approved):
this is definitely an annoying failure mode 😒

PR #1886, opened 2026-05-27
[codegen] Rework how we handle tables with generic fields

cmyr:
Specifically this adds a 'Discriminant' trait that we implement for
tables that have an inline discriminant that determines the type
of a subtable, such as in GPOS/GSUB lookups.

Previously the 'smarts' for reading the discriminant was declared
outside of the containing table itself, and that was going to be
annoying to get working in the sanitize code.

This new approach feels generally cleaner.


dfrg (approved):
👍

PR #1889, opened 2026-06-02
Disable write-fonts "table" feature from skera and incremental-font-transfer

wmedrano:
- This requires setting `--no-default-features` at the workspace level since features are additive. Additive means that crates can not remove any features.
- `incremental-font-transfer` still uses "tables" in test so the feature is added as a dev-dependency
- `fuzz` is fine since it sets `write-fonts.default-features = true`


cmyr (approved):
👍

PR #1891, opened 2026-06-03
[subset] a few bug fixes

qxliu76:
We need those fixes to make `layout.notonastaliqurdu.tests` clean, thus combine them in one PR. Each commit only  contains a small fix


cmyr (approved):
looks good!

garretrieger (approved):

PR #1899, opened 2026-06-05
[subset] bug fixes

qxliu76:


skera/src/cmap.rs
```diff
@@ -248,36 +275,43 @@ fn serialize_encoding_record(
...
-    let init_len = s.length();
-    cmap_subtable.serialize(s, plan, cp_to_new_gid_list)?;
-    let mut obj_idx = None;
-    if s.length() > init_len {
-        obj_idx = s.pop_pack(true);
-        if obj_idx.is_none() {
-            s.revert_snapshot(snap);
-            return Err(s.error());
-        }
+    let subtable_idx = if let Some(idx) = obj_idx {
```
wmedrano, line 285: Nit:

Style is similar, but `let_else` is slightly quicker https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

```
let Some(subtable_idx) = obj_idx else {
    ...
};
```

wmedrano (approved):

skera/src/cmap.rs
qxliu76: Thanks! I tried but it doesn't work if I want to return an idx from `pop_pack()` and assign it to `subtable_idx` in the `else `clause. Original is like this:
```
    let subtable_idx = if let Some(idx) = obj_idx {
        idx
    } else {
       Do something and return an idx
    };
```

wmedrano (approved):

garretrieger (approved):

PR #1901, opened 2026-06-07
Towards implementing beyond-64k glyphs in the font

behdad:
This just adds implementation of 24-bit GlyphIDs and Coverage/ClassDef's format 3 & 4.

First, new formats are done by duplication of the logic. Then a commit to use traits to merge the logic, but the trait commit adds more lines than it deletes. I'm concerned by the size of this change, which is a tiny fraction of what needs to be done for the whole beyond-64k to work, so sharing early to gather feedback.

### DIFF
--- a/otexplorer/src/print.rs
+++ b/otexplorer/src/print.rs
@@ -258,6 +258,7 @@ impl<'a> PrettyPrinter<'a> {
             FieldType::Fixed(val) => self.print_hex(&val.to_be_bytes())?,
             FieldType::LongDateTime(val) => self.print_hex(&val.to_be_bytes())?,
             FieldType::GlyphId16(val) => self.print_hex(&val.to_be_bytes())?,
+            FieldType::GlyphId24(val) => self.print_hex(&val.to_be_bytes())?,
             FieldType::BareOffset(offset) => self.print_offset_hex(*offset)?,
             _ => (),
         }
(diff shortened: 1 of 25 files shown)


font-codegen/src/parsing/attrs.rs
```diff
@@ -708,7 +708,9 @@ impl Count {
     pub(crate) fn count_expr(&self) -> TokenStream {
         match self {
             Count::All(_) => unreachable!("'all' count handled separately"),
-            Count::SingleArg(CountArg::Field(arg)) => quote!(#arg as usize),
+            Count::SingleArg(CountArg::Field(arg)) => {
+                quote!(transforms::to_usize(#arg))
```
cmyr: just a quick note: if this is a change we want in general, can it be PR'd outside of this patch? It's really adding a lot of noise to this diff...

behdad: Right. I'll do that separately. It's needed to unify conversion for 24bit types and other scalar types.

cmyr: makes sense. I've been thinking recently that we should be doing all this math in 32-bits and not really using usize anywhere, but that is orthogonal.

behdad: https://github.com/googlefonts/fontations/pull/1925

PR #1903, opened 2026-06-08
read-fonts: fix oob panic in fvar.rs user_to_normalized

Mrmaxmeier:
Hi,
I'm experimenting with some fuzzing techniques on the fuzzing harnesses in this repo and ran into a crash:
`thread panicked at read-fonts/src/tables/fvar.rs:147:47: range end index 65 out of range for slice of length 64`

This PR applies the same pattern that's already used with `heap_fixed_coords` a few lines above: Use a heap allocation if the amount of coords exceeds the `MAX_INLINE_AVAR2_AXES` limit.

Let me know if you'd like to see a regression test for this. I've haven't included one since the fix feels fairly uncontroversial :slightly_smiling_face: 
Thanks!

### DIFF
--- a/read-fonts/src/tables/fvar.rs
+++ b/read-fonts/src/tables/fvar.rs
@@ -143,8 +143,14 @@ impl<'a> Fvar<'a> {
         let var_store = avar.var_store();
         let var_index_map = avar.axis_index_map();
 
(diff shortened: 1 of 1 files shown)


cmyr (approved):
makes sense, thank you!

PR #1906, opened 2026-06-08
fix(package): Remove test files from package

wmedrano:
- Skera is over the 10MB cargo package limit ([doc](https://doc.rust-lang.org/cargo/reference/publishing.html#packaging-a-crate)) so we need to exclude the test data
- Added incremental-font-transfer since it has 8MB of testdata as well.


behdad:
While at it, I also noticed that `cargo test` now takes ~90s on my laptop, mostly in the skera integration tests? This is not the best developer experience.

wmedrano:
> While at it, I also noticed that `cargo test` now takes ~90s on my laptop, mostly in the skera integration tests? This is not the best developer experience.

I would leave that for a separate PR. But seems possible, will look into it.

- I prefer `cargo nextest run -p skera`, still slow here but easier to read
- We could probably disable the integration tests and require a cli flag to enable them

PR #1911, opened 2026-06-08
Skip slow skera integration test by default

wmedrano:
- The tests are still run in the CI
- Developers need to run `cargo test -p skera -- --ignored` to trigger a local manual run

### DIFF
--- a/skera/tests/integration_test.rs
+++ b/skera/tests/integration_test.rs
@@ -676,16 +676,30 @@ fn compare_with_expected(output_dir: &Path, output_file: &Path, expected_file: &
     }
 }
 
+const LAYOUT_NOTONASTALIQURDU_TEST: &str = "tests/layout.notonastaliqurdu.tests";
+
 #[rstest]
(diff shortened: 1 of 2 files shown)


skera/tests/integration_test.rs
```diff
@@ -695,6 +700,25 @@ fn run_all_tests() {
     }
 }
 
+#[test]
+#[ignore = r#"Slow integration test, see https://github.com/googlefonts/fontations/issues/1910.
+To run manually: cargo test -p skera -- --ignored"#]
```
wmedrano, line 696: I need to make sure the error messages are fine with the multithreading in https://github.com/googlefonts/fontations/compare/wm-skera2

But if this should give a 4x speedup to layout.notonastaliqurdu.tests if that's good enough for you. I still sent this out since ~90s -> ~20s still isn't great

behdad: ~90s -> ~13s for me. Nice.

behdad (approved):

cmyr (approved):
as long as we're still running the slow tests in CI this lgtm :)

PR #1913, opened 2026-06-08
fix(build): Add GlyphId24 support to otexplorer

wmedrano:


behdad:
Lgtm. Thanks. Sorry about this.

PR #1915, opened 2026-06-09
[skrifa] report adjusted advances for CFF(2) fonts

dfrg:
We previously punted on computing (hinted) advances widths during drawing for CFF and CFF2 fonts, forcing the client to use linear glyph metrics.

This changes our CFF drawing path to populate `AdjustedMetrics::advance_width` using the width from the charstring if available, falling back to DefaultWidthX from the Private DICT or to hmtx/HVAR.


skrifa/src/outline/mod.rs
```diff
@@ -163,6 +163,13 @@ pub struct AdjustedMetrics {
     /// If present, an adjusted advance width value generated by the
     /// scaler.
     ///
+    /// Note that, like FreeType, this value is always rounded when hinting
```
dfrg, line 166: @drott you asked for additional clarity in the docs. Please let me know if this is sufficient

drott: Further above:

```
 // If charstring eval returned a width, add the nominal width
            // from the Private DICT
            .map(|w| w + subfont.nominal_width)
            // Otherwise, try the default width from the Private DICT
            .or(subfont.default_width)
            // If all else fails, fall back to hmtx/HVAR tables
```

Is the linear advance from `[`GlyphMetrics`](crate::metrics::GlyphMetrics)` also accessing the charstring eval results and private dict based computation, or is that only `hmtx/HVAR`?

drott (approved):
LGTM, with a question below. Docs look good, good that the rounding is mentioned. I am wondering if the raw value from CFF glyph processing (after dict summation, charstring eval) is accessible before rounding, but perhaps that's not needed if hinting is requested.

skrifa/src/outline/mod.rs
dfrg: `GlyphMetrics` only uses hmtx/HVAR. In a properly constructed font, the unscaled widths in charstrings and hmtx should be identical anyway. I chose to use charstring widths when loading outlines because 1) that’s what FreeType does and 2) they’re essentially free since we have to read them anyway to correctly parse charstrings.

dfrg:
> I am wondering if the raw value from CFF glyph processing (after dict summation, charstring eval) is accessible before rounding, but perhaps that's not needed if hinting is requested.

We have lower level accessors in read-fonts that provide this. `CffFontRef::evaluate_charstring` will return the width including dict adjustments and `ps::cs::evaluate` returns the raw width as encoded in the charstring.

For Type1 and pure CFF fonts, widths must be read from charstrings because there are no metrics tables. AFAIK, CFF2 only exists in sfnt containers and widths are never included in charstrings.

PR #1916, opened 2026-06-09
[read-fonts] use real segment start code in cmap4 iterator

metsw24-max:
**Wrong glyph for codepoints in an overlapping cmap4 segment**

When two format 4 segments overlap, `Cmap4Iter` clamps the later segment's iteration range up to the previous end so a codepoint is not emitted twice, but it was also taking that clamped value as the segment start code. For `idRangeOffset` segments `lookup_glyph_id` needs the real `startCode[i]` to index the glyph id array, so codepoints in the clamped tail resolved to the wrong glyph and disagreed with `map_codepoint`, which always uses the true start code. The clamp now applies only to the iteration range and the real start code is kept, which is how the format 12/13 iterator already handles the same case.


metsw24-max:
Pushed a fix for the Rustfmt failure, just ran the formatter over the test assertion. CI should be green now.

cmyr (approved):
good catch, thank you!

PR #1917, opened 2026-06-09
[subset] bug fix: always generate advMap in HVAR/VVAR even it's empty

qxliu76:


cmyr (approved):
It would be useful for these bug fix patches to include in the commit message or the PR itself some context about what the bug is, and how the patch fixes it? I really can't tell much from the diff itself what is going on, and the test cases also don't really tell me much...

behdad:
I also am curious.

qxliu76:
I updated commit message. For context: in fonttools, avdMap is not always generated, fonttools built once without AdvMap, and once with it, and choose the smaller one. HB always generates advMap.

For this fix: previously serialize_subset() call would panic if original advMap is empty because subset trait requires a non-empty `&self `input. The fix is to use/implement serialize trait instead of subset trait, so that we could still serialize an advMap as long as we have a non-empty plan

behdad:
Does this mean subsetting a font with the equivalent of `--keep-everything` can enlarge the `HVAR` table?

qxliu76:
> Does this mean subsetting a font with the equivalent of `--keep-everything` can enlarge the `HVAR` table?

yeah, it's likely in some cases.

PR #1918, opened 2026-06-10
[read-fonts] read the full flag run in read_points_fast

metsw24-max:
read_points_fast reads the simple-glyph flags as a slice capped at num_points bytes, but the flag run uses up to two bytes per point (a flag plus a repeat count), so the encoded flags can be longer than the point count. when a glyph stores each point as its own one-element repeat run the flags iterator runs dry mid-decode and read_points_fast returns OutOfBounds, even though the points() iterator decodes the same glyph without complaint. reading across the available glyph data instead lets the existing point counter end the loop, and read_flags_bytes still marks where the x/y coordinates begin, so the fast path lines back up with the slow one. the regression builds a three-point glyph whose flags are stored as length-one repeat runs.

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -219,8 +219,10 @@ impl<'a> SimpleGlyph<'a> {
             return Err(ReadError::InvalidArrayLen);
         }
         let mut cursor = FontData::new(self.glyph_data()).cursor();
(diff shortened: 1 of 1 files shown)


metsw24-max:
Pushed a fix for the Rustfmt failure. It was just an over-long line in the new test that needed wrapping, no logic changes.

cmyr (approved):
good catch, thank you!

PR #1924, opened 2026-06-15
Unify FontRead[WithArgs]

cmyr:
This is something that had been suggested by @rsheeter a while ago, and while I'm making other significant changes in read-fonts it felt like a good time to explore making this change as well. The specific motivation was that having these two separate reading traits complicated the trait hierarchy of the sanitize code.

It's still a bit of a scary patch? Ultimately this is _mostly_ internal API, and this patch should avoid most breakage; the only thing that this does break is that custom impls of FontRead, if they exist, will now require specifying `Args=()` and implementing `read_with_args` instead of `read`.

a `read` method is still available, provided via the trait when Args = ().

main commit message:

This combines these three traits into a single FontRead trait.
`read_with_args` becomes the default constructor, and a `read`
constructor is provided for free for the (general case) where Args=().

This fixes a longstanding source of headaches, where we had a blanket impl
of `FontReadWithArgs<Args=()>` for all `T: FontRead` types; we needed
this because we want there to be a single trait that represent anything
that can be read, for use in types like `ArrayOfOffsets`.

The annoyance with this, though, is that it could produce annoying
errors elsewhere (you couldn't impl FontRead outside of read-fonts,
because the blanket impl introduced a coherence restriction), and generally
complicated the trait heirarchy.

We also fold ReadArgs into this trait. ReadArgs existed so that ComputeSize
could share the Args type without a lifetime; ComputeSize instead gains a
(vestigial) lifetime and FontRead as a supertrait, which lets bounds like
`T: FontRead<'a> + ComputeSize` collapse to `T: ComputeSize<'a>`.

Downsides:

The one downside of this trait is that all FontRead impls now need to
specify the Args associated type, even though in the majority of cases
Args is (). Default associated types are not supported yet, so this is
boilerplate.


read-fonts/src/read.rs
```diff
@@ -8,57 +8,49 @@ use crate::font_data::FontData;
...
-    fn read_with_args(data: FontData<'a>, _: &Self::Args) -> Result<Self, ReadError> {
-        Self::read(data)
-    }
+/// A trait for a type that needs additional arguments to be read.
+///
+/// Types that do not require any external state use `()` as their args.
+///
+/// This is separate from [`FontRead`] so that it can also be a supertrait of
+/// [`ComputeSize`], which does not need a lifetime.
+pub trait ReadArgs {
```
dfrg, line 52: _Could_ we just just make `ComputeSize<'a>: FontRead<'a>` and drop the `ReadArgs` trait? I assume we don't need to compute sizes for anything that can't be read. Even even it makes `ComputeSize` impls a bit uglier, that might be worth deleting the additional args trait?

dfrg (approved):
One small question about `ReadArgs`, but otherwise code changes LGTM and I really like this unification.

read-fonts/src/read.rs
cmyr: It becomes tricky later, because we also need the args for Sanitize, and in the Sanitize world FontRead is only impl'd in terms of Sanitize, and then where do the Args go?

dfrg: ah, good point. carry on!

PR #1926, opened 2026-06-16
[subset] add additional sorting after extension promotion

qxliu76:
assign_spaces() requires correct topological ordering to find space roots, so add an additional topo sorting after extension promotion and table splitting to keep thing sorted. Ported from https://github.com/harfbuzz/harfbuzz/pull/6033


dfrg (approved):
LGTM!

wmedrano (approved):

PR #1929, opened 2026-06-19
Reduce SubsetDefinition clones in IFT

wmedrano:
# Benchmark Results

| Benchmark                                | Time      | Change   | Status    |
|------------------------------------------|-----------|----------|-----------|
| select-all/roboto/2-patches              | 353.14 µs | -6.7629% | Improved  |
| select-all/notosanshigh/877-patches      | 1.4781 ms | -4.4606% | Improved  |
| apply-all/roboto/7-rounds                | 8.3287 ms | -2.0637% | Improved  |
| apply-all/notosanshigh/4-rounds          | 30.008 ms | -1.0130% | Noise     |
| apply-incremental/roboto/initial         | 184.60 µs | -10.191% | Improved  |
| apply-incremental/roboto/+vietnamese     | 898.99 µs | -6.2517% | Improved  |
| apply-incremental/roboto/+greek-cyrillic | 1.3571 ms | -4.1702% | Improved  |
| incremental/notosanshigh/initial         | 4.4191 ms | -0.9452% | No change |
| incremental/notosanshigh/+full-article   | 5.1852 ms | -0.5272% | Noise     |

### DIFF
--- a/incremental-font-transfer/src/patchmap.rs
+++ b/incremental-font-transfer/src/patchmap.rs
@@ -123,7 +123,7 @@ fn add_intersecting_format1_patches(
             })?;
         let intersection_info = if PatchFormat::is_invalidating_format(map.patch_format()) {
             IntersectionInfo::from_subset(
-                subset_def,
+                &subset_def,
                 // For format 1 the entry index is the "order",
                 // see: https://w3c.github.io/IFT/Overview.html#font-patch-invalidations
                 index.into(),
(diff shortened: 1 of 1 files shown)


incremental-font-transfer/src/patchmap.rs
```diff
@@ -384,17 +384,37 @@ fn merge_intersecting_entries<const RECORD_INTERSECTION: bool>(
...
+    }
+
     /// Returns true if the target_subset_definition intersects the entry at index.
     fn intersects(&mut self, index: usize) -> bool {
-        if let Some(result) = self.cache.get(&index) {
+        if let Some(CacheEntry {
+            intersects: Some(result),
+            ..
+        }) = self.cache.get(index)
+        {
```
cmyr: if we have no cache entry then the index is invalid no?

incremental-font-transfer/src/patchmap.rs
```diff
@@ -384,17 +384,37 @@ fn merge_intersecting_entries<const RECORD_INTERSECTION: bool>(
     }
 }
 
+#[derive(Clone, Default)]
+struct CacheEntry {
+    intersects: Option<bool>,
```
cmyr, line 390: this deserves a bit of docs, as it's a little counter intuitive? I'd normally expect the `CacheEntry` itself to be missing if there was no entry, instead of us always having an entry and the individual values being unset...

cmyr (approved):
couple little thoughts but overall looks good!

incremental-font-transfer/src/patchmap.rs
wmedrano: We make a `CacheEntry` for each `Entry` so this would be invalid.

This feature is not productionized yet so we can tweak the spec still. On a well formed font, this is not an issue. However, an IFT font can have an entry that references an out of bounds index.

incremental-font-transfer/src/patchmap.rs
wmedrano: Done. Renamed to EntryCaches since there are 2 caches on the inside.

PR #1931, opened 2026-06-19
[skrifa]: improve bitmap glyph search

fundon:
No clone required


cmyr:
I don't think this is a meaningful improvement, the type we're cloning is comprised of copy-friendly types, and removing that single clone should not have meaningful performance implications. More importantly adding the `&'a self` bound meaningfully changes the semantics, making this a breaking change.

PR #1934, opened 2026-06-21
Sanitize Part I

cmyr:
_this PR is based on #1924_ 

This adds the `Sanitize` trait, and generates impls for a bunch of our test inputs.

The sanitize behaviour is opt-in, and this does not opt-in for any real tables; that will happen in a follow-up.

The trait and impl are very simple; it is currently only concerned with validating table data, not dealing with things like cycle detection or mitigating DOS surface area. That work can also come as a followup.

### DIFF
--- a/font-codegen/src/parsing/attrs.rs
+++ b/font-codegen/src/parsing/attrs.rs
@@ -97,6 +97,8 @@ pub(crate) struct FieldAttrs {
     pub(crate) discriminant: Option<syn::Path>,
     /// During sanitize, only check the length of this field (don't recurse).
     pub(crate) sanitize_len_only: Option<syn::Path>,
+    /// Custom sanitize fn, like compile_with but for sanitize.
+    pub(crate) sanitize_with: Option<Attr<SanitizeWith>>,
 }
 
 #[derive(Clone, Debug, Default)]
@@ -125,6 +127,12 @@ pub(crate) struct TableReadArg {
 
 #[derive(Debug, Clone)]
 pub(crate) struct FieldReadArgs {
+    pub(crate) inputs: Vec<syn::Ident>,
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct SanitizeWith {
+    pub(crate) fn_name: syn::Ident,
     pub(crate) inputs: Vec<syn::Ident>,
 }
 
@@ -282,6 +290,7 @@ static TO_OWNED: &str = "to_owned";
 static VALIDATE: &str = "validate";
 static DISCRIMINANT: &str = "discriminant";
 static SANITIZE_LEN_ONLY: &str = "sanitize_len_only";
+static SANITIZE_WITH: &str = "sanitize_with";
 
 static MATCH_IF: &str = "match_if";
 static WRITE_FONTS_ONLY: &str = "write_fonts_only";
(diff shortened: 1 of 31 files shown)


font-codegen/src/parsing/attrs.rs
```diff
@@ -95,6 +95,10 @@ pub(crate) struct FieldAttrs {
     pub(crate) validate: Option<Attr<FieldValidation>>,
     /// Marks this field as the discriminant for a generic offset type.
     pub(crate) discriminant: Option<syn::Path>,
+    /// During sanitize, only check the length of this field (don't recurse).
+    pub(crate) sanitize_len_only: Option<syn::Path>,
```
behdad, line 99: If it's of any use, in HB we call this `sanitize_shallow`.

read-fonts/src/sanitize.rs
```diff
@@ -0,0 +1,364 @@
...
+    state: &'a mut SanitizeState,
+}
+
+/// State tracked during during a sanitize pass
+#[derive(Clone, Debug, Default)]
+pub(crate) struct SanitizeState {
+    // only used in COLRv1
+    _recursion_depth: u32,
+    // gpos/gsub
+    _subtable_depth: u32,
```
behdad, line 27: Not sure off the top of my head what this would be used for. For sanitizing, we don't need to recurse into lookups from contextual types. And extension lookups do not allow their contained lookup itself to be an extension. So, the tree structure in GSUB/GPOS is quite shallow and not unbounded. COLRv1 on the other hand is an unbounded graph structure.

read-fonts/src/sanitize.rs
```diff
@@ -0,0 +1,350 @@
...
+        count: u16,
+        len_only: bool,
+        f: F,
+    ) -> Result<(), ReadError>
+    where
+        O: Offset + Scalar,
+        T: AnyBitPattern + FixedSize,
+        F: Fn(&T, &mut SanitizeContext) -> Result<(), ReadError>,
+    {
+        let offset = self.read::<O>()?;
```
dfrg, line 148: Could we just call `self.sanitize_resolved_offset_to_array(offset)` here to avoid the duplicated code or do we expect the logic in these two methods to diverge at some point?

read-fonts/src/sanitize.rs
```diff
@@ -0,0 +1,350 @@
...
+        &mut self,
+        count: usize,
+        recurse: bool,
+    ) -> Result<(), ReadError>
+    where
+        T: VarSize + SanitizeStruct<Args = ()> + FontRead<'a>,
+    {
+        let remaining = self.cursor.remaining().ok_or(ReadError::OutOfBounds)?;
+        let total_len = T::total_len_for_count(remaining, count)?;
+        // FIXME: do we actually ever recurse here?
```
dfrg: The `morx` table uses `VarLenArray` with table elements. In fact, `Morx` has `VarLenArray<Chain>` and `Chain` has `VarLenArray<Subtable>` so we definitely have nested data inside these.

read-fonts/src/sanitize.rs
```diff
@@ -0,0 +1,350 @@
...
+        //about DDOS that doesn't touch offsets?
+        self.cursor.position().map(|_| ())
+    }
+}
+
+// okay so: the sanitize context _does_ kinda want to mutate, and it wants to mutate
+// alongside access to the table data? probably?
+//
+// the annoying bit here is going to be figuring out how we keep track of our
+// position as we... do stuff?
```
dfrg: Are these still open questions?

dfrg (approved):
A few questions but otherwise looks good!

read-fonts/src/sanitize.rs
cmyr: good catch!

read-fonts/src/sanitize.rs
cmyr: good call!

PR #1935, opened 2026-06-22
Parse base32hex and base64url encodings on demand

wmedrano:
- Encodings (split off into `encode_id32` and `encode_id64` are run and cached only when needed.
  - This helps as most template expansions only use one. Caching used as there are instances where and id is needed twice and this avoids the edge case anyways.
- The interpreter loop was also inlined. Seems more readable as the instruction set is relatively small.

# Benchmark Results

| Benchmark | Time | Change | Result |
| :--- | :--- | :--- | :--- |
| select-all/roboto/2-patches | 339.30 µs | -5.4227% | Improved |
| select-all/notosanshigh/877-patches | 1.4533 ms | -2.6854% | Improved |
| apply-all/roboto/7-rounds | 8.1691 ms | -1.4598% | Improved |
| apply-all/notosanshigh/4-rounds | 29.511 ms | -1.1423% | Noise |
| apply-incremental/roboto/initial | 168.36 µs | -15.816% | Improved |
| apply-incremental/roboto/+vietnamese | 838.59 µs | -6.9244% | Improved |
| apply-incremental/roboto/+greek-cyrillic | 1.2760 ms | -5.6375% | Improved |
| incremental/notosanshigh/initial | 4.2660 ms | -2.7645% | Improved |
| incremental/notosanshigh/+full-article | 5.1075 ms | -1.6747% | Improved |

### DIFF
--- a/incremental-font-transfer/src/url_templates.rs
+++ b/incremental-font-transfer/src/url_templates.rs
@@ -69,25 +69,45 @@ impl TryFrom<u8> for Variable {
     }
 }
 
-fn encode_id32(patch_id: &PatchId) -> String {
-    match patch_id {
-        PatchId::Numeric(id) => {
-            let id = id.to_be_bytes();
-            let id = &id[count_leading_zeroes(&id)..];
-            BASE32HEX_NO_PADDING.encode(id)
-        }
-        PatchId::String(id) => BASE32HEX_NO_PADDING.encode(id),
-    }
+struct CachedEncoder<'a> {
+    patch_id: &'a PatchId,
+    id32: Option<String>,
+    id64: Option<String>,
 }
 
-fn encode_id64(patch_id: &PatchId) -> String {
-    match patch_id {
-        PatchId::Numeric(id) => {
-            let id = id.to_be_bytes();
-            let id = &id[count_leading_zeroes(&id)..];
-            BASE64URL.encode(id)
+impl<'a> CachedEncoder<'a> {
+    fn new(patch_id: &'a PatchId) -> Self {
+        CachedEncoder {
+            patch_id,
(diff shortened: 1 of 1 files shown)


cmyr (approved):
The numbers are nice, but I find the new control flow to be much harder to follow. I'd much prefer an API that looked something like,

```rust
/// Each encoding is used at most once; an encoding can be unused.
#[derive(Default)]
struct CachedEncoder {
    cached_id32: Option<String>,
    cached_id64: Option<String>,
}

impl CachedEncoder {
    fn get_id32(&mut self, patch_id: &PatchId) -> &str {
        fn encode_id32(patch_id: &PatchId) -> String {
            match patch_id {
                PatchId::Numeric(id) => {
                    let id = id.to_be_bytes();
                    let id = &id[count_leading_zeroes(&id)..];
                    BASE32HEX_NO_PADDING.encode(id)
                }
                PatchId::String(id) => BASE32HEX_NO_PADDING.encode(id),
            }
        }
        self.cached_id32
            .get_or_insert_with(|| encode_id32(patch_id))
            .as_str()
    }

    fn get_id64(&mut self, patch_id: &PatchId) -> &str {
        fn encode_id64(patch_id: &PatchId) -> String {
            match patch_id {
                PatchId::Numeric(id) => {
                    let id = id.to_be_bytes();
                    let id = &id[count_leading_zeroes(&id)..];
                    BASE64URL.encode(id)
                }
                PatchId::String(id) => BASE64URL.encode(id),
            }
        }
        self.cached_id64
            .get_or_insert_with(|| encode_id64(patch_id))
            .as_str()
    }
}
```

and then we can get rid of the macro at the call site?

PR #1937, opened 2026-06-22
`font-types`: serde fixed floats as their value not repr

### LINKED ISSUE #1936: `font-types`: `Fixed` serialises as `i32` with `serde`
`Fixed` currently is serialised as an `i32` with `serde`, instead of the underlying fixed point number that the `i32` is used to represent.

This results in things like `post.italic_angle` of -851968.

Would you be open to a PR that adds a custom `Serialize`/`Deserialize` impl that stores the numbers as their underlying value, instead of their binary representation?

If there are any other types that would need similar work, let me know.

RickyDaMa:
Fixes #1936 for all fixed-point types

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -632,11 +632,27 @@ mod tests {
         }
 
         #[test]
-        fn one_is_one() {
+        fn one_is_one_f2dot14() {
             roundtrip_one!(F2Dot14);
+        }
+
+        #[test]
+        fn one_is_one_f4dot12() {
             roundtrip_one!(F4Dot12);
+        }
+
+        #[test]
+        fn one_is_one_f6dot10() {
             roundtrip_one!(F6Dot10);
+        }
+
+        #[test]
+        fn one_is_one_fixed() {
             roundtrip_one!(Fixed);
+        }
+
+        #[test]
+        fn one_is_one_f26dot6() {
             roundtrip_one!(F26Dot6);
         }
     }
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -606,4 +606,38 @@ mod tests {
         assert_eq!(F26Dot6::ONE / F26Dot6::ZERO, F26Dot6(0x7FFFFFFF));
         assert_eq!(-F26Dot6::ONE / F26Dot6::ZERO, F26Dot6(-0x7FFFFFFF));
     }
+
+    #[cfg(feature = "serde")]
+    mod serde {
+        use super::*;
+
+        macro_rules! roundtrip_one {
+            ($fixed:ident) => {{
+                let before = <$fixed>::ONE;
+                let serialized = ::serde_json::to_string(&before).expect("should serialize");
+                assert_eq!(
+                    &serialized,
+                    "1.0",
+                    "{}::ONE didn't serialize to 1.0",
+                    ::std::stringify!($fixed),
+                );
+                let after = ::serde_json::from_str(&serialized).expect("should deserialize");
+                assert_eq!(
+                    before,
+                    after,
+                    "{}::ONE doesn't round-trip",
+                    ::std::stringify!($fixed),
+                );
+            }};
+        }
+
+        #[test]
+        fn one_is_one() {
+            roundtrip_one!(F2Dot14);
+            roundtrip_one!(F4Dot12);
+            roundtrip_one!(F6Dot10);
+            roundtrip_one!(Fixed);
+            roundtrip_one!(F26Dot6);
+        }
+    }
 }
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -288,7 +288,7 @@ macro_rules! float_conv {
     // explicitly opt out of Display/Default (for types that get both f32 & f64)
     ($name:ident, $to:ident, $from:ident, $ty:ty, no_fmt) => {
         impl $name {
-            #[doc = concat!("Creates a fixed point value from a", stringify!($ty), ".")]
+            #[doc = concat!("Creates a fixed point value from a ", stringify!($ty), ".")]
             ///
             /// This operation is lossy; the float will be rounded to the nearest
             /// representable value.
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -655,5 +655,57 @@ mod tests {
         fn one_is_one_f26dot6() {
             roundtrip_one!(F26Dot6);
         }
+
+        macro_rules! roundtrip_all {
+            ($fixed:ident, $ty:ty) => {
+                for raw in <$ty>::MIN..=<$ty>::MAX {
+                    let fixed = $fixed(raw);
+                    let fixed_float = fixed.to_f64();
+
+                    let json_value = ::serde_json::to_value(&fixed).expect("should serialize");
+                    let Some(json_float) = json_value.as_f64() else {
+                        panic!(
+                            "serde didn't serialize {} as a float",
+                            ::std::stringify!($fixed)
+                        );
+                    };
+
+                    // Normally directly comparing floats is flawed, but these
+                    // should have been converted to float using the exact same
+                    // method each, so I wouldn't expect them to be different
+                    assert_eq!(
+                        fixed_float, json_float,
+                        "failed on {raw}: {json_float} != {fixed_float}"
+                    );
+                }
+            };
+        }
+
+        #[test]
+        fn roundtrip_all_f2dot14() {
+            roundtrip_all!(F2Dot14, i16);
+        }
+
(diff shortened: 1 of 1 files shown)


font-types/src/fixed.rs
```diff
@@ -586,4 +606,38 @@ mod tests {
...
+
+        #[test]
+        fn one_is_one() {
+            roundtrip_one!(F2Dot14);
+            roundtrip_one!(F4Dot12);
+            roundtrip_one!(F6Dot10);
+            roundtrip_one!(Fixed);
+            roundtrip_one!(F26Dot6);
+        }
+    }
```
cmyr, line 699: just for my edification I did,

```rust
macro_rules! roundtrip_all {
            ($fixed:ident, $ty:ty) => {
                let mut buf = Vec::new();
                for raw in <$ty>::MIN..=<$ty>::MAX {
                    let val = $fixed(raw);
                    serde_json::to_writer(&mut buf, &val).unwrap();
                    let de = serde_json::from_slice(&buf).unwrap();
                    assert_eq!(val, de, "failed on {raw}: {val:?} != {de:?}");
                    buf.clear();
                }
            };
        }

        #[test]
        fn roundtrip_all_f2dot14() {
            roundtrip_all!(F2Dot14, i16);
        }

        #[test]
        fn roundtrip_all_f4dot12() {
            roundtrip_all!(F4Dot12, i16);
        }

        #[test]
        fn roundtrip_all_f6dot10() {
            roundtrip_all!(F6Dot10, i16);
        }

        // disabled because they're slow, but they do both pass
        //#[test]
        //fn roundtrip_all_fixed() {
        //roundtrip_all!(Fixed, i32);
        //}
        //#[test]
        //fn roundtrip_all_f26dot6() {
        //roundtrip_all!(F26Dot6, i32);
        //}
```

but this _is_ slow (like ~40m slow) because json is slow; round-tripping just to/from floats is ~100x faster, and does capture the main thing we're actually worried about here.

font-types/src/fixed.rs
```diff
@@ -586,4 +606,38 @@ mod tests {
...
+                    before,
+                    after,
+                    "{}::ONE doesn't round-trip",
+                    ::std::stringify!($fixed),
+                );
+            }};
+        }
+
+        #[test]
+        fn one_is_one() {
```
cmyr: i'd personally probably prefer to have a test case for each type, which will also clean up the macro a bit? see below for not-exactly-the-same-idea.

cmyr (approved):
this is definitely a bug and should fix it; my only nit is that I find the testing a bit... well both a bit fancy and also a bit unconvincing?

ultimately though to be confident in this we really just need to be convinced of two things:
- that float -> $FIXED_TYPE round-tripping works correctly,
- and that serde is actually serializing floats.

in any case this looks good, there are some nits below but i'm happy to merge.

font-types/src/fixed.rs
RickyDaMa: I've split them out into separate tests. Kept them along with yours, but I can remove my set if you want.

Initially I just wanted to smash out a basic test so I had _something_ to go along with the PR, but didn't want to invest super hard in thorough tests until I knew that this was an issue we wanted to fix.

font-types/src/fixed.rs
RickyDaMa: I've included all of these (but ignored the `i32` ones by default). I did make one change which speeds things up a lot: serialising to `serde_json::Value` instead of all the way to bytes. This makes the slow tests take 3 minutes for me instead of your 40. Let me know if you're happy with how they're handled compared to your vision

font-types/src/fixed.rs
```diff
@@ -586,4 +606,108 @@ mod tests {
...
+
+        macro_rules! roundtrip_one {
+            ($fixed:ident) => {{
+                let before = <$fixed>::ONE;
+                let serialized = ::serde_json::to_string(&before).expect("should serialize");
+                assert_eq!(
+                    &serialized,
+                    "1.0",
+                    "{}::ONE didn't serialize to 1.0",
+                    ::std::stringify!($fixed),
```
cmyr: last nit: the advantage of having each in its own test case is we can remove this all and just use bare assert_eq!

font-types/src/fixed.rs
```diff
@@ -586,4 +606,108 @@ mod tests {
...
+                );
+                let after = ::serde_json::from_str(&serialized).expect("should deserialize");
+                assert_eq!(
+                    before,
+                    after,
+                    "{}::ONE doesn't round-trip",
+                    ::std::stringify!($fixed),
```
cmyr: ditto

font-types/src/fixed.rs
```diff
@@ -586,4 +606,108 @@ mod tests {
...
+                for raw in <$ty>::MIN..=<$ty>::MAX {
+                    let fixed = $fixed(raw);
+                    let fixed_float = fixed.to_f64();
+
+                    let json_value = ::serde_json::to_value(&fixed).expect("should serialize");
+                    let Some(json_float) = json_value.as_f64() else {
+                        panic!(
+                            "serde didn't serialize {} as a float",
+                            ::std::stringify!($fixed)
+                        );
```
cmyr: this is just telling us the type, but we get that info from the name of the test case? Honestly i think just an unwrap is fine.

font-types/src/fixed.rs
```diff
@@ -586,4 +606,108 @@ mod tests {
...
+                    };
+
+                    // Normally directly comparing floats is flawed, but these
+                    // should have been converted to float using the exact same
+                    // method each, so I wouldn't expect them to be different
+                    assert_eq!(
+                        fixed_float,
+                        json_float,
+                        "failed on {raw} ({fixed_type}({raw:#X})): {json_float} != {fixed_float}",
+                        fixed_type = ::std::stringify!($fixed),
```
cmyr: again just plain assert_eq should tell us all we need!

cmyr (approved):
great, going to value instead of str/bytes makes a lot of sense.

just a last couple nits but this looks good.

font-types/src/fixed.rs
RickyDaMa: I disagree on this one, if this assertion panicked the first thing I'd want to know is what `raw` or `fixed` is to reproduce without enumerating every `i16`. Admittedly I did go quite ham on making the message as helpful as possible, so I've pared it down to just `fixed`.

cmyr: I believe by default assert_eq! debug prints both of the arguments on failure?

tested:
```rust
fn main() {
    let (one, two) = (1, 2);
    assert_eq!(one, two);
}
```
outputs,

```
  left: 1
 right: 2
```

(this is the main justification for having these be standalone macros, instead of just using `assert!(left == right)`)

font-types/src/fixed.rs
RickyDaMa: Changed to an `expect`, I think the message here is valuable to someone who accidentally makes this test break but isn't aware of the wider serialisation context

font-types/src/fixed.rs
RickyDaMa: Removed

font-types/src/fixed.rs
RickyDaMa: ditto :)

font-types/src/fixed.rs
RickyDaMa: Yeah that's fair, we agree then that knowing `raw`/`fixed` is the important part then 👍

cmyr: yes that makes total sense :)

font-types/src/fixed.rs
```diff
@@ -586,4 +606,90 @@ mod tests {
...
+
+                    let json_value = ::serde_json::to_value(&fixed).expect("should serialize");
+                    let json_float = json_value
+                        .as_f64()
+                        .expect("serde didn't serialize the value to a float");
+
+                    // Normally directly comparing floats is flawed, but these
+                    // should have been converted to float using the exact same
+                    // method each, so I wouldn't expect them to be different
+                    assert_eq!(fixed_float, json_float, "inconsistent value for {fixed:?}");
```
cmyr: note that having your own custom message here suppresses the default message; and if we _do_ want a custom message what is most important is the actual raw value, and the `Debug` impl here converts to a float and prints that, which might be confusing 😅 

if we're trying to be maximally useful, I'd probably do,

```rust
assert_eq!(fixed_float, json_float, "failed on {raw}: {fixed_float} != {json_float}"
```

(that said i'm happy to just get this merged, you tell me)

RickyDaMa: I've restored my original super duper fancy assertion message for this one, left the others untouched.

cmyr (approved):
thanks!

PR #1940, opened 2026-06-23
Fix trivial clippy errors

wmedrano:
- Clippy doesn't like `Bool::then(..).unwrap_or(..)`
- `repeat(v).take(n)` replaced with `repeat_n(v, n)`
- `slice.iter().any(|x| x == &const)` replaced with `slice.contains(&const)`


cmyr (approved):
looks good. I want to replace our use of Range<usize> with some u32-based type that can actually be `Copy` and can implement some helper methods to handle some of the simple things we do with ranges, but that is currently just a dream...

PR #1941, opened 2026-06-23
Fix clippy lint

wmedrano:
Avoid Option::is_some() + Option::unwrap() pattern

# Setup

The first branch only executed if `prev_segment_ix` is `None` or it it meets some other condition.

```rust
                   if prev_segment_ix.is_none()
                        || axis.segments[segment_ix].first_ix
                            != axis.segments[prev_segment_ix.unwrap()].last_ix
```

# Fix

`prev_segment_ix` is set to `None` if the precondition is met. Then we do a match with the usual `None` and `Some` arms

```rust
    prev_segment_ix.take_if(|idx| {
                        axis.segments[segment_ix].first_ix != axis.segments[*idx].last_ix
                    });
                    match prev_segment_ix {
```

### DIFF
--- a/skrifa/src/outline/autohint/topo/segments.rs
+++ b/skrifa/src/outline/autohint/topo/segments.rs
@@ -353,20 +353,12 @@ fn build_segments(outline: &mut Outline, axis: &mut Axis) -> bool {
                     state.max_on_coord = state.max_on_coord.max(point.v);
                 }
                 if point.out_dir != segment_dir || point_ix == last_ix {
-                    if prev_segment_ix.is_none()
-                        || axis.segments[segment_ix].first_ix
-                            != axis.segments[prev_segment_ix.unwrap()].last_ix
-                    {
-                        // The points are different signifying that we are
-                        // leaving an edge, so create a new segment
-                        let segment = &mut axis.segments[segment_ix];
-                        segment.last_ix = point_ix as u16;
-                        state.apply_to_segment(segment, flat_threshold);
-                        prev_segment_ix = Some(segment_ix);
-                        prev_state = state;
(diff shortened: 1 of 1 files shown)


skrifa/src/outline/autohint/topo/segments.rs
```diff
@@ -353,63 +353,68 @@ fn build_segments(outline: &mut Outline, axis: &mut Axis) -> bool {
...
-                            if (prev_state.max_coord - prev_state.min_coord).abs()
-                                > (state.max_coord - state.min_coord).abs()
-                            {
-                                // Discard current segment
-                                prev_state.min_pos = prev_state.min_pos.min(state.min_pos);
-                                prev_state.max_pos = prev_state.max_pos.max(state.max_pos);
+                    prev_segment_ix.take_if(|idx| {
+                        axis.segments[segment_ix].first_ix != axis.segments[*idx].last_ix
+                    });
+                    match prev_segment_ix {
```
cmyr: I think this is worse control flow. Specifically the fact that we are now mutating prev_segment_ix in a fairly obscure way makes this harder to track, and it is hard to reason about whether or not this is actually equivalent.

additionally this block is already heavily nested, and the match adds an additional level.

Additionally since this code is largely (iirc) a fairly faithful port from the C, and because chad isn't here to give a +1, I'd prefer here to match the existing logic and control flow as closely as possible.



...Actually playing around, I think the best bet here is to keep your `take_if`, but then just use an `if let Some(_)` instead of a match, so that at least we avoid the extra level of right drift? 🤷

cmyr (approved):
one nit, broadly looks good :)

PR #1947, opened 2026-06-24
[subset] retain extension subtable

qxliu76:
Some implementation code for `Intersect` trait is changed as well, because we check `intersect()` result before we subset each subtable, which requires all subtables implement `Intersect` Trait.


garretrieger (approved):

cmyr (approved):
looks good!

PR #1954, opened 2026-07-03
[read-fonts] Make FloatItemDelta float accessors public

behdad:
Expose FloatItemDelta::to_f32 (new) and to_f64 (was pub(crate)) so downstream crates can read the unrounded item-variation delta.

Previously the only reachable accessor was the integer-rounded ItemVariationStore::compute_delta.  Callers that need HarfBuzz-compatible positioning must keep the delta fractional and round only once, after scaling to output units; that requires the raw float value from compute_float_delta.

Assisted by Claude Opus 4.8 (1M context)

### DIFF
--- a/read-fonts/src/tables/variations.rs
+++ b/read-fonts/src/tables/variations.rs
@@ -1554,7 +1554,8 @@ pub struct FloatItemDelta(f64);
 impl FloatItemDelta {
     pub const ZERO: Self = Self(0.0);
 
-    pub(crate) fn to_f64(self) -> f64 {
+    /// Returns the (unrounded) delta value as a 64-bit float.
+    pub fn to_f64(self) -> f64 {
         self.0
     }
 }


behdad:
Downstream user: https://github.com/harfbuzz/harfrust/pull/403

read-fonts/src/tables/variations.rs
```diff
@@ -1554,7 +1554,13 @@ pub struct FloatItemDelta(f64);
 impl FloatItemDelta {
     pub const ZERO: Self = Self(0.0);
 
-    pub(crate) fn to_f64(self) -> f64 {
+    /// Returns the (unrounded) delta value as a 32-bit float.
+    pub fn to_f32(self) -> f32 {
+        self.0 as f32
+    }
```
cmyr: is this genuinely useful, versus just exposing `to_f64` and letting the caller convert if they want?

behdad: Sure, that works.

PR #1960, opened 2026-07-08
Support more fields when writing glyphs

simoncozens:
Currently people using write-fonts to create glyph structures do not have mutable access to

* components on composite glyphs (if you have obtained a CompositeGlyph, you can't edit the flags on its components without rebuilding it from scratch)
* instructions on composite glyphs
* the overlaps bit on simple glyphs

This allows access to these three fields.

### DIFF
--- a/write-fonts/src/tables/glyf/composite.rs
+++ b/write-fonts/src/tables/glyf/composite.rs
@@ -191,6 +191,10 @@ impl CompositeGlyph {
         &self.components
     }
 
+    pub fn components_mut(&mut self) -> &mut [Component] {
+        &mut self.components
+    }
+
     pub fn add_instructions(&mut self, instructions: impl Into<Vec<u8>>) {
         self._instructions.extend(instructions.into());
     }
--- a/write-fonts/src/tables/glyf/composite.rs
+++ b/write-fonts/src/tables/glyf/composite.rs
@@ -190,6 +190,14 @@ impl CompositeGlyph {
     pub fn components(&self) -> &[Component] {
         &self.components
     }
+
+    pub fn add_instructions(&mut self, instructions: impl Into<Vec<u8>>) {
+        self._instructions.extend(instructions.into());
+    }
+
+    pub fn instructions(&self) -> &[u8] {
+        &self._instructions
+    }
 }
 
 impl FontWrite for CompositeGlyph {
--- a/write-fonts/src/tables/glyf/composite.rs
+++ b/write-fonts/src/tables/glyf/composite.rs
@@ -16,7 +16,7 @@ pub use read_fonts::tables::glyf::{Anchor, Transform};
 pub struct CompositeGlyph {
     pub bbox: Bbox,
     components: Vec<Component>,
-    _instructions: Vec<u8>,
+    instructions: Vec<u8>,
 }
 
 /// A single component glyph (part of a [`CompositeGlyph`]).
@@ -76,7 +76,7 @@ impl FromObjRef<read_fonts::tables::glyf::CompositeGlyph<'_>> for CompositeGlyph
         Self {
             bbox,
             components,
-            _instructions: from
+            instructions: from
                 .instructions()
                 .map(|v| v.to_owned())
                 .unwrap_or_default(),
@@ -148,7 +148,7 @@ impl CompositeGlyph {
         Self {
             bbox: bbox.into(),
             components: vec![component],
-            _instructions: Default::default(),
+            instructions: Default::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


write-fonts/src/tables/glyf/composite.rs
```diff
@@ -190,6 +190,18 @@ impl CompositeGlyph {
     pub fn components(&self) -> &[Component] {
         &self.components
     }
+
+    pub fn components_mut(&mut self) -> &mut [Component] {
+        &mut self.components
+    }
+
+    pub fn add_instructions(&mut self, instructions: impl Into<Vec<u8>>) {
```
cmyr: Genuine question, is 'append' useful semantics for instructions? I would naively expect that you generally set them in full?


on a separate note I'm not sure `impl Into<Vec<u8>>` makes a ton of sense as the argument type: unless I'm missing something it's less general than `&[u8]` and forces an unnecessary allocation in some cases.

write-fonts/src/tables/glyf/composite.rs
```diff
@@ -190,6 +190,18 @@ impl CompositeGlyph {
...
+        self._instructions.extend(instructions.into());
```
cmyr: while we're in here we might as well rename `_instructions` to `instructions`? I think it was unused when this was first written.

write-fonts/src/tables/glyf/simple.rs
```diff
@@ -283,7 +285,12 @@ impl FontWrite for SimpleGlyph {
         (self.instructions.len() as u16).write_into(writer);
         self.instructions.write_into(writer);
 
-        let deltas = self.compute_point_deltas().collect::<Vec<_>>();
+        let mut deltas = self.compute_point_deltas().collect::<Vec<_>>();
+        if self.overlaps {
+            if let Some((flags, _, _)) = deltas.first_mut() {
+                flags.insert(SimpleGlyphFlags::OVERLAP_SIMPLE);
+            }
+        }
```
cmyr, line 296: why don't we just set this in `compute_point_deltas` itself?

In a dream world we would compute whether or not there are overlaps ourselves and not require storing the flag, but I can understand that not being with the complication.

write-fonts/src/tables/glyf/simple.rs
```diff
@@ -19,6 +19,7 @@ pub struct SimpleGlyph {
     pub bbox: Bbox,
     pub contours: Vec<Contour>,
     pub instructions: Vec<u8>,
+    pub overlaps: bool,
```
cmyr, line 25: does this need to be public? Are there cases where we genuinely expect the user to manually set this?

If we think it _should_ be public than we should document that it means we will set the OVERLAP_SIMPLE bit on the first flag in the glyph.

cmyr (commented):
A few notes inline, but makes sense that these need more public API as they see more use.

write-fonts/src/tables/glyf/simple.rs
simoncozens: I think that's (a) out of scope for this library - would you *really* want to be doing vector operations inside of write-fonts?, and (b) at the wrong level; this is about low-level control of the font structure.

write-fonts/src/tables/glyf/simple.rs
simoncozens: Yes, absolutely; see rationale above. I'll add documentation.

write-fonts/src/tables/glyf/simple.rs
cmyr: agree about not doing the vector ops here.

What about my first point, setting this in `compute_point_deltas`?

simoncozens: If I want to use this to write fonts as an end-user, I want to be able to have control over whether the flag is set or not. At the moment, I don't have that control. Your alternative suggestion is that I still don't get control. It's better than not being able to set the flag at all but only slightly. What about *my* idea, allowing the user to decide? The spec says that the use of the flag is not required, i.e. it is valid to have overlapping contours with the flag set, and with the flag unset - your approach would be to disallow one of those valid options. I don't understand why.

cmyr: I think you've misunderstood me? I'm not suggesting removing the `overlaps` flag, I'm just suggesting that instead of checking `self.overlaps` and mutating the flags here, we instead set the flag correctly inside the `compute_point_deltas` method. This is a very minor difference and certainly not a blocker, it just seems slightly more natural.

cmyr (approved):
I thought I'd approved this already, but lgtm in any case :)

PR #1963, opened 2026-07-09
[skrifa] tt hint: limit number of skipped instructions

dfrg:
We currently limit the number of executed instructions but don't track the number of instructions skipped when handling conditional regions. A maliciously designed font can exhaust an excessive number of cycles by carefully constructing IF/ELSE regions.

This adds a counter to track the number of skipped instructions and reports an error when we surpass the limit. The budget is the same as the execution budget (1M instructions) but tracked separately, effectively limiting our instruction _decode_ count to 2M per run of the interpreter.


cmyr (approved):
makes sense. Any context on this? Did it come out of fuzzing?

dfrg:
This comes from Google internal vulnerability analysis.

PR #1964, opened 2026-07-09
[skrifa] tt hint: overflow/OOB fixes

dfrg:
A small assortment of changes to prevent panic on overflowing arithmetic or out of bounds indexing. This certainly isn't the last of these bugs but we'll fix 'em and we find 'em.

These were identified by an internal vulnerability analysis tool.


cmyr (approved):
🫡

PR #1974, opened 2026-07-21
Make vhea version default, not computed

### LINKED ISSUE #1972: [write-fonts] vhea version is fixed
When writing a `vhea` table with `write-fonts`, the version number is predetermined to be 1.1:

https://github.com/googlefonts/fontations/blob/bbedcb2167dbd3a0d887325fdb720808072624df/resources/codegen_inputs/vhea.rs#L6-L8

But version 1.0 is [also available](https://learn.microsoft.com/en-us/typography/opentype/spec/vhea#table-formats), and while the structure is the same, the interpretation of the fields is subtly different. More to the point, calling `.to_owned_table()` on a version 1.0 table "upgrades" it silently to 1.1, which is surprising for anything which compares fonts.

simoncozens:
Fixes #1972.

### DIFF
--- a/read-fonts/generated/generated_vhea.rs
+++ b/read-fonts/generated/generated_vhea.rs
@@ -61,7 +61,7 @@ impl<'a> Vhea<'a> {
         + u16::RAW_BYTE_LEN);
     basic_table_impls!(impl_the_methods);
 
-    /// The major/minor version (1, 1)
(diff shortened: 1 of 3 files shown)


cmyr (approved):
Seems reasonable, I don't recall why I assumed this would always be 1.1.

PR #1980, opened 2026-07-23
[skrifa] avoid infinite loop in autohinter

dfrg:
The long blues calculation was missing a break condition causing an infinite loop.

Fixed the bug and added a test that ensures the operation completes within a reasonable time.

internal bug ref: b/537782944

### DIFF
--- a/skrifa/src/outline/mod.rs
+++ b/skrifa/src/outline/mod.rs
@@ -776,6 +776,7 @@ mod tests {
     use super::*;
     use crate::{instance::Location, outline::pen::SvgPen, MetadataProvider};
     use kurbo::{Affine, BezPath, PathEl, Point};
+    use raw::model::pen::NullPen;
     use read_fonts::{types::GlyphId, FontRef, TableProvider};
 
     use pretty_assertions::assert_eq;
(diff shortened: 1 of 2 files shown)


rsheeter (approved):

skrifa/src/outline/mod.rs
```diff
@@ -1620,4 +1621,49 @@ mod tests {
...
+        )
+        .unwrap();
+        let gid = font.charmap().map('\u{05D1}').unwrap();
+        let outline = outlines.get(gid).unwrap();
+        let (tx, rx) = std::sync::mpsc::channel();
+        std::thread::spawn(move || {
+            outline.draw(&hinter, &mut NullPen).unwrap();
+            let _ = tx.send(());
+        });
+        let budget = std::time::Duration::from_secs(5);
```
rsheeter, line 1661: One day I'll grow up and be a flaky test. Ideally we'd have a work counter or something and not have to do this. For now ... I'm assuming 5 seconds is *way* too long so it shouldn't often naturally overrun?

dfrg: If it ever takes us longer than 5s to load a single glyph then something is very broken :)

drott: We do not run skrifa tests in chrome, but indeed timeouts lead to flakiness as in Chrome the bot configuration runs test highly parallel, and scheduling can pause a test for a long time, without necessarily much anything being broken, but just not on the CPU while other tests run. So I'd second the wish to have a non time-based threshold here.

dfrg: Ack. Filed #2030 to track adding a work budget for the autohinter as a better strategy for catching bugs like this.

PR #1983, opened 2026-07-24
[read-fonts] optimize post glyph name iter

dfrg:
Custom glyph names in the `post` table are stored as an array of indices which reference a position in sequence of length prefixed (Pascal) strings. The "naive" approach to iteration, calling `glyph_name` for each gid, is quadratic over number of glyphs and leads to poor performance in the general case and potentially multi-second delays in the pathological case (all indices point to last name).

FreeType and HarfBuzz both use a precomputed, heap allocated offset buffer to avoid the cost. Since we'd prefer to avoid allocations, this patch introduces a sparse offset "checkpoint" buffer (currently with 16 entries) that is filled lazily during iteration and limits the linear search space to `num_glyphs / 17` (the 0 offset is implicit). It also caches the index and offset for the last glyph which is used for the start of the linear scan for the next glyph if it is closer than the nearest checkpoint. This leads to very fast iteration when the indices are (mostly) monotonic.

Also updates skrifa to use this.

Benchmarks for "naive" vs new cached iterator with 8k glyphs):

| Benchmark | Naive (ms) | Cached (ms) | Speedup |
| :--- | :---: | :---: | :---: |
| **Monotonic** | 45.241 | 0.164 | **276×** |
| **Mostly Monotonic** | 43.016 | 0.177 | **243×** |
| **Pathological** | 92.672 | 0.170 | **545×** |

Increasing the glyph count to 20k shows the scaling behavior:

| Benchmark | Naive (ms) | Cached (ms) | Speedup |
| :--- | :---: | :---: | :---: |
| **Monotonic** | 257.760 | 0.404 | **638×** |
| **Mostly Monotonic** | 260.030 | 0.501 | **519×** |
| **Pathological** | 530.350 | 0.413 | **1,285×** |

Note that the worst-case for the two iterators is now different. The cached iterator easily handles the "all indices point to last" structure due to the monotonic optimization. The absolute worst-case timing is triggered by a structure where the first index points to the last glyph, while remaining indices alternate between the ends of the first and second buckets. This layout forces repeated scans across an entire bucket width.

Comparing worst-case performance at 65k glyphs (all indices point to last for naive and the above layout for cached):

| Benchmark | Naive (ms) | Cached (ms) | Speedup |
| :--- | :---: | :---: | :---: |
| **Worst Case** | 6,184.20 | 329.82 | **18.8×** |

internal bug ref: b/537782685

### DIFF
--- a/font-test-data/src/lib.rs
+++ b/font-test-data/src/lib.rs
@@ -158,6 +158,105 @@ pub mod closure {
 }
 
 pub mod post {
+    use crate::bebuffer::BeBuffer;
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub enum GlyphNameOrder {
+        /// Glyph name indices in strictly increasing order.
+        Monotonic,
+        /// Glyph name indices in mostly increasing order, with some back
+        /// references to previous names.
+        ///
+        /// This likely matches most sane font data with duplicate glyph names.
+        MostlyMonotonicWithBackrefs,
+        /// All glyph name indices point to the last glyph name.
+        ///
+        /// This is the patholical case, requiring a full scan for each glyph
+        /// during iteration.
+        AllPointToLast,
+    }
+
+    /// Build a synthetic post v2 table with custom names for each glyph.
+    ///
+    /// Returns the table bytes and the generated glyph names in glyph id order.
+    pub fn v2_with_varied_glyph_names(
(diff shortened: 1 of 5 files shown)


behdad:
Design sounds good to me.

qxliu76:
nice optimization, LGTM

qxliu76 (approved):

PR #1984, opened 2026-07-28
[skera] tweak benchmark suite

qxliu76:
1. use designs from harfbuzz subset benchmark suite
2. move plan creation out of iteration loop, this will match harfbuzz and make binary runtime comparison fairer


skera/benches/skera-benchmarks.rs
```diff
@@ -13,16 +13,17 @@ fn read_test_font(file_name: &str) -> Vec<u8> {
     std::fs::read(&path).expect("failed to read test font file")
 }
 
-fn set_from_ranges<I>(ranges: I) -> IntSet<u32>
```
wmedrano, line 16: I think its fine to delete the code since the history will live in git and this PR.

wmedrano (approved):

skera/benches/skera-benchmarks.rs
qxliu76: Done

PR #1989, opened 2026-07-29
[skrifa] autohint: ensure indices fit in u16

dfrg:
Add some sanity checks to ensure our 16-bit indices fit our data before additional processing. Use wrapping_add for one place with potential overflow-- speculatively setting `point.next_ix = ix + 1`-- this can only overflow for the last point and this is patched at the end of the loop.

internal bug ref: b/525070808

### DIFF
--- a/skrifa/src/outline/autohint/outline.rs
+++ b/skrifa/src/outline/autohint/outline.rs
@@ -14,7 +14,7 @@ use crate::collections::SmallVec;
 use core::ops::Range;
 use raw::{
     tables::glyf::{PointFlags, PointMarker},
-    types::{F26Dot6, F2Dot14},
+    types::{F26Dot6, F2Dot14, GlyphId},
 };
 
 /// Hinting directions.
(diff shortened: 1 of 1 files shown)


skrifa/src/outline/autohint/outline.rs
```diff
@@ -689,6 +704,57 @@ mod tests {
...
+    fn fill_sanity_checks_u16_index_bounds() {
+        let mut outline = Outline::default();
+        let gid = GlyphId::new(42);
+        outline
+            .points
+            .resize_and_fill(u16::MAX as usize + 1, Point::default());
+        outline
+            .contours
+            .resize_and_fill(u16::MAX as usize + 1, Contour::default());
+        assert!(outline.check_u16_index_bounds(gid).is_ok());
```
wmedrano: Can we test the public `fill` functcion or is would the setup be too complicated?

dfrg: The public entry point basically requires a fully functioning font which makes that a challenge. Even worse, we have guards in place that actually prevent constructing a TrueType glyph with enough points to trigger this. I could conceivably build a CFF font that produces more than 64k points but that would indeed require a lot of setup.

As a compromise, I extracted all the analysis and validation logic into a `analyze_and_validate` function that is called by both `fill` and the test code. Does that seem sufficient?

wmedrano (approved):

PR #1992, opened 2026-07-30
[glyph closure] try to reduce allocations by splitting borrows

qxliu76:
```
ClosureCtx API signature is like below: 
impl ClosureCtx {
    pub fn add_glyphs(&mut self) {}
    pub fn parent_active_glyphs(&self) { self.glyphs }
}

And in simple table glyph_closure(), we want to do something like:
   while iterate coverage table {
        // The entire ClosureCtx is already borrowed as immutable here
        filter out those glyphs that're not included in parent_active_glyphs()

         //  ❌ ERROR: Cannot borrow `self` as mutable 
         add_glyphs(retained_glyphs)
   }

This is impossible due to multiple references issue in rust, so previous code collects
retained glyphs into a vector first and then call add_glyphs(), like:

   while iterate coverage table {
       // The entire ClosureCtx is already borrowed as immutable here
        filter out those glyphs that're not included in parent_active_glyphs()
       
      add retained glyphs into a new Vec
    }

    iterate above Vec and call add_glyphs()

This works but Vec alloc is expensive
```
This PR makes some members in ClosureCtx public so that we can access the struct fields directly, which makes it possible to immutably iterate over one member while mutably update another member. Also known as splitting borrows in rust.

about ~5% runtime gain on Roboto-Regular(maybe in the noise range). I'm still working to improve (Chain)Context tables. they are more complicated, so changes will be in another PR.

### DIFF
--- a/read-fonts/src/tables/gsub/closure.rs
+++ b/read-fonts/src/tables/gsub/closure.rs
@@ -450,7 +450,7 @@ impl GlyphClosure for SingleSubstFormat2<'_> {
         }
         let coverage = self.coverage()?;
         let glyph_set = if let Some(glyph_set) = ctx.active_glyphs_stack.last() {
-            &glyph_set
+            glyph_set
         } else {
             &*ctx.glyphs
         };
@@ -492,7 +492,7 @@ impl GlyphClosure for MultipleSubstFormat1<'_> {
         }
         let coverage = self.coverage()?;
         let glyph_set = if let Some(glyph_set) = ctx.active_glyphs_stack.last() {
-            &glyph_set
+            glyph_set
         } else {
             &*ctx.glyphs
         };
@@ -544,7 +544,7 @@ impl GlyphClosure for AlternateSubstFormat1<'_> {
         }
         let coverage = self.coverage()?;
         let glyph_set = if let Some(glyph_set) = ctx.active_glyphs_stack.last() {
-            &glyph_set
+            glyph_set
         } else {
             &*ctx.glyphs
         };
@@ -598,7 +598,7 @@ impl GlyphClosure for LigatureSubstFormat1<'_> {
         let coverage = self.coverage()?;
         let ligs = self.ligature_sets();
         let glyph_set = if let Some(glyph_set) = ctx.active_glyphs_stack.last() {
-            &glyph_set
+            glyph_set
         } else {
             &*ctx.glyphs
         };
@@ -661,7 +661,7 @@ impl GlyphClosure for ReverseChainSingleSubstFormat1<'_> {
         let coverage = self.coverage()?;
         let sub_glyphs = self.substitute_glyph_ids();
         let glyph_set = if let Some(glyph_set) = ctx.active_glyphs_stack.last() {
-            &glyph_set
+            glyph_set
         } else {
             &*ctx.glyphs
         };
--- a/read-fonts/src/tables/gsub/closure.rs
+++ b/read-fonts/src/tables/gsub/closure.rs
@@ -42,9 +42,9 @@ mod ctx {
 
     pub(super) struct ClosureCtx<'a> {
         /// the current closure glyphs. This is updated as we go.
-        glyphs: &'a mut IntSet<GlyphId>,
-        active_glyphs_stack: Vec<IntSet<GlyphId>>,
-        output: IntSet<GlyphId>,
+        pub(super) glyphs: &'a mut IntSet<GlyphId>,
+        pub(super) active_glyphs_stack: Vec<IntSet<GlyphId>>,
+        pub(super) output: IntSet<GlyphId>,
         lookup_count: u16,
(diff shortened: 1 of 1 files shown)


dfrg (approved):
This LGTM.

Optional: a common alternative to exposing the fields would be to add something like so to `ClosureCtx`:

```rust
pub(super) fn with_active_glyphs_and_output<F, T>(&mut self, f: F) -> T
where
    F: FnOnce(&IntSet<GlyphId>, &IntSet<GlyphId>, &mut IntSet<GlyphId>) -> T,
{
    let glyph_set = if let Some(glyph_set) = self.active_glyphs_stack.last() {
        glyph_set
    } else {
        &*self.glyphs
    };
    f(&self.glyphs, glyph_set, &mut self.output)
}
```

Then you could reduce the repetition by wrapping all the new code in the closure:
```rust
ctx.with_active_glyphs_and_output(|glyphs, active_set, output| {
    // use glyphs and active_set to generate new output
});
```

qxliu76:
> This LGTM.
> 
> Optional: a common alternative to exposing the fields would be to add something like so to `ClosureCtx`:
> 
> ```rust
> pub(super) fn with_active_glyphs_and_output<F, T>(&mut self, f: F) -> T
> where
>     F: FnOnce(&IntSet<GlyphId>, &IntSet<GlyphId>, &mut IntSet<GlyphId>) -> T,
> {
>     let glyph_set = if let Some(glyph_set) = self.active_glyphs_stack.last() {
>         glyph_set
>     } else {
>         &*self.glyphs
>     };
>     f(&self.glyphs, glyph_set, &mut self.output)
> }
> ```
> 
> Then you could reduce the repetition by wrapping all the new code in the closure:
> 
> ```rust
> ctx.with_active_glyphs_and_output(|glyphs, active_set, output| {
>     // use glyphs and active_set to generate new output
> });
> ```

Thank you! It looks cleaner, I'll try it and see if we could use it for (Chain)Contextual tables

qxliu76:
@dfrg Sorry to ask here, I just wonder if there's a cleaner way to do for (Chain)Contextual tables, ideally code structure should look like this:

```rust
for rule_set in table.filter_by(ctx.parent_active_glyphs) { // immutable
    if ctx.lookup_limit_exceed() {  // immutable
        return;
    }

    for rule in rul_set {
        if ctx.lookup_limit_exceed() { // immutable
            return;
        }
        if !rule.intersects(ctx.glyphs) { // immutable
            continue;
        }

        for lookup_record in rule {
            collect active_glyphs // unavoidable alloc, fine
            // mutable borrow, recurse into next lookup
            // ❌ ERROR if we don't replace above filter_by() with a separate set(needs alloc)
            ctx.recurse(looukp_idx, active_glyphs)
        }
    }
}

impl ClosureCtx {
    fn recurse(&mut self, lookup_idx) {
        mixed immutable/mutable things
    }
}
```
Currently to avoid multiple reference issues, I collected intersected classes/glyphs into a separate set, my plan is to collapse `fn recurse()` but we'll then have code repetitions

dfrg:
In the general case, you can break a type into multiple pieces based on how they will be borrowed. 

In this case, however, I’d just do a first pass and collect all the lookup indices into a vec and then process them sequentially. This lets you reuse a single allocation for the index list and also potentially sidesteps your unavoidable allocation for glyph collection.

qxliu76:
sorry I didn't mention that recurse fn is like: `ctx.recurse(looukp_idx, active_glyphs)`,  during recursion, collected active glyphs will be the new parent_active_glyphs for child lookups. Because only the glyphs that the parent lookup can apply the recursion to can participate in recursing the lookup. We need to take into account the context of what we recursed from. Otherwise we would end up more glyphs

garretrieger (approved):

dfrg:
Ah, I see. So `ClosureCtx` can’t own the active glyph set (or the stack of active sets). I’d wrap this state in a new type and pass it as a separate parameter. You can probably drop the explicit stack entirely and just pass the current active set to functions that need it since you have to compute these anyway. If you want to reduce heap allocations, maybe add a limited size pool of `IntSet<GlyphId>` to the context so they can be reused while traversing the graph. If you pop a set from the pool when needed and push it back when complete, this transfers ownership temporarily so doesn’t hold an immutable borrow on the context, allowing you to recurse mutably.

qxliu76:
> Ah, I see. So `ClosureCtx` can’t own the active glyph set (or the stack of active sets). I’d wrap this state in a new type and pass it as a separate parameter. You can probably drop the explicit stack entirely and just pass the current active set to functions that need it since you have to compute these anyway. If you want to reduce heap allocations, maybe add a limited size pool of `IntSet<GlyphId>` to the context so they can be reused while traversing the graph. If you pop a set from the pool when needed and push it back when complete, this transfers ownership temporarily so doesn’t hold an immutable borrow on the context, allowing you to recurse mutably.

Thank you!

PR #1993, opened 2026-07-31
[read-fonts] type1: remove unbounded recursion

### LINKED ISSUE #1995: Type1 font fails to load with skrifa
Spun off from https://github.com/LaurenzV/hayro/issues/1334.

The following font: https://drive.proton.me/urls/98WSHDVZH8#kX5pnHSNaFVN

seemingly fails to parse with skrifa, while it works fine with FreeType.

With a script like this one:
```rs
#!/usr/bin/env -S cargo +nightly -Zscript
---
[package]
edition = "2024"

[dependencies]
freetype-rs = "0.38"
skrifa = { path = "../../fontations/skrifa" }
---

use freetype::{Library, face::LoadFlag, ffi};
use skrifa::raw::ps::type1::Type1Font;

const FONT: &[u8] = include_bytes!("TeX-matha10.pfa");

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let library = Library::init()?;
    let face = library.new_memory_face(FONT.to_vec(), 0)?;
    let charmap = (0..face.num_charmaps())
        .map(|index| face.get_charmap(index as isize))
        .find(|charmap| charmap.encoding() == ffi::FT_ENCODING_ADOBE_CUSTOM)
        .expect("font has no Adobe Custom charmap");
    face.set_charmap(&charmap)?;

    let code = 112;
    let freetype_gid = face.get_char_index(code).expect("FreeType did not map code 112");
    assert_eq!(face.get_name_index("p"), Some(freetype_gid));
    face.load_glyph(freetype_gid, LoadFlag::NO_SCALE)?;
    let freetype_points = face
        .glyph()
        .outline()
        .expect("FreeType did not produce an outline")
        .points()
        .len();
    println!(
        "FreeType:  accepted {} glyphs; code 112 -> gid {freetype_gid} /p ({freetype_points} points)",
        face.num_glyphs(),
    );

    let fontations = match Type1Font::new(FONT) {
        Ok(font) => font,
        Err(error) => {
            eprintln!("Fontations: rejected the same font: {error:?}");
            return Err("Fontations rejected a font that FreeType accepted".into());
        }
    };
    let fontations_gid = fontations
        .encoding()
        .and_then(|encoding| encoding.map(code as u8))
        .expect("Fontations did not map code 112");
    assert_eq!(fontations_gid.to_u32(), freetype_gid);
    assert_eq!(fontations.glyph_name(fontations_gid), Some("p"));
    println!(
        "Fontations: accepted {} glyphs; code 112 -> gid {} /p",
        fontations.num_glyphs(),
        fontations_gid.to_u32(),
    );

    Ok(())
}
```

I get:
```
FreeType:  accepted 61 glyphs; code 112 -> gid 43 /p (51 points)
Fontations: rejected the same font: InvalidFontFormat
Error: "Fontations rejected a font that FreeType accepted"
```

# Cause

This is what AI thinks, I haven't double-checked this:

The font contains this valid PostScript procedure:

```postscript
dup 3 {3} executeonly put
```

FreeType accepts the font. The Fontations procedure parser consumes the `3` before tokenizing it, starts its next token at `}`, and eventually rejects the font as `InvalidFontFormat`.

dfrg:
Fixes two cases:

1. Binary blob parsing: an integer followed by RD or -| is tokenized as a blob. This used `peek` which called `next` meaning that a series of integers would recurse through `next` calls until a non-integer was detected, leading to potential stack overflow and generally bad performance. Fixed by adding a specialized `accept_blob_start` function that looks for the specific tokens.

2. Procedure parsing: procs can be nested and this was also handled with a recursive call to `next`. Changed to an iterative approach, tracking nesting depth as we do for strings.

Additionally: removed `peek` entirely and changed `accept` to just copy the current position of the cloned parser rather than parsing the token twice.

Added benchmarks for type1 parsing. This patch improves performance for our test PFA and PFB fonts by 3.6% and 10%, respectively.

internal bug ref: b/537783114

### DIFF
--- a/read-fonts/src/ps/type1.rs
+++ b/read-fonts/src/ps/type1.rs
@@ -816,10 +816,7 @@ impl<'a> Parser<'a> {
                             // but Type1 fonts define /RD procs and this
                             // pattern is used by FreeType.
                             // <https://gitlab.freedesktop.org/freetype/freetype/-/blob/80a507a6b8e3d2906ad2c8ba69329bd2fb2a85ef/src/type1/t1load.c#L1351>
-                            if matches!(
-                                self.peek(),
-                                Some(Token::Raw(b"RD")) | Some(Token::Raw(b"-|"))
-                            ) {
+                            if self.peek_is_rd() {
                                 // skip the token
                                 self.next();
                                 // and a single space
@@ -840,13 +837,21 @@ impl<'a> Parser<'a> {
         }
     }
 
-    fn peek(&self) -> Option<Token<'a>> {
-        self.clone().next()
+    /// Special lookahead case for RD or -| which indicates a binary blob in
+    /// Type1 fonts.
+    fn peek_is_rd(&self) -> bool {
+        let mut p = self.clone();
+        p.skip_whitespace();
+        let start = p.pos;
+        p.skip_until(is_special_or_whitespace);
+        let end = p.pos;
+        matches!(self.data.get(start..end), Some(b"RD") | Some(b"-|"))
     }
 
     fn accept(&mut self, token: Token) -> bool {
-        if self.peek() == Some(token) {
-            self.next();
+        let mut p = self.clone();
+        if p.next() == Some(token) {
+            self.pos = p.pos;
             true
         } else {
             false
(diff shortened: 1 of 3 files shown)


read-fonts/src/ps/type1.rs
```diff
@@ -940,7 +959,6 @@ impl<'a> Parser<'a> {
             return None;
         }
         let end = self.pos;
-        self.pos += 1;
```
cmyr, line 943: sneaking in another little fix? 😈

cmyr (approved):
makes sense!

read-fonts/src/ps/type1.rs
dfrg: and I would have gotten away with it too...

PR #1994, opened 2026-08-01
[read-fonts] apply imageDataOffset to format 4 sbit offsets

metsw24-max:
**Format 4 bitmap offsets ignore imageDataOffset**

Every other arm of `BitmapSize::location` adds the index subtable's `imageDataOffset` to the sbit offsets, so a glyph served by a format 4 subtable resolves relative to the start of EBDT/CBDT instead of the subtable's image data block. The size stays right because it is a difference of two offsets, so the read quietly returns a correctly sized window of unrelated bytes rather than erroring; format 5 is the nearest analogue and already applies the base.


dfrg (approved):
Good catch, thanks!

PR #1996, opened 2026-08-01
[skera] name: don't panic on out-of-range name record count

sagikazarmark:
The `count` field in the `name` table header is attacker-controlled and is not validated when the table is parsed: `Name::read` only checks that the table is at least MIN_SIZE (6) bytes, and `name_record_byte_range()` derives its range directly from `count`. A font declaring more name records than the table contains produces a range past the end of the table data, and the unchecked `.unwrap()` in `serialize_name_records` panicked.

Return `SubsetError::SubsetTableError(Name::TAG)` instead, matching the `record_bytes` binding four lines below.

### DIFF
--- a/skera/src/name.rs
+++ b/skera/src/name.rs
@@ -80,7 +80,9 @@ fn serialize_name_records(
 ) -> Result<(), SubsetError> {
     let data = name.offset_data().as_bytes();
     let name_records = name.name_record();
-    let name_records_bytes = data.get(name.name_record_byte_range()).unwrap();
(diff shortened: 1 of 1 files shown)


cmyr (approved):
looks good to me, will defer to @qxliu76 to merge in case there's anything conflicting in flight.

sagikazarmark:
Thanks! Lemme know if anything needs changing.

qxliu76:
It looks good. Thank you!

PR #1997, opened 2026-08-02
Lower MSRV to 1.85 and add MSRV CI check

nicoburns:
The MSRV was bumped to 1.89 in #1967 to match the rustc version running in CI, but this bump is not actually required (entire workspace builds with 1.85).

Changes:
- Reduce MSRV back to 1.85
- Add a CI job that runs 'cargo check --workspace' on the MSRV toolchain.
- Patch-bump all publishable crates so the lowered MSRV can be released.

(Servo is still on MSRV 1.88, so can't upgrade to the latest versions atm)

We could also just reduce the rustc version used for all of CI. https://github.com/googlefonts/fontations/pull/1942 bumped to 1.89 to match the version supported by Skia, but it is unclear to me why this repo would need to run CI with the rustc version used by Skia.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -64,6 +64,20 @@ jobs:
       - name: cargo clippy incremental-font-transfer (1.89)
         run: cargo clippy -p incremental-font-transfer --all-features --all-targets -- -D warnings
 
+  check-msrv:
+    name: cargo check msrv
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: install stable toolchain
+        uses: dtolnay/rust-toolchain@4360b52568e2003a75bf9bc1d59f33a8e3fc893c # stable
+        with:
+          toolchain: 1.85
+
+      - name: cargo check
+        run: cargo check --workspace
+
   test-stable:
     name: cargo test stable
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
 edition = "2021"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/googlefonts/fontations"
-rust-version = "1.89"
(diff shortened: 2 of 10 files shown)


cmyr:
there was a bit of a hiccup with that 1.89 bump, where CI and MSRV were bumped separately. We should definitely be keeping them in sync.

I forget the exact rationale for wanting to go to 1.89, maybe @wmedrano can elaborate?

I'm personally not sure how to think about MSRV. I like to be able to use new language features as they become available, but understand that as core library software we need to be somewhat conservative. I'm not sure how we should decide on where to draw the line, though? I would be very happy for some guidance on this... my best idea would just be to keep track of our major consumers (skia, servo, ??) and just track the lower of their MSRV, but if there is some more general guidance on this question I'd love to hear it.

wmedrano:
Bumping down is ok. I initially wanted it for the smol-str dependency but I ended up doing a custom short string type

cmyr (approved):
okay, no complaints here.

there's a new CI check about how we specify action revisions, so rebase this on main and update the actions to match what I did in #2002, and we can merge when CI is green.

I'd normally want to have the version changes be a separate patch but happy to take this as-is, I assume that you would like a release?

cmyr:
@nicoburns I assume you would like releases?

nicoburns:
@cmyr I ideally would. Unfortunately it seems like https://github.com/googlefonts/fontations/commit/40cd87d21b7d0f030e60f239cc60dc814d548a0e is a breaking change, so the patch version bumps I've done here won't be sufficient.

(sorry - I should have thought to check for breaking changes from other patches when bumping versions)

So either we need bump minor versions for read-fonts/write-fonts and dependencies, or temporarily revert https://github.com/googlefonts/fontations/commit/40cd87d21b7d0f030e60f239cc60dc814d548a0e, publish a patch release, and then reinstate https://github.com/googlefonts/fontations/commit/40cd87d21b7d0f030e60f239cc60dc814d548a0e.

Happy to put up a PR for the former if that's helpful. The latter might be best done locally?

cmyr:
happy to just bump minor again, I'll do it.

PR #2000, opened 2026-08-04
[read-fonts] limit work in charstring eval

dfrg:
Add a work counter to the charstring evaluator to limit computation time and prevent DoS attacks from malicious fonts.

internal bug ref: b/537782659


cmyr (approved):
👍

PR #2001, opened 2026-08-04
Remove support for IFT Patch Map Format 1.

garretrieger:
It has been removed from the specification: https://github.com/w3c/IFT/pull/311

This leaves what was previously called Format 2 as the sole format. So also drop the format enum and remove all mentions of format 2.


resources/codegen_inputs/ift.rs
```diff
@@ -5,134 +5,14 @@ extern record U8Or16;
...
-  #[compile(skip)] // TODO remove this once write fonts side is implemented.
-  last_entry_index: U8Or16,
-}
-
-/// [Patch Map Format Format 2](https://w3c.github.io/IFT/Overview.html#patch-map-format-2)
-table PatchMapFormat2 {
+/// [Patch Map Table](https://w3c.github.io/IFT/Overview.html#patch-map-table)
+table IftPatchMap {
   /// Format identifier: format = 2
-  #[format = 2]
```
cmyr, line 135: is the comment above this still correct?

cmyr (approved):
one little question inline :)

resources/codegen_inputs/ift.rs
garretrieger: Yes, we've kept format = 2 to maintain compatibility with existing encodings.

PR #2002, opened 2026-08-05
[chore] Better manage github actions deps

cmyr:
This makes sure that we are pinning actions to git shas, and additionally configures dependabot to bother us monthly if there are updates to our actions.

This was motivated by a failing CI check on #1997; this check is new (afaik) and there is a go/ link in the failure log that I don't have access to, if someone wants to sanity check this patch against go/github-zizmor-help?


wmedrano (approved):

wmedrano:
There's not much to the go link, its similar to the zizmor docs https://docs.zizmor.sh/audits/

PR #2006, opened 2026-08-05
[glyph closure] try to reduce allocs + add caches for hot methods

qxliu76:
With these changes, NotoNastaliqUrdu subsetting runtime improved by 40%. 
This PR adds a FnvHashMap into read-fonts, since we need to use it as a cache to speed up glyph closure. Not sure where is the best place for it.


read-fonts/src/tables/gsub/closure.rs
```diff
@@ -133,8 +133,12 @@ mod ctx {
...
         // Return true if we have visited this lookup with current set of glyphs
         pub(super) fn is_lookup_done(&mut self, lookup_index: u16) -> bool {
-            let mut cur_active_glyphs = IntSet::empty();
-            cur_active_glyphs.union(self.parent_active_glyphs());
+            let cur_active_glyphs = if let Some(cur_active_glyphs) = self.active_glyphs_stack.last()
+            {
+                cur_active_glyphs
+            } else {
+                &*self.glyphs
+            };
```
cmyr: i think this can just be,

```suggestion
            let cur_active_glyphs = self.active_glyphs_stack.last().unwrap_or(self.glyphs);
```

read-fonts/src/tables/gsub/closure.rs
```diff
@@ -829,6 +829,21 @@ impl GlyphClosure for ContextFormat1<'_> {
...
 }
 
+fn intersected_class_glyphs(
+    class_def: &ClassDef,
+    glyphs: &IntSet<GlyphId>,
+    class: u16,
+    cache: &mut FnvHashMap<u16, IntSet<GlyphId>>,
+) -> IntSet<GlyphId> {
+    if let Some(cached_set) = cache.get(&class) {
+        return cached_set.clone();
```
cmyr, line 834: each cache hit here is an alloc, so if we think that's a bottleneck you could use `Rc<IntSet<GlyphId>>` and then a cache hit would just be bumping a ref count.

read-fonts/src/tables/layout/closure.rs
```diff
@@ -728,47 +728,103 @@ impl Format2Rule<'_> {
...
-        input_classes: &IntSet<u16>,
-        backtrack_classes: &IntSet<u16>,
-        lookahead_classes: &IntSet<u16>,
+        glyphs: &IntSet<GlyphId>,
+        input_class_def: &ClassDef,
+        backtrack_class_def: Option<&ClassDef>,
+        lookahead_class_def: Option<&ClassDef>,
+        input_cache: &mut FnvHashMap<u16, bool>,
+        backtrack_cache: &mut FnvHashMap<u16, bool>,
+        lookahead_cache: &mut FnvHashMap<u16, bool>,
```
cmyr: another option would be something like,

```rust
struct SeqCache {
    input: FnvHashMap<u16, bool>,
    backtrack: FnvHashMap<u16, bool>,
    lookahead: FnvHashMap<u16, bool>,
}
```

so that we only need to pass around a single argument.

read-fonts/src/tables/layout/closure.rs
```diff
@@ -728,47 +728,103 @@ impl Format2Rule<'_> {
...
+    class: u16,
+    cache: &mut FnvHashMap<u16, bool>,
+) -> bool {
+    if let Some(v) = cache.get(&class) {
+        return *v;
+    }
+
+    let v = class_def.intersects_class_glyphs(glyphs, class);
+    cache.insert(class, v);
+    v
```
cmyr: fwiw there's an 'entry' API on collections that does what you're trying to do here, with the (small) advantage that it only requires a single key lookup.


```suggestion
    *cache.entry(class).or_insert_with(|| class_def.intersects_class_glyphs(glyphs, class))
```

cmyr (approved):
some notes inline but looks good!

read-fonts/src/tables/gsub/closure.rs
qxliu76: Thanks! This one is not shown in the profile so I'll just keep it as is for now.

PR #2008, opened 2026-08-05
Add is_subset() method to IntSet.

garretrieger:


qxliu76 (approved):

read-fonts/src/collections/int_set/bitpage.rs
```diff
@@ -60,6 +60,28 @@ impl BitPage {
...
+    pub(crate) fn is_subset(&self, other: &BitPage) -> bool {
+        if self.len() > other.len() {
+            return false;
+        }
+        for (a, b) in self.storage.iter().zip(other.storage.iter()) {
+            if (*a & *b) != *a {
+                return false;
+            }
+        }
+        true
```
cmyr: i'd write this using `all`? it's probably a wash in this case but my general sense is that in general the iterator syntax is more amenable to compiler optimizations..
```suggestion
        self.storage.iter().zip(other.storage.iter()).all(|(a, b)| *a & *b == *a)
```

cmyr (approved):
looks good!

read-fonts/src/collections/int_set/bitpage.rs
garretrieger: Good idea, also updated the intersects method to use .any()

PR #2009, opened 2026-08-05
[skrifa] COLRv1: remove nested traversal optimization and limit total work

dfrg:
The first commit replaces the nested traversal that detects the `clip transform* fill` pattern with an explicit check, avoiding multiple traversals every time we encounter a clip paint.

The second commit limits the total number of paints processed for a single glyph.

There's some unfortunate churn here to enable these changes:
* The transform and fill node processing is moved to instance.rs as methods on `ResolvedPaint` so that they can be used from multiple call sites (the optimization check and the normal pass).
* Some parameters from `traverse_with_callbacks` have been moved into a new `TraversalState` struct. We need to track visited nodes and this reduces the parameter count while also giving us a nice place to stash the combined resolve + limit check.

internal bug ref: b/537783593


skrifa/src/color/traversal.rs
```diff
@@ -70,92 +76,39 @@ impl From<ResolvedColorStop> for ColorStop {
...
-        self.optimization_success = false;
-    }
-
-    fn pop_layer(&mut self) {
-        self.optimization_success = false;
+    pub(crate) fn resolve_paint(
+        &mut self,
+        paint: &Paint<'a>,
+    ) -> Result<ResolvedPaint<'a>, PaintError> {
+        self.nodes_left = self
```
cmyr, line 100: so we track nodes, but iiuc a 'node' can correspond to quite a large range of possible work. Are we worried about a malicious font where a smaller number of nodes are performing a larger number of operations?

skrifa/src/color/traversal.rs
```diff
@@ -165,369 +118,95 @@ pub(crate) fn traverse_with_callbacks(
...
                             if let Some(rect) = clipbox {
-                                painter.push_clip_box(rect);
+                                state.painter.push_clip_box(rect);
                             }
 
                             let result = traverse_with_callbacks(
-                                &resolve_paint(instance, &base_glyph)?,
-                                instance,
-                                painter,
+                                &state.resolve_paint(&base_glyph)?,
```
cmyr, line 203: reviewing with opus and it points out that this error path results in us potentially leaving a clip box on the paint context that is never popped. I'm not sure if this is a concern in practice, but seems worth pointing out?

cmyr (approved):
okay i've spent thirty minutes digging into this and do find the patch very hard to follow as a consequence of all the churn, but thanks to my tireless assistant I've left a few comments inline.

skrifa/src/color/traversal.rs
dfrg: This is a good point, particularly with regard to gradients with variable color stops. It might be worth adding an additional limit for those though I’m not sure what a reasonable value might be. Napkin math: ~4k nodes * 65k stops * 2 values = 4*10^8 IVS lookups for a single glyph. 

@behdad I don’t think HB has a limit for these? Should we?

skrifa/src/color/traversal.rs
dfrg: We can fail at any point during traversal so this isn’t new. Clients that want to avoid partial rendering (or inconsistent graphics state) for broken glyphs will need to render into an offscreen buffer or accumulate the commands into some intermediate data structure.

PR #2011, opened 2026-08-05
[skrifa] avoid more overflow panics in TT hinting

dfrg:
Use wrapping arithmetic in a few more places:
* CVT index + 1 in MIRP instruction (b/537782380)
* Everywhere in `RoundState::round`


cmyr (approved):
👍

PR #2012, opened 2026-08-05
[skrifa] Don't panic drawing a VARC glyph with a null MultiItemVariationStore

### LINKED ISSUE #2010: skrifa: panic drawing a VARC glyph when the MultiItemVariationStore offset is null
## Summary

`skrifa`'s VARC outline renderer panics (`called Option::unwrap() on a None value`)
when asked to draw a glyph from a VARC font whose `multiVarStore` offset is null.
The `multiVarStore` offset is defined as nullable, so such a font is spec-valid, yet
`Outlines::draw` unconditionally `.unwrap()`s the scalar cache that is only produced
when a variation store is present. Rendering an untrusted font through the public
outline API is therefore a denial-of-service vector (panic / process abort).

## Affected

- Crate: `skrifa` 0.45.1 (reproduced on `main` @ `f95a2db`)
- Location: `skrifa/src/outline/varc/mod.rs:269` (twin at `:403`)
- Memory-safe panic (no UB). CWE-248 / CWE-617 (reachable panic / unhandled condition).

## Root cause

`scalar_cache_from_store` returns `Ok(None)` precisely when there is no variation
store:

```rust
fn scalar_cache_from_store(
    &self,
    store: Option<&MultiItemVariationStore<'a>>,
) -> Result<Option<ScalarCache>, DrawError> {
    let Some(store) = store else {
        return Ok(None);           // <- null / absent store
    };
    let region_count = store.region_list()?.region_count() as usize;
    Ok(Some(ScalarCache::new(region_count)))
}
```

`Outlines::draw` then unwraps that `Option` at the top of drawing, before any glyph
component is processed:

```rust
let mut scalar_cache = self
    .scalar_cache_from_store(self.var_store.as_ref())?
    .unwrap();                     // <- panics when var_store is None
```

`self.var_store` is `None` whenever the VARC table's `multiVarStore` offset is null.
That offset is declared nullable (`resources/codegen_inputs/varc.rs`):

```
coverage_offset: Offset32<CoverageTable>,
#[nullable]
multi_var_store_offset: Offset32<MultiItemVariationStore>,
```

so `varc.multi_var_store()` resolves to `None`, `Outlines::from_base` stores
`var_store: None`, and the very first thing `draw` does is unwrap it. A VARC font that
uses composite glyphs without a variation store (or simply omits the store) is enough;
the panic fires before any component-level variation logic runs.

There is a latent twin at `:403`: once `:269` no longer panics, a null-store font with
a nested, axis-bearing component reaches `child_scalar_cache = scalar_cache_from_store(None)?`
(→ `None`) followed by `child_scalar_cache.as_mut().unwrap()`. A complete fix must
address both.

## Reproduction

Take the real `font-test-data` VARC fixture `varc-6868.ttf` and zero the 4-byte
`multiVarStore` offset in the VARC header (version 4B + coverage 4B, then the store
offset at table byte 8), then drive the public API:

```rust
let mut bytes = font_test_data::varc::CJK_6868.to_vec();
let varc_off = sfnt_table_offset(&bytes, b"VARC").unwrap();
for b in &mut bytes[varc_off + 8..varc_off + 12] { *b = 0; } // null multiVarStore

let font = FontRef::new(&bytes).unwrap();
let outlines = font.outline_glyphs();
let gid = font.cmap().unwrap().map_codepoint(0x6868_u32).unwrap();
let glyph = outlines.get(gid).unwrap();
let mut pen = Vec::<PathElement>::new();
let _ = glyph.draw(
    DrawSettings::unhinted(Size::unscaled(), LocationRef::default()),
    &mut pen,
); // panics: called `Option::unwrap()` on a `None` value  @ varc/mod.rs:269:14
```

Observed:

```
thread '...' panicked at skrifa/src/outline/varc/mod.rs:269:14:
called `Option::unwrap()` on a `None` value
```

The existing `fuzz_skrifa_outline` target already calls `glyph.draw(...)`, but all
three checked-in VARC fixtures ship *with* a variation store, so the null-store path
is never exercised by the corpus.

## Impact

Denial of service (panic → abort) when `skrifa`'s public outline API draws an
attacker-controlled font. This is an upstream / non-web surface: it affects callers
that render untrusted or system-provided fonts directly with skrifa. (In Chromium the
VARC table is stripped by OTS before skrifa sees it, so the browser path is not
affected.) Severity: Medium, memory-safe.

## Suggested fix

Treat an absent store as an empty scalar cache instead of a hard `unwrap`. Making
`scalar_cache_from_store` return `ScalarCache` (empty when the store is `None`) fixes
both the primary site and the twin, and is behavior-preserving: a component that
actually needs variation data still fails gracefully with `ReadError::NullOffset` when
it tries to resolve the missing store. A patch with a regression test is attached
(all 393 `skrifa` lib tests + 11 doc-tests pass).

---
Found with rust-in-peace (https://github.com/scadastrangelove/rust-in-peace).

scadastrangelove:
## What

`Outlines::draw` in the VARC renderer unwraps the scalar cache returned by
`scalar_cache_from_store`, which is `None` whenever the VARC table has no variation
store. Since the `multiVarStore` offset is `#[nullable]`, a spec-valid font can omit
the store, and drawing any covered glyph from such a font panics:

```
thread '...' panicked at skrifa/src/outline/varc/mod.rs:269:14:
called `Option::unwrap()` on a `None` value
```

reachable through the public API (`FontRef::outline_glyphs().get(gid).draw(...)`) on an
untrusted font.

## Fix

`scalar_cache_from_store` now returns a `ScalarCache` directly, yielding an empty cache
(zero regions) when the store is absent instead of `Some`/`None`. This removes the
`.unwrap()` at the primary call site and the latent twin in `draw_glyph` (which would
otherwise panic at `child_scalar_cache.as_mut().unwrap()` for a null-store font with a
nested axis-bearing component).

The behavior is unchanged for well-formed fonts: an empty cache is only ever produced
when there are no regions, and any component that references variation data still
returns a graceful `ReadError::NullOffset` when it resolves the missing store — no more
panicking.

## Test

Adds `draw_varc_with_null_var_store_does_not_panic`, which nulls the `multiVarStore`
offset of the `varc-6868` fixture and draws through the public outline API. It fails
(panics) before this change and passes after.

```
cargo test -p skrifa
test result: ok. 393 passed; 0 failed; ...
   Doc-tests skrifa
test result: ok. 11 passed; 0 failed; ...
```

Found with rust-in-peace (https://github.com/scadastrangelove/rust-in-peace).

Closes #2010

### DIFF
--- a/skrifa/src/outline/varc/mod.rs
+++ b/skrifa/src/outline/varc/mod.rs
@@ -264,9 +264,7 @@ impl<'a> Outlines<'a> {
         expand_coords(&mut font_coords, self.axis_count, coords);
         let mut stack = GlyphStack::new();
         let pen: &mut dyn OutlinePen = pen;
-        let mut scalar_cache = self
-            .scalar_cache_from_store(self.var_store.as_ref())?
-            .unwrap();
+        let mut scalar_cache = self.scalar_cache_from_store(self.var_store.as_ref())?;
         let mut scratch = Scratchpad::new();
         let ctx = VarcSharedContext {
             font_coords: &font_coords,
@@ -389,7 +387,8 @@ impl<'a> Outlines<'a> {
                             if let Some(ref mut cache) = child_scalar_cache {
                                 cache.values.fill(ScalarCache::INVALID);
                             } else {
-                                child_scalar_cache = self.scalar_cache_from_store(ctx.var_store)?;
+                                child_scalar_cache =
+                                    Some(self.scalar_cache_from_store(ctx.var_store)?);
                             }
                             self.draw_glyph(
                                 component_gid,
@@ -733,12 +732,17 @@ impl<'a> Outlines<'a> {
     fn scalar_cache_from_store(
         &self,
         store: Option<&MultiItemVariationStore<'a>>,
-    ) -> Result<Option<ScalarCache>, DrawError> {
-        let Some(store) = store else {
-            return Ok(None);
+    ) -> Result<ScalarCache, DrawError> {
+        // The VARC `multiVarStore` offset is nullable, so a valid font may omit the
+        // store entirely. In that case there are no variation regions and an empty
+        // cache is correct: any component that references variation data will fail
+        // gracefully with `ReadError::NullOffset` when it resolves the missing store,
+        // rather than us panicking here.
+        let region_count = match store {
+            Some(store) => store.region_list()?.region_count() as usize,
+            None => 0,
         };
-        let region_count = store.region_list()?.region_count() as usize;
-        Ok(Some(ScalarCache::new(region_count)))
+        Ok(ScalarCache::new(region_count))
     }
 }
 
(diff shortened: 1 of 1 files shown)


scadastrangelove:
@googlebot I signed it!

skrifa/src/outline/varc/mod.rs
```diff
@@ -1259,4 +1263,50 @@ mod tests {
             ]
         );
     }
+
+    /// Locate the file offset of a top-level sfnt table by tag.
+    fn sfnt_table_offset(data: &[u8], tag: &[u8; 4]) -> Option<usize> {
```
dfrg: Should be able to use `FontRef::table_directory()` and then `TableDirectory::table_records()` to find a table offset without manually parsing.

dfrg (commented):
The fix LGTM. Just one nit about simplifying some of the test setup.

dfrg:
Looks good now, thanks!

PR #2017, opened 2026-08-05
[read-fonts] aat: use wrapping to avoid panics on 32-bit

dfrg:
Security analysis pointed at that these ops could overflow only on 32-bit builds. So use wrapping to avoid the cost on target platforms. Any values that overflow will be caught by subsequent checked memory reads and return an error anyway.

JMM

internal bug ref: b/537783431


cmyr (approved):
👍

PR #2020, opened 2026-08-06
Fix failure to read EBLC glyphs from Proggy font

### LINKED ISSUE #2019: Skrifa fails to read EBLC glyphs from Proggy font
The Proggy font:

- Tested with: https://github.com/servo/webrender/blob/main/wrench/reftests/text/Proggy.ttf
- Source: https://proggyfonts.net/download

AI analysis:

> skrifa 0.44/0.45 returns no bitmap for ANY Proggy glyph at any size: Proggy's EBLC declares indexTablesSize=56 (the index-subtable array only, not the subtables), read-fonts bounds all subtable reads by that field → every lookup is OutOfBounds, while FreeType bounds by the end of the EBLC table and tolerates it.

nicoburns:
Fixes https://github.com/googlefonts/fontations/issues/2019

Generated with Fable.

### DIFF
--- a/read-fonts/src/tables/bitmap.rs
+++ b/read-fonts/src/tables/bitmap.rs
@@ -128,12 +128,14 @@ impl BitmapSize {
         offset_data: FontData<'a>,
     ) -> Result<IndexSubtableList<'a>, ReadError> {
         let start = self.index_subtable_list_offset() as usize;
-        let end = start
-            .checked_add(self.index_subtable_list_size() as usize)
-            .ok_or(ReadError::OutOfBounds)?;
-        let data = offset_data
-            .slice(start..end)
-            .ok_or(ReadError::OutOfBounds)?;
+        // FreeType ignores the declared size and bounds reads by the end of the
+        // parent table. Some fonts, such as ProggyClean, rely on this behavior
+        // because they set indexTablesSize to the size of the index subtable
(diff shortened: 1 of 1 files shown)


read-fonts/src/tables/bitmap.rs
```diff
@@ -128,12 +128,13 @@ impl BitmapSize {
...
-        let end = start
-            .checked_add(self.index_subtable_list_size() as usize)
-            .ok_or(ReadError::OutOfBounds)?;
-        let data = offset_data
-            .slice(start..end)
-            .ok_or(ReadError::OutOfBounds)?;
+        // The list is nominally bounded by indexTablesSize, but some fonts
+        // (e.g. ProggyClean) set it to the size of the index subtable array
+        // alone, excluding the subtables it points to, which would make all
+        // subtable offsets resolve out of bounds. Like FreeType, ignore the
```
rsheeter: Can we cite where in FreeType this is done?

nicoburns: Added. Bit of a sloppy comment now. I can clean up tomorrow.

nicoburns: Done. It's still kind of long. But I thought it was worth including the history?

dfrg (approved):
LGTM. Thanks Nico.

PR #2022, opened 2026-08-07
[sanitize] Implement max offset depth

cmyr:
I thought we could put this off until colrv1, but it turns out to also be relevant for Condition tables, and there's a test in varc that explicitly builds a highly nested graph and tries to read it, which overflows.

Interestingly this is okay on harfbuzz, because in harfbuzz the calls to sanitize child offsets of this table are tail calls, and so at standard optimization levels it doesn't end up growing the stack; because we call ctx.finish() after we finish with the offsets, this isn't available to us.


behdad:
We fixed this last week in HarfBuzz: https://github.com/harfbuzz/harfbuzz/issues/6116

PR #2024, opened 2026-08-07
[skera-repack] single parent optimization

qxliu76:
If a node has only a single parent then we store it directly and don't allocate a hash map. This helps reduce allocations
 and speed things up.  Improves NotoNastaliqUrdu subsetting time by ~10%

### DIFF
--- a/skera/src/graph.rs
+++ b/skera/src/graph.rs
@@ -65,6 +65,7 @@ pub(crate) struct Vertex {
     has_incoming_virtual_edges: bool,
     parents: FnvHashMap<ObjIdx, usize>,
     virtual_parents: IntSet<u32>,
+    single_parent: Option<ObjIdx>,
 }
 
 impl Vertex {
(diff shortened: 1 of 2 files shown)


skera/src/graph.rs
```diff
@@ -65,6 +65,7 @@ pub(crate) struct Vertex {
     has_incoming_virtual_edges: bool,
     parents: FnvHashMap<ObjIdx, usize>,
```
garretrieger: This is probably a good case to use an enum type that hold either the single index or the hash map. Should simplify the handling a bit since it ensures we only have one or the other at any time.

garretrieger: Could also have a None type value too for the case where there are no parents

garretrieger: Though we'll want to be careful to retain the allocated parent hash map during update_parents()

qxliu76: good idea, I have pushed a commit, can you help take a look? about 2% increase in runtime on my desktop, maybe in the noise range

garretrieger (approved):

PR #2025, opened 2026-08-09
Do not add final segment in cmap4 if already present

stephaneyfx:
- Windows treats a font file with a repeated final segment as invalid. Fontconfig allows such a font to be used.
- `0xFFFF` is not a valid Unicode character, but it is a Unicode scalar value valid for a Rust `char`.
- When building a new Cmap4 table from the pairs returned by `Cmap4::iter`, the final segment ends up being repeated if one does not filter it out from the input.

Thank you for the very useful crates!

### DIFF
--- a/write-fonts/src/tables/cmap.rs
+++ b/write-fonts/src/tables/cmap.rs
@@ -74,10 +74,15 @@ impl CmapSubtable {
         }
 
         // add the final segment:
-        end_code.push(0xFFFF);
-        start_code.push(0xFFFF);
-        id_deltas.push(1);
-        id_range_offsets.push(0);
+        if !matches!(
+            (start_code.as_slice(), end_code.as_slice()),
+            ([.., 0xFFFF], [.., 0xFFFF])
+        ) {
+            end_code.push(0xFFFF);
+            start_code.push(0xFFFF);
+            id_deltas.push(1);
+            id_range_offsets.push(0);
+        }
 
         Some(Self::format_4(
             0,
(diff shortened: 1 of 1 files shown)


write-fonts/src/tables/cmap.rs
```diff
@@ -74,10 +74,15 @@ impl CmapSubtable {
...
-        id_range_offsets.push(0);
+        if !matches!(
+            (start_code.as_slice(), end_code.as_slice()),
+            ([.., 0xFFFF], [.., 0xFFFF])
+        ) {
+            end_code.push(0xFFFF);
+            start_code.push(0xFFFF);
+            id_deltas.push(1);
+            id_range_offsets.push(0);
+        }
```
cmyr, line 85: so this is a bit tricky. At the top of this fn we assume that we will always be adding this segment, and use that to set `n_segments = segments.len() + 1`, which is assuming that we're always adding a final segment. the `n_segments` value is used to compute the range offsets, and so those end up subtly wrong if we _don't_ add that final segment.


I think the correct fix is to determine up above if we're adding this extra segment, and then compute n_segments based on that, like:


```rust
        let add_final_segment = segments
            .last()
            .is_none_or(|seg| mappings[seg.end_ix].0 != '\u{FFFF}');
        let n_segments = segments.len() + add_final_segment as usize;
```

and then use `if add_final_segment {` as your check down here.

cmyr (commented):
good catch on the bug, but the fix is slightly incorrect; left more detail inline.

write-fonts/src/tables/cmap.rs
stephaneyfx: Thank you for explaining. I added a slightly modified version of your suggestion to handle the case where the last generated segment is `(start, 0xFFFF)` and `start != 0xFFFF`, so that a final segment is appended in that case. I think the spec requires that the list of segments ends with `(0xFFFF, 0xFFFF)`. Please let me know if that's incorrect.

> For the search to terminate, the final startCode and endCode values must be 0xFFFF. [...] the segment must be present.

cmyr (approved):
great, thanks!

PR #2028, opened 2026-08-09
[read-fonts] avoid panics in classdef intersection 

dfrg:
These are mostly minimal validate things where checks were conditional on a count field rather than the length of the available array data.

### DIFF
--- a/read-fonts/src/tables/layout.rs
+++ b/read-fonts/src/tables/layout.rs
@@ -463,15 +463,18 @@ impl<'a> ClassDefFormat1<'a> {
         }
 
         let start_glyph = self.start_glyph_id().to_u32();
-        let glyph_count = self.glyph_count();
-        let end_glyph = start_glyph + glyph_count as u32 - 1;
+        let class_values = self.class_value_array();
+        if class_values.is_empty() {
+            out.insert(0);
+            return out;
+        }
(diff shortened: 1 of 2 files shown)


qxliu76 (approved):

read-fonts/src/tables/layout.rs
```diff
@@ -463,15 +463,18 @@ impl<'a> ClassDefFormat1<'a> {
         }
 
         let start_glyph = self.start_glyph_id().to_u32();
-        let glyph_count = self.glyph_count();
-        let end_glyph = start_glyph + glyph_count as u32 - 1;
+        let class_values = self.class_value_array();
+        if class_values.is_empty() {
```
dfrg, line 467: @qxliu76 I’m not sure if this is the correct return value for the empty case.

qxliu76: I'm also not sure for this edge case. Returning an empty set also makes sense in my opinion.Since it's not breaking any of our tests, I think it's fine?

qxliu76: spec says: Any glyph not included in the range of covered glyph IDs is assigned to Class 0. If the range is empty and glyph set is not, returning 0 seems right. :)

dfrg: Great, thanks for confirming!

PR #2029, opened 2026-08-09
Validate VariationRegionList axis counts

### LINKED ISSUE #744: validation of VariationRegionList axis_count
We should be validating that any set of regions added to the builder have the correct axis count.

I also believe that if we know the number of axes ahead of time there may be possible improvements in the implementation, but I may be confusing axis count and region count.

> If we don't already we should probably validate that the regions match this count since that's possible to check with access to only self?

_Originally posted by @rsheeter in https://github.com/googlefonts/fontations/pull/742#discussion_r1428604171_

alloutflo:
## Summary

`VariationRegionList::axis_count` determines how many `RegionAxisCoordinates` every region must contain, but write-fonts did not validate that relationship. A malformed builder value could therefore be serialized even when a region had too few or too many axes.

Fixes #744.

- Add a custom `validate_axis_count` annotation to the codegen input.
- Validate every variation region against `axis_count`, reporting errors at the corresponding `variation_regions[index]` path.
- Regenerate the checked-in write-fonts implementation from the codegen plan.
- Cover both matching and mismatched lists, including multiple bad region indices.

## Verification

- `cargo test -p write-fonts --all-targets --all-features` — 288 passed
- `cargo clippy -p write-fonts --all-features --all-targets -- -D warnings` with the CI-pinned Rust 1.89 toolchain — passed
- `cargo fmt --check` — passed
- Focused validation tests — 2 passed
- `cargo run --bin=codegen resources/codegen_plan.toml` — rerun successfully with no additional generated changes
- `git diff --check` — passed

AI disclosure: I used an AI coding assistant to help inspect the validation/codegen patterns, implement this focused validator and its tests, and run the checks above. I reviewed the behavior, generated diff, and test results before submitting.

### DIFF
--- a/resources/codegen_inputs/variations.rs
+++ b/resources/codegen_inputs/variations.rs
@@ -91,6 +91,7 @@ flags u8 EntryFormat {
 }
 
 /// The [VariationRegionList](https://learn.microsoft.com/en-us/typography/opentype/spec/otvarcommonformats#variation-regions) table
+#[validate(validate_axis_count)]
 table VariationRegionList {
     /// The number of variation axes for this font. This must be the
     /// same number as axisCount in the 'fvar' table.
@@ -159,4 +160,3 @@ table ItemVariationData {
     #[count(item_variation_data_len($item_count, $word_delta_count, $region_index_count))]
     delta_sets: [u8],
 }
-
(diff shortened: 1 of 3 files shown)


alloutflo:
The Google CLA check revealed a contributor-side legal prerequisite that I cannot complete within this contribution workflow. I’m closing this rather than leaving maintainers a PR they cannot review yet. The focused, fully tested commit remains available and can be reopened after the contributor has independently completed the CLA.

PR #2031, opened 2026-08-10
[skrifa] COLRv1: avoid unbalanced stack state on error 

dfrg:
When handling nodes that push/pop states (transforms, clips, layers), capture nested paint results and return after popping the relevant state. This ensures that clients always get a pop callback for every push.

internal bug ref: b/538461733


cmyr (approved):
If there's one thing I've learned from LLMs it's that an error path inside a push/pop pair is _unacceptable_.

dfrg:
As I’ve now learned! Didn’t reach this internal issue until after I responded to your comment on the earlier PR.

PR #2032, opened 2026-08-10
[skrifa] VARC: limit number of edges processed

dfrg:
Add a work budget for both outline memory computation and drawing to prevent exponential fan-out.

internal bug ref: b/537783570

### DIFF
--- a/skrifa/src/outline/varc/mod.rs
+++ b/skrifa/src/outline/varc/mod.rs
@@ -18,7 +18,7 @@ use crate::{
     instance::Size,
     outline::{cff, glyf, metrics::GlyphHMetrics, pen::PathStyle, DrawError, OutlinePen},
     provider::MetadataProvider,
-    GLYF_COMPOSITE_RECURSION_LIMIT,
+    GLYF_COMPOSITE_RECURSION_LIMIT, MAX_GRAPH_EDGES,
 };
 
 #[cfg(feature = "libm")]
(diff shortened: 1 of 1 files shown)


behdad:
As it happens, right now I'm adding unified work budget for VARC AND the recursed-to tables, such that it wouldn't explode by VARC limit multiplied by glyf/CFF(2) limit.

dfrg:
Ah, good call! that's slightly more difficult here but I'll see what I can do.

behdad:
Same issue with COLR table. That's harder to fix I think.

dfrg:
I assume you're tracking two counters: one for edges (shared among all VARC edges and TT components) and one for total CFF ops?

behdad:
> I assume you're tracking two counters: one for edges (shared among all VARC edges and TT components) and one for total CFF ops?

Yeah my current PR doesn't consolidate CFF ops count (each leaf glyph still gets a full CFF ops budget), I just limit number of segments emitted and stop progress when limit exceeded.

https://github.com/harfbuzz/harfbuzz/pull/6147

behdad:
A later iteration can limit the number of CFF ops done between any two consecutive drawing ops to 16 or something.

dfrg:
Ok, I think I'll defer that to another PR. Does this one look correct for limiting edges in VARC?

behdad (approved):

dfrg:
Thanks! Filed #2033 to track unified budget

PR #2035, opened 2026-08-11
[skera-repack] replace BTreeMap with vector

qxliu76:
Turns out small BTreeMap(under 20 elements) is slow. This change improves NotoNastaliqUrdu runtime by ~9%


skera/src/graph/pairpos_graph.rs
```diff
@@ -605,17 +605,21 @@ fn get_device_table_indices(val: u16) -> Vec<u8> {
...
     let mut size = 0;
     let record_start_pos = PairPosFormat2::MIN_SIZE as u32 + value_record_index * 2;
-    for &i in device_table_indices {
-        let pos = record_start_pos + i as u32 * 2;
-        if let Some(&link) = links.get(&pos) {
-            size += graph.find_subgraph_size(link.obj_idx(), visited, u16::MAX)?;
+    for l in links {
+        let link_pos = l.position();
+        // maximum num of device_table_indices is 4
+        for &i in device_table_indices {
```
garretrieger: I believe device_table_indices is always sorted, so if links was sorted as well then this could be processed in O(n) instead of O(n^2) by walking bother iterators simultaneously.

skera/src/graph.rs
```diff
@@ -1403,8 +1407,12 @@ impl Graph {
     // within vertices[idx].
     pub(crate) fn index_for_position(&self, idx: ObjIdx, position: u32) -> Option<ObjIdx> {
         let v = self.vertices.get(idx)?;
-        let link = v.real_links.get(&position)?;
-        Some(link.obj_idx())
+        for l in v.real_links() {
```
garretrieger: If real_links() is kept sorted then this could use binary search to find the link.

garretrieger (commented):
I've noted a couple of places that could be sped up if links was sorted. What you could do is:
- Track whether links is sorted or not in the vertex (could use an enum Sorted(Vec<...>), Unsorted(Vec<...>))
- Have a method to convert it to sorted if it's not already.
- Then do on demand sorting anywhere we require sorted order (including the few existing spots + the new ones I mentioned).

dfrg (commented):
I do wonder if this would be improved with a new type that maintains invariants so you can't forget to sort before update.

Code golfing to avoid fixing more overflow bugs for a bit:
```rust
struct OrderedLinks {
    links: Vec<Link>,
    is_sorted: bool,
}

impl OrderedLinks {
    pub fn new() -> Self {
        Self {
            links: Vec::new(),
            is_sorted: true,
        }
    }

    pub fn push(&mut self, link: Link) {
        self.is_sorted = self.is_sorted && self.links.last().is_none_or(|last| link.position() > last.position());
        self.links.push(link);
    }

    pub fn iter_ordered(&mut self) -> impl Iterator<Item = &Link> {
        self.sort_if_needed();
        self.links.iter()
    }

    pub fn iter_unordered(&self) -> impl Iterator<Item = &Link> {
        self.links.iter()
    }

    pub fn update_all(&mut self, mut pos: u32, each_size: u32) {
        self.sort_if_needed();
        // guaranteed to remain sorted
        for link in self.links.iter_mut() {
            link.update_position(pos);
            pos += each_size;
        }
    }

    pub fn update_obj_idx(&mut self, pos: u32, obj_idx: u32) -> bool {
        let link = if self.is_sorted {
            self.links.binary_search_by_key(&pos, |a| a.position()).ok().and_then(|idx| self.links.get_mut(idx))
        } else {
            self.links.iter_mut().find(|a| a.position() == pos)
        };
        if let Some(link) = link {
            link.update_obj_idx(obj_idx);
            return true;
        }
        false
    }

    fn sort_if_needed(&mut self) {
        if !self.is_sorted {
            self.links.sort_by_key(|a| a.position());
            self.is_sorted = true;
        }
    }
}

**Comment by dfrg** [comment]
oh, Garret beat me to it :)

**Comment by dfrg** [comment]
Note also: `sort_unstable_by_key` avoids a potential heap allocation if you don't need stable sort (which I'm assuming you don' t here)

**Comment by qxliu76** [comment]
Thank you so much! I'll address these issues

**Comment by qxliu76** [comment]
I pushed a new commit, can you help take another look?

File: skera/src/graph.rs [hunk 3]
```diff
@@ -53,13 +52,157 @@ pub(crate) enum Parents {
...
+            RealLinks::Unsorted(v) | RealLinks::Sorted(v) => v.iter(),
+        }
+    }
+
+    /// Returns a mutable iterator over the links.
+    fn iter_mut(&mut self) -> std::slice::IterMut<'_, Link> {
+        match self {
+            RealLinks::Unsorted(v) | RealLinks::Sorted(v) => v.iter_mut(),
+        }
+    }
```
garretrieger, line 86: Suggest adding a sorted_iter() and sorted_iter_mut() which calls sort_if_needed(), will make calls that want to iterate in sorted order a little cleaner by avoiding a separate sort_if_needed() call.

skera/src/graph.rs
```diff
@@ -53,13 +52,157 @@ pub(crate) enum Parents {
     Multiple(FnvHashMap<ObjIdx, usize>),
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum RealLinks {
+    Unsorted(Vec<Link>),
+    Sorted(Vec<Link>),
```
dfrg, line 58: This isn't public API so it's not a huge problem, but since variants are accessible at the same visibility as the enum, it's possible to construct an invalid sorted list anywhere in the crate with something like:
```rust
RealLinks::Sorted(vec![link_with_pos(100), link_with_pos(0)])
```
This doesn't necessarily need to change but just something to keep in mind.

skera/src/graph.rs
```diff
@@ -53,13 +52,157 @@ pub(crate) enum Parents {
...
+}
+
+impl Default for RealLinks {
+    fn default() -> Self {
+        RealLinks::Unsorted(Vec::new())
```
dfrg: An empty vec is always sorted :)

skera/src/graph.rs
```diff
@@ -53,13 +52,157 @@ pub(crate) enum Parents {
...
+impl RealLinks {
+    /// Returns an immutable iterator over the links.
+    fn iter(&self) -> std::slice::Iter<'_, Link> {
+        match self {
+            RealLinks::Unsorted(v) | RealLinks::Sorted(v) => v.iter(),
+        }
+    }
+
+    /// Returns a mutable iterator over the links.
+    fn iter_mut(&mut self) -> std::slice::IterMut<'_, Link> {
```
dfrg, line 78: If this is necessary, perhaps add a comment that calling `Link::update_position` can potentially invalidate the order of `Sorted` links?

But if possible, I suggest adding helper methods to this type that avoid the need for the mutable iterator.

skera/src/graph.rs
```diff
@@ -53,13 +52,157 @@ pub(crate) enum Parents {
...
+
+    fn sort_if_needed(&mut self) {
+        if let RealLinks::Unsorted(_) = self {
+            let tmp_state = std::mem::take(self);
+
+            if let RealLinks::Unsorted(mut v) = tmp_state {
+                v.sort_unstable_by_key(|link| link.position());
+                *self = RealLinks::Sorted(v);
+            }
+        }
```
dfrg: This could be simplified a bit:
```rust
if let Self::Unsorted(v) = self {
    v.sort_unstable_by_key(|link| link.position());
    *self = Self::Sorted(std::mem::take(v));
}
```

skera/src/graph/markbasepos_graph.rs
```diff
@@ -407,28 +416,13 @@ fn shrink_mark_array(
...
-    {
-        let old_pos = old_idx * 4 + 4;
-        let Some((_, mut l)) = links.remove_entry(&old_pos) else {
-            continue;
-        };
+    mark_array_v.real_links.sort_if_needed();
 
-        let new_pos = new_idx as u32 * 4 + 4;
+    let mut new_pos = MarkArray::MIN_SIZE as u32 + 2;
+    for l in mark_array_v.real_links.iter_mut() {
```
garretrieger: This doesn't seem correct, the old code was locating the record at the old index and changing it's position plus removing links as needed, but this just assigns the updated positions real_links in their current ordering with no respect to what's retained or not retained and doesn't remove any links, so real_links is going to contain more links than new_mark_count.

skera/src/graph.rs
garretrieger, line 78: If links are mutable then their positions may be changed which may cause the links array to become unsorted. So probably want to pre-emptively change the internal state to Unsorted when this is called.

skera/src/graph.rs
```diff
@@ -479,11 +627,7 @@ impl Graph {
             pos += 1;
 
             let next_v = &self.vertices[next_id];
-            for link in next_v
-                .real_links
-                .values()
-                .chain(next_v.virtual_links.iter())
-            {
+            for link in next_v.real_links.iter().chain(next_v.virtual_links.iter()) {
```
dfrg: This pattern appears more than a few times. Perhaps add a `Vertex::all_links()` method that returns the chained iterator?

skera/src/graph/markbasepos_graph.rs
garretrieger: I think roughly what you want to do is iterate real_links and retained_mark_classes at the same time advancing past dropped links until you find things with matching positions to the next retained index. Then build up the new link vector that way (this is assuming old_idx's are in sorted order).

qxliu76: previous approach allocates a new vector of links which are removed from real_links, we actually assumed that the num of left real links should equals to the num of retained mark classes, see: https://github.com/googlefonts/fontations/blob/d28db340e438eed1120120b2fda94827911fbeec/skera/src/graph/markbasepos_graph.rs#L427, since we then do a re-mapping from old class-> new class, and the new position is based on the new class, I think it's safe to just update position in sorted order?

skera/src/graph.rs
qxliu76: Done

skera/src/graph.rs
qxliu76: yeah it's much better.:) DOne

skera/src/graph.rs
qxliu76: Unfortunately iter_mut()  is necessary, but I changed internal state to Unsorted when it is called and added a doc comment about it.

skera/src/graph.rs
qxliu76: yeah a good catch! Thanks!

skera/src/graph.rs
qxliu76: Done.

skera/src/graph.rs
qxliu76: Done. I only added sorted_iter_mut(). sorted_iter() is not used anywhere.

skera/src/graph/markbasepos_graph.rs
qxliu76: uh..  thinking about it twice, you might be right

qxliu76: Ok, I reverted this change for shrinking MarkArray, just use the new API for real links, since the old idxes are not always in sorted order

dfrg (approved):

PR #2038, opened 2026-08-11
[read-fonts] fvar: use different normalize path for large number of axes

dfrg:
To support non-linear interpolation, axis coordinate normalization employed a `O(nm)` loop where `n = axis count` and `m = input user coord count`. This is fine for most (well-behaved) fonts but malicious fonts can contain 64k axes making this computation expensive.

For fonts with more than 64 axes, build an axis index array sorted by tag and use that to look up the range for each user coord. Cost is one sort and two binary searches per user coord, roughly `O(nlogn + mlogn)`. Spills to heap for more than 128 axes (max 128k heap allocation).

### DIFF
--- a/read-fonts/src/tables/fvar.rs
+++ b/read-fonts/src/tables/fvar.rs
@@ -11,6 +11,9 @@ use alloc::vec::Vec;
 pub use instance_record::InstanceRecord;
 
 const MAX_INLINE_AVAR2_AXES: usize = 64;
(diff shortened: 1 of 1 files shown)


qxliu76 (approved):

qxliu76:
Seems HB might have the same issue

dfrg:
> Seems HB might have the same issue

Nice to know we're in good company :)

PR #2039, opened 2026-08-11
[skrifa] tthint: avoid overflow in line_vector

dfrg:
Simply replaces a negation with wrapping_neg()

internal bug ref: b/538461591


cmyr (approved):
👍

PR #2040, opened 2026-08-11
[skrifa] tthint: ignore fractional bits in parity checks

dfrg:
The even and odd instructions were masking the full fractional component rather than just the first integral bit.

This originally matched FreeType but that code was fixed in https://gitlab.freedesktop.org/freetype/freetype/-/commit/47d80cf27b05d1062fbe5d3d368c1bac7ff1a20c

internal bug ref: b/538461019


cmyr (approved):
👍

PR #2041, opened 2026-08-11
[font-types] remove assert from Version16Dot16::new

dfrg:
I'm not sure this buys us anything?

internal bug ref: b/525061507

### DIFF
--- a/font-types/src/version.rs
+++ b/font-types/src/version.rs
@@ -181,10 +181,4 @@ mod tests {
         assert_eq!(Version16Dot16::new(0, 5).0, 0x00005000);
         assert_eq!(Version16Dot16::new(1, 1).0, 0x00011000);
     }
-
-    #[test]
-    #[should_panic]
-    fn minor_version_out_of_range_test() {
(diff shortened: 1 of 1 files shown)


cmyr (approved):
hmm, I guess we blow up on a suitably malformed version?

In any case one funny thing about this is that if minor is > 16 it'll bleed into the major, so we could also just fix this by clamping minor to 9?

font-types/src/version.rs
```diff
@@ -54,14 +54,7 @@ impl Version16Dot16 {
     pub const VERSION_3_0: Version16Dot16 = Version16Dot16::new(3, 0);
 
     /// Create a new version with the provided major and minor parts.
-    ///
-    /// The minor version must be in the range 0..=9.
```
cmyr, line 58: oh I'd also keep this? and just remove the panics section...

dfrg:
Ah, thanks for forcing me to read the spec on this :) I didn't realize how, uhm, "creative" this type is. Clamping minor to 9 sounds good to me. I'll note that we do so in the docs.

PR #2043, opened 2026-08-11
[font-types] add normalize method to bbox

dfrg:
To ensure min values are less than max values.

internal bug ref: b/525061546


cmyr (commented):
👍

cmyr (approved):

PR #2044, opened 2026-08-11
[skrifa] autohint: avoid panic on font with 0 glyphs

dfrg:
Simply early out of GsubHandler::finish() if we have any empty glyph style slice.

internal bug ref: b/538461890


cmyr (approved):
👍

PR #2047, opened 2026-08-12
[read-fonts] sign-extend packed device deltas in iter_packed_values

metsw24-max:
**Wrong sign extension for packed Device deltas**

`iter_packed_values` shifts each 2/4/8-bit field down to the low bits, but the two's-complement step for negative values still masks with the shifted mask, so it ANDs a low-aligned value against a high-aligned mask. A negative delta only decodes correctly in the least significant slot of its word, which is the one position the existing tests happen to cover; anywhere else it comes out wrong, and a leading -128 additionally overflows the negation. These are ordinary signed values from any GPOS/GDEF Device table, so no malformed input is needed to hit it.

Reworked the loop to pull the field down first and sign extend against the unshifted mask, which keeps the decode in one spot and drops the fragile negation. Added a regression test covering a negative delta outside the last slot for each format.


dfrg (approved):
Looks good, thanks!

PR #2049, opened 2026-08-13
[read-fonts] fix COLRv1 closure overflow

dfrg:
internal bug ref: b/537783358


cmyr (approved):
sorry for letting these sit!

dfrg:
no worries! I imagine we won’t do a release until end of week anyway.

PR #2051, opened 2026-08-13
[read-fonts] bound VarLenArray::get() reads

dfrg:
The `VarLenArray::get()` function was computing the offset for the requested index and then returning the entire tail of the buffer following that offset. This was unlike `VarLenArray::iter()` which bounded each item by its length. This changes `get` to work like `iter` so that they're consistent.

internal bug ref: b/538459419

### DIFF
--- a/read-fonts/src/array.rs
+++ b/read-fonts/src/array.rs
@@ -145,12 +145,13 @@ impl<'a, T: FontRead<'a, Args = ()> + VarSize> VarLenArray<'a, T> {
         if self.data.is_empty() {
             return None;
         }
-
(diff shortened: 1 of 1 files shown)


cmyr (approved):
would be curious to see a summary of the issue this caused!

dfrg:
Analysis didn’t like that it “leaked” trailing bytes and was inconsistent with the iterator. Both seem like reasonable complaints.

Thinking about it again, if we changed the iterator to leak instead then we could remove a hack from kern that wants the remaining bytes for the last subtable but I think this is probably better.

PR #2052, opened 2026-08-13
[read-fonts] fix PointIter::advance_flags overflow

dfrg:
Just widens the count to u16 to absorb the + 1.

internal bug ref: b/537783651


cmyr (approved):
makes sense

PR #2053, opened 2026-08-14
[read-fonts] fix potential bitmap overflows

dfrg:
Standard fare of using checked arithmetic to catch overflows.

However, this patch introduces two new helpers to assist with this in the future without so much boilerplate:

1. `offset::CheckedOffset` allows computing an offset with chained `add` and `mul` calls that internally use checked arithmetic. The final result can be extracted as an option, user specified result or `ReadError::OutOfBounds` (with `ok_or_oob`).
2. `array::get_pair` is a helper to retrieve elements at both `idx` and `idx + 1` handling potential overflow from the addition and checked reads from the slice.

Would have been nice to have these from the start!

internal bug ref: b/537782625

### DIFF
--- a/read-fonts/src/offset.rs
+++ b/read-fonts/src/offset.rs
@@ -88,3 +88,37 @@ impl<O: Offset> ResolveOffset for O {
             .and_then(|data| T::read_with_args(data, args))
     }
 }
+
+/// Helper for performing checked sequences of arithmetic operations on
+/// offsets.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(transparent)]
+pub(crate) struct CheckedOffset(Option<usize>);
+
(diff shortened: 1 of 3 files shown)


read-fonts/src/offset.rs
```diff
@@ -88,3 +88,37 @@ impl<O: Offset> ResolveOffset for O {
             .and_then(|data| T::read_with_args(data, args))
     }
 }
+
+/// Helper for performing checked sequences of arithmetic operations on
+/// offsets.
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(transparent)]
+pub(crate) struct CheckedOffset(Option<usize>);
```
cmyr, line 96: yea, this is a nice idea.

cmyr (approved):
👍

PR #2054, opened 2026-08-14
[codegen] traversal: don't unwrap versioned fields

dfrg:
Since minimal validate, we don't check presence of versioned fields but traversal still assumed they exist. Replaces unwrap with `?` for generated traversal code ensuring we don't panic when encountering these.

internal bug ref: b/537782897


cmyr (approved):
makes sense, and maybe it's time to reconsider whether or not this API is worth holding onto.

dfrg:
a rust implementation of ttx might be more useful and could serve a similar purpose… whenever someone has the time to do it 🤷‍♂️

PR #2055, opened 2026-08-14
[read-fonts] avar: cap to start of trimmed range

dfrg:
We skip duplicate segments but fail to check if we're at the start of the trimmed range. In this case, if `start > 0`, we miss the check at the boundary condition and fall into a case where we try to interpolate between ranges with duplicate `from` values leading to division by 0.

Seems to only occur with inputs < -1.0

@behdad please double check me on this one


behdad:
Good catch. Fixing it in HarfBuzz as well. LGTM.

PR #2056, opened 2026-08-15
[read-fonts] add Cmap14::iter_with_limits

dfrg:
Unconditionally skip non-monotonic selector records since the spec requires them to be sorted. If limits are supplied, stop iteration when we've yielded `max_char + 1` defaults or `glyph_count` variants.

Update skrifa to use the limited version of the iterator.

internal bug ref: b/537782883

### DIFF
--- a/read-fonts/src/tables/cmap.rs
+++ b/read-fonts/src/tables/cmap.rs
@@ -805,33 +805,13 @@ impl<'a> Cmap14<'a> {
     /// Returns an iterator over all (codepoint, selector, mapping variant)
     /// triples in the subtable.
     pub fn iter(&self) -> Cmap14Iter<'a> {
-        Cmap14Iter::new(self.clone())
+        Cmap14Iter::new(self.clone(), None)
     }
 
-    fn selector(
-        &self,
-        index: usize,
-    ) -> (
-        Option<VariationSelector>,
-        Option<DefaultUvs<'a>>,
-        Option<NonDefaultUvs<'a>>,
-    ) {
-        let selector = self.var_selector().get(index).cloned();
-        let default_uvs = selector.as_ref().and_then(|selector| {
-            selector
-                .default_uvs(self.offset_data())
-                .transpose()
-                .ok()
-                .flatten()
-        });
-        let non_default_uvs = selector.as_ref().and_then(|selector| {
-            selector
-                .non_default_uvs(self.offset_data())
-                .transpose()
-                .ok()
-                .flatten()
-        });
-        (selector, default_uvs, non_default_uvs)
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/cmap.rs
```diff
@@ -805,33 +805,13 @@ impl<'a> Cmap14<'a> {
     /// Returns an iterator over all (codepoint, selector, mapping variant)
     /// triples in the subtable.
     pub fn iter(&self) -> Cmap14Iter<'a> {
-        Cmap14Iter::new(self.clone())
+        Cmap14Iter::new(self.clone(), None)
```
cmyr, line 816: is it worth updating docs here to point to `iter_with_limits`? I'm not totally clear on the use case, but given that is public API I presume it is preferable in some cases...

read-fonts/src/tables/cmap.rs
```diff
@@ -862,22 +842,62 @@ impl<'a> Cmap14<'a> {
...
+    default_uv_left: u32,
+    non_default_uv_left: u32,
 }
 
 impl<'a> Cmap14Iter<'a> {
-    fn new(subtable: Cmap14<'a>) -> Self {
-        let (selector_record, default_uvs, non_default_uvs) = subtable.selector(0);
+    fn new(subtable: Cmap14<'a>, limits: Option<CmapIterLimits>) -> Self {
+        let (default_uv_left, non_default_uv_left) = if let Some(limits) = limits {
+            (limits.max_char.saturating_add(1), limits.glyph_count)
```
cmyr, line 865: a funny little thing here: in the other places where we use `CmapIterLimits` `max_char` is a maximum codepoint value, but here we're using it as a 'max number of things we will return'? That seems fine, but it is a little confusing.

cmyr (approved):
looks good, a few little notes inline..

read-fonts/src/tables/cmap.rs
dfrg: Indeed! Copied the "use iter_with_limits" suggestion from `Cmap12` to both `Cmap13` and `Cmap14`. I'd say it's preferable in almost all cases (skrifa only uses the limited iterators) except that sometimes it may be useful to extract the actual data in the font even if malformed? Perhaps we can revisit this at some point.

read-fonts/src/tables/cmap.rs
dfrg: Good catch. This was the remediation suggested by security analysis and, after some thinking, it made sense. The theoretical maximum count for default UVS is really `(char::MAX + 1) * n` where `n` is the number of codepoints designated as variation selectors but dropping the extra factor seems reasonable.

PR #2057, opened 2026-08-18
[skera] bug fixes

qxliu76:
No need to retain a VariationDevice if its` varidx = NO_VARIATION_INDEX `, and fix GPOS/GDEF/BASE tables accordingly so we don't err out in case of empty VariationDevice


skera/src/gpos/anchor.rs
```diff
@@ -93,14 +93,19 @@ impl SubsetTable<'_> for AnchorFormat3<'_> {
...
-            Offset16::serialize_subset(
+            match Offset16::serialize_subset(
                 &x_device,
                 s,
                 plan,
                 &plan.layout_varidx_delta_map,
                 x_device_offset_pos,
-            )?;
+            ) {
+                Ok(()) => downgrade_to_format1 = false,
```
dfrg: optional: might be worth considering adding a method for this:

```rust
impl Result<(), SerializeErrorFlags> {
    pub(crate) fn is_empty(self) -> Result<bool, SerializeErrorFlags> {
        match self {
            Ok(()) => Ok(false),
            Err(SerializeErrorFlags::SERIALIZE_ERROR_EMPTY) => Ok(true),
            Err(e) => Err(e),
        }
    }
}
```

Then this becomes `downgrade_to_format1 = !x.is_empty()?`

and similar for other uses of this pattern

dfrg (approved):
LGTM with one optional suggestion for the repetitive empty check pattern.

skera/src/gpos/anchor.rs
qxliu76: good idea! I'll make this change, I have a lot more places that have this pattern.

qxliu76: Done.

garretrieger (approved):

PR #2059, opened 2026-08-19
[skrifa] raise COLRv1 paint node budget for complex flags

wmedrano:
The Ecuador and El Salvador flags in Noto Color Emoji fail with `PaintError::DepthLimitExceeded`. Raise MAX_NODES to the `8192` and add a regression test that paints these two flags.

Test data subset created with:

```
pyftsubset NotoColorEmoji-Regular.ttf --glyphs=glyph01852,glyph01986,u1F1EA,u1F1E8,u1F1F8,u1F1FB --layout-features='*' --drop-tables='SVG ,vhea,vmtx'
```


skrifa/src/color/traversal.rs
```diff
@@ -53,10 +53,10 @@ const MAX_TRAVERSAL_DEPTH: usize = 64;
 
 /// Maximum number of nodes visited during a single traversal.
 ///
-/// Prevents excessive execution time on graphs with high fan-out. Set to 4096
+/// Prevents excessive execution time on graphs with high fan-out. Set to 8192
```
dfrg: Let’s just bump to 16k as originally suggested.

dfrg (approved):
Nice catch!

skrifa/src/color/traversal.rs
wmedrano: Done.

behdad:
Thanks. I bumped HB as well.

PR #2065, opened 2026-08-21
[int-set] Reuse the last-page cache when inserting

behdad:
`ensure_page_index_for_major` (the path under `insert`, `extend_unsorted`, etc.) goes straight to a binary search of the page map on every call, while `contains` consults the `last_page_map_index` cache first. This checks and maintains the same cache on the insert path — mirroring `hb_bit_set_t`, which caches page lookups on both reads and writes. Staleness is handled the same way `contains` already handles it: verify the major value, fall back to the search on mismatch.

First commit adds an ordered-insert benchmark: the existing `BM_SetInsert_1000` inserts hash-scattered values (the always-miss case), and nothing covered runs of nearby values — the pattern range-ish workloads (closure, cmap collection, buffer glyph sets) actually produce.

Measured (criterion, `--baseline` A/B):
- `BM_SetInsertOrdered_1000`: **−30% to −83%**, larger sets gain more (e.g. 65536/64: 12.0 µs → 2.0 µs).
- `BM_SetInsert_1000` (scattered): **+7–9%** on small/mid sets (~0.5 ns/insert for the extra compare); large-set configs unchanged within noise.

That trade-off matches HarfBuzz's long-standing choice for the same structure. All read-fonts tests pass.

### DIFF
--- a/read-fonts/benches/int_set_benchmark.rs
+++ b/read-fonts/benches/int_set_benchmark.rs
@@ -56,6 +56,30 @@ pub fn insert_benchmark(c: &mut Criterion) {
     }
 }
 
+pub fn insert_ordered_benchmark(c: &mut Criterion) {
+    const NUM_INSERTS: u32 = 1000;
+    let inputs = set_parameters();
+
+    for input in inputs {
+        c.bench_with_input(
+            BenchmarkId::new("BM_SetInsertOrdered_1000", &input),
+            &input,
+            |b, p: &SetTest| {
+                let set = random_set(p.set_size, p.max_value());
+                b.iter_batched(
+                    || set.clone(),
+                    |mut s| {
+                        for i in 0..NUM_INSERTS {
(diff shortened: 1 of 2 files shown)


behdad:
Actually with this I'm measuring 1% slowdown in shaping Menlo with HR. Investigating.

behdad:
Closing after further measurement: in HarfRust's shaping parity benchmarks (buffer glyph sets, 1–3 pages, ~6 ns/insert), every formulation of this cache — the page-map-index probe, a page-count-gated variant, and a plain (major, page-index) field pair — showed up as a measurable regression (~1% on Menlo), because at that insert cost any added per-insert instruction is visible. The ordered-insert wins for large sets are real, but `U32SetBuilder` (used by `Extend`) already provides the same memoization opt-in at the call site for bulk workloads without taxing plain `insert()`. Callers inserting runs of clustered values should prefer `extend` — no library change needed.

PR #2066, opened 2026-08-21
[font-types] derive bytemuck::NoUninit for glyph ids

dfrg:
see https://github.com/harfbuzz/harfrust/pull/434#discussion_r3832544092


behdad:
LGTM. Thanks.

PR #2068, opened 2026-08-23
[int_set] Streamline the insert path

behdad:
Two changes to the hot insert chain, found while profiling HarfRust's AAT shaping (the buffer glyph set is rebuilt per line — ~74K unsorted inserts per benchmark run, ~95 instructions each before this):

1. **Inline hints + hot/cold split**: hints along the insert chain, and the page-allocation miss path of `ensure_page_index_for_major` split into a `#[cold]` `#[inline(never)]` helper so the search hit path inlines into per-value loops.
2. **Loop-local memo in `extend_unsorted`**: glyph sequences are unsorted but cluster within a few pages, so remembering the last (major, page index) in the loop skips the page-map search on most iterations. The memo lives in the loop, not the set — per-insert cost is unchanged for non-clustered input (a miss adds one compare), unlike the set-resident cache of #2065 which regressed.

Microbench (extend_unsorted, 300 values / 4 pages, clear() between rounds, fat-LTO release): **51.9 → 29.7 instructions, 22.4 → 8.2 cycles per insert** clustered; non-regressing on random and single-page input. End-to-end HarfRust shaping (clang release rig): instructions −4.4% GeezaPro, −3.9% Menlo, −2.7% Lucida Grande, −0.8% Devanagari Sangam; shaping outputs byte-identical. `cargo test -p read-fonts` passes (637).

### DIFF
--- a/read-fonts/src/collections/int_set/bitpage.rs
+++ b/read-fonts/src/collections/int_set/bitpage.rs
@@ -124,6 +124,7 @@ impl BitPage {
     }
 
     /// Marks `(val % page width)` a member of this set and returns `true` if it is newly added.
+    #[inline(always)]
     pub(crate) fn insert(&mut self, val: u32) -> bool {
         let el_mut = self.element_mut(val);
         let mask = elem_index_bit_mask(val);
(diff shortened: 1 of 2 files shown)


dfrg (approved):
LGTM

PR #2069, opened 2026-08-23
[aat] Slim the per-transition entry() parse

behdad:
`entry()` is the per-transition hot path of state-machine-driven AAT shaping (called up to 3× per glyph by HarfRust's drive loops). Three behavior-identical changes:

- Cache `state_array_offset` as a plain field on `StateTable` — the legacy new-state conversion was re-reading it through the header accessor on every call.
- Hand-roll the entry parse in both `entry()` impls: one bounds check for the whole entry, then fixed big-endian reads, replacing the generic `StateEntry::read` cursor machinery. Formulation matters here: a `split_first_chunk` variant made LLVM materialize the entry through a stack slot (narrow stores + wide reload = store-forwarding stall, 2× cycles on the 8-byte-entry path); the committed constant-index form keeps everything in registers.
- Replace the new-state `try_into` with an equivalent range check.

Equivalence is locked in by tests that keep the old bodies verbatim as reference implementations and sweep every byte-truncation × (state, class) combination — including out-of-range indices — over all four entry payload types.

Per-call (100M-transition drive loop, fat LTO): legacy 37.1→36.1 instr, extended u16 payload 35.2→34.2, extended 4-byte payload 43.2→37.2. End-to-end HarfRust shaping: −1.3% instructions GeezaPro, −0.6% Devanagari Sangam, within noise elsewhere; shaping outputs byte-identical. `cargo test -p read-fonts` passes (639); fmt and clippy clean.

### DIFF
--- a/read-fonts/src/tables/aat.rs
+++ b/read-fonts/src/tables/aat.rs
@@ -349,6 +349,10 @@ pub struct StateTable<'a> {
     class_array: &'a [u8],
     state_array: &'a [u8],
     entry_table: &'a [u8],
+    /// Byte offset from the start of the table to the state array, cached
+    /// from the header so the per-transition new-state conversion doesn't
+    /// re-read it through the header accessor on every call.
+    state_array_offset: i32,
     /// floor(2^32 / n_classes) + 1: exact reciprocal for dividends < 2^16,
     /// so the per-transition new-state conversion avoids a hardware divide.
     n_classes_magic: u64,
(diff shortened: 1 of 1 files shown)


behdad:
@dfrg This looks like noise level gain. Happy to close.

dfrg:
Code is quite a bit harder to read and it wasn’t nice to begin with so if we’re not measuring a gain then probably not worth merging.

PR #2070, opened 2026-08-24
[read-fonts] Move gvar delta computation out of skrifa

### LINKED ISSUE #1263: Moving variable font logic into `read-fonts`

dfrg:
Also expands test coverage significantly.

Big patch but no functional changes. Bulk of the diff is new tests which are here to support the next PR which does some nice refactoring to simplify this code.

And after two years, fixes #1263


cmyr (approved):
would've loved to see this in two parts (at least two commits, maybe two PRs), where first we just move the code and then we add the tests, just for readability. otherwise LGTM

dfrg:
totally fair, anything git tracks as a rename is an easy review. will do next time!

PR #2071, opened 2026-08-24
[read-fonts] Refactor gvar deltas

dfrg:
Simplify the interpolation code by working on a single contour at a time with equal sized slices rather than trying to walk different ranges of multiple slices in nested loops.

Based on #2070 which must land first


behdad:
Looks good to me.

PR #2072, opened 2026-08-24
[read-fonts] Simplify `array::get_pair()`.

kpreid:
I happened to be reading this code and thought that it could be made simpler and clearer. This PR has no significance other than that.

The change has no effect on the optimized machine code (presumably because all uses of `get_pair()` get inlined), but simplifies the Rust code, with only two error paths instead of three and no checked (or unchecked) arithmetic.


dfrg:
This looks better, thanks!

PR #2073, opened 2026-08-24
[read-fonts] Add lazily-read GPOS value records

dfrg:
`ValueRecord` is an owned struct built eagerly from a cursor, so anything that walks value records pays for fields it never looks at. A subsetter scanning PairPosFormat2 builds two of them per class pair just to OR together a value format, and a PairSet binary search builds a full `PairValueRecord` at every probe to compare one glyph id.

Add `ValueRecordRef`, which stores the containing table's data, the record's offset within it, and its format. Constructing one performs no reads; fields are located by a popcount over the lower format bits, so a single field can be read without touching its neighbours.

The data is the containing table's rather than a slice at the record because the device and variation index offsets in a value record resolve relative to that table, so slicing to the record would lose the base.

Alongside it, add accessors that locate records without reading them: `SinglePosFormat1/2::value_record_ref`, `PairSet::pair_value_record_refs` (strided, with a `find` that binary searches on second glyph), and `PairPosFormat2::class_value_records`, which reads the class counts and formats once so callers walking many class pairs don't repeat that work.

Nothing here is wired into codegen and the existing `ValueRecord` is untouched, so this is additive. That is a deliberately temporary state: a follow-up will replace `ValueRecord` with this and update its users, at which point `Value` and `ValueContext` are expected to go away too. No conversion between the two is offered in the meantime.

### DIFF
--- a/read-fonts/src/tables/gpos.rs
+++ b/read-fonts/src/tables/gpos.rs
@@ -16,7 +16,7 @@ pub use super::layout::{
     Lookup, ScriptList,
 };
 use super::layout::{ExtensionLookup, LookupFlag, Subtables};
-pub use value_record::{Value, ValueContext, ValueRecord};
+pub use value_record::{Value, ValueContext, ValueRecord, ValueRecordRef};
 
 #[cfg(test)]
 #[path = "../tests/test_gpos.rs"]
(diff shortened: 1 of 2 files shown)


read-fonts/src/tables/gpos.rs
```diff
@@ -154,6 +154,267 @@ impl<'a> PositionLookup<'a> {
     }
 }
 
+impl<'a> SinglePosFormat1<'a> {
+    /// Returns the value record for this subtable, without reading it.
+    #[inline]
+    pub fn value_record_ref(&self) -> ValueRecordRef<'a> {
```
qxliu76: In format1, the offset data is from parent PairSet table, maybe we can remove this fn

read-fonts/src/tables/value_record.rs
```diff
@@ -143,6 +143,195 @@ impl Value {
...
+    /// `data` must be the offset data of the table containing the record, and
+    /// not merely the bytes of the record itself: the device and variation
+    /// index offsets in a value record are resolved relative to that table.
+    #[inline]
+    pub fn new(data: FontData<'a>, offset: usize, format: ValueFormat) -> Self {
+        Self {
+            data,
+            // an offset that doesn't fit is out of bounds by definition, and
+            // saturating here lets every subsequent read fail cleanly
+            offset: u32::try_from(offset).unwrap_or(u32::MAX),
```
qxliu76, line 174: can we return a ReadError instead?

read-fonts/src/tables/gpos.rs
dfrg, line 160: Do you mean `PairPosFormat1`? This is just single pos. Keep in mind, the intention for the next stage is to completely remove our current `ValueRecord` and it's associated methods, renaming `ValueRecordRef` -> `ValueRecord` and renaming all these methods to match the originals, so we need them for every subtable.

read-fonts/src/tables/value_record.rs
dfrg: We can. HarfRust just uses a `usize` for the offset but the intention here is to keep the `ValueRecordRef` struct smallish, 24 bytes instead of 32. I think just going `usize` might be a win here. All offsets are `u16` so are guaranteed to be in bounds.

qxliu76:
oops, I realized that I looked at SinglePosFormat2 so deleted that comment. Otherwise LGTM.

read-fonts/src/tables/gpos.rs
```diff
@@ -154,6 +154,267 @@ impl<'a> PositionLookup<'a> {
...
+        )
+    }
+}
+
+impl<'a> SinglePosFormat2<'a> {
+    /// Returns the value record at `index`, without reading it.
+    ///
+    /// `index` is a coverage index; returns `None` if it is out of range.
+    #[inline]
+    pub fn value_record_ref(&self, index: usize) -> Option<ValueRecordRef<'a>> {
```
cmyr, line 174: ideally we would have some intermediate collection-like type here that would provide access and iteration? If that is too cumbersome I understand, but it could be as simple as like,


```rust
struct ValueRecordArray<'a> {
    parent_data: FontData<'a>,
    format: ValueFormat,
    count: u32,
    /// offset of the first value record in the array
    array_offset: u32,
}
```

read-fonts/src/tables/gpos.rs
```diff
@@ -154,6 +154,267 @@ impl<'a> PositionLookup<'a> {
...
+}
+
+/// The pair value records of a [`PairSet`], located but not read.
+///
+/// Records are addressed by a fixed stride derived from the two value formats,
+/// so locating one costs a multiply and no reads. This makes the array cheap to
+/// scan and cheap to binary search on
+/// [`second_glyph`](PairValueRecordRef::second_glyph).
+#[derive(Copy, Clone)]
+pub struct PairValueRecordRefs<'a> {
```
cmyr, line 232: right this makes sense, maybe I'd just quibble about the name? But nothing much better is coming to mind..

cmyr (approved):
this looks excellent... one little note inline for something i'm sure you thought of and have a good reason for omitting.

dfrg:
closing this in favor of #2075 and #2076 which use codegen and plug a lot of holes here

PR #2074, opened 2026-08-24
[read-fonts] Add mort table support

behdad:
## Summary

- generalize legacy AAT state tables to support typed entry payloads
- add generated mort table, chain, and subtable parsing
- implement readers for all five mort subtable types
- expose mort through TableProvider and the experimental FontTables API

## Testing

- cargo run --bin=codegen resources/codegen_plan.toml
- cargo check -p read-fonts --all-features
- cargo test -p read-fonts --all-features


read-fonts/src/tables/aat.rs
```diff
@@ -342,21 +342,42 @@ where
...
+    /// callers now that payload-bearing legacy state tables are supported.
+    pub fn read(data: FontData<'a>) -> Result<Self, ReadError> {
+        <Self as FontRead>::read(data)
+    }
+
+    /// Reads a state table with no entry payload and explicit empty arguments.
+    pub fn read_with_args(data: FontData<'a>, args: ()) -> Result<Self, ReadError> {
+        <Self as FontRead>::read_with_args(data, args)
+    }
+}
```
dfrg: This should be a `FontRead` impl which should only need `read_with_args`.

dfrg (approved):
one nit, otherwise LGTM

read-fonts/src/tables/aat.rs
behdad: Fixed in df7ba9db: removed the inherent wrappers and kept the generic `FontRead` implementation as the sole parser. The unconstrained unit test now explicitly selects `NoPayload`.

behdad: PTAL.

dfrg:
LGTM

PR #2075, opened 2026-08-24
[codegen] Add positioned records

dfrg:
Some records hold offsets that are measured from the start of the table containing them, not from the record itself. A GPOS value record is the motivating case: its device and variation index offsets are relative to the enclosing subtable. `FontRead` hands a record data already sliced to its start, which throws away everything the offset is measured against, so such a record cannot resolve its own fields.

Add a second arrangement for these. An extern record declared with a lifetime is *positioned*: it is given the enclosing table's data plus its own offset within it, and keeps both.

    extern record Foo<'a>;

Positioned-ness is closed over containment during resolution, so a record holding a positioned field, or an array of such records, becomes positioned too and passes the table's data down. This is what lets the arrangement survive nesting, where the enclosing table is several levels up from the field that needs it.

Supporting pieces in read-fonts: a `FontReadAt` trait, a ~~`PositionedArray`~~ (no longer, see below) that addresses items by offset rather than by slicing, two `Cursor` methods, and a traversal case. The traversal hook for struct fields is now keyed off positioned-ness rather than only a list of type names, so one of the two hardcoded names in that match can go away when GPOS moves over.

Nothing declares a positioned record yet outside the codegen tests; GPOS follows separately. The tests cover a table field, an array, an array nested inside a positioned record, a stride wider than the fields it holds, and the truncated and out-of-range cases.


dfrg:
Updated this to drop the new `PositionedArray` type, folding the logic into `ComputedArray`. This now requires an impl of `FontReadAt` for every type that is used as a computed array element. A macro is provided to simplify this for records that are not positioned. We can probably fold this together into one read trait that takes an offset and decides whether to slice or not, effectively pushing offset traversal into read rather than before (the type decides whether it needs to hold a reference to the parent table), but that's work for later.

font-codegen/src/parsing/fields.rs
```diff
@@ -35,6 +35,13 @@ pub(crate) struct Field {
     /// These fields must be present, which means reads can unwrap (and could even
     /// be unsafe.)
     pub(crate) validated_at_parse: bool,
+    /// `true` if this field is read at an offset within the enclosing table's
+    /// data, rather than from data sliced to the field.
+    ///
+    /// This is required by fields whose type holds offsets that resolve against
+    /// the enclosing table; see the `FontReadAt` trait in read-fonts. It is
+    /// computed during resolution, from the extern records declared positioned.
```
cmyr: >     /// computed during resolution, from the extern records declared positioned.


I find the second clause here quite confusing

font-codegen/src/fields.rs
```diff
@@ -424,9 +424,13 @@ fn traversal_arm_for_field(
             let data = fld.offset_getter_data_src();
             quote!(Field::new(#name_str, traversal::FieldType::var_array(#typ_str, self.#name()#maybe_try, #data)))
         }
+        // a positioned record is hand-written and supplies its own traversal;
+        // the two named types are the same arrangement predating that.
         // See if there are better ways to handle these hardcoded types
         // <https://github.com/googlefonts/fontations/issues/659>
-        FieldType::Struct { typ } if typ == "ValueRecord" || typ == "SbitLineMetrics" => {
+        FieldType::Struct { typ }
+            if fld.positioned || typ == "ValueRecord" || typ == "SbitLineMetrics" =>
```
cmyr, line 432: 1) can we fold everything into the 'positioned' flag?
2) ... can you remind me why SbitLineMetrics needs special treatment? it doesn't contain offsets does it??

qxliu76:
Thank you! I'm not familiar with codegen code but those generated_*.rs LGTM. Other code changes look nice.

qxliu76 (approved):

cmyr (commented):
okay so i'm going to avoid details and start high-level: my main concern with this patch is that it adds a lot of cognitive complexity, and I want to see if there's any room to simplify that.

Records are already a bit of a pain in the ass, and already require some special casing in codegen. I'd like to just step back for a second and summarize how codegen thinks about records currently, and then we can discuss how to make these changes.


## right now
The OpenType spec has something genuinely useful to say about what a record is, and how it differs from a table. I think this bit is worth quoting in full; [source is here](https://learn.microsoft.com/en-us/typography/opentype/spec/otff#:~:text=Structures%20are%20characterized%20as%20either%20records%20or%20tables.):

>
> Within this specification, many structures are defined in terms of the data types listed above. Structures are characterized as either records or tables. The distinction between records and tables is based on these general criteria:
> - Tables are referenced by offsets. If a table contains an offset to a sub-structure, the offset is normally from the start of that table.
> - Records occur sequentially within a parent structure, either within a sequence of table fields or within an array of records of a given type. If a record contains an offset to a sub-structure, that structure is logically a subtable of the record’s parent table and the offset is normally from the start of the parent table.

So _most_ records are fixed-size, and we just declare them as packed structs; codegen expects them to generally exist in arrays, and we read them by byte-casting. Some records are not fixed-size. `ValueRecord` is the most significant of these, but there are a number of others. Because these do not have a fixed size, we cannot do any byte casting; this is why we have the `ComputedArray` type, which pretends to be an array but is actually just dynamically reading items from the bytes on access, using `FontRead`.

Regardless of whether or not they're fixed-size, currently all records follow a single pattern for dealing with offsets: the caller has to manually pass in the data from the parent table, since the record does not hold onto it. This isn't great! It has been a source of confusion in the past, and there have been proposals to try and improve this API (#1475). 


What I don't like about this PR is that it leaves us in a half-way state: we now have _some_ records that actually hold on to the parent table's data and can resolve offsets on their own, and then we have _other_ records that don't have this luxury, and require external data to be passed in.

Ideally, we would figure out a way to unify these? I have an idea I'm going to sketch out in a moment for how we could add parent-table data as a sort of sidecar to fixed-size records, and for non-fixed-size records there's no reason we can't always include it.

This also potentially leads us to another improvement: currently non-fixed-size records always implement `FontRead`, and we instantiate some type from the bytes when we read them. One other thing to consider is if we should generalize the approach taken by this patch; instead of copying bytes greedily, we could treat non-fixed-size records more like tables, where they are always just (parent data, offset) couplings, and they read their fields dynamically?

qxliu76:
I don't think it's a big issue. There's no API change and we get significant performance gains. Can we get this change in and simplify/unify the code later if necessary?

dfrg:
> my main concern with this patch is that it adds a lot of cognitive complexity

Fair, but I'd argue that codegen already has very high complexity. It's probably the most dense functional code in the repo-- cool stuff but difficult to follow if you're not already familiar with it.

> Regardless of whether or not they're fixed-size, currently all records follow a single pattern for dealing with offsets: the caller has to manually pass in the data from the parent table, since the record does not hold onto it. This isn't great! It has been a source of confusion in the past, and there have been proposals to try and improve this API (https://github.com/googlefonts/fontations/pull/1475).

> Ideally, we would figure out a way to unify these? I have an idea I'm going to sketch out in a moment for how we could add parent-table data as a sort of sidecar to fixed-size records, and for non-fixed-size records there's no reason we can't always include it.

Perhaps this is solved by some type like:

```rust
pub struct WithParent<'a, R> {
    offset_data: FontData<'a>,
    record: &'a R,
}

impl<R> Deref for WithParent<'_, R> {
    type Target = R;
    fn deref(&self) -> &Self::Target {
        self.record
    }
}

// for each R containing an offset, codegen methods for WithParent<R> that
// automatically resolves those offsets
```

Then handle all the embedded, array, nesting, etc. complexity in codegen. This still doesn't save ValueRecord, which is a different beast entirely, but helps for the others.

> One other thing to consider is if we should generalize the approach taken by this patch; instead of copying bytes greedily, we could treat non-fixed-size records more like tables, where they are always just (parent data, offset) couplings, and they read their fields dynamically?

Records are the one place where we get a nice codegen win: one bounds check, transmute, and then free field access. I don't think we should give that up.

dfrg:
> What I don't like about this PR is that it leaves us in a half-way state

True, but all of our codegen improvements have essentially been of this form which is why we're here today :)

I'd be very happy to work with you on more holistic improvements to the whole thing but blocking on that _now_ would be unfortunate since this lays the groundwork for 1) faster ValueRecord parsing for two consumers and 2) embedded records which cleans up codegen hacks and unblocks math table support (#1269 by @simoncozens and would also be helpful for typst)

font-codegen/src/fields.rs
dfrg: SbitLineMetrics is a record embedded in a table. Both of these exceptions go away in the follow up PRs.

font-codegen/src/parsing/fields.rs
dfrg: 🫤

cmyr:
> > my main concern with this patch is that it adds a lot of cognitive complexity
> 
> Fair, but I'd argue that codegen already has very high complexity. It's probably the most dense functional code in the repo-- cool stuff but difficult to follow if you're not already familiar with it.

I agree that codegen is complex (maybe too complex) but what I'm trying to say is slightly different: this patch adds to the cognitive complexity of read-fonts _itself_, by making it so that some records don't act like all the rest.


> Perhaps this is solved by some type like:
> 
> ```rust
> pub struct WithParent<'a, R> {
>     offset_data: FontData<'a>,
>     record: &'a R,
> }
> 
> impl<R> Deref for WithParent<'_, R> {
>     type Target = R;
>     fn deref(&self) -> &Self::Target {
>         self.record
>     }
> }
> 
> // for each R containing an offset, codegen methods for WithParent<R> that
> // automatically resolves those offsets
> ```

yes, this is exactly what i was thinking, I'll have a draft of something like this up in a bit. :)

> > One other thing to consider is if we should generalize the approach taken by this patch; instead of copying bytes greedily, we could treat non-fixed-size records more like tables, where they are always just (parent data, offset) couplings, and they read their fields dynamically?
> 
> Records are the one place where we get a nice codegen win: one bounds check, transmute, and then free field access. I don't think we should give that up.

I'm talking specifically about the records which are not currently zero copy.

In any case I'm not interested in blocking this, I just want to make sure that I actually understand it. I'm going to give it another read through, and i'll try to identify the bits I find hard to follow.

read-fonts/src/read.rs
```diff
@@ -78,6 +78,22 @@ pub trait ComputeSize: ReadArgs {
...
 }
 
+/// A trait for types that are read at a position within enclosing data.
+///
+/// [`FontRead`] receives data already sliced to the start of the item, which
+/// discards everything before it. That is the wrong shape for an item holding
+/// offsets that resolve relative to the enclosing table rather than to the item
+/// itself: a GPOS value record's device offsets are measured from the start of
+/// the table containing the record. Such a type is given the enclosing data and
+/// its own position within it, and keeps both.
```
cmyr, line 88: I find this a bit confusing? in the majority of cases if a record contains an offset it is resolved relative to the parent table, so that isn't really the thing that distinguishes if we use this trait, is it?

This might actually be the main thing I'm confused by: when is this new 'positioned' stuff used? If it's opt-in, when would we expect to want to opt-in?

read-fonts/src/read.rs
```diff
@@ -78,6 +78,22 @@ pub trait ComputeSize: ReadArgs {
...
+///
+/// You should not generally need to call this directly; it is intended to be
+/// used from generated code.
+pub trait FontReadAt<'a>: Sized + ReadArgs {
```
cmyr, line 92: do we want to have `ComputeSize` as a bound as well?

font-codegen/src/parsing/fields.rs
```diff
@@ -35,6 +35,13 @@ pub(crate) struct Field {
...
+    pub(crate) positioned: bool,
```
cmyr, line 45: I do still find this unclear: is it expected that this is true for any record that contains an offset? I don't think it is, but as worded it sort of implies so?

cmyr (approved):
okay so my only real issue here is just that I don't find it very clear what types get this new pattern, and how that is determined. I guess it is spelled out, but maybe i would spell it out a bit more: something is positioned if it is explicitly declared as such, or if it contains a field that was so declared?

read-fonts/src/read.rs
dfrg: The trigger is an extern record with a lifetime. That's not the best signal but I think it does actually indicate 1) this record must borrow and 2) we can't declare a fixed layout for it. Those two things in combination suggest that the constructor requires parent data and offset, which is what this does.

read-fonts/src/read.rs
dfrg: Possibly, but it's not strictly required for this so I'd like to defer that to future work.

font-codegen/src/parsing/fields.rs
dfrg: yes, this comment is overly broad because it definitely isn't a general solution that captures every record that holds an offset. That's a desirable change but also huge and breaking which is what sent me down the rabbit whole of "breaking the world" :)

dfrg: Changed to
```rust
    /// This is required by fields whose type holds offsets that resolve against
    /// the enclosing table; see the `FontReadAt` trait in read-fonts. It is
    /// computed during resolution: seeded from the extern records declared with
    /// a lifetime, then closed over containment. See `build_positioned_set`.
```

The "closed over containment" is probably the context that was missing.

PR #2076, opened 2026-08-24
[read-fonts] Make GPOS value records positioned

dfrg:
A value record's device and variation index offsets are measured from the start of the table containing the record, so a record read from data sliced to itself cannot resolve them. Declare it positioned, and replace the owned struct with one that keeps the enclosing table's data, the record's offset within it, and the format.

Constructing a record now performs no reads. Fields are located by a popcount over the lower format bits, so a caller can read one field without touching its neighbours, and code that only wants to know which fields are present reads nothing at all. That is most of the win for the subsetter: computing a subset value format no longer builds two records per class pair.

`PairValueRecord` and `Class2Record` become positioned by containment, and `Class1Record` by holding an array of them.

`Value` and `ValueContext` go with the owned record. They resolved deltas to i32, which is not what a shaper wants: HarfBuzz scales the fractional delta and rounds once, so anything built on them would have to redo the work. Nothing in-tree used them.

Two hand-written helpers stay in gpos.rs, both hoisting reads out of a loop: `ClassValueRecords` reads a PairPosFormat2's counts and formats once and then addresses the 2D array arithmetically, and `PairSet::find_pair_value_record` binary searches on second glyph, reading only the glyph id of each probed record.

GPOS subsetting is 5-48% faster depending on the font. Subsetting every font in skera's test data, under three flag combinations, produces byte identical output.

Since GPOS value records were the only user of `Cursor::read_with_args` and `read_computed_array`, both are now dead; they are kept because codegen still emits them for any record added in that shape. The traversal hook no longer needs to know the name `ValueRecord`.


read-fonts/generated/generated_gpos.rs
```diff
@@ -1585,11 +1583,9 @@ impl<'a> SinglePosFormat2<'a> {
     }
 
     /// Array of ValueRecords — positioning values applied to glyphs.
-    pub fn value_records(&self) -> ComputedArray<'a, ValueRecord> {
+    pub fn value_records(&self) -> PositionedArray<'a, ValueRecord<'a>> {
```
dfrg: @cmyr I think this is what you're looking for from the previous PR, now codegen'd

skera/src/gpos/value_record.rs
```diff
@@ -9,71 +9,66 @@ use write_fonts::{
...
+            {
+                effective -= field;
+            }
         }
     }
 
-    if let Some(y_advance) = value_record.y_advance {
-        if !strip_empty || y_advance.get() != 0 {
-            value_format |= ValueFormat::Y_ADVANCE;
+    if strip_empty {
```
qxliu76: I think the code missed to remove flags with value 0 when strip_empty is true.

qxliu76: let me think twice:)

qxliu76: uh, I think it's a bug in Skera that I'll fix later in a separate PR. This change keeps the same logic with current Skera code

qxliu76 (approved):

skera/src/gpos/value_record.rs
qxliu76: or I wonder if AI could help fix this bug, the correct logic should be: for all fields, if strip_empty is true, strip fields with offset/value=0. If strip_hints is true, strip all device fields. Otherwise keep the original format. HB ref is here: https://github.com/harfbuzz/harfbuzz/blob/ebdb43f4d91a8cf5db0298f8305dd985a4988664/src/OT/Layout/GPOS/ValueFormat.hh#L160

qxliu76: or nvm, it's a tiny change anyway.:) This PR LGTM

dfrg: so strip_hints looks correct but strip_empty isn't handling devices?

dfrg: wait, this seems correct: if strip_hints is true, we drop all devices, otherwise we drop devices that are null. Then we just check the remaining fields. So we essentially always drop device fields with NULL offsets regardless of any strip_ setting

dfrg: or do you mean that when no strip_ is true then we should keep NULL devices?

dfrg: pushed a fix

qxliu76: Thank you!

qxliu76:
LGTM, nice change!

read-fonts/src/font_data.rs
```diff
@@ -274,6 +274,11 @@ impl<'a> Cursor<'a> {
         temp
     }
 
+    // emitted by codegen for record fields that take read args and are not
+    // positioned. GPOS value records were the only such fields, so nothing
+    // calls this today; kept because codegen still emits it for any record
+    // added in that shape.
+    #[allow(dead_code)]
```
cmyr: this was a hack that was really just for ValueRecord, and I don't expect us to want to handle some future type in the same way; we should just delete this.

cmyr (approved):
This follows naturally from 2075, no real notes except that we don't need to hold onto the now-unused library code (imo)

read-fonts/src/font_data.rs
dfrg: Done

PR #2077, opened 2026-08-25
[codegen] Support records embedded in tables

dfrg:
A table field whose type is a record only worked when the field carried read args, which is how GPOS value records are declared. A plain record field fell through to the scalar branch and tried to read it with `FontData::read_at`, which needs `Scalar`, so it did not compile.

The workaround was to declare a single record as an array of one. That leaves callers doing `.first().ok_or(OutOfBounds)?` for something that is structurally always present, and it hides the record from the table's minimum size, so a table too short to hold it still parsed. Both IndexSubtable2 and IndexSubtable5 did this.

Let such a field be declared directly. A record whose size is known at compile time (a zerocopy record) is borrowed straight out of the table's data with the existing `FontData::read_ref_at`, so the getter hands back `&'a Record` and the field counts toward MIN_SIZE like any scalar.

That last part is what makes the borrow sound: there is no default to fall back on when the data is short, so the getter unwraps, and it may only do that if MIN_SIZE covers the field. An embedded record must therefore precede any variable-length field, which is checked, with an error saying so. Working out a field's size now happens before fields are marked as validated at parse, since the marking asks for that size.

Traversal needs nothing hand-written: a generated record already has a `SomeRecord` impl. The two type names hardcoded in that match keep their existing path rather than moving to the new one.

The motivating case is MATH, which is largely embedded records: MathConstants alone holds fifty one of them.

Based on #2075 which must land first.


font-codegen/src/parsing.rs
```diff
@@ -504,20 +504,15 @@ impl Items {
...
-            {
-                field.validated_at_parse = true;
-            }
         }
 
-        // now that field types are resolved we can work out which of them are
-        // read at an offset within their enclosing table
+        // Now that field types are resolved we can work out which of them are
+        // read at an offset within their enclosing table, and which are records
+        // of a known fixed size. This has to precede the `validated_at_parse`
```
cmyr, line 511: we should have a test input that is a fixed-size structure after an optional field (if you haven't included one already)

font-codegen/src/table.rs
```diff
@@ -354,6 +354,18 @@ fn generate_to_owned_impl(item: &Table, parse_module: &syn::Path) -> syn::Result
...
+            // A table reads its fields on demand, so an embedded record is
+            // borrowed out of the table's data. A borrow has no default to
+            // fall back on when the data is short, so the getter unwraps,
+            // which is only sound if MIN_SIZE covers the field. (Records
+            // hold their fields outright and are unaffected.)
+            if matches!(phase, Phase::Analysis) && fld.fixed_size_record && !fld.validated_at_parse
+            {
+                return Err(logged_syn_error(
+                    fld.name.span(),
+                    "an embedded record must precede any variable-length field, so that it is covered by the table's minimum size",
```
cmyr, line 366: could we also return Option<&T>? Any reason to prefer one option over the other?

cmyr (approved):
nice, this feels like something that we mostly just didn't get around to because it didn't show up in the initially supported tables; seems like a reasonable gap to fill.

font-codegen/src/parsing.rs
dfrg: This is statically denied by codegen which I think you caught below :)

font-codegen/src/table.rs
dfrg: we could but I don't believe this sort of thing exists in the spec (how many times have we said this?)

cmyr: I prefer to leave these sorts of things as achievements that can be unlocked by future spec changes.

PR #2080, opened 2026-08-25
[read-fonts] Decode glyph coordinates through a slice iterator

dfrg:
Profiling a glyph draw puts read_points_fast at 58% of loading, which is more than everything else in the loader put together.

Its two coordinate passes read through a cursor, so each of the up to four bytes a point takes costs a bounds check, a Result and a saturating add of the offset. Both passes now walk one iterator over the bytes the flags did not use: next is a pointer comparison, and the y pass carries on from where the x pass stopped rather than tracking an offset of its own.

Point decoding, and loading a whole glyph, both measured interleaved over three rounds at 16ppem:

                decode          load
    arial      -13.1%          -5.6%
    times      -11.8%          -8.4%
    calibri    -11.2%          -7.7%
    segoeui     -9.8%          -5.0%
    consola    -15.7%

Two other shapes were tried and were worse. Counting each axis's byte length during the flags pass, to size the coordinate slices exactly and bounds check once, costs more in the flags pass than it saves: 20% to 25% slower than the cursor it replaced. Reading whole i16s rather than byte pairs makes no difference either way.

Decoding is unchanged. Every glyph of 145 fonts draws to the same path as before, and skrifa's own tests pass untouched.

### DIFF
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -260,11 +260,12 @@ impl<'a> SimpleGlyph<'a> {
                 break;
             }
         }
-        // Both coordinate passes walk one iterator over the bytes the flags
-        // did not use. A cursor would bounds check, build a `Result` and
-        // advance a saturating offset for each of the up to four bytes a point
-        // takes; `next` is a pointer comparison, and the y pass simply carries
-        // on from where the x pass stopped.
+        // This used to use a `Cursor` but that implies saturating
+        // arithmetic, bounds checking and `Result` building for each byte
+        // read.
+        //
+        // A byte slice iterator is just a pointer comparison and is
+        // significantly faster.
         let coords = self
             .glyph_data()
             .get(read_flags_bytes..)
--- a/read-fonts/src/tables/glyf.rs
+++ b/read-fonts/src/tables/glyf.rs
@@ -260,19 +260,29 @@ impl<'a> SimpleGlyph<'a> {
                 break;
             }
         }
-        let mut cursor = FontData::new(self.glyph_data()).cursor();
-        cursor.advance_by(read_flags_bytes);
+        // Both coordinate passes walk one iterator over the bytes the flags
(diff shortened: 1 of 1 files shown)


behdad:
FWIW, HB's glyph / variations load is super fast, in case you care to look at, or run the benchmarks.

dfrg:
Looks fairly similar, as to be expected. The two major differences are the branchless sign and the two byte read. Both ended up being slight regressions on this code. For the former, the multiply cost more than the branch (which LLVM emits as neg + cmov) and the latter required giving up the iterator which cost more than the two byte read saved. Don't think we can match HB without unsafe.

read-fonts/src/tables/glyf.rs
```diff
@@ -260,19 +260,29 @@ impl<'a> SimpleGlyph<'a> {
                 break;
             }
         }
-        let mut cursor = FontData::new(self.glyph_data()).cursor();
-        cursor.advance_by(read_flags_bytes);
+        // Both coordinate passes walk one iterator over the bytes the flags
```
drott: I have trouble understanding this comment, suggest to rephrase.

cmyr (approved):
Good catch; the `Cursor` type had a very specific purpose which has basically disappeared (doing the full validation pass during table construction) and it is probably not the best design for most other uses.

read-fonts/src/tables/glyf.rs
dfrg: changed to
``` rust
        // This used to use a `Cursor` but that implies saturating
        // arithmetic, bounds checking and `Result` building for each byte
        // read.
        //
        // A byte slice iterator is just a pointer comparison and is
        // significantly faster.
```

dfrg:
yeah, I think the saturating arithmetic really bites us in tight loops. I'll keep an eye out for similar cases

read-fonts/src/tables/glyf.rs
drott: Thank you, much better.

PR #2081, opened 2026-08-26
Take on smallvec in read-fonts and skrifa

dfrg:
We've had a hand-rolled partial implementation in skrifa for some time. We've since grown another, even more partial, implementation as well as several open coded copies of this pattern (conditional stack alloc or vec) in read-fonts.

This dependency is ubiquitous in the rust ecosystem and available in all of our target environments. I've chosen the floor of those versions (1.13.1) here.


cmyr (approved):
as long as it doesn't introduce vendoring issues it lgtm.

PR #2082, opened 2026-08-26
[read-fonts] Push DICT operands straight onto the stack

dfrg:
Every operand in a DICT went through a 40 byte Token -- 8 bytes of number and 32 of binary coded decimal state that only BlueScale ever reads -- built by parse_token behind a cursor length check, matched apart again by entries, and only then pushed. The entries loop now reads the byte and dispatches itself: integers parse and push directly, a real number pushes and records its components for BlueScale, and the operator paths are unchanged. parse_token stays for the public tokens iterator.

Subfont construction, which is two DICT parses and today runs per glyph (see the subfont caching notes), interleaved over eight rounds:

    AdobeVFPrototype   1548ns ->  526ns
    SourceSansPro       515ns ->  207ns
    cmunrm              416ns ->  174ns
    overall            -60.7%

End to end unhinted CFF drawing is 18.4% faster, five faces of five; hinted drawing is unchanged, as expected, because HintingInstance already builds its subfonts once. Glyph output -- unhinted and hinted, static and variable -- is unchanged across the corpus, and the dict unit tests assert the exact entry sequences.

### DIFF
--- a/read-fonts/src/ps/cff/dict.rs
+++ b/read-fonts/src/ps/cff/dict.rs
@@ -301,85 +301,113 @@ pub fn entries<'a>(
         if cursor.remaining_bytes() == 0 {
             return None;
         }
-        let token = match parse_token(&mut cursor) {
-            Ok(token) => token,
-            Err(Error::InvalidDictOperator(_)) => {
-                // Some buggy fonts have invalid dict operators. Clear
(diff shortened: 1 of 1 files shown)


cmyr (approved):
nice!

PR #2086, opened 2026-08-27
[skera-repack] only recompute coverage size when included glyphs set changes

qxliu76:
Port from: https://github.com/harfbuzz/harfbuzz/pull/6212


cmyr (approved):
makes sense!

PR #2087, opened 2026-08-27
Drop traversal

dfrg:
Traversal was an untyped, reflection-style view of a parsed font: a `SomeTable` trait object per table, `get_field(idx)` returning a `FieldType`, offsets resolved into further trait objects. It has been experimental since it landed, behind `experimental_traverse` and off by default.

Its only consumer was otexplorer, which goes with it. Everything else in the workspace builds and passes without it, and the two crates that named the feature did not use it: skrifa's `traversal` feature only forwarded it, and write-fonts enabled it in dev-dependencies for four assertion messages.

```
  read-fonts      -12,525    9,436 of it generated
  otexplorer         -889    the whole crate
  font-codegen       -311    the emitters
  docs, CI            -30
```

**Tables no longer implement `Debug`.** The generated impl deferred to `SomeTable`, so it was gated too and never existed in a default build. Nothing needs it: the only uses were `"{cmap:?}"` in three write-fonts tests, which now print the encoding records they are actually asserting about, and a stray `dbg!` left in the meta round-trip test. Emitting a name-only `Debug` for several hundred generated types would be a `fmt` function and a string literal each, and no caller for any of them.

Flags and enums keep the `Debug` that prints their set members -- that is hand-rolled in the emitter and has nothing to do with traversal. Records keep their derived `Debug`; they are plain structs.

Format groups and generic groups lose their `dyn_inner` dispatch along with the `SomeTable` impl it existed to serve.


rsheeter (approved):

cmyr:
oops was merged while i was looking. Only note is that there are a few remaining references to otexplorer in the script that checks for println?

dfrg:
ah, you're too thorough. will scrap those

PR #2088, opened 2026-08-27
Remove vestigial traces of traversal

dfrg:
The otexplorer println checks, the traverse_with codegen attribute and the parsing functionality for it.


cmyr (approved):
👋 🥲

PR #2089, opened 2026-08-28
[read-fonts] ps: clean up pass on error reporting

dfrg:
tl;dr Remove `Result`s and `Error` variants that provide no signal. 

----

The PostScript module returned `Result` in a lot of places where the only thing that could go wrong was that the data did not make sense, and the callers said so: `charset.string_id(gid).ok()?` inside `charset.rs` itself, `RawEncoding::new(..).ok()?` in `font.rs`, `.ok()` in skrifa's glyph names.

**Lookups return `Option`.** `charset.rs` constructed nineteen errors and every one was `ReadError::OutOfBounds`. `index.rs`'s `size_in_bytes` was the same, and reached it by `map_err`ing a richer error away. A glyph maps to a string id or it does not; there is nothing for a caller to act on.

Converting deletes code rather than adding it -- the bodies were already `Option` chains with `.ok_or(ReadError::OutOfBounds)` bolted on the end.

**`Index::get_offset` and `get` too, and their variants go with them.** `ZeroOffsetInIndex` and `InvalidIndexOffsetSize` looked like real signal, but they say "this INDEX is malformed" -- an offset of zero where the spec requires one or more, an offset size outside 1..=4 -- which is what every other failure in this module now answers with `None`.

**The encodings, for the same reason.** `CustomEncoding::new` reported `InvalidFormat` for a format byte that is not 0 or 1, and no caller read it.

**`Operator::read` as well.** It reported `InvalidCharstringOperator` with the offending opcode, and its one caller discards the error and clears the stack -- deliberately, because FreeType ignores reserved operators.

**An unknown DICT operator becomes a token, not an error.** `entries` already skips one and clears the stack, as FreeType does; `tokens` -- the raw view -- ended the walk instead, so a caller could not do the same thing from it. That made it useless as a general DICT parser at the first buggy font. `Token` gains `InvalidOperator(u8)` and `InvalidTwoByteOperator(u8)` (the first byte is always escape), the decision moves to the caller, and `InvalidDictOperator` goes.

**`Cff::charset` flattens.** It returned `Result<Option<Charset>, Error>`, where the `Ok(None)` arms were a CID-keyed font and a font with no charset, and the `Err` arms were all malformed data. One `Option` says both.

**`Error::Read` becomes `Error::Malformed`, and loses its payload.** It was the only variant named for an operation rather than a condition, which was defensible while it wrapped a `ReadError` and is not now that it also covers eleven direct constructions meaning "these bytes do not make sense".

Those eleven built a `ReadError::OutOfBounds` only for `From` to discard it a line later, so they say `Error::Malformed` directly. What still names `ReadError` here is the two `FontRead` impls, whose signature the trait fixes, and the `From` impl -- which stays: removing it turns 28 `?` on the existing API's reads into `.map_err(..)`, for no information.

**Six variants go, and two payloads.** `InvalidStackAccess` has not been constructed since #1686 relaxed the stack checks. `InvalidFontFormat` named two things that are simply malformed data. `ExpectedI32StackEntry(usize)` becomes `ExpectedInt`: a stack position is not something a caller can act on. `InvalidVariationStoreIndex` drops an index its caller passed in.

The one payload left, `InvalidSeacCode(i32)`, names a value read out of the font rather than one the caller supplied or a position in our own state. Every one of the fourteen variants is constructed.

`ReadError` references in `ps` go from 71 to 13, the module has 25 fewer `Result`-returning signatures, and `Error` goes from 20 variants to 14.

### DIFF
--- a/incremental-font-transfer/src/glyph_keyed.rs
+++ b/incremental-font-transfer/src/glyph_keyed.rs
@@ -1006,8 +1006,8 @@ impl GlyphDataOffsetArray for CFFAndCharStrings<'_> {
     fn offset_for(&self, gid: GlyphId) -> Result<u32, PatchingError> {
         self.charstrings
             .get_offset(gid.to_u32() as usize)
-            .map_err(|_| PatchingError::FontParsingFailed(ReadError::OutOfBounds))
             .map(|offset| offset as u32)
+            .ok_or(PatchingError::FontParsingFailed(ReadError::OutOfBounds))
     }
 
     fn all_offsets_are_ascending(&self) -> bool {
--- a/read-fonts/src/ps/cff/blend.rs
+++ b/read-fonts/src/ps/cff/blend.rs
@@ -95,7 +95,7 @@ impl<'a> BlendState<'a> {
         let variation_data = store.item_variation_data();
         let data = variation_data
             .get(self.store_index as usize)
-            .ok_or(Error::InvalidVariationStoreIndex(self.store_index))??;
+            .ok_or(Error::InvalidVariationStoreIndex)??;
         let region_indices = data.region_indexes();
         let regions = self.store.variation_region_list()?.variation_regions();
         // Precompute scalars for all regions up to MAX_PRECOMPUTED_SCALARS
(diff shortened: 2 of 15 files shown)


cmyr (approved):
looks reasonable!

PR #2090, opened 2026-08-28
[skera] use flat vector which outperforms hashmap lookup

qxliu76:
port from HB: https://github.com/harfbuzz/harfbuzz/pull/6187, but I use this vector for all layout tables, about 5% improvement on NotoNastaliqUrdu.


dfrg (approved):
Classic trade of space for time. LGTM!

PR #2091, opened 2026-08-28
[font-types] Add F48Dot16

dfrg:
This is the first of a three PR chain that optimizes IVS work. Next is #2092

A 64 bit fixed point number with 16 fractional bits, existing for what item variation delta accumulation actually computes: the sum of raw integer deltas times 16.16 scalars, which is exact in 48.16 and in nothing narrower. Fixed::mul_i32 produces one such term, wrapping rather than panicking per this module's style -- though the widened 32 bit product is at most 2^62 and cannot actually wrap -- and to_i64 rounds off the fraction the way the integer delta path always has.

Each variable value type gains apply_delta, taking the accumulated 48.16 delta and returning the varied value in single precision: deltas are raw counts of the target's own quantum, so Fixed scales by 1/65536, F2Dot14 by 1/16384, and the font unit types apply it unscaled. The arithmetic reproduces read-fonts' FloatItemDeltaTarget impls exactly -- the tests replicate those expressions and hold the results equal -- so migrating that trait onto these methods cannot move a value.

### DIFF
--- a/font-types/src/fixed.rs
+++ b/font-types/src/fixed.rs
@@ -339,6 +339,7 @@ fixed_impl!(F4Dot12, 16, 12, i16);
 fixed_impl!(F6Dot10, 16, 10, i16);
 fixed_impl!(Fixed, 32, 16, i32);
 fixed_impl!(F26Dot6, 32, 6, i32);
+fixed_impl!(F48Dot16, 64, 16, i64);
 
 fixed_mul_div_assign!(Fixed);
 fixed_mul_div_assign!(F26Dot6);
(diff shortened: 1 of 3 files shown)


qxliu76:
Code changes LGTM, though the PR comment is pretty AI that I got a headache while reading it.

qxliu76 (approved):

dfrg:
The worst part about that is I actually wrote most of this code 😂 (just not the commit message… shh)

qxliu76:
> The worst part about that is I actually wrote most of this code 😂 (just not the commit message… shh)

 😂 so hilarious

PR #2092, opened 2026-08-28
[read-fonts] Compute variation deltas exactly, in 48.16

dfrg:
Second in a 3 PR chain to optimize IVS work. Depends on #2091. Next is #2093

compute_delta now returns the exact F48Dot16 sum of raw deltas times 16.16 scalars, and the integer and float endings become the callers' one liners: to_i32 for the classic rounded delta, to_f64 behind compute_float_delta for targets that take deltas unrounded. The float path previously ran a separate f32 scalar pipeline; it is now the integer path's value before rounding, so the two can never disagree beyond that rounding. The sum cannot overflow even for hostile data: scalars are products of ratios the range guards keep at most one.

The shared loop also collects the measured micro work:

- the region scalar checks the pinned axis first, on the raw 2.14 values, skipping the start and end reads and three conversions for the common case of an instance on a region corner
- the region indices and the delta set are walked as the parallel arrays they are, dropping a per item bounds check that could never fail
- delta sets are read through a byte slice iterator rather than a cursor

VariationRegion::compute_scalar_f32 loses its last caller and is removed.

The float delta path serves shaping, COLR and avar2 -- HarfRust applies these to every varied GPOS value record -- and main computed it with f32 scalars and none of the loop work. Against main, resolved tables held once so the computation dominates, four variable faces, interleaved over eight rounds:

    float deltas, axes at maximum      -38.4%  (4/4 faster)
    float deltas, interior instance    -24.7%  (4/4 faster)
    integer deltas, same instances     -12.5% / -3.0%

Advances are unchanged across the corpus at both kinds of instance, and float results move only where the old f32 scalars disagreed with the 16.16 ones: at most 3.3e-3 raw delta units, around 2e-7 after 2.14 target scaling, far below one quantum.

### DIFF
--- a/font-codegen/README.md
+++ b/font-codegen/README.md
@@ -224,10 +224,6 @@ The following annotations are supported on top-level objects:
 - `#[validate(arg)]`: arg is either the literal 'skip' or the name of a method.
   If the name of a method, that method will be called during validation, and can
   implement custom validation logic.
-- `#[traverse_with(method name)]`: uncommon/hacky: provides a method name that
-  will be called in traversal code to get the `FieldType` for this field.
-  To skip traversing this field, you can use the 'skip' keyword
-  (`#[traverse_with(skip)]`).
 - `#[to_owned(expr)]`: uncommon/hacky: provide an expression that will be used
   in `FromObjRef` to convert the parse type to the compile type.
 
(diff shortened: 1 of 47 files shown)


dfrg:
Note that the fixed point delta computation was already done in 48.16. This work just wraps it up in a type.

The real change is that the floating point version now reuses the fixed point computation, keeps the fractional bits and converts at the end. Turns out that the conditional fixed point mul_div is actually faster than doing the int to float conversions for all values per region.

behdad:
LGTM.

PR #2093, opened 2026-08-28
[read-fonts] Drop the floating point delta path

dfrg:
Final in the IVS optimization work chain. Modest but measurable gains for COLRv1. Depends on #2092 

Delete compute_float_delta, FloatItemDelta and FloatItemDeltaTarget. compute_delta returns the exact 48.16 sum and F48Dot16 carries the conversions the float path existed for: to_f64() yields the same unrounded value (exact, since the sum has fewer than 53 significant bits) and the inherent apply_delta methods on Fixed, F2Dot14, FWord and UfWord reproduce the former trait impls' arithmetic bit for bit. The colrv1 resolver applies deltas to concrete types only, so no trait replaces FloatItemDeltaTarget; skrifa's color instance now holds [F48Dot16; N] delta arrays and calls apply_delta directly.

Migration for external users (e.g. HarfRust's GPOS value records): compute_float_delta(ix, coords)?.to_f64() becomes
compute_delta(ix, coords)?.to_f64(), value identical.

COLRv1 paint resolution on test_glyphs-glyf_colr_1_variable.ttf (200 glyphs at 5 varied instances): -2.3% vs main, outputs bit identical.

### DIFF
--- a/fauntlet/src/font/skrifa.rs
+++ b/fauntlet/src/font/skrifa.rs
@@ -6,7 +6,7 @@ use skrifa::{
     raw::{
         ps::{cff::CffFontRef, type1::Type1Font},
         types::F2Dot14,
-        FontRef, ReadError, TableProvider,
+        FontRef, TableProvider,
     },
     GlyphId, MetadataProvider, OutlineGlyphCollection,
 };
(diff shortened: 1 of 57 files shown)


behdad:
LGTM.

cmyr (approved):
👍

PR #2094, opened 2026-08-28
[read-fonts] COLR's hand-written accessors answer with Option

dfrg:
Replace `Result` with `Option` for hand written accessors in the `COLR` table. None of these results returned anything useful.


dfrg:
oops, not standalone

PR #2095, opened 2026-08-28
Rework some skrifa errors

dfrg:
The first commit swaps out `Result` for `Option` in the hand-written `COLR` table accessors in read-fonts. Nothing in our must-maintain-compat set touches them directly.

The second is a skrifa patch that replaces error variants that hold `ReadError` with a payload free `Malformed` variant, like was done for the postscript work. No functional changes, all just mechanical swapping of error types. Final result is that skrifa contains no reference to `ReadError` at all.


cmyr (approved):
looks good!

PR #2099, opened 2026-09-01
[ci] Compile Chromium's font code against this tree 

dfrg:
Check that Chrome's rust code compiles. This one also runs in CI since it's sparse pull + cargo check.

Includes a tiny fix for the fontconfig action: the apt version of meson wasn't new enough so grab meson from pypi instead.

edit: fontconfig run completed successfully in 2m1s (https://github.com/googlefonts/fontations/actions/runs/33545973337) so this also patches that config to run on push.


resources/scripts/check_chromium_compat.sh
```diff
@@ -0,0 +1,100 @@
...
+EOF
+
+{
+    echo '// Generated: every file in the checkout that reads fonts.'
+    for src in "${sources[@]}"; do
+        name=$(basename "$src" .rs)
+        echo "#[path = \"$src\"]"
+        echo "mod $name;"
+    done
+} > "$harness/src/lib.rs"
```
cmyr, line 91: hmm, so we add a mod for each file we find, but what if there are submodules? wouldn't that break?

Also how does this handle the same filename appearing in different projects? Easy to imagine with, say, lib.rs?

cmyr (approved):
will definitely be nice to know when we're breaking the world. One little comment inline, but happy to see this get in :)

resources/scripts/check_chromium_compat.sh
dfrg: Good catch. It's not ideal but so far, every module in chrome is a simple single file helper. Given the recent "resource reallocation" I don't expect that to change in the near future. When it does, this will scream at us and we can fix it :)

PR #2101, opened 2026-09-03
[read-fonts] add Loca::get that returns Option<LocaGlyph>

dfrg:
Adds a new `LocaGlyph` type so we can represent the empty glyph and then adds `Loca::get` which returns `Option<LocaGlyph>`. The current `get_glyf` method is used by Skia so we'll need to defer removing it but new code should use this one.

I looked at modifying codegen to add an `Empty` variant to `glyf::Glyph` but that required new complexity for one case and would have made the bbox accessors return `Option` which seemed like a bad trade all around.

### DIFF
--- a/read-fonts/src/tables/loca.rs
+++ b/read-fonts/src/tables/loca.rs
@@ -3,9 +3,9 @@
 //! [loca]: https://docs.microsoft.com/en-us/typography/opentype/spec/loca
 
 use crate::{
-    read::{FontRead, ReadArgs, ReadError},
+    read::{FontRead, ReadArgs},
     table_provider::TopLevelTable,
-    FontData,
+    FontData, ReadError,
 };
 use types::{BigEndian, GlyphId, Tag};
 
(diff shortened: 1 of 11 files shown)


dfrg:
```
error[E0599]: no method named `ok` found for enum `Option<T>` in the current scope
   --> /home/runner/work/fontations/fontations/skia/src/ports/fontations/src/bitmap.rs:198:10
    |
194 |       let glyph = font_ref
    |  _________________-
195 | |         .loca(None)
196 | |         .ok()?
197 | |         .get_glyf(glyph_id, &glyf_table)
198 | |         .ok()??;
    | |_________-^^
    |
``` 

and the Skia check is already paying off. I'll use a different approach.

read-fonts/src/tables/loca.rs
```diff
@@ -59,24 +59,64 @@ impl<'a> Loca<'a> {
...
+
+/// What a glyph's entry in the `loca` table leads to.
+///
+/// A glyph with no contours — a space, say — is described by an empty range,
+/// and finding one is a success. It is separate from the glyph being
+/// unreadable, which [`Loca::get`] reports by answering `None`.
+#[derive(Clone)]
+pub enum LocaGlyph<'a> {
+    /// The glyph has no outline.
+    Empty,
```
cmyr, line 101: yes, this is definitely better API (aka the current API was not intentional)

cmyr (approved):
agree this is an improvement!

PR #2102, opened 2026-09-03
[read-fonts] svg, trak, cvar and dsig helpers return Option

dfrg:
Just a batch of changes to hand written accessors to swap the `Result<T, ReadError>` out for `Option<T>`.


cmyr (approved):
👍

PR #2103, opened 2026-09-03
[skera] refactor ValueRecord related subsetting code

qxliu76:
rewrite of subsetting code for SinglePos/PairPos tables. About 5% runtime imrpovement on Roboto, compared to HB, it's improved from 1.25X->1.16X

### DIFF
--- a/skera/src/gpos/value_record.rs
+++ b/skera/src/gpos/value_record.rs
@@ -8,10 +8,22 @@ use crate::{
 use write_fonts::{
     read::{
         collections::IntSet,
-        tables::gpos::{ValueFormat, ValueRecord},
+        tables::{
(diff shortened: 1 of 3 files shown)


skera/src/gpos/value_record.rs
```diff
@@ -21,47 +33,50 @@ const DEVICE_FIELDS: [ValueFormat; 4] = [
     ValueFormat::Y_ADVANCE_DEVICE,
 ];
 
+#[inline]
+fn popcount8(v: u8) -> u8 {
```
garretrieger, line 38: Can you add a comment explaining what this method does (ie. returns the number of 1 bits in a u8)?

garretrieger (approved):

PR #2105, opened 2026-09-05
[read-fonts] Add MATH table and supporting functions.

dfrg:
Broken into two commits for reviewer convenience. The first is just the codegen boilerplate. The second contains all the helper accessor methods. These were specifically designed to map nicely to the `hb_ot_math_*` functions because we'll need to implement those to replace HarfBuzz in Chrome. They should generalize to all consumers of this table though and the raw fields/methods are still there.

Plan for implementing on the HarfRust side: https://github.com/harfbuzz/harfrust/issues/466


cmyr:
does this put to rest (to my great shame) https://github.com/googlefonts/fontations/pull/1269 ?

dfrg:
The first rule of #1269 is that we do not talk about #1269.

read-fonts/src/tables/math.rs
```diff
@@ -1,14 +1,439 @@
 //! The [MATH](https://learn.microsoft.com/en-us/typography/opentype/spec/math) table
 
-use super::layout::{CoverageTable, DeviceOrVariationIndex};
+use super::layout::{CoverageTable, Device, DeviceOrVariationIndex};
 
 include!("../../generated/generated_math.rs");
 
+/// Which corner of a glyph a [`MathKern`] applies to.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum MathKernSide {
```
cmyr: 'side' seems like a funny word here, is it our choice or are we matching prior art?

read-fonts/src/tables/math.rs
```diff
@@ -1,14 +1,439 @@
...
+pub struct MathKernEntry {
+    /// The height the band reaches up to, or `None` for the last band, which
+    /// is unbounded.
+    pub max_height: Option<MathValue>,
+    /// The kern to apply within the band.
+    pub kern: MathValue,
+}
+
+impl Math<'_> {
+    /// Whether this is one of the Cambria Math builds that stores
```
cmyr, line 57: 🤢

cmyr (approved):
nice and easy now that we've got records-in-tables...

read-fonts/src/tables/math.rs
dfrg: fair! renaming to `MathKernCorner`

read-fonts/src/tables/math.rs
dfrg: same

PR #2109, opened 2026-09-06
[read-fonts] Inline AAT lookup format dispatch

behdad:
## Motivation

HarfRust repeatedly reconstructs AAT class-lookup views while applying
`morx` subtables. The generated `Lookup::read_with_args` format switch is
small, but without an inline hint it remains opaque across the
`read-fonts` crate boundary.

Add an opt-in `#[read_inline]` codegen annotation and use it only for the
AAT `Lookup` format group. I tested adding the hint to every format group,
but that caused a measurable regression in an OpenType workload, so this
keeps the change narrowly scoped.

## Performance

With release LTO builds of HarfRust and `benchmark-harf`:

| Workload | Before | After |
| --- | ---: | ---: |
| GeezaPro / `fa-words.txt` | 11.2 ms | 10.7 ms |
| LucidaGrande / `en-thelittleprince.txt` | 4.58 ms | 4.51 ms |
| Menlo / `en-thelittleprince.txt` | 3.35 ms | 3.32 ms |

On Devanagari Sangam / `hi-words.txt`, fixed-iteration hardware counters
dropped from 891,866,354 to 879,745,089 instructions. Cycles are sensitive
to final code layout but were about 1–3% lower in repeated runs.

## Testing

`cargo test --workspace`

### DIFF
--- a/font-codegen/src/format_group.rs
+++ b/font-codegen/src/format_group.rs
@@ -24,6 +24,7 @@ pub(crate) fn generate(item: &TableFormat, items: &Items) -> syn::Result<TokenSt
         });
 
     let format = &item.format;
+    let read_inline = item.attrs.read_inline.as_ref().map(|_| quote!(#[inline]));
     // if we have any fancy match statement we disable a clippy lint
     let mut has_any_match_stmt = false;
     let match_arms = item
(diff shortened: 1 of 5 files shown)


dfrg:
Interesting that adding inline on all format groups causes a regression. I can only guess that LTO is choosing not to inline coverage and classdef reads as I believe those are the only ones we touch on the OT hot path. Was the regression on Amiri?

behdad:
> Interesting that adding inline on all format groups causes a regression. I can only guess that LTO is choosing not to inline coverage and classdef reads as I believe those are the only ones we touch on the OT hot path. Was the regression on Amiri?

Yeah as I pushed agents to close more gaps between HB & HR, code gen & inlining decisions showed more and more. I'm open to either option.

PR #2110, opened 2026-09-07
[read-fonts] Add exact variation deltas to HVAR, VVAR and MVAR

dfrg:
Expose 48.16 delta methods for the metrics variation tables. `MVAR` gets a helper that resolves the IVS once for batch queries.

### DIFF
--- a/read-fonts/src/tables/vvar.rs
+++ b/read-fonts/src/tables/vvar.rs
@@ -1,17 +1,19 @@
 //! The [VVAR (Vertical Metrics Variation)](https://docs.microsoft.com/en-us/typography/opentype/spec/vvar) table
 
 use super::variations::{self, DeltaSetIndexMap, ItemVariationStore};
+use types::F48Dot16;
 
 include!("../../generated/generated_vvar.rs");
 
 impl Vvar<'_> {
-    /// Returns the advance height delta for the specified glyph identifier and
-    /// normalized variation coordinates.
-    pub fn advance_height_delta(
-        &self,
-        glyph_id: GlyphId,
-        coords: &[F2Dot14],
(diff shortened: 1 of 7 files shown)


read-fonts/src/tables/vvar.rs
```diff
@@ -52,4 +53,135 @@ impl Vvar<'_> {
...
+    }
+
+    /// Returns the change a location makes to the vertical origin of a glyph.
+    ///
+    /// The value carries every bit the item variation store computed. It is
+    /// a caller that decides how to round it into a whole design unit, and
+    /// implementations differ on that.
+    ///
+    /// Returns `None` where the table says nothing readable about the glyph.
+    pub fn v_org(&self, glyph_id: GlyphId, coords: &[F2Dot14]) -> Option<F48Dot16> {
```
cmyr: I do generally wonder if these names aren't a bit _too_ concise, but I understand how tsb and bsb would follow from lsb/rsb, which are names already in use, and if we're using those than I can see how you end up here? I'd probably still prefer vert_origin/v_origin though, personally..

cmyr (approved):
looks good!

read-fonts/src/tables/vvar.rs
dfrg: Agreed that this became worse when the `_delta` was dropped. Also, I realized that these aren't in our compatibility set so I'm just going to reuse the old names but extend `v_org_delta` to `v_origin_y_delta` which is what this actually is.

Should note that when I get to the later glyph metrics work on the new font API, `lsb` will be spelled `h_bearing_x`. Perhaps we can make similar changes to these later.

PR #2111, opened 2026-09-07
[read-fonts] experimental Font API: add GlobalMetrics

dfrg:
Adds font-wide metrics to `Font` and `FontInstance`. The list of metrics is slightly different here than in `skrifa::metrics::Metrics` because this slices the set to the values that don't require the `post` table (those will follow later in `StyleMetrics`). The reasoning is that none of those are required for shaping so we don't want to pull large `post` 2.0 tables unnecessarily.

Also removes any notion of size from `FontInstance` (that will come later as some `Scaler` type), so `FontInstance` becomes internally `Arc`'d and we can now create an instance at the default location for free.

Requires #2110

### DIFF
--- a/read-fonts/src/model/font/instance.rs
+++ b/read-fonts/src/model/font/instance.rs
@@ -1,6 +1,7 @@
 //! Font instance representation.
 
 use super::Font;
+use crate::model::{metrics::GlobalMetrics, once::Once};
 use crate::{
     tables::{
         avar::Avar,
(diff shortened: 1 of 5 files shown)


read-fonts/src/model/font/instance.rs
```diff
@@ -9,55 +10,117 @@ use crate::{
...
-pub struct FontInstance {
+/// A font at one location in its design space.
+///
+/// Cloning is cheap: clones share one instance rather than copying it.
+#[derive(Clone)]
+pub struct FontInstance(Repr);
+
+#[derive(Clone)]
+enum Repr {
+    Default(Font),
```
cmyr, line 28: Is this only used for variable fonts, or also statics?

read-fonts/src/model/font/instance.rs
```diff
@@ -9,55 +10,117 @@ use crate::{
...
+        }
+    }
+}
+
+impl From<&Font> for FontInstance {
+    /// A font is an instance of itself, at the default location.
+    fn from(font: &Font) -> Self {
+        Self(Repr::Default(font.clone()))
+    }
+}
```
cmyr, line 112: by-ref only?

read-fonts/src/model/metrics/global.rs
```diff
@@ -0,0 +1,356 @@
...
+    },
+    TableProvider,
+};
+use types::{BoundingBox, F2Dot14, F48Dot16, Fixed, Tag};
+
+/// Widens a `Fixed` to `F48Dot16`. Both hold 16 fractional bits, so this is
+/// exact.
+fn widen(value: Fixed) -> F48Dot16 {
+    F48Dot16::from_bits(value.to_bits() as i64)
+}
```
cmyr: should this be a method on `Fixed`? (maybe not, if it's only internal)

cmyr (approved):
looks good, a few ignorable questions inline

read-fonts/src/model/font/instance.rs
dfrg: Exactly the right question, and the motivation for this change. This allows future code to accept only `FontInstance` while providing a free (sans atomic counter bump) conversion from `Font` for both statics and variable fonts at default location.

read-fonts/src/model/font/instance.rs
dfrg: Good catch. Both conversions make sense here.

read-fonts/src/model/metrics/global.rs
dfrg: and it already is! Fixed (pun intended?)

PR #2113, opened 2026-09-08
[read-fonts] new Font API: add glyph metrics, using yoke to cache lifetime bound data

dfrg:
This introduces the `yoke` crate for safely caching data with lifetimes. Combined with onces, the result is fast, lazy loading... pretty much equivalent to HarfBuzz.

The caching is demonstrated with a very minimal `GlyphMetrics` view over `Font` currently only supporting unvaried advances. The full `HVAR`/`gvar` fallback ladder comes in the next patch, with additional metrics to follow. Note that this does add a new fallback: a font with no horizontal metrics table yields `units_per_em / 2` for all glyphs, which matches what HB does.

### DIFF
--- a/read-fonts/src/model/font/cache.rs
+++ b/read-fonts/src/model/font/cache.rs
@@ -1,0 +1,34 @@
+//! Tables parsed once for a font and kept.
+//!
+//! A parsed table borrows the bytes it came from, so it cannot simply be
+//! stored beside them. [`Yoke`] makes that expressible by holding the parsed
+//! value together with the tables it borrows, behind an [`Arc`] that outlives
+//! both however they are dropped.
+//!
+//! Every table gets its own cell, so reading one never drags in another. That
+//! matters for `glyf` and `loca`, which outlines and metrics in both
+//! directions all read, and for `gvar`, which is typically about half a
+//! variable font and must not be read to answer what another table can.
+
+use super::FontTables;
+use alloc::sync::Arc;
+use yoke::{Yoke, Yokeable};
+
+/// A parsed table held beside the tables it borrows.
+pub(crate) struct TableCache<Y: for<'a> Yokeable<'a>>(Yoke<Y, Arc<FontTables>>);
+
+impl<Y: for<'a> Yokeable<'a>> TableCache<Y> {
+    /// Parses once, with `read`, and keeps the result.
+    pub(crate) fn read<F>(tables: Arc<FontTables>, read: F) -> Self
+    where
+        F: for<'a> FnOnce(&'a FontTables) -> <Y as Yokeable<'a>>::Output,
+    {
+        Self(Yoke::attach_to_cart(tables, read))
+    }
+
+    /// Returns the parsed table, borrowed for no longer than this.
+    #[inline]
+    pub(crate) fn get(&self) -> &<Y as Yokeable<'_>>::Output {
+        self.0.get()
+    }
+}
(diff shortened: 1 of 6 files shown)


read-fonts/src/model/font/cache.rs
```diff
@@ -0,0 +1,34 @@
...
+//! both however they are dropped.
+//!
+//! Every table gets its own cell, so reading one never drags in another. That
+//! matters for `glyf` and `loca`, which outlines and metrics in both
+//! directions all read, and for `gvar`, which is typically about half a
+//! variable font and must not be read to answer what another table can.
+
+use super::FontTables;
+use alloc::sync::Arc;
+use yoke::{Yoke, Yokeable};
```
cmyr, line 15: I trust we know this dep is already vendored, but just to call it out...

read-fonts/src/model/metrics/glyph.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Writes the advance width of each glyph to its slot, in order.
+    ///
+    /// Each width passes through `convert`, so a caller working in another
+    /// number type writes into that type directly.
+    #[inline]
+    pub fn h_advance_batched<'o, V: 'o>(
+        &self,
+        convert: impl Fn(F48Dot16) -> V,
+        glyphs: impl Iterator<Item = (GlyphId, &'o mut V)>,
```
cmyr, line 63: curious API, but I assume this is perf motivated?

read-fonts/src/model/metrics/glyph.rs
```diff
@@ -0,0 +1,436 @@
...
+            return;
+        }
+        raw.run(self.num_glyphs, convert, glyphs)
+    }
+
+    /// The Type 1 half of [`h_advance_batched`](Self::h_advance_batched).
+    ///
+    /// A Type 1 charstring states its own width, so reading one means
+    /// running it. Out of line because that has nothing in common with
+    /// reading a table, and no `sfnt` should pay for its presence.
```
cmyr, line 89:
```suggestion
    /// A Type 1 charstring states its own width, so reading one means
    /// running it. `#[inline(never)]` because that has nothing in common with
    /// reading a table, and no `sfnt` should pay for its presence.
```

cmyr (approved):
looks good!

read-fonts/src/model/font/cache.rs
dfrg: yes, please continue to call these out. there's a comment in Cargo.toml above the dep reminding us to check versions with links to the chrome and android trees. I verified google3 myself.

read-fonts/src/model/metrics/glyph.rs
dfrg: yeah, we use something similar in HarfRust so we need to be able to emulate it here. We have two arrays of structs where the glyph id is in one and the advance in another, typed as `i32`, so this allows us to do a fast fill of those, converting types and applying scale as we go. Just for reference, the batched call is about twice as fast as loading each one independently, and we're fighting for cycles on the shaper

read-fonts/src/model/font/cache.rs
behdad: Yeah it is by the venerable Manish from Google i18n team.

dfrg: Colin, just want to give credit where it's due because I think you envisioned almost this exact path years ago :) I fought for borrow only as long as I could but the format eventually defeated me...

cmyr: a stopped clock is right twice a day etc etc ;)

PR #2114, opened 2026-09-08
[skera] Convert repacker graph traversals to BFS

qxliu76:
Port from: https://github.com/harfbuzz/harfbuzz/pull/6130

None of the repacker traversals require DFS order, so BFS can be used instead to eliminate the possibility of stack overflow on deep graphs. All directed graph traversal methods (find_subgraph_nodes, find_subgraph_nodes_incoming_edges, find_subgraph_size, find_32bit_roots, and duplicate_subgraph) are unified into a single traverse_directed_bfs implementation.

 Additionally:
    - find_connected_nodes is converted from recursive DFS to an iterative DFS using an explicit stack.
    - Graph::from_serializer now validates that object count does not exceed MAX_VERTICES upfront.
    - Fixes a bug in find_32bit_roots() caused by DFS traversal order: when a node was reachable via both 16-bit and 32-bit links, DFS would explore the 16-bit link first and mistakenly recurse into child 32-bit links, incorrectly marking them as top-level roots. BFS visits level-by-level and avoids recursing through already-marked roots. 
    
Tests added:
    - test_deep_graph_traversal
    - test_32bit_roots_traversal

### DIFF
--- a/skera/src/graph.rs
+++ b/skera/src/graph.rs
@@ -557,6 +557,10 @@ impl Graph {
     pub(crate) fn from_serializer(s: &Serializer) -> Result<Self, RepackError> {
         let packed_obj_idxs = s.packed_idxs();
         let count = packed_obj_idxs.len();
+        if count > MAX_VERTICES {
+            return Err(RepackError::ErrorMaxOperationsExceeded);
+        }
+
         let mut this = Graph {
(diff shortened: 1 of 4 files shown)


dfrg (approved):
LGTM

garretrieger (approved):

PR #2115, opened 2026-09-08
[read-fonts] new Font API: variable advance widths

dfrg:
Based on #2113 but split out so the first one focuses on the yoke caching mechanism. This adds the fallback ladder for variable advances: first `HVAR`, then `gvar` phantom points. The variation batch function is intentionally `inline(never)` because the inline code slows down the fast path for static fonts. The hot loops are unswitched in each path, leaning on `run` and `run_varied` in `RawGlyphMetrics` to avoid code duplication.

### DIFF
--- a/read-fonts/src/model/font/cache.rs
+++ b/read-fonts/src/model/font/cache.rs
@@ -5,12 +5,14 @@
 //! value together with the tables it borrows, behind an [`Arc`] that outlives
 //! both however they are dropped.
 //!
-//! Every table gets its own cell, so reading one never drags in another. That
-//! matters for `glyf` and `loca`, which outlines and metrics in both
+//! Every table gets its own cell, so reading one never drags in another.
+//! That matters for `glyf` and `loca`, which outlines and metrics in both
 //! directions all read, and for `gvar`, which is typically about half a
 //! variable font and must not be read to answer what another table can.
 
 use super::FontTables;
+use crate::tables::{glyf::Glyf, gvar::Gvar, hvar::Hvar, loca::Loca};
+use crate::TableProvider;
 use alloc::sync::Arc;
 use yoke::{Yoke, Yokeable};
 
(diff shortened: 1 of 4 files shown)


read-fonts/src/model/font/cache.rs
```diff
@@ -5,12 +5,17 @@
 //! value together with the tables it borrows, behind an [`Arc`] that outlives
 //! both however they are dropped.
 //!
-//! Every table gets its own cell, so reading one never drags in another. That
+//! Every table gets its own cell, so reading one never drags in another.
+//! That matters for `glyf` and `loca`, which outlines and metrics in both
+//! directions all read, and for `gvar`, which is typically about half a
+//! variable font and must not be read to answer what another table can. That
```
cmyr: this is a dupe?

read-fonts/src/model/metrics/glyph.rs
```diff
@@ -176,6 +227,28 @@ impl<'a> RawGlyphMetrics<'a> {
...
+        &self,
+        num_glyphs: u32,
+        delta: impl Fn(GlyphId) -> F48Dot16,
+        convert: impl Fn(F48Dot16) -> V,
+        glyphs: impl Iterator<Item = (GlyphId, &'o mut V)>,
+    ) {
+        for (gid, out) in glyphs {
+            let advance = self
+                .stored_advance(num_glyphs, gid)
+                .saturating_add(delta(gid));
```
cmyr: the tireless assistant points out that this will return non-zero advances for glyphs that are OOB, because we compute/add the delta unconditionally. I will let you decide if this is a concern or not.

cmyr (approved):
a few Qs inline :)

read-fonts/src/model/font/cache.rs
dfrg: *my* tireless assistant apparently got tired... and I fixed that particular comment for different reasons *twice*

read-fonts/src/model/metrics/glyph.rs
dfrg: confirmed, and while it's probably not a problem in practice, I think it's worth fixing anyway

PR #2118, opened 2026-09-09
[write-fonts] fix orphaned duplicate when isolating a subgraph root

anthrotype:
When a space root is linked from both 16 and 32-bit space, `isolate_subgraph_hb` duplicates it and then has to move the wide links over to the duplicate. It looked for those links among the parents of the *duplicate*, but [`duplicate_subgraph`](https://github.com/googlefonts/fontations/blob/e3cbfdd2769e8b051fc7255ab24992768b132bb3/write-fonts/src/graph.rs#L1091-L1093) creates it with an empty parent list, so nothing was ever remapped: the original kept both its parents and the copy was left orphaned, still linking to its (possibly shared) children. The next `sort_shortest_distance` then panics with ["cycle or something?"](https://github.com/googlefonts/fontations/blob/e3cbfdd2769e8b051fc7255ab24992768b132bb3/write-fonts/src/graph.rs#L490-L495), because the orphan is still in those children's parent lists, so the sort never sees all of their incoming links and mistakes that for a cycle.

I hit this compiling a large GPOS with fontc where extension promotion promoted almost every lookup, so a few subtables deduplicated across a promoted and an unpromoted lookup ended up as space roots with one 32-bit and one 16-bit parent. The existing `duplicate_shared_root_subgraph` test has exactly that shape but only counts nodes, so it passed with the orphan in place (my test also sorts and checks that the wide link moved and that nothing is unreachable).

Here I do what harfbuzz's [isolate_subgraph](https://github.com/harfbuzz/harfbuzz/blob/bc678801ce4be8e83c5b7c013fca805ea71a38e4/src/graph/graph.hh#L998-L1046) does: i.e. collect the wide parents of the roots up front, before anything is duplicated, and afterwards retarget any wide link from those parents whose target was duplicated. This also covers a wide link from one of those parents to a duplicated node other than the root, which the old loop never considered.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -695,14 +695,13 @@ impl Graph {
             // for the roots, we set the edge count to the number of long
             // incoming offsets; if this differs from the total number of
             // incoming offsets it means we need to dupe the root as well.
-            let inbound_wide_offsets = self.nodes[root]
-                .parents
-                .iter()
-                .filter(|(_, len)| !matches!(len, OffsetLen::Offset16))
-                .inspect(|(parent_id, _)| {
+            let mut inbound_wide_offsets = 0;
+            for (parent_id, len) in &self.nodes[root].parents {
+                if !matches!(len, OffsetLen::Offset16) {
                     wide_parents.insert(*parent_id);
-                })
-                .count();
+                    inbound_wide_offsets += 1;
+                }
+            }
             subgraph.insert(*root, inbound_wide_offsets);
             self.find_subgraph_map_hb(*root, &mut subgraph);
         }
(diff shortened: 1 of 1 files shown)


anthrotype:
reproducer here: https://github.com/googlefonts/fontations/actions/runs/34352853170/job/102470122108?pr=2118#step:8:458

write-fonts/src/graph.rs
```diff
@@ -696,6 +699,9 @@ impl Graph {
                 .parents
                 .iter()
                 .filter(|(_, len)| !matches!(len, OffsetLen::Offset16))
+                .inspect(|(parent_id, _)| {
+                    wide_parents.insert(*parent_id);
```
cmyr, line 701: using `inspect` just for the side effect opportunity feels a bit bad here, I'd be more inclined to just redo this as a loop? but not a blocker 🤷

cmyr (approved):
good catch... one note inline but overall looks good :)

PR #2119, opened 2026-09-09
[read-fonts] new Font API: vertical glyph advances

dfrg:
This adds support for vertical advances which mostly just mirrors the horizontal path with the exception that the fallback requires the horizontal line metrics (ascent and descent). This comes from a new `h_line` method on `GlobalMetrics`. The selection logic there matches FreeType/Skrifa and *almost* matches HarfBuzz but with an additional guard tries to avoid zeroed metrics which actually occurs in Arial Narrow Bold.


read-fonts/src/model/metrics/global.rs
```diff
@@ -160,6 +160,53 @@ impl GlobalMetrics {
...
+                _ => match (self.win_ascent, self.win_descent) {
+                    (Some(ascent), Some(descent)) => Some(LineBox {
+                        ascender: ascent,
+                        descender: -descent,
+                        line_gap: F48Dot16::ZERO,
+                    }),
+                    _ => hhea,
+                },
+            },
+        }
```
cmyr: found this logic hard to follow and wondered if it would be easier to express as a match on a tuple, results are ambiguous:


```suggestion
match (self.hhea_line, self.typo_line) {
            (_, Some(typo)) if self.use_typo_metrics => Some(typo),
            (Some(hhea), _)
                if hhea.ascender != F48Dot16::ZERO || hhea.descender != F48Dot16::ZERO =>
            {
                Some(hhea)
            }
            (_, Some(typo))
                if typo.ascender != F48Dot16::ZERO || typo.descender != F48Dot16::ZERO =>
            {
                Some(typo)
            }
            (hhea, _) => match (self.win_ascent, self.win_descent) {
                (Some(ascent), Some(descent)) => Some(LineBox {
                    ascender: ascent,
                    descender: -descent,
                    line_gap: F48Dot16::ZERO,
                }),
                _ => hhea,
            }
        }
```

cmyr (approved):
👍

read-fonts/src/model/metrics/global.rs
dfrg: Thanks, this looks better.

PR #2120, opened 2026-09-09
[ci] Read only the archive the packages come from

dfrg:
The fontconfig job installs six packages, all from Ubuntu's own archive, and was failing on a repository it installs nothing from:

    E: Failed to fetch https://dl.google.com/linux/chrome-stable/... Hash Sum mismatch
       Last modification reported: Wed, 09 Sep 2026 09:41:12 +0000
       Release file created at:    Wed, 09 Sep 2026 17:16:59 +0000

The runner image carries Chrome's apt repository. Those two timestamps are that repository mid-publish, offering a Release file newer than the `Packages.gz` it points at, which apt reports as a hash sum mismatch and which fails the whole update rather than the one source. Retrying does not help while it lasts, and refetching the index makes it likelier by discarding a copy that still agreed with itself.

Dropping the third-party sources leaves the archive every package here comes from. The image keeps Ubuntu's own in `ubuntu.sources` rather than beside them, so removing these by name takes nothing that is needed. Microsoft's is removed alongside Chrome's: it is the other one the image carries and it can publish just as unluckily.

### DIFF
--- a/.github/workflows/fontconfig-compat.yml
+++ b/.github/workflows/fontconfig-compat.yml
@@ -62,8 +62,17 @@ jobs:
       # meson comes from PyPI rather than apt: fontconfig asks for >= 1.11.0
       # and the runner image ships 1.3.2. pipx because Ubuntu marks its
       # Python externally managed, so a plain pip install is refused.
+      #
+      # The runner image carries third-party apt repositories that nothing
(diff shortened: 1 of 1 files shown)


cmyr (commented):
👍

PR #2122, opened 2026-09-09
[read-fonts] new Font API: scaling metrics

dfrg:
Supports scaling global and glyph metrics. I was not intending to do this, but supporting the HB/HR model without duplicating code requires *merging* font functions callbacks that provide already scaled values into the fallback ladder for vertical metrics. For example, vertical advances fallback to line height (ascender - descender) when `vmtx` isn't present, and *those* values can be overridden in HB by providing a `get_font_extents` callback which gets called from HB's own vertical advance query. The same applies for vertical origins: they lean on a potentially overridden glyph extents for fallback.

So this provides an API that allows us to match what HB does without having to duplicate the fallback functionality everywhere. We never really offered these in either read-fonts or skrifa so matching HB's semantics breaks nothing and makes sense.

Built on #2119

### DIFF
--- a/read-fonts/src/model/metrics/scaled.rs
+++ b/read-fonts/src/model/metrics/scaled.rs
@@ -1,0 +1,874 @@
+//! Metrics in a caller's own units.
+
+use super::{GlobalMetrics, GlyphExtents, GlyphMetrics, LineBox, LineExtents};
+use types::{BoundingBox, F26Dot6, F48Dot16, Fixed, GlyphId};
+
+/// Converts a measurement in design units into a caller's own units.
+///
+/// Written by whatever owns scaling in a text stack: a shaper, or the bridge
+/// between one and this crate. Callers that want design units read them from
+/// [`GlyphMetrics`] and [`GlobalMetrics`] and never implement this.
+///
+/// Scaling is more than a multiplication, which is why it is a trait rather
+/// than a factor this crate would apply. HarfBuzz scales each axis by its own
+/// fixed point number, in a format the caller picks, and is particular about
+/// how it rounds and in what order it applies things. An implementation keeps
+/// all of that, while the rules for combining several measurements into one
+/// metric stay here.
+///
+/// The arithmetic is on this trait rather than bounds on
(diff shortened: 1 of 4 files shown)


wmedrano (approved):

read-fonts/src/model/metrics/scaled.rs
```diff
@@ -0,0 +1,649 @@
...
+
+use super::{GlobalMetrics, GlyphExtents, GlyphMetrics, LineBox, LineExtents};
+use types::{BoundingBox, F48Dot16, GlyphId};
+
+/// Converts a measurement in design units into a caller's own units.
+///
+/// Design units carry no size and a caller laying out text has one. A metric
+/// built from several measurements must combine them after they reach that
+/// size rather than before, so the arithmetic belongs where the caller can
+/// see it. A scale is how that happens without this crate choosing a size.
```
cmyr: this isn't very clear to me. What is... a size?

cmyr (approved):
I find the Scale stuff confusing because it really isn't clear who the intended implementor is? Maybe that doesn't matter (if the consumer is specialized and is expected to know what's up) but it _is_ confusing as general API.

read-fonts/src/model/metrics/scaled.rs
dfrg: A size is just a ppem, which is what FontInstance used to carry. For better or worse, HB doesn’t scale that way. It takes x_scale and y_scale in arbitrary fixed point and is particular about how it rounds and in what order the operations apply. This API was about my fifth attempt to thread that needle so that we could hold the complicated logic in one place but allow the user complete control over scaling. A follow up could provide types like Scale26Dot6 and ScaleF32 so other users have something to reach for and to act as examples.

I’m considering HB normative here because everything either uses its metrics directly or supplies their own wholesale.

dfrg: Reworked the comment and added `Scale` impls for 26.6 and f32 which should cover all the non-HB cases.

PR #2123, opened 2026-09-09
[read-fonts] Read a charstring's width without drawing it

dfrg:
We currently have to evaluate a whole charstring to grab the advance width but those are usually encoded up front so this adds an early out when only the advance is needed. This only matters for type1 and bare cff fonts but those are read by PDFium and we don't want to regress performance there.

    CFF     1.269us -> 223ns per glyph
    Type 1  4.249us -> 166ns per glyph


cmyr (approved):
👍

PR #2124, opened 2026-09-09
[read-fonts] Compute charstring bounds

dfrg:
Adds a `ControlBoundsSink` to match `ControlBoundsPen` and some helper functions to compute glyph extents for CFF and Type1.


cmyr (approved):
okay, sure, fine, whatever you need 🤡

PR #2125, opened 2026-09-09
Reserve capacity for output data

wmedrano:
-15% runtime improvement for:

```sh
hyperfine --warmup 2 \
  -L target main,wm-opt \
  'target/release/skera-{target} --path /usr/local/google/home/wmedrano/Downloads/FruityGirl-Regular.ttf -u "0-ffff" -n 10000 -o /tmp/out-{target}.ttf'
```

| Command | Mean ± σ          | Min … Max         |
|---------|-------------------|-------------------|
| wm-opt  | 1.139 s ± 0.016 s | 1.125 s … 1.180 s |
| main    | 1.341 s ± 0.048 s | 1.295 s … 1.458 s |

### DIFF
--- a/write-fonts/src/font_builder.rs
+++ b/write-fonts/src/font_builder.rs
@@ -182,10 +182,10 @@ impl<'a> FontBuilder<'a> {
         let table_order = self.ordered_tags();
 
         let mut position = header_len as u32;
-        let mut checksums = Vec::new();
+        let mut checksums = Vec::with_capacity(self.tables.len() + 1);
         let head_tag = Tag::new(b"head");
 
-        let mut table_records = Vec::new();
+        let mut table_records = Vec::with_capacity(self.tables.len());
         for tag in table_order.iter() {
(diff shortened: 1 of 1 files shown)


behdad:
Nothing against it: but does this reserve the full size of the source table?

In HarfBuzz we have a simple estimator for target subset table sizes:

https://github.com/harfbuzz/harfbuzz/blob/f5efbbef3841bdcdc6427bde7c2971f62ac8c8f1/src/hb-subset-table.hh#L74-L108

wmedrano:
> Nothing against it: but does this reserve the full size of the source table?
> 
> In HarfBuzz we have a simple estimator for target subset table sizes:
> 
> https://github.com/harfbuzz/harfbuzz/blob/f5efbbef3841bdcdc6427bde7c2971f62ac8c8f1/src/hb-subset-table.hh#L74-L108

These tables are already transformed. Added a small check that emits a warning if we don't allocate properly.

cmyr (approved):
👍

PR #2127, opened 2026-09-09
Speed up full-font subsetting in Skera

behdad:
Stacked on #2125.

## Summary

- iterate cmap mappings directly instead of materializing intermediate `IntSet` and `HashMap` collections
- remove the unused codepoint-to-glyph map from `Plan`
- serialize simple glyphs directly from borrowed slices instead of allocating and copying one `Vec<u8>` per glyph
- let `FontBuilder` accept a trusted precomputed checksum for unchanged raw tables
- pass canonical, identity-mapped `CBLC`/`CBDT` tables through unchanged, avoiding a large temporary allocation, an intermediate bitmap-data copy, and a redundant checksum scan

## Performance

Release builds, five interleaved runs of 50,000 full-font subsets using `FruityGirl-Regular.ttf`:

| Implementation | Median |
| --- | ---: |
| Skera on `wm-opt` | 4.14 s |
| First two commits in this PR | 3.99 s |
| This PR | 1.31 s |
| HarfBuzz `hb-subset` | 1.63 s |

The bitmap fast path makes Skera approximately 68% faster than the base branch and approximately 20% faster than HarfBuzz for this font. The resulting Skera file is byte-for-byte identical before and after the bitmap optimization. Skera and HarfBuzz both produce a 711,664-byte font with 96 glyphs and the same table set and table sizes.

The earlier benchmark font contained `glyf`, so the simple-glyph change helped there. FruityGirl is bitmap-only: its 710,072-byte `CBDT` table accounts for almost the whole font. Profiling showed Skera allocating a serializer for `CBDT` despite handling it through `CBLC`, copying all bitmap bytes through an intermediate buffer, and then spending most of the remaining time checksumming those unchanged bytes. The new conservative fast path removes all three costs when the glyph mapping and bitmap layout prove the tables will be emitted unchanged.

Benchmark arguments:

```sh
--unicodes="*" --num-iterations=50000
```

## Validation

- `cargo test -p write-fonts -p skera --all-features`
- `cargo +stable clippy -p write-fonts --lib --all-features --no-deps -- -D warnings` (with allowances for two pre-existing lints from the newer local stable toolchain)
- `cargo +stable clippy -p skera --all-targets --all-features --no-deps -- -D warnings` (with allowances for pre-existing lints from the newer local stable toolchain)
- `cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/write-fonts/src/font_builder.rs
+++ b/write-fonts/src/font_builder.rs
@@ -182,10 +182,10 @@ impl<'a> FontBuilder<'a> {
         let table_order = self.ordered_tags();
 
         let mut position = header_len as u32;
-        let mut checksums = Vec::new();
+        let mut checksums = Vec::with_capacity(self.tables.len() + 1);
         let head_tag = Tag::new(b"head");
 
-        let mut table_records = Vec::new();
+        let mut table_records = Vec::with_capacity(self.tables.len());
         for tag in table_order.iter() {
             // safe to unwrap as ordered_tags() guarantees that all keys exist
             let data = self.tables.get_mut(tag).unwrap();
@@ -212,6 +212,8 @@ impl<'a> FontBuilder<'a> {
         let mut writer = TableWriter::default();
         directory.write_into(&mut writer);
         let mut data = writer.into_data().bytes;
+        data.reserve_exact((position as usize).saturating_sub(data.len()));
+        let expected_data_len = data.capacity();
         checksums.push(read_fonts::tables::compute_checksum(&data));
 
         // Summing all the individual table checksums, including the table directory's,
@@ -234,6 +236,12 @@ impl<'a> FontBuilder<'a> {
             let rem = round4(table.len()) - table.len();
             let padding = [0u8; 4];
             data.extend_from_slice(&padding[..rem]);
+        }
+        if expected_data_len != data.len() {
+            log::warn!(
+                "suboptimal: allocated {expected_data_len} bytes but needed {actual_len}",
+                actual_len = data.len()
+            );
         }
         data
     }
--- a/write-fonts/src/font_builder.rs
+++ b/write-fonts/src/font_builder.rs
@@ -20,6 +20,7 @@ const CFF2: Tag = Tag::new(b"CFF2");
 #[derive(Debug, Clone, Default)]
 pub struct FontBuilder<'a> {
     tables: BTreeMap<Tag, Cow<'a, [u8]>>,
+    checksums: BTreeMap<Tag, u32>,
 }
 
 impl TableDirectory {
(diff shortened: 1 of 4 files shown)


qxliu76:
so I did measure perf on both commits separately(based on Will's PR), neither has perf gains:

| Implementation | Median |
| --- | ---: |
| Skera on `wm-opt` | 3.425 s |
| first commit on Plan | 3.945 s |
| second commit on glyf |  3.496 s |

All runs used `--num-iterations=50000`
Apparently your machine is faster than mine:)

behdad:
Interesting.

qxliu76:
I guess the font file you used is different from mine.

behdad:
Ah maybe. Please email me yours.

qxliu76:
I still see 3X slowdown with this font file on main branch: 
[fruity-girl.zip](https://github.com/user-attachments/files/32029430/fruity-girl.zip)

qxliu76:
can you download the file from my commet?

behdad:
Yes. Got it. Lemme try.

behdad:
Retried with FruityGirl and confirmed the reported gap: the earlier commits measured ~3.99s versus ~1.63s for HarfBuzz over 50k iterations. This font is bitmap-only, and profiling exposed redundant allocation, copying, and checksumming of its 710KB CBDT table. The two new commits add a conservative unchanged-bitmap pass-through; Skera now measures ~1.31s median versus ~1.63s for HarfBuzz, with byte-identical Skera output before/after the optimization.

qxliu76:
Thanks, nice improvement on this case! 
Changes in write-fonts would need @dfrg or @cmyr 's review first. 
It seems just handle this special scenario in skera, I'll take a look if write-fonts change is accepted.

behdad:
It's a bit of a cheat, since it reuses table checksum if full table not changed...

dfrg:
> Thanks, nice improvement on this case! Changes in write-fonts would need @dfrg or @cmyr 's review first. It seems just handle this special scenario in skera, I'll take a look if write-fonts change is accepted.

I don’t see a problem with adding an escape hatch to avoid computing a checksum.

write-fonts/src/font_builder.rs
```diff
@@ -105,6 +106,22 @@ impl<'a> FontBuilder<'a> {
     /// A builder method to add raw data for the provided tag
     pub fn add_raw(&mut self, tag: Tag, data: impl Into<Cow<'a, [u8]>>) -> &mut Self {
```
cmyr, line 107: worth updating the docs here to say, like: "see [`add_raw_with_checksum`] as an alternative if the table checksum is also known."

qxliu76 (approved):

qxliu76:
LGTM. The glyf table optimization is very nice, subsetting SimpleGlyph is pretty similar to HB now. 
Also nice to add a fast path for this special scenario in CBLC! 

I feel it's still interesting to check if we could improve write-fonts with a more general approach. I think if we use` --unicodes=* --unicodes-=0X41`, skera would still be much slower than HB. I cannot test it now because skera doesn't support `-=` in command line option parsing. I'll try ask my AI  Agent

PR #2128, opened 2026-09-10
[write-fonts] remap VariationIndex inside Extension lookups

anthrotype:
RemapVarStore for PositionLookup treated Extension the same as the contextual lookups, i.e. as having no metrics of its own, and never descended into the wrapped subtable. So any value record or anchor with deltas inside a `useExtension` lookup kept its temporary delta-set id and blew up at write time with "Attempted to write PendingVariationIndex". In fea-rs terms, this is enough to reproduce it on a variable font:

```
lookup K useExtension { pos a b (wght=400:-20 wght=900:-40); } K;
```

The same lookup without `useExtension` compiles fine. AFDKO-style sources put kern in extension lookups routinely, so this isn't exotic.

So here I added RemapVarStore impls for ExtensionSubtable and ExtensionPosFormat1 that recurse into the inner subtable, plus a test that round-trips an extension-wrapped PairPosFormat1 through the remap and checks the resolved index. The test is a separate first commit so it can be seen failing before the fix.

### DIFF
--- a/write-fonts/src/tables/gpos.rs
+++ b/write-fonts/src/tables/gpos.rs
@@ -412,6 +412,82 @@ mod tests {
 
     use super::*;
 
+    #[test]
+    fn remap_variation_indices_through_extension() {
+        use crate::tables::variations::{
+            ivs_builder::VariationStoreBuilder, RegionAxisCoordinates, VariationRegion,
+        };
+        use font_types::F2Dot14;
+
+        let region = VariationRegion::new(vec![RegionAxisCoordinates {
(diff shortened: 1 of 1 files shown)


anthrotype:
repro at https://github.com/googlefonts/fontations/actions/runs/34487832389/job/102906558678?pr=2128#step:8:458

```
thread 'tables::gpos::tests::remap_variation_indices_through_extension' (6507) panicked at write-fonts/src/tables/layout.rs:496:9:
Attempted to write PendingVariationIndex.
```

dfrg (approved):

PR #2131, opened 2026-09-10
[skera] Optimize glyf and loca subsetting performance

qxliu76:
Improve performance in glyf and loca subsetting by reducing allocations, avoiding repetitive vector resizing, and eliminating branchy flag decoding:

- Preallocate loca_out with exact table size: Instead of relying on estimate_subset_table_size heuristics (which can under-estimate and trigger reallocations), allocate (num_output_glyphs + 1) * entry_size upfront and write entries directly into slices, eliminating repeated extend_from_slice calls and capacity checks on every glyph.

- Single serializer write for unmodified simple glyphs: In SimpleSubsetGlyph, when no hinting or overlap flag changes are requested, pass the contiguous slice (header + instructions + glyph data) to the serializer in a single embed_bytes call rather than three separate ones.

- Branchless coordinate length decoding: Replace 4 conditional branches per coordinate flag in trim_simple_glyph_padding with a precomputed 256-byte lookup table (COORD_BYTES).

- Accurate subset_glyphs preallocation: Allocate subset_glyphs capacity using plan.new_to_old_gid_list.len() and iterate using direct zip traversal instead of index lookups.

Benchmark results (skera-benchmarks):
  MPLUS1-Variable-all: -7.3% (3.47 ms -> 3.23 ms) 
  Mplus1p-Regular-all: -3.8% (3.07 ms -> 2.93 ms)
  Fraunces-all:        -2.3% (287 µs -> 280 µs)

Assissted by Gemini

### DIFF
--- a/skera/src/glyf_loca.rs
+++ b/skera/src/glyf_loca.rs
@@ -1,6 +1,5 @@
 //! impl subset() for glyf and loca
 use crate::{
-    estimate_subset_table_size,
     serialize::Serializer,
     Plan, Subset,
     SubsetError::{self, SubsetTableError},
@@ -36,8 +35,7 @@ impl Subset for Glyf<'_> {
         let loca = font.loca(None).or(Err(SubsetTableError(Loca::TAG)))?;
         let head = font.head().or(Err(SubsetTableError(Head::TAG)))?;
 
-        let num_output_glyphs = plan.num_output_glyphs;
-        let mut subset_glyphs = Vec::with_capacity(num_output_glyphs);
+        let mut subset_glyphs = Vec::with_capacity(plan.new_to_old_gid_list.len());
         let mut max_offset: u32 = 0;
 
         for (new_gid, old_gid) in &plan.new_to_old_gid_list {
@@ -70,7 +68,7 @@ impl Subset for Glyf<'_> {
 
         //TODO: support force_long_loca in the plan
         let loca_format: u8 = if max_offset < 0x1FFFF { 0 } else { 1 };
-        let loca_out = write_glyf_loca(font, plan, s, loca_format, &subset_glyphs)?;
+        let loca_out = write_glyf_loca(plan, s, loca_format, &subset_glyphs)?;
 
         let head_out = subset_head(&head, loca_format);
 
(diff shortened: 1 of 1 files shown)


skera/src/glyf_loca.rs
```diff
@@ -223,24 +219,32 @@ impl<'a> SimpleSubsetGlyph<'a> {
...
             .contains(SubsetFlags::SUBSET_FLAGS_NO_HINTING);
-        let instructions = if drop_hinting {
-            &[]
+        let set_overlaps_flag = plan
+            .subset_flags
+            .contains(SubsetFlags::SUBSET_FLAGS_SET_OVERLAPS_FLAG);
+
+        let (header, instructions, glyph_data) = if !drop_hinting && !set_overlaps_flag {
+            let total_len = header_len + g.instruction_length() as usize + glyph_data.len();
+            (&[][..], &[][..], glyph_bytes.get(..total_len)?)
```
wmedrano, line 231: `&[][..]` should just be `&[]`?

skera/src/glyf_loca.rs
```diff
@@ -86,38 +84,33 @@ fn padded_size(len: usize) -> usize {
...
         let mut value = 0_u16.to_be_bytes();
-        for ((new_gid, _), i) in plan.new_to_old_gid_list.iter().zip(0u16..) {
-            let gid = new_gid.to_u32();
+        for ((new_gid, _), g) in plan.new_to_old_gid_list.iter().zip(subset_glyphs) {
+            let gid = new_gid.to_u32() as usize;
 
             while last < gid {
-                loca_out.extend_from_slice(&value);
+                let pos = (last + 1) * 2;
+                loca_out[pos..pos + 2].copy_from_slice(&value);
```
wmedrano, line 106: These will panic if we get the initial size wrong.

Would be nice if there was a way to make it more obvious that we are always within `0..total_len` but I'm not sure of a any good ways to do that so 🤷🏽 . Good thing our test suite is good.

dfrg (approved):
LGTM

wmedrano (approved):

skera/src/glyf_loca.rs
dfrg: I think `&[]` sometimes gets typed as `&[T; 0]` which then fails to unify. Not sure if that's happening here though.

skera/src/glyf_loca.rs
qxliu76: We pre-allocated exact size needed for loca table, I think this should be fine

skera/src/glyf_loca.rs
qxliu76: yeah `&[]` doesn't work here, `&[][..] `works since it explicitly tells the compiler to treat the empty array as a slice reference rather than an array reference.

PR #2132, opened 2026-09-10
Move skrifa's outline to path conversion to read-fonts

dfrg:
First commit copies the path converter to read-fonts with the following changes:
* Takes an array with a projection function rather than an iterator so the user doesn't have to provide the last point.
* Drops the noisy error type that nobody ever checked. Path conversion only fails if the font data is malformed (invalid sequence of flags on a simple glyph) so skrifa now says that.

Second commit drops the original code from skrifa and replaces it with calls to read-fonts instead.

All in the service of supporting TrueType outline loadingin read-fonts.


qxliu76:
For trading code reviews: since I don't have enough context about these fns, did a high level review, code changes LGTM

dfrg:
Thanks :) Luckily this is well tested code and we have fauntlet in CI that compares the final result against FreeType so if it passes a high level review, then all should be good.

PR #2133, opened 2026-09-10
Move skrifa's scratch memory helpers to read-fonts

dfrg:
This is the bucketed stack allocator with fallback to heap along with the function for carving up bytes into typed slices.

Simply moves them (with better names) to read-fonts as public API under a new `mem` module so they can be used by both read-fonts and skrifa.


cmyr (approved):
looks good!

PR #2135, opened 2026-09-10
[skera] Serialize composite glyphs without temporary buffers

qxliu76:
Keep composite glyphs as borrowed slices rather than allocating a temporary Vec for each composite glyph during subsetting.

- Change SubsetGlyph::Composite from Vec<u8> to &'a [u8].
- In trim_composite_glyph, determine the trimmed glyph length by scanning component records and instructions directly on the borrowed font data.
- Rename write() to serialize() on SimpleSubsetGlyph and SubsetGlyph.
- In serialize_composite_glyph, copy the trimmed slice into the Serializer once and mutate flags (OVERLAP_COMPOUND, clearing WE_HAVE_INSTRUCTIONS when dropping hints) and remapped glyph IDs in-place in the serializer buffer.
- Remove obsolete helper functions (subset_glyph, subset_simple_glyph, and subset_composite_glyph) and update unit tests to use SubsetGlyph::new and serialize() directly.
- Add unit tests for composite glyph hinting removal and overlap flags.

### DIFF
--- a/skera/src/glyf_loca.rs
+++ b/skera/src/glyf_loca.rs
@@ -111,7 +111,7 @@ fn write_glyf_loca(
             value = ((offset >> 1) as u16).to_be_bytes();
             let pos = (last + 1) * 2;
             loca_out[pos..pos + 2].copy_from_slice(&value);
-            g.write(s)?;
+            g.serialize(s, plan)?;
             if padded_len > g.len() {
                 s.embed_bytes(&[0])
                     .map_err(|_| SubsetError::SubsetTableError(Glyf::TAG))?;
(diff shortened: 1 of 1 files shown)


dfrg (approved):
LGTM. 

One nit: these patches are starting to grow a lot of manual u16be reads and writes. Might be worth having an LLM look into replacing these with some kind of cursor type.

qxliu76:
> LGTM.
> 
> One nit: these patches are starting to grow a lot of manual u16be reads and writes. Might be worth having an LLM look into replacing these with some kind of cursor type.

good idea!

PR #2136, opened 2026-09-11
[write-fonts] split subtables of Extension lookups when packing

### LINKED ISSUE #2129: [write-fonts] packer never splits subtables of a lookup that is already an Extension
When a GPOS lookup arrives at the packer already wrapped in Extension subtables (e.g. because the FEA said `useExtension`), `split_subtables_if_needed` only matches the bare PairPos and MarkToBase lookup types, so the type 9 lookup falls through and its inner subtables are never split. Promotion doesn't help either, the lookup is already an extension. If the inner PairPos is too big for 16-bit offsets on its own the pack just fails.

https://github.com/googlefonts/fontations/blob/05c2b229bf4305dbbbaef2f80e6ac03db9e231f1/write-fonts/src/graph.rs#L1032-L1041

I hit this building Source Serif 4 as a variable font with fontc (9 masters, kern in a `lookup KERN useExtension` block, as Adobe sources tend to do): the merged PairPosFormat2 ends up with ~5000 VariationIndex children and the packer reports 10720 overflows from that one subtable. The same source with `useExtension` stripped packs fine, the log shows the pair lookup being split into 5 subtables and then promoted to extension by the packer itself (until #2128 it panicked on the unresolved VariationIndex first).

hb-repacker handles this in its own `split_subtables_if_needed`:

https://github.com/harfbuzz/harfbuzz/blob/f5efbbef3841bdcdc6427bde7c2971f62ac8c8f1/src/graph/gsubgpos-graph.hh#L127-L156

We should do the same in write-fonts: unwrap the extension subtable before dispatching to `split_pair_pos` / `split_mark_to_base`, and wrap each of the split results back in its own `ExtensionPosFormat1`.

anthrotype:
Fixes #2129

An Extension lookup is now splittable when the lookup type it wraps is: the existing PairPos / MarkToBase splitters run on the subtables the extensions point to, and each new piece gets its own ExtensionPosFormat1, like hb-repacker does. The `make_extension` helper that promotion used is hoisted to a `Graph` method so both paths share it.

The tests fail on main and are a separate first commit.

With this, Source Serif 4, whose kern lookup uses `useExtension`, builds with fontc to a GPOS identical to the one you get when the lookup is left bare and the packer promotes it itself.

### DIFF
--- a/write-fonts/src/graph.rs
+++ b/write-fonts/src/graph.rs
@@ -838,15 +838,6 @@ impl Graph {
     }
 
     fn actually_promote_subtables(&mut self, to_promote: &[ObjectId]) {
-        fn make_extension(type_: LookupType, subtable_id: ObjectId) -> TableData {
-            const EXT_FORMAT: u16 = 1;
-            let mut data = TableData::new(TableType::Named("ExtensionPosFormat1"));
-            data.write(EXT_FORMAT);
-            data.write(type_.to_raw());
-            data.add_offset(subtable_id, 4, 0);
-            data
-        }
-
         for id in to_promote {
             // 'id' is a lookup table.
             // we need to:
@@ -857,7 +848,7 @@ impl Graph {
             let mut lookup = self.objects.remove(id).unwrap();
             let lookup_type = lookup.type_.to_lookup_type().expect("validated before now");
             for subtable_ref in &mut lookup.offsets {
-                let ext_table = make_extension(lookup_type, subtable_ref.object);
+                let ext_table = Self::make_extension(lookup_type, subtable_ref.object);
                 let ext_id = self.add_object(ext_table);
                 subtable_ref.object = ext_id;
             }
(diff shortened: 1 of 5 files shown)


anthrotype:
repros here:

https://github.com/googlefonts/fontations/actions/runs/34622563520/job/103339822139#step:8:459

write-fonts/src/graph.rs
```diff
@@ -1043,6 +1057,17 @@ impl Graph {
         }
     }
 
+    /// For an Extension lookup, the lookup type of the wrapped subtables.
+    ///
+    /// Returns `None` for an empty lookup.
+    fn extension_subtable_type(&self, lookup: ObjectId) -> Option<u16> {
+        let first_subtable = self.objects[&lookup].offsets.first()?.object;
```
cmyr: kind of speculative, but if we happen to have bad data (multiple extension subtables that have different extension types) this will cause us to treat them all as the type of the first, which would lead to confusing errors later?

It's hard to imagine this happening by accident so I'm not _too_ concerned, but it might be worth either writing it down or asserting that the inner types are consistent, if it isn't too expensive..

cmyr (approved):
looks good, one little note inline that you can probably ignore :)

write-fonts/src/graph.rs
anthrotype: Done in 2fe6b374, `extension_subtable_type` now debug_asserts that all the extensions agree.
FWIW read-fonts makes the same assumption in `PositionLookup::subtables`: it takes the type from the first parseable extension and reads the rest as that.

PR #2137, opened 2026-09-11
[release] push only the tags cargo-release just created

anthrotype:
Same as googlefonts/fontc#2141 -- release.sh here ends with the same `git push --tags`, so it gets the same fix.


cmyr (approved):
👍

PR #2139, opened 2026-09-11
[skera] 2 bug fixes

qxliu76:
see commit messages for details

### DIFF
--- a/skera/src/gpos/single_pos.rs
+++ b/skera/src/gpos/single_pos.rs
@@ -506,8 +506,10 @@ mod test {
         plan.glyphset_gsub.insert(GlyphId::from(30_u32));
 
         // Map old varidx -> new varidx
-        plan.layout_varidx_delta_map.insert(0x00030004, (0x00070008, 0));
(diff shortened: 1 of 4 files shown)


skera/src/gpos/single_pos.rs
```diff
@@ -312,7 +312,7 @@ impl CollectVariationIndices for SinglePosFormat2<'_> {
         };
         let glyph_set = &plan.glyphset_gsub;
         let value_count = self.value_count();
-        let record_size = compute_record_len(value_format);
+        let record_size = 2 * compute_record_len(value_format);
```
dfrg: If we never need the actual value count then I’d suggest `compute_record_len` should just return the number of bytes directly so callers don’t need to remember to multiply by two.

dfrg (approved):
LGTM with one suggestion.

skera/src/gpos/single_pos.rs
qxliu76: good point! Done.

PR #2140, opened 2026-09-11
[skera] Track visited nodes in repacker duplicate_subgraph

qxliu76:
In HarfBuzz, `duplicate()` is invoked inline during the BFS traversal, which inserts the cloned index into `index_map` immediately and prevents revisiting nodes with multiple incoming paths.

In Skera, because mutating vertices during graph traversal violates borrow rules, nodes to duplicate are first collected into a `to_duplicate` vector and then cloned afterwards. Without a visited set during the BFS traversal, shared descendant nodes in DAG structures were visited multiple times,
 leading to redundant enqueuing and creating multiple clone vertices for
the same original node.

Add a `visited: &mut IntSet<u32>` parameter to `duplicate_subgraph` to ensure each node is visited and queued for duplication at most once. Pass and reuse the visited scratch set through `isolate_subgraph` (from `assign_spaces` and `resolve_overflows`) to avoid per-call allocations.


skera/src/graph.rs
```diff
@@ -1437,7 +1444,8 @@ impl Graph {
             }
         }
 
-        if !self.isolate_subgraph(&mut roots_to_isolate)? {
+        let mut visited_subgraph_nodes = IntSet::empty();
+        if !self.isolate_subgraph(&mut roots_to_isolate, &mut visited_subgraph_nodes)? {
```
wmedrano: You can also pass in `&mut IntSet::empty()` directly if you want.

wmedrano (approved):

garretrieger (approved):

wmedrano (approved):

skera/src/graph.rs
qxliu76: Done

PR #2144, opened 2026-09-15
[skera] Fix overflow panic on empty ClassDef

qxliu76:
An empty `ClassDef` assigns class 0 to every glyph, but several places computed an inclusive end glyph as `start_glyph + glyph_count - 1`. That underflows when `glyphCount` is 0 *and* `startGlyphID` is 0.

See also https://github.com/harfbuzz/harfbuzz/pull/6262 for read-fonts fix


cmyr (approved):
looks good!

garretrieger (approved):

PR #2145, opened 2026-09-16
[skera] Fix gvar short offset overflow when glyphs are remapped

qxliu76:
The choice between Offset16 and Offset32 for glyphVariationDataOffsets is made by summing the variation data of the retained glyphs, but that sum indexed the source gvar with the new gid instead of the old one. The write loop below it correctly uses the old gid, so the estimate bore no relation to what was actually emitted.

When the mapping is not the identity, the estimate is typically far too small, and the short format gets selected for data that does not fit. Short offsets store offset/2 in a uint16 and so top out at 0x1FFFE bytes; past that every offset silently wraps around and the table is corrupt.

Add tests covering both offset formats under a non-identity gid mapping. The bug is only observable when the gids are remapped and the retained data crosses the 0x1FFFE limit, so the long offset test builds its gvar programmatically rather than from a byte literal.


cmyr (approved):
👍

PR #2148, opened 2026-09-17
[write-fonts] Compute cmap14 length and record count

cmyr:
Previously these were fields the caller had to fill in. The length has to cover the UVS tables as well as the header: fonttools slices the subtable by it before decompiling, and ots requires it to be exact. Since the packer doesn't promise to lay offset targets out contiguously (or to keep duplicates), we write the UVS tables inline ourselves, so the layout and the length are correct by construction.

see https://github.com/googlefonts/fontc/issues/1543 and https://github.com/googlefonts/fontc/pulls/2158


anthrotype (approved):
LGTM

PR #2149, opened 2026-09-18
[read-fonts] glyf part 1: Add the outline error and coordinate modes

dfrg:
First of six patches adding a `glyf` outline loader beside the existing one. The files land without a `mod` declaration, so nothing compiles them yet and each patch can be read on its own; the last one turns the module on and CI takes over.

`error.rs` is what can go wrong building an outline. `scale.rs` is the `Scale` trait and its three implementations: `Scale26Dot6` reproduces FreeType, `ScaleF32` reproduces HarfBuzz, and `Unscaled` stays in font units for an autohinter. Loading takes the same steps whichever is in use; the trait is the arithmetic that differs.

The trait is sealed, since one method takes a type this crate does not export.


read-fonts/src/tables/glyf/outline/scale.rs
```diff
@@ -0,0 +1,969 @@
...
+        offset: Point<i32>,
+        transform: &Transform,
+        scale_offset: bool,
+        delta: Option<Point<Self::Delta>>,
+        round_y: bool,
+    ) -> Point<Self::Coord>;
+}
+
+/// Everything [`Scale::load_simple_points`] needs beyond the glyph itself.
+pub struct SimplePass<'a, 'v, S: Scale> {
```
cmyr, line 125: this is the type that's sealed?

read-fonts/src/tables/glyf/outline/scale.rs
```diff
@@ -0,0 +1,969 @@
...
+        F26Dot6::from_bits((Fixed::from_bits(value) * self.scale).to_bits())
+    }
+
+    fn mul_point(&self, value: Point<i32>) -> Point<F26Dot6> {
+        Point::new(self.mul(value.x), self.mul(value.y))
+    }
+
+    /// Scales a point in font units that carries a delta, given in 26.6.
+    ///
+    /// The scale factor has an i32 to 26.26 conversion built into it, so the
```
cmyr:
```suggestion
    /// The scale factor has an i32 to 26.6 conversion built into it, so the
```

read-fonts/src/tables/glyf/outline/scale.rs
```diff
@@ -0,0 +1,969 @@
...
+//!
+//! [`Scale`] holds the arithmetic that differs between target coordinate
+//! types. Loading takes the same steps whichever one is in use.
+//!
+//! Three are provided:
+//!
+//! - [`Scale26Dot6`] matches FreeType: 26.6 coordinates, 16.16 deltas, and
+//!   rounding where FreeType rounds.
+//! - [`ScaleF32`] matches HarfBuzz: floating point throughout, no rounding.
+//! - [`Unscaled`] stays in font units, which is what an autohinter needs. It
```
cmyr, line 11: was confusing to me exactly what this is used for...

cmyr (approved):
looks good, a couple little notes inline :)

read-fonts/src/tables/glyf/outline/scale.rs
```diff
@@ -0,0 +1,969 @@
...
+        let mut offset = offset;
+        if scale_offset {
+            let t = FixedTransform::new(transform);
+            offset = Point::new(
+                (Fixed::from_bits(offset.x) * hypot_fixed(t.xx, t.xy)).to_bits(),
+                (Fixed::from_bits(offset.y) * hypot_fixed(t.yy, t.yx)).to_bits(),
+            );
+        }
+        if let Some(delta) = delta {
+            offset += delta.map(Fixed::to_i32);
```
cmyr, line 698: also is this a possible overflow?

read-fonts/src/tables/glyf/outline/scale.rs
dfrg: Yeah, this type isn’t exported so it’s not possible to name it to implement the trait. As the docs mention, this isn’t a hard constraint but makes sense for a first pass at the API.

read-fonts/src/tables/glyf/outline/scale.rs
dfrg: oof, good catch

read-fonts/src/tables/glyf/outline/scale.rs
dfrg: The unscaled impl? This produces the same results as unsized 26.6 but uses one less buffer and avoids all the intermediate conversions. And happens to be exactly what the autohinter wants, so we shave off a bit of memory/time in the one component where we’re not significantly faster than FreeType.

read-fonts/src/tables/glyf/outline/scale.rs
dfrg: Delta is guaranteed to be 16 bits and I believe offset is derived from a 16 bit value so it should be safe but definitely worth checking.

dfrg: Confirmed this one can’t overflow

PR #2150, opened 2026-09-18
[read-fonts] glyf part 2: Add the outline context and hinter hook

dfrg:
`OutlineContext` is everything loading needs from a font: glyphs, variation data, and the metrics behind the phantom points. A caller that already holds those tables implements it and hands them over, rather than having them read a second time. It also states the variation location, so a context that carries one cannot be given a different one. `OutlineTables` is the implementation that reads a font.

`Hinter` is the hook a TrueType interpreter plugs into, and `GlyphZone` the points it works on. Nothing in this crate implements it.

Vertical metrics are deliberately left to the ascender, descender and bounding box, which is what FreeType does for a font without `vmtx`. A context that caches, and can afford two more tables, is where reading them belongs.

Based on #2149


read-fonts/src/tables/glyf/outline/hinter.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Scaled points, including phantom points, which hinting moves.
+    pub points: &'a mut [Point<C>],
+    /// The same points in font units.
+    pub unscaled: &'a mut [Point<i32>],
+    /// The flag for each point, which hinting also marks up.
+    pub flags: &'a mut [PointFlags],
+    /// Contour end points, relative to the start of `points`.
+    pub contours: &'a [u16],
+    /// The four phantom points, which hinting may move.
+    pub phantom: &'a mut [Point<C>; PHANTOM_POINT_COUNT],
```
cmyr, line 42: just for my own confusion, why do we have phantoms here and also included in `points`, above?

cmyr (approved):
looks good. The clanker also raises a concern I can't determine the validity of, so I will reproduce it here:


> hinter.rs — no equivalent of skrifa's original_scaled → zone 1 org (read by MD[0], MIRP cut-in,
> SHP/SHC, IP). A hinter's only option is points.to_vec() on entry, which is wrong for the last four
> because they're already rounded.

read-fonts/src/tables/glyf/outline/hinter.rs
dfrg: In skrifa, the hinter writes to the phantom points conditionally based on the backward compatibility flag which can be flipped by the glyph itself. 

For the new read-fonts code, the phantom array also serves as input allowing the hinter to build the original_scaled buffer internally… this requires unrounded phantom points but the points array carries them rounded so the hinter will splice them.

It’s a slightly different design but separates hinting memory from outline memory which skrifa currently conflates.

PR #2151, opened 2026-09-18
[read-fonts] glyf part 3: Add the outline buffers and the loaded outline

dfrg:
`Buffers` is the scratch space a load fills, all of it supplied by the caller. `BufferLengths` says how long each one has to be, and `Buffers::from_bytes` carves them from a single block for a caller that would rather make one allocation than seven. A buffer shorter than asked for is reported, never overrun.

`OutlineRef` is the result: the loaded points, flags and contour end points, borrowing those buffers.

Based on #2150


read-fonts/src/tables/glyf/outline/buffers.rs
```diff
@@ -0,0 +1,275 @@
...
+    OutlineError,
+};
+use crate::mem::take_slice;
+use bytemuck::{AnyBitPattern, NoUninit};
+
+/// Scratch space for [`OutlinePlan::load`](super::OutlinePlan::load).
+///
+/// Every buffer comes from the caller. Size them with
+/// [`OutlinePlan::buffer_lengths`](super::OutlinePlan::buffer_lengths); a buffer
+/// shorter than that is reported as [`OutlineError::InsufficientMemory`], not
```
cmyr, line 15: biggest nit in the world, but the docs here are inconsistent in using inline vs anchored links. I'd generally prefer anchored, because the prose reads better? but extremely non-blocking

cmyr (approved):
looks good

read-fonts/src/tables/glyf/outline/buffers.rs
dfrg: Fair enough. I don’t have a preference so I’m happy to do a pass after these land converting links to anchored and using that style going forward.

PR #2152, opened 2026-09-18
[read-fonts] glyf part 4: Add the outline plan and the loading pass

dfrg:
`OutlinePlan` walks a glyph and the composite tree below it without reading point data, which is cheap enough to do per draw. It says how much scratch space the result needs through `buffer_lengths`, and `load` fills it. Whether variations apply is settled when the plan is built, against the context it was built from.

`load.rs` is the pass itself: the tree walk, buffer bookkeeping, phantom points, component transforms and anchors, and the call into the hinter. Everything numeric belongs to a `Scale`; the structure here is shared by all three.

The pass checks the recursion depth that building the plan already checked, because the two take the context separately and nothing makes the tree walked the one that was surveyed.

Based on #2151


read-fonts/src/tables/glyf/outline/load.rs
```diff
@@ -0,0 +1,587 @@
...
+                for (phantom, delta) in self
+                    .phantom
+                    .iter_mut()
+                    .zip(&deltas[count - PHANTOM_POINT_COUNT..])
+                {
+                    *phantom = self.scale.add_phantom_delta(*phantom, *delta);
+                }
+                have_deltas = true;
+            }
+            self.component_delta_count += count;
```
cmyr, line 411: "`component_delta_count += count` is unconditional once `var_data` is `Some`, but the restore (L481) is guarded by `have_deltas`. Malformed packed gvar deltas leak the stack pointer, so siblings start from a higher base." – a friend of mine

cmyr (commented):
👍

read-fonts/src/tables/glyf/outline/load.rs
dfrg: A real bug, also in skrifa. This causes a nested composite to fail with an out of memory error when an earlier composite has bad variation data. Not something we’ve seen in production and the only real effect is failing to load a glyph with malformed data but fixed here for good measure.

PR #2153, opened 2026-09-18
[read-fonts] glyf part 5: Add the outline that owns its buffers

dfrg:
`OutlinePlan::load` allocates nothing, so a caller has to size and supply seven buffers. `Outline` takes the other side of that trade: it owns them, grows them as needed and reuses them across glyphs, so loading stops allocating after the first few. A caller supplies a glyph and a size.

Loading returns an `OutlineRef` borrowing those buffers; the same data is reachable through the accessors once that borrow ends.

Based on #2152


cmyr (approved):
looks good!

PR #2154, opened 2026-09-18
[read-fonts] glyf part 6: Turn on the new outline loader

dfrg:
Declares the modules the last five patches added and re-exports what they make public, so the loader compiles, lints and documents for the first time. The unit tests in those files come on with them.

The two skrifa tests draw every glyph of every font they are given through both this loader and skrifa's own, and compare the results. They live there because that is where both implementations can be seen. The loader will sit unused for months while `glyf`, `gvar` and `PointCoord` move underneath it, and those tests are what will catch the rot.

Nothing calls this yet. The existing loader is untouched.

Based on #2153


skrifa/tests/glyf_parity.rs
```diff
@@ -0,0 +1,375 @@
...
+        } else if fixed.adjusted_lsb() != lsb || fixed.adjusted_advance_width() != advance {
+            self.fail(&format!(
+                "{at}: metrics differ, lsb {:?}/{lsb:?} advance {:?}/{advance:?}",
+                fixed.adjusted_lsb(),
+                fixed.adjusted_advance_width()
+            ));
+        }
+
+        // The f32 mode has to track the 26.6 mode, which is now pinned above.
+        if let Ok(float) = mine.float.load(*tables, glyph_id, ppem) {
```
cmyr: what does it mean if a load fails?

skrifa/tests/glyf_path_parity.rs
```diff
@@ -0,0 +1,164 @@
...
+                let drawn = if forward {
+                    float
+                        .load(&tables, glyph_id, size.ppem())
+                        .map(|o| o.to_path(start, &mut mine))
+                } else {
+                    fixed
+                        .load(&tables, glyph_id, size.ppem())
+                        .map(|o| o.to_path(start, &mut mine))
+                };
+                let Ok(true) = drawn else { continue };
```
cmyr: similar, are we okay with discarding false/Err here?

cmyr (approved):
some sanity check Qs inline, but otherwise lgtm :)

skrifa/tests/glyf_parity.rs
dfrg: Good catch. This silently discards checks where skrifa succeeds and read-fonts fails which is clearly not the intention. Will fix.

skrifa/tests/glyf_path_parity.rs
dfrg: Needs a fix as above.

PR #2155, opened 2026-09-18
[read-fonts] Add a `glyf` outline loader

dfrg:
Builds an outline from `glyf` data: walks the composite tree, works out how much scratch space the result needs, and fills it. Nothing calls this yet and the existing loader is untouched.

There are two ways in. `Outline` owns its buffers and grows them as needed, so a caller supplies a glyph and a size. `OutlinePlan` allocates nothing: it surveys the glyph, reports how long each buffer has to be, and loads into buffers the caller supplies. Use it to control where the memory comes from, or to supply a hinter.

Both read through an `OutlineContext`, which states the tables and the variation location. A caller that already holds those tables implements it rather than having them read a second time; `OutlineTables` reads a font for callers that do not. Coordinates and rounding belong to a `Scale`: `Scale26Dot6` reproduces FreeType, `ScaleF32` reproduces HarfBuzz, and `Unscaled` stays in font units for an autohinter.

`Hinter` is the hook a TrueType interpreter plugs into. Nothing here implements it, but it has been driven from skrifa's interpreter over 13,486 fonts, 26 million hinted glyphs, with no disagreement.

The two skrifa tests draw every glyph of every font they are given through both this loader and skrifa's own and compare the results. The loader will sit unused for months while `glyf`, `gvar` and `PointCoord` move underneath it, and those tests are what will catch the rot.

This is the same change as the `cut/1` through `cut/6` series, squashed. The individual PRs:

#2149 glyf part 1: Add the outline error and coordinate modes
#2150 glyf part 2: Add the outline context and hinter hook
#2151 glyf part 3: Add the outline buffers and the loaded outline
#2152 glyf part 4: Add the outline plan and the loading pass
#2153 glyf part 5: Add the outline that owns its buffers
#2154 glyf part 6: Turn on the new outline loader


nicoburns:
I'm curious: what is the motivation for this second `glyf` loader? Performance?

dfrg:
The primary motivation is for HarfRust which needs accurate variable glyf extents to implement correct vertical metrics fallback. Additional benefits are improved consistency (read-fonts can load CFF outlines but not glyf), structural refactoring (skrifa duplicates almost all of the logic for the 26.6 and f32 paths) and yes, some performance (the autohinter wants unscaled outlines and pays for unnecessary memory, copies and conversions).

dfrg:
This landed from the partial PRs so closing.

PR #2157, opened 2026-09-20
[skrifa] use outer index for implicit COLR variation mapping

metsw24-max:
**Implicit COLR variation mapping drops the outer index**

With no `VarIndexMap`, the [spec](https://learn.microsoft.com/en-us/typography/opentype/spec/colr#colr-table-and-opentype-font-variations) treats `varIndexBase + n` directly as a delta-set index with the outer index in the high word and the inner in the low word, but `ColrInstance::var_deltas` kept only the low word and always used outer 0, so any index at or above 0x10000 quietly picked up a delta from the wrong `ItemVariationData` subtable. Split the index the same way `DeltaSetIndexMap::get` already does for an empty map, which should only affect fonts whose store has more than one subtable; regression test included.


dfrg (approved):
I suppose we haven't seen a font that is both missing a delta set index map and uses outer indices beyond 0. Fix looks good, thanks!

